PR #1137, opened 2014-11-03
Fixing CAT-1061 issue

qianlvable:
Fixing approach : When insert operator in number token divide the number token into another two number token instead of placing the number token by the operator. Example: if current formula is 2* 23, and I want this formula transform into 2 * 2+3 by insert an "+" in 23 the previous version will turn the formula into "2*+" which will  cause the issue cat-1061


linsm:
please reopen your pull request

use your forked repository
target of pull request should be 'develop' branch
link THIS pull request in the new one

PR #1139, opened 2014-11-11
ActionBarSherlock is replaced with ActionBarCompat

iboyko:
Build: https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/509/

appcompat lib v19.1.0 is used.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,8 +48,7 @@ configurations {
 }
 
 dependencies {
-    compile 'com.actionbarsherlock:actionbarsherlock:4.4.0@aar'
-    compile 'com.android.support:support-v4:19.1.0'
+    compile 'com.android.support:appcompat-v7:19.1.0'
     compile 'com.google.guava:guava:17.0'
     compile fileTree(include: '*.jar', dir: 'catroid/libs')
     compile fileTree(include: '*.jar', dir: 'catroid/libs-natives')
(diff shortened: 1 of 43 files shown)


avihric:
obsolete with #1162

iboyko:
Cannot be obsolete on my opinion.
CAT-1242 says to use build level 16 but it says nothing about Action bar implementation. 
Replacing ActionBarSherlock with ActionBarCompat allows fast and native switch to Lollipop UI.

aried3r:
I agree, we should use AppCompat instead.

Another question, can we delete `libraryProjects/` or is anything still using it?

obusher:
With #1162 the targetVersion is increased to 21 so we could use `com.android.support:appcompat-v7:21+`. Was your testrun green ? It seems it got deleted already. :octocat:

iboyko:
Pull request was created on Nov 11, 2014. It was not possible to get stable build at that time because of the problems with master branch. I have not started testrun after master became stable.

iboyko:
@avihric What do you think about it?

avihric:
Yes we can restore the branch and do a new PR on development branch. But first I want to process the other PRs before making new ones, except it is a critical bugfix.

iboyko:
OK, I will create an Issue.

PR #1147, opened 2015-02-03
CAT-1237 changed to English only

thmq:


robertpainsi:
So, is StringsTest still doing something?

avihric:
fixed with #1161 
CAT-1237 #fixed fixed by CAT-1252

PR #1159, opened 2015-02-22
CAT-1157 Physics Issue

anddman:
calculation of shapes / their resulting dimensions should be more accurate now
scaling of shapes in PhysicsShapeBuilder should be more accurate now aswell
Adjust tests for new scaling calculations

### DIFF
--- a/catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
+++ b/catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
@@ -124,9 +124,9 @@ public class PhysicsObject {
 		}
 
 		Vector2 dimensions = getBoundaryBoxDimensions();
-		float a = dimensions.x / 2;
-		float b = dimensions.y / 2;
-		circumference = PhysicsWorldConverter.convertNormalToBox2dCoordinate((float) Math.sqrt(a * a + b * b));
+		float w = dimensions.x / 2;
+		float h = dimensions.y / 2;
+		circumference = PhysicsWorldConverter.convertNormalToBox2dCoordinate((float) Math.sqrt(w * w + h * h));
 	}
 
 	public Type getType() {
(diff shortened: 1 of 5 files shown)


catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
```diff
@@ -123,12 +123,10 @@ private void calculateCircumference() {
...
-		calculateAABB();
-		float max1 = Math.max(getMassCenter().dst(fixtureAABBUpperRight), getMassCenter().dst(fixtureAABBLowerLeft));
-		Vector2 aabbLowerRight = new Vector2(fixtureAABBUpperRight.x, fixtureAABBLowerLeft.y);
-		Vector2 aaabbUpperLeft = new Vector2(fixtureAABBLowerLeft.x, fixtureAABBUpperRight.y);
-		float max2 = Math.max(getMassCenter().dst(aabbLowerRight), getMassCenter().dst(aaabbUpperLeft));
-		circumference = Math.max(max1, max2);
+		Vector2 dimensions = getBoundaryBoxDimensions();
+		float a = dimensions.x / 2;
+		float b = dimensions.y / 2;
+		circumference = PhysicsWorldConverter.convertNormalToBox2dCoordinate((float) Math.sqrt(a * a + b * b));
```
robertpainsi: Vector2 provides a function for the euclidean distance, ```len```
Variable names a and b... no no :smile_cat:. If this should be the center, the name center should be in the variable name.

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilder.java
```diff
@@ -106,4 +109,27 @@ private String getKey(LookData lookData, float scaleFactor) {
...
+		}
+		return vertex;
+	}
+
+	public static float scaleCoordinate (float coord, float factor) {
+		if (Math.abs(factor) < 0.001f) {
+			coord = coord*factor;
+		} else if (Math.abs(factor - 1.0f) > 0.001f) {
+			float offset;
+			offset = coord < -0.001f ? -OFFSET_VALUE : coord > 0.001f ? OFFSET_VALUE : 0.0f;
```
robertpainsi: A method for that would be better. It's used at least 3 times and a good name would clarify what it does.

PR #1165, opened 2015-02-25
CAT-1157 Physics Issue

anddman:
calculation of shapes / their resulting dimensions are more accurate now
scaling of shapes in PhysicsShapeBuilder should be more accurate now aswell
Adjust tests for new scaling calculations
Comment unused private method in PhysicsShapeBuilder

### DIFF
--- a/catroidTest/src/org/catrobat/catroid/test/physics/PhysicsActiveStageAreaTest.java
+++ b/catroidTest/src/org/catrobat/catroid/test/physics/PhysicsActiveStageAreaTest.java
@@ -51,8 +51,8 @@ public class PhysicsActiveStageAreaTest extends PhysicsBaseTest {
 	}
 
 	public void testCircumferenceCalculation() {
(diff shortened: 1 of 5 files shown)


catroidTest/src/org/catrobat/catroid/test/physics/PhysicsActiveStageAreaTest.java
```diff
@@ -52,7 +52,7 @@ protected void setUp() throws Exception {
 
 	public void testCircumferenceCalculation() {
 		assertTrue("calculated rectangle circumference not equal to expected value",
-				Math.abs(Math.round(physicsObject.getCircumference() * PhysicsWorld.RATIO) - EXPECTED_CIRCUMFERENCE_125X125) <= 1);
+				Math.abs(physicsObject.getCircumference() - EXPECTED_CIRCUMFERENCE_125X125) <= 1);
```
robertpainsi: assertEquals-with-delta to the rescue! :smile_cat:

PR #1166, opened 2015-02-25
rebased/updated languageTest

thmq:
seems to be a problem with: BrickValueParameterTest.testSoundBricksDefaultValues, everything in the Jenkins run: https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/697/


SimonStefan:
Why are you using toLowerCase?

thmq:
because there have been problems while comparing the strings and this fixes it and still assures the correctness of the strings

robertpainsi:
If so, you should consider to add a note to the commit message about the problem.

SimonStefan:
We have a green build, so could you provide a link to a testrun with faulty comparison? I am asking, because these tests should probably also ensure that the title is correctly written.
Is this part of CAT-1070?

P.S.: We should always use the name written in the code.xml and not the filename. Currently we are renaming the filename correctly, so there should be a difference only when using characters that are not allowed in a filename.

thmq:
ok to test

thmq:
ok to test

thmq:
ok to test

thmq:
ok to test

thmq:
ok to test

thmq:
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/734/

PR #1155, opened 2015-02-19
CAT-1243 Adjust gradle files to make Catroid work with Android Studio 1....

amix73:
...x

PR #1167, opened 2015-02-25
Cat-1243

kaufi:
substitutes #1155 

we definately need a ticket for the Handlerleak Lint warning!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,8 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
-        classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.10.+'
+        classpath 'com.android.tools.build:gradle:1.0.1'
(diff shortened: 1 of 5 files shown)


kaufi:
it seems, that with the new version, all tasks for the tests have been renamed...
assembleDebugTest->assembleDebugAndroidTest and so on!
so we have to be careful, since this requires a change on the Jenkins side

kaufi:
@avihric @SimonStefan shall we forget about the gradle plugin update (at least to the latest version) for now and use the latest possible version which is compatible with our current jenkins configuration?

SimonStefan:
yes, I think we can postpone the update to version 1.1.x

SimonStefan:
ok to test

kaufi:
something is still not working - have to have a look into it; also strange, that on my local machine "gradle check" reports 0 lint warnings/errors, and here we have 2 of them :)

avihric:
ok to test

avihric:
ok to test

avihric:
ok to test

linsm:
ok to test

PR #1196, opened 2015-03-25
Remove duplicate test util classes

robertpainsi:
Remove uitest Reflection class because there is an identical copy in
the package org.catrobat.catroid.test.utils. Also remove the Reflection
test which assures that both copies of the Reflection class are the
same.


obusher:
Nice! :) Looks like there are some other duplicate classes in these directories too. (`XMLValidationUtil, SimulatedSensorManager, SimulatedSoundRecorder ...`)


Maybe we should create a `global-util` directory (package) to resolve the dependency of `uitest` from `test`.

robertpainsi:
Ah, thanks for the tip @obusher !
I will take a look over all the duplicate utils classes. So please don't merge atm.

robertpainsi:
@obusher: There is already a common package for the Bluetooth stuff. This would be a good place for all the common util classes too.

However, I don't like it :pouting_cat: because the classes
```org.catrobat.catroid.common.CatroidService``` and
```org.catrobat.catroid.common.SoundInfo``` aren't in the same project nor in the same folder even if it looks so.

Other developer should jump in and discuss it with you. \**slowly sneaking away*\*

obusher:
Looks fine for now! :)

robertpainsi:
The test.utils package it is... *for now*! :smile_cat:
Thanks!

I'll create a ticket to create a global/common/whatever package for the tests und uitests later in the evening.

aried3r:
Most of this would be a lot easier with the intended directory structure of
Gradle, so I'd rather work on that.

robertpainsi:
@aried3r: This is the current project structure in AS

screenshot from 2015-03-27 16 45 52

I don't see any difference to a newly created AS project (beside the internal representation on the file system itself). Smiley with question mark not found, using :confused:

PR #1201, opened 2015-03-30
CAT-1157 Physics Issue

anddman:
calculate physicsObject base shapes according to accuracyLevels
calculate base shapes in separate thread, adjust PhysicsShapeBuilderStrategyFastHull for simultaneous calls
remove public boolean visible in class which extends another with private boolean with same name
add invisible conditions to physicsObjects
handling physicsObjectStates for very small objects
update PhysicsObjectStates after every call to PhysicsObject.setType
downscale large Pixmaps when creating physicsObject Shapes to increase performance
adjust physics tests
correct calculation of PhysicsObjectStateHandler's positionCondition


Gorphi:
Bitte rebasen und PR neu öffnen.

PR #1202, opened 2015-03-30
Current Implementation CAT-1216

iboyko:
Request contains current implementation of Issue CAT-1216. 
BrickViewProvider and BrickView are central classes. BrickViewProvider manages creation of BrickViews. 
*Only Logic implementation is to mention. All tests should be checked/rewritten and do not work for now.*


obusher:
Tests should be refactored ASAP... :zap:

iboyko:
@obusher is correct. At the implementation will follow. 
This branch is also used as info for all contributors and for public development. 
@avihric please apply Pull request

PR #1203, opened 2015-03-30
Cat 1254

X4mp:
Fixes 
- CAT-1314
- CAT-1258
- CAT-1259
- CAT-1282
- CAT-1261
- CAT-1263


aried3r:
@robertpainsi, please consider not using the *AvoidStaticImport* rule, nothing is gained in these cases.

robertpainsi:
@aried3r: The lowest I'll get, is removing the *AvoidStaticImport* but still fix all cases except the StorageHandler (even those static imports of the methods buildPath and buildProjectPath in the StorageHandler which I don't like at all). Deal?

(The cucumber stuff will be postponed. If those are just test methods like the junit ones, I'm fine with static imports and/or star imports in these cases too)

aried3r:
> The lowest I'll get, is removing the AvoidStaticImport but still fix all cases except the StorageHandler

I'm a fan of static imports if it makes the code more readable, having `Constants.` and `Utils.` in so many places just clutters up code if it's clear what is being done. I'm not saying every call to the utility classes should be made static, but in some cases I think it's justified.

Another example is
```Java
Toast.makeText(context, text, Toast.LENGTH_SHORT);
myView.setVisibility(View.INVISIBLE);
```
nothing is gained by having `Toast.` or `View.` here, the context is clear and I guess most people would not wonder where this constant is coming from, because of the context it is used in.

Here's what Google does in two of its official apps:
[iosched](https://github.com/google/iosched/search?p=1&q=%22import+static%22&type=Code&utf8=%E2%9C%93)
[UniversalMusicPlayer](https://github.com/googlesamples/android-UniversalMusicPlayer/search?utf8=%E2%9C%93&q=%22import+static%22) (here, they are used very scarcely)

> Deal?

You're the maintainer, in the end, you decide what goes into the repository and what doesn't. I'm just not a fan of blindly adding static checker rules and cluttering up code (see also #1205 wrt to `super()`).

> or star imports

Please, no!

robertpainsi:
@aried3r Hell yeah, these are good examples! :+1: Thanks. I take that advice.
@X4mp Please drop the commit about the AvoidStaticImport

Btw: I always use star imports for junit (only case atm). ```import junit.framework.Assert.*``` I don't see any disadvantages there.
So, teach me master! :smiley_cat:

X4mp:
will be done asap

Robert Painsi <notifications@github.com> schrieb am Di., 31. März 2015 um
01:34 Uhr:

X4mp:
retest this please

PR #1204, opened 2015-03-30
Changes from Catroweb

thomasmauerhofer:
Testrun: https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/1186/

PR #1206, opened 2015-03-30
Update OkHttp to 2.3.0, kill ConnectionWrapper

aried3r:
This is just a test, because I had problems using the custom branch job of Jenkins, but the code is 90% there. This would supersede #1204.


aried3r:
Ok, I might need some [help here](https://jenkins.catrob.at/job/Catroid-Build-SourceTests-StaticChecks/1589/console), what went wrong?

PR #1214, opened 2015-04-10
CAT-1157 Physics Issue

anddman:
calculate physicsObject base shapes according to accuracyLevels
calculate base shapes in separate thread, adjust PhysicsShapeBuilderStrategyFastHull for simultaneous calls
remove public boolean visible in class which extends another with private boolean with same name
add invisible conditions to physicsObjects
handling physicsObjectStates for very small objects
update PhysicsObjectStates after every call to PhysicsObject.setType
downscale large Pixmaps when creating physicsObject Shapes to increase performance
adjust physics tests
correct calculation of PhysicsObjectStateHandler's positionCondition
fix not all fixtures of PhysicsObject being destroyed

### DIFF
--- a/catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
+++ b/catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
@@ -23,6 +23,7 @@
 package org.catrobat.catroid.physics;
 
 import com.badlogic.gdx.graphics.g2d.Batch;
+import com.badlogic.gdx.math.Vector2;
 
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.common.LookData;
@@ -33,24 +34,21 @@ import java.util.LinkedList;
 
 public class PhysicsLook extends Look {
 
-	private final transient PhysicsObject physicsObject;
+	public static final float SCALE_FACTOR_ACCURACY = 10000.0f;
+	public static final float MINIMUM_PHYSICS_OBJECT_SIZE = 8.0f;
 
-	private PhysicsObjectStateHandler physicsObjectStateHandler = new PhysicsObjectStateHandler();
+	private final PhysicsObject physicsObject;
+	private final PhysicsObjectStateHandler physicsObjectStateHandler = new PhysicsObjectStateHandler();
 
 	public PhysicsLook(Sprite sprite, PhysicsWorld physicsWorld) {
 		super(sprite);
 		physicsObject = physicsWorld.getPhysicsObject(sprite);
 	}
 
-	@Override
-	public void setVisible(boolean visible) {
-		super.setVisible(visible);
-		updatePhysicsObjectState(true);
-	}
 
 	@Override
 	public void setTransparencyInUserInterfaceDimensionUnit(float percent) {
-		super.setTransparencyInUserInterfaceDimensionUnit(percent);
+		updateTransparency(percent, false);
 		updatePhysicsObjectState(true);
 	}
 
@@ -59,6 +57,7 @@ public class PhysicsLook extends Look {
 		super.setLookData(lookData);
 		PhysicsWorld physicsWorld = ProjectManager.getInstance().getCurrentProject().getPhysicsWorld();
 		physicsWorld.changeLook(physicsObject, this);
+		updatePhysicsObjectState(true);
 	}
 
 	@Override
@@ -136,10 +135,24 @@ public class PhysicsLook extends Look {
 
 	@Override
 	public void setScale(float scaleX, float scaleY) {
+		Vector2 oldScales = new Vector2(getScaleX(), getScaleY());
+		if (scaleX < 0.0f || scaleY < 0.0f) {
+			scaleX = 0.0f;
+			scaleY = 0.0f;
+		}
+
+		int scaleXComp = Math.round(scaleX * SCALE_FACTOR_ACCURACY);
+		int scaleYComp = Math.round(scaleY * SCALE_FACTOR_ACCURACY);
+		if (scaleXComp == Math.round(oldScales.x * SCALE_FACTOR_ACCURACY) && scaleYComp == Math.round(oldScales.y * SCALE_FACTOR_ACCURACY)) {
+			return;
+		}
+
 		super.setScale(scaleX, scaleY);
-		if (null != physicsObject) {
+
+		if (physicsObject != null) {
 			PhysicsWorld physicsWorld = ProjectManager.getInstance().getCurrentProject().getPhysicsWorld();
 			physicsWorld.changeLook(physicsObject, this);
+			updatePhysicsObjectState(true);
 		}
 	}
 
@@ -147,8 +160,23 @@ public class PhysicsLook extends Look {
 		physicsObjectStateHandler.update(record);
 	}
 
+	@Override
+	public boolean isVisible() {
+		return super.isVisible() && !physicsObjectStateHandler.isInvisible();
+	}
+
+	@Override
+	public void setVisible(boolean visible) {
+		super.setVisible(visible);
+		physicsObjectStateHandler.update(true);
+	}
+
 	public boolean isHangedUp() {
 		return physicsObjectStateHandler.isHangedUp();
+	}
+
+	public void setNonColliding(boolean nonColliding) {
+		physicsObjectStateHandler.setNonColliding(nonColliding);
 	}
 
 	public void startGlide() {
@@ -172,18 +200,19 @@ public class PhysicsLook extends Look {
 	private class PhysicsObjectStateHandler {
 		//private final String TAG = PhysicsObjectStateHandler.class.getSimpleName();
 
-		private LinkedList<PhysicsObjectStateCondition> hangupConditions =
-				new LinkedList<PhysicsObjectStateCondition>();
-		private LinkedList<PhysicsObjectStateCondition> nonCollidingConditions =
-				new LinkedList<PhysicsObjectStateCondition>();
+		private LinkedList<PhysicsObjectStateCondition> invisibleConditions = new LinkedList<PhysicsObjectStateCondition>();
+		private LinkedList<PhysicsObjectStateCondition> hangupConditions = new LinkedList<PhysicsObjectStateCondition>();
+		private LinkedList<PhysicsObjectStateCondition> nonCollidingConditions = new LinkedList<PhysicsObjectStateCondition>();
 		private LinkedList<PhysicsObjectStateCondition> fixConditions = new LinkedList<PhysicsObjectStateCondition>();
 
+		private PhysicsObjectStateCondition sizeCondition;
 		private PhysicsObjectStateCondition positionCondition;
 		private PhysicsObjectStateCondition visibleCondition;
 		private PhysicsObjectStateCondition transparencyCondition;
 		private PhysicsObjectStateCondition glideToCondition;
-		private boolean glideToIsActive = false;
 
+		private boolean invisible = false;
+		private boolean glideToIsActive = false;
 		private boolean hangedUp = false;
 		private boolean fixed = false;
 		private boolean nonColliding = false;
@@ -210,14 +239,20 @@ public class PhysicsLook extends Look {
 
 				private boolean isXOutsideActiveArea() {
 					return Math.abs(PhysicsWorldConverter.convertBox2dToNormalCoordinate(physicsObject.getMassCenter().x))
-							- PhysicsWorldConverter.convertBox2dToNormalCoordinate(physicsObject.getCircumference())
-							> PhysicsWorld.activeArea.x / 2.0f;
+							- physicsObject.getCircumference() > PhysicsWorld.activeArea.x / 2.0f;
 				}
 
 				private boolean isYOutsideActiveArea() {
 					return Math.abs(PhysicsWorldConverter.convertBox2dToNormalCoordinate(physicsObject.getMassCenter().y))
-							- PhysicsWorldConverter.convertBox2dToNormalCoordinate(physicsObject.getCircumference())
-							> PhysicsWorld.activeArea.y / 2.0f;
+							- physicsObject.getCircumference() > PhysicsWorld.activeArea.y / 2.0f;
+				}
+			};
+
+			sizeCondition = new PhysicsObjectStateCondition() {
+				@Override
+				public boolean isTrue() {
+					Vector2 dimensions = physicsObject.getBoundaryBoxDimensions();
+					return dimensions.x < MINIMUM_PHYSICS_OBJECT_SIZE || dimensions.y < MINIMUM_PHYSICS_OBJECT_SIZE;
 				}
 			};
 
@@ -225,7 +260,7 @@ public class PhysicsLook extends Look {
 				@Override
 				public boolean isTrue() {
 					//Log.d(TAG, "visibleCondition:"+!isVisible);
-					return !visible;
+					return !isVisible();
 				}
 			};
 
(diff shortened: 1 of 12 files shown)


catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -42,15 +45,16 @@ public PhysicsLook(Sprite sprite, PhysicsWorld physicsWorld) {
...
 
 	@Override
 	public void setTransparencyInUserInterfaceDimensionUnit(float percent) {
-		super.setTransparencyInUserInterfaceDimensionUnit(percent);
+		if (percent < 0f) {
+			percent = 0f;
+		} else if (percent >= 100f) {
+			percent = 100f;
+		}
+		alpha = (100f - percent) / 100f;
```
robertpainsi: Duplicate code :scream_cat: (I know why you copied it but should still remove and fix it)

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -59,6 +63,7 @@ public void setLookData(LookData lookData) {
 		super.setLookData(lookData);
 		PhysicsWorld physicsWorld = ProjectManager.getInstance().getCurrentProject().getPhysicsWorld();
 		physicsWorld.changeLook(physicsObject, this);
+		updatePhysicsObjectState(true);
```
robertpainsi: The updatePhysicsObjectState parameter is always true :wink: just saying

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -128,10 +133,26 @@ public void setRotation(float degrees) {
...
+
+		int scaleXComp = Math.round(scaleX * SCALEFACTOR_ACCURACY);
+		int scaleYComp = Math.round(scaleY * SCALEFACTOR_ACCURACY);
+		if (scaleXComp == Math.round(oldScales.x * SCALEFACTOR_ACCURACY) && scaleYComp == Math.round(oldScales.y * SCALEFACTOR_ACCURACY)) {
+			return;
+		}
+
+		scaleX = scaleXComp / SCALEFACTOR_ACCURACY;
+		scaleY = scaleYComp / SCALEFACTOR_ACCURACY;
 		super.setScale(scaleX, scaleY);
```
robertpainsi: This is not a good idea. The scale values for the look should always be the values specified by the user.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -128,10 +133,26 @@ public void setRotation(float degrees) {
...
-		if (null != physicsObject) {
+
+		if (physicsObject != null) {
```
robertpainsi: All those nasty null checks are screaming for a NullObject!

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -33,6 +34,8 @@
 
 public class PhysicsLook extends Look {
 
+	public static final float SCALEFACTOR_ACCURACY = 1000.0f;
+	public static final float MINIMUM_PHYSICSOBJECT_SIZE = 8.0f;
```
robertpainsi: Scale factor and physics object are two separate words.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -33,6 +34,8 @@
...
 	private final transient PhysicsObject physicsObject;
```
robertpainsi: This doesn't have to be transient since the whole Look will never be saved.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -33,6 +34,8 @@
...
 
 	private PhysicsObjectStateHandler physicsObjectStateHandler = new PhysicsObjectStateHandler();
```
robertpainsi: This can be final I guess

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -139,10 +160,25 @@ public void updatePhysicsObjectState(boolean record) {
 		physicsObjectStateHandler.update(record);
 	}
 
+	@Override
+	public boolean isVisible() {
+		return super.isVisible() && (physicsObjectStateHandler != null ? !physicsObjectStateHandler.isInvisible() : true);
```
robertpainsi: There should be no need to check for null, I guess

robertpainsi:
Even if you didn't write the code for some of the code review comments, you should consider to fix them. Since they aren't related to the ticket itself too, you should at least fix them in a new commit.

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilder.java
```diff
@@ -44,70 +49,164 @@
...
+			this.lookData = lookData;
+			this.sizeAdjustmentScaleFactor = sizeAdjustmentScaleFactor;
+		}
+
+		@Override
+		public void run() {
+			physicsShapeBuilder.buildBaseShapes(lookData, sizeAdjustmentScaleFactor);
+		}
+
+		public static void main(String[] args) {
```
robertpainsi: A main method? :scream_cat:

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilder.java
```diff
@@ -44,70 +49,164 @@
...
+			(new Thread(new BaseShapeBuilder())).start();
```
robertpainsi: Unnecessary braces

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilderStrategyFastHull.java
```diff
@@ -35,26 +35,26 @@
 
 public final class PhysicsShapeBuilderStrategyFastHull implements PhysicsShapeBuilderStrategy {
 
-	private final Stack<Vector2> convexHull = new Stack<Vector2>();
+	public static final int MINIMUM_PIXEL_ALPHA_VALUE = 1;
```
robertpainsi: private

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilderStrategyFastHull.java
```diff
@@ -35,26 +35,26 @@
...
 
 	@Override
 	public Shape[] build(Pixmap pixmap, float scale) {
-
+		Stack<Vector2> convexHull = new Stack<Vector2>();
```
robertpainsi: A few lines done with it. Declare variables as close as possible to the first use.

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilderStrategyFastHull.java
```diff
@@ -35,26 +35,26 @@
...
 		if (pixmap == null) {
 			return null;
 		}
 
 		int width = pixmap.getWidth();
 		int height = pixmap.getHeight();
+		float coordinateAdjustmentValue = 1.0f;
 		convexHull.clear();
```
robertpainsi: This isn't needed anymore.

catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
```diff
@@ -333,6 +328,19 @@ protected void setCollisionBits(short categoryBits, short maskBits) {
...
+			updateNonCollidingState();
+		}
+	}
+
+	private void updateNonCollidingState() {
+		if (body.getUserData() != null && body.getUserData() instanceof Sprite) {
+			Object look = ((Sprite) body.getUserData()).look;
+			if (look != null && look instanceof PhysicsLook) {
+				((PhysicsLook) look).setNonColliding(collisionMaskRecord == PhysicsWorld.MASK_NOCOLLISION);
+			}
```
robertpainsi: I don't like the whole method... :pouting_cat:
Not sure what this method does nor why the PhysicsObject has to know the sprite or look. Maybe you can do some magic and get rid of it...

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
anddman: I assume you're referring to the code of the overridden method in Look.java
If so, how would you suggest solving this issue - the overriding method may not update the "visible" boolean which the superclass' method does.

I was thinking of implementing another method in Look.java which contains the code in question and takes a boolean as second parameter which is used to determine whether "visible" should be updated. So both methods in question would call this new method passing only a different boolean parameter.

Would this be acceptable? Thought I'd ask before starting a trial-and-error process :)

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
anddman: Would you like me to remove the parameter from that method's implementation? I don't know what whoever implemented it at the time was thinking so I'm somewhat hesitant to do so.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
anddman: You're right, of course. I'll remove this but am undecided as to whether a check should be performed which would prevent "re-setting" the same scale (as in line 144). This would require a float comparison delta which would then also disallow minuscule changes in scale (which could of course add up to larger numbers). What's your recommendation?

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
robertpainsi: PhysicsObjectStateHandler.update(boolean) method doesn't need a parameter as well.
@Gorphi did some refactoring related to these methods. Maybe he can help.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
robertpainsi: I don't know how the scaling for the physics stuff works and I don't have the time to dig into that right now. :crying_cat_face: 
But setScale should definitely call super.setScale with the original scaleX and scaleY values and without any guard clauses (no early returns). If and how you forward the scale changes to the physics engine is completely up to you. :wink:

catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
Gorphi: It is just an idea, but maybe its an option to pass the PhysicsObjectStateHandler Object through the userdata?

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
Gorphi: if I remember correctly, that parameter was used for a staterecording (especially for the hangup state, that had to be recorded to restart if the hangup is over)
That whole concept should probably be mentioned during the refactor/extention of the PhysicsObjectStateHandler.

anddman: soo.. I guess that means we should keep them?

catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
anddman: For what it does: when collision bits of a physicsObject are changed, it passes this information  to the physicsLook's objectStateHandler which has some conditions depending on whether it is a colliding or non-colliding object. 

It is called mostly from the setType method and I couldn't find a better way of accessing the physicsObject's respective physicsLook.

What do you mean exactly by "passing it through the userdata"?

Gorphi: Maybe its possible to save the PhysicsObjectStateHandler instead of the Sprite into the bodies userdata. But I am afraid that the Sprite is necessary anywhere else.
On the other side, the body knows the sprite anyway, so why not use it?

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
Gorphi: For now, I would say yes. (should be a refactoring ticket)

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
anddman: I get what you're saying but the problem is that re-scaling physicsObjects uses the scaleX variable of the look object (defined in Actor -> Widget -> Image -> Look) so there are no separate scale values for physicsLooks and the values passed to super.setScale are the exact same which are used by the physics engine.

Should I just call super.setScale with the original values and remove the lines from 142 to 149 ? I don't think allowing scales of a negative value would be a good idea and would therefore keep lines 136-140.

catroid/src/org/catrobat/catroid/physics/PhysicsObject.java
anddman: As far as I can tell, the body knows the sprite only via the "userData", which is what I am using at the moment. Is there something I'm missing here?

I'm sorry but I can't think of another/a better way to access the object's respective physicsLook - at least not without changing/adding significant amount of code - perhaps adding a direct reference but I think that's out of the question.

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -33,24 +34,21 @@
 
 public class PhysicsLook extends Look {
 
-	private final transient PhysicsObject physicsObject;
+	public static final float SCALEFACTOR_ACCURACY = 10000.0f;
```
robertpainsi: SCALE_FACTOR_ACCURACY

catroid/src/org/catrobat/catroid/physics/PhysicsLook.java
```diff
@@ -237,6 +273,10 @@ public boolean isTrue() {
 				}
 			};
 
+
```
robertpainsi: Remove me

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilder.java
```diff
@@ -37,6 +37,11 @@
 import java.util.Map;
 
 public class PhysicsShapeBuilder {
+
+	public static final int BASESHAPEBUILDER_THREAD_PRIORITY = 3;
```
robertpainsi:
```java
private ... BASE_SHAPE_BUILDER_THREAD_PRIORITY
```

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilder.java
```diff
@@ -44,70 +49,164 @@
...
+			}
+		}
+		return sizeAdjustmentFactor;
 	}
 
-	private float getScaleLevel(float scaleFactor) {
-		if (scaleFactor >= 0.25f) {
-			return 0.25f;
+	public void buildBaseShapes(LookData lookData, float sizeAdjustmentScaleFactor) {
+		for (int i = ACCURACY_LEVELS.length-1; i >= 0; i--) {
```
robertpainsi: Spaces between ```length - 1``` are missing.
I guess this has to be in reverse order?

catroid/src/org/catrobat/catroid/physics/shapebuilder/PhysicsShapeBuilderStrategyFastHull.java
```diff
@@ -35,26 +35,25 @@
 
 public final class PhysicsShapeBuilderStrategyFastHull implements PhysicsShapeBuilderStrategy {
 
-	private final Stack<Vector2> convexHull = new Stack<Vector2>();
+	private static final int minimumPixelAlphaValue = 1;
```
robertpainsi: Upper case

PR #1215, opened 2015-04-12
CAT-1332

obusher:
Fixes a test-job-aborting timeout.

* prepareStageForTest() call in setUp() for each stage testing UI-test.
* solo.goBack() in StageTestSimple after stage testing.

[Test Green] (https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/1475/)

### DIFF
--- a/catroidTest/src/org/catrobat/catroid/uitest/util/UiTestUtils.java
+++ b/catroidTest/src/org/catrobat/catroid/uitest/util/UiTestUtils.java
@@ -134,6 +134,7 @@ import org.catrobat.catroid.formulaeditor.FormulaElement;
 import org.catrobat.catroid.formulaeditor.InternToken;
 import org.catrobat.catroid.formulaeditor.InterpretationException;
 import org.catrobat.catroid.io.StorageHandler;
+import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.stage.StageListener;
 import org.catrobat.catroid.test.utils.Reflection;
 import org.catrobat.catroid.ui.MainMenuActivity;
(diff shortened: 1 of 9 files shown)


obusher:
ok to test

obusher:
ok to test

catroidTest/src/org/catrobat/catroid/uitest/util/UiTestUtils.java
```diff
@@ -2132,4 +2133,12 @@ public boolean isSatisfied() {
 
 		return wantedState;
 	}
+
+	public static void leaveStageSafeButSlow(Solo solo) {
+		if(solo.getCurrentActivity().getClass() == StageActivity.class){
```
aried3r: Missing space, also, do you want to compare like this instead of using `instanceof`? Just checking.

obusher: Yea the `instanceof` operator would be also an option, but in this case i favour comparing classes. ([instanceof vs. getClass( )](http://stackoverflow.com/questions/4989818/instanceof-vs-getclass)) :ocean:

obusher:
ok to test

obusher:
ok to test

thmq:
remove TODOs, either fix the problem or if it has nothing to do with the problems mentioned in CAT-1332, please create a new Jira-issue which explains why the TODOs are necessary.

obusher:
This test-job-aborting issue does not occur anymore.

PR #1205, opened 2015-03-30
[CAT-1276] Introduce pmd java-controversial ruleset

cheidenreich:
JIRA issue:
https://jira.catrob.at/browse/CAT-1276

Jenkins run:
https://jenkins.catrob.at/job/CatroidDeviceTest/1396/

have fun reviewing :P

PR #1216, opened 2015-04-13
[CAT-1276] Introduce pmd java-controversial ruleset

cheidenreich:
reopen of #1205 (after force push)

JIRA issue:
https://jira.catrob.at/browse/CAT-1276

Jenkins run:
https://jenkins.catrob.at/job/CatroidDeviceTest/1676/

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/XmlHeader.java
+++ b/catroid/src/org/catrobat/catroid/content/XmlHeader.java
@@ -67,94 +67,96 @@ public class XmlHeader implements Serializable {
 	@SuppressWarnings("unused")
 	private String userHandle = "";
 
+	@SuppressWarnings("PMD.UnnecessaryConstructor")
 	public XmlHeader() {
+		super();
 	}
 
-	String getProgramName() {
+	public String getProgramName() {
 		return programName;
 	}
 
-	void setProgramName(String programName) {
+	public void setProgramName(String programName) {
 		this.programName = programName;
 	}
 
-	String getDescription() {
+	public String getDescription() {
 		return description;
 	}
 
-	void setDescription(String description) {
+	public void setDescription(String description) {
 		this.description = description;
 	}
 
-	float getCatrobatLanguageVersion() {
+	public float getCatrobatLanguageVersion() {
 		return catrobatLanguageVersion;
 	}
 
-	void setCatrobatLanguageVersion(float catrobatLanguageVersion) {
+	public void setCatrobatLanguageVersion(float catrobatLanguageVersion) {
 		this.catrobatLanguageVersion = catrobatLanguageVersion;
 	}
 
-	String getPlatform() {
+	public String getPlatform() {
 		return platform;
 	}
 
-	void setPlatform(String platform) {
+	public void setPlatform(String platform) {
 		this.platform = platform;
 	}
 
-	String getApplicationBuildName() {
+	public String getApplicationBuildName() {
 		return applicationBuildName;
 	}
 
-	void setApplicationBuildName(String applicationBuildName) {
+	public void setApplicationBuildName(String applicationBuildName) {
 		this.applicationBuildName = applicationBuildName;
 	}
 
-	int getApplicationBuildNumber() {
+	public int getApplicationBuildNumber() {
 		return applicationBuildNumber;
 	}
 
-	void setApplicationBuildNumber(int applicationBuildNumber) {
+	public void setApplicationBuildNumber(int applicationBuildNumber) {
 		this.applicationBuildNumber = applicationBuildNumber;
 	}
 
-	String getApplicationName() {
+	public String getApplicationName() {
 		return applicationName;
 	}
 
-	void setApplicationName(String applicationName) {
+	public void setApplicationName(String applicationName) {
 		this.applicationName = applicationName;
 	}
 
-	String getApplicationVersion() {
+	public String getApplicationVersion() {
 		return applicationVersion;
 	}
 
-	void setApplicationVersion(String applicationVersion) {
+	public void setApplicationVersion(String applicationVersion) {
 		this.applicationVersion = applicationVersion;
 	}
 
-	String getDeviceName() {
+	public String getDeviceName() {
 		return deviceName;
 	}
 
-	void setDeviceName(String deviceName) {
+	public void setDeviceName(String deviceName) {
 		this.deviceName = deviceName;
 	}
 
-	double getPlatformVersion() {
+	public double getPlatformVersion() {
 		return platformVersion;
 	}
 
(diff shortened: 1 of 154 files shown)


robertpainsi:
I guess you could have reopened the other PR as well. :wink: 

Please remove all added super calls and exclude the CallSuperInConstructor rule. (https://github.com/Catrobat/Catroid/pull/1205#issuecomment-87746232) (I already updated the jira ticket as well)

~~Also restore all deleted default constructor implementing Serializable (directly or indirectly). https://github.com/Catrobat/Catroid/pull/1205#issuecomment-87750178
LookData, SoundInfo and all XYZBricks implement Serializable and therefore should have a default constructor.~~

I'm not sure about that anymore. First of all it works without default constructors without a problem (https://jenkins.catrob.at/job/Catroid-PullRequest-TestJob/181/)
Second, I finally read http://rolandtapken.de/blog/2010-08/let-xstream-call-default-constructor-where-possible ^^ We don't use this approach. Also it will just handle final transient problems and not even all of them

> This reduces the "final transient" problem to classes without a default constructor.

Third, and the most important one

> XStream is not calling the default constructor during deserialization.

http://xstream.codehaus.org/faq.html#Serialization_no_ctor_running

So, keep the UnnecessaryConstructor and remove all unnecessary constructors?
(I didn't do any more research :wink:)

Pinging all the people! (Do we have an "xstream guy"?)
@avihric @obusher @aried3r @sjaindl @SimonStefan

cheidenreich:
@robertpainsi unfortunately reopening of the old PR was not possible since i force pushed after squashing.

robertpainsi:
@cheidenreich Ah, got it! Didn't know that, thanks for the info!

obusher:
ok to test

obusher:
ok to test

obusher:
ok to test

obusher:
ok to test

cheidenreich:
any update on XStream?

catroidTest/src/org/catrobat/catroid/test/utils/Reflection.java
```diff
@@ -127,8 +127,6 @@ public ParameterList(Object... parameters) {
 				objectClass = int.class;
 			} else if (objectClass == Long.class) {
 				objectClass = long.class;
-			} else if (objectClass == Short.class) {
-				objectClass = short.class;
 			}
```
obusher: Which rule coplained about those two lines? :ear_of_rice:

config/pmd_controversial.xml
```diff
@@ -0,0 +1,53 @@
...
+  -->
+
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         name="Catroid PMD unnecessary ruleset"
+         xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
+         xsi:schemaLocation="http://pmd.sourceforge.net/ruleset/2.0.0 http://pmd.sourceforge.net/ruleset_2_0_0.xsd" >
+	<description >PMD Java Controversial Ruleset for Catroid/PocketCode</description >
+	<rule ref="rulesets/java/controversial.xml" >
+		<exclude name="AssignmentInOperand" />
+        <exclude name="AtLeastOneConstructor" />
```
obusher: Reformat the `<exclude />` tags, most of them are too much on the left side. They should be like the happy one in `line:31`. :walking:

obusher:
Lets open a separate ticket which treats the `~DefaultConstructor` rule, Xstream related research and testing, so we can merge the other rules swifter. :mag:

catroidTest/src/org/catrobat/catroid/test/utils/Reflection.java
robertpainsi: http://pmd.sourceforge.net/pmd-4.3.0/rules/controversial.html#AvoidUsingShortType

@cheidenreich
Please undo these changes and exclude the rule. The physics engine (and box2d) is using shorts for collision filtering.

catroid/src/org/catrobat/catroid/content/XmlHeader.java
```diff
@@ -67,94 +67,96 @@
 	@SuppressWarnings("unused")
 	private String userHandle = "";
 
+	@SuppressWarnings("PMD.UnnecessaryConstructor")
 	public XmlHeader() {
+		super();
 	}
 
-	String getProgramName() {
+	public String getProgramName() {
```
robertpainsi: Hmm, I can't find the pmd rule in the docu at the moment, but
- Adding private if the access modifier is missing is a good idea :wink: 
- Adding public if the access modifier is missing isn't a good idea :wink: because there is also package-private.

robertpainsi:
Finally I took a look at the xstream stuff. All approaches have it pros and cons.

Calling the default constructor if there is one, otherwise don't call any constructor, seems a little bit random. This is the approach mentioned in http://rolandtapken.de/blog/2010-08/let-xstream-call-default-constructor-where-possible
I don't like that.

Using the PureJavaReflectionProvider (also mentioned in http://rolandtapken.de/blog/2010-08/let-xstream-call-default-constructor-where-possible), all serializable classes need a default constructor which is called when deserializing an object. This is consistent.

Also the default constructor has to be empty imo. Because when deserializing an object, the default constructor is called first and after that all serialized members are set. So calling a method in the constructor using a serialized member is a paddlin. Setting a serialized member in the constructor is a paddlin.

The last option is the current one. Don't call any constructor when deserializing. This also means that final transient variables won't be initialized, so we can't use the final modifier for transient variables.

In all cases, readResolve is the way to go to do stuff when deserializing an object.

Summarized:
- We can use transient final variables but have to assure that the constructors are and stay empty.

or

- We can not use transient final variables for objects but also don't have to care about constructors.

Any thoughts? opinions? Am I missing something?
ping @avihric

catroid/src/org/catrobat/catroid/content/XmlHeader.java
robertpainsi: Found it ^^ It's the DefaultPackage rule. You should exclude it.
http://pmd.sourceforge.net/pmd-5.1.1/rules/java/controversial.html#DefaultPackage

robertpainsi:
> Lets open a separate ticket which treats the ~DefaultConstructor rule, Xstream related research and testing, so we can merge the other rules swifter.

Done, https://jira.catrob.at/browse/CAT-1475

obusher:
@robertpainsi Thank you very much!
@cheidenreich Would be nice if you could polish your PR. ( Rebasing, Squashing to 1 commit....)

cheidenreich:
Too much to rebase and change, i will start over

PR #1223, opened 2015-04-14
Physics cat 1158 remove unnecessary physics look init

Gorphi:


catroidTest/src/org/catrobat/catroid/test/physics/PhysicsSpriteCloneTest.java
```diff
@@ -188,7 +187,7 @@ public void testSpriteClonePhysicsLookAndPhysicsObject() {
 		sprite.addScript(startScript);
 
 		PhysicsWorld physicsWorld = project.getPhysicsWorld();
-		sprite.look = new PhysicsLook(sprite, physicsWorld);
+		sprite.look = new Look(sprite);
```
robertpainsi: A test and also the code for cloning a physics look is missing I guess? See https://github.com/Catrobat/Catroid/blob/feature/physics/catroid/src/org/catrobat/catroid/content/Look.java#L103-L113

If so, I'll write a jira ticket :wink:

Gorphi: Every Sprite has a Look-Object. A PhysicsLook is generated at the first "stage-entry". So after that first "state-entry" there is a PhysicsLook in the sprite. If we copy the sprite the PhysicsLook will convert back to a Look (with the function you mentioned). So it is back in the "initial state" and will be converted into a PhysicsLook again at the first "stage-entry".
So there is no real need for a copyPhysicsLookForSprite Methode ;-)

Pls correct me if I don't concern any important point. :-)

robertpainsi: If you clone a sprite, all it's data is cloned too, including the look data values.

So, if you create a sprite with a set transparency to 50% brick, start and leave the program, copy the sprite, go into that newly copied sprite and compute the transparency, it displays 50.

However, the method mentioned in my previous comment is missing some values to copy (like x, y, etc.), and tests. I'll write a jira ticket for that.

catroidTest/src/org/catrobat/catroid/test/physics/PhysicsLookTest.java
```diff
@@ -230,6 +231,118 @@ public void testSetLookDataWithNullPixmap() {
 			Log.e(TAG, "unexpected exception", exception);
 			fail("unexpected exception");
 		}
+	}
+
+	public void testDefaultValueEqualityOfPhysicsLookAndLook()
+	{
```
robertpainsi: Get in line you sneaky braces! :wink:

catroidTest/src/org/catrobat/catroid/test/physics/PhysicsSpriteCloneTest.java
Gorphi: I get your point. Thx 4 creating the tickets.

meiserm:
Reviewed

PR #1240, opened 2015-05-03
Lego NXT Active Light Sensor added

MightyMendor:
Added NXT sensor-type Active Light Sensor
Deactivate all sensors on stage-destroy and Mindstorm-disconnect
Fixed duration of PlayTone

### DIFF
--- a/catroid/src/org/catrobat/catroid/devices/mindstorms/nxt/sensors/NXTSensorService.java
+++ b/catroid/src/org/catrobat/catroid/devices/mindstorms/nxt/sensors/NXTSensorService.java
@@ -30,6 +30,9 @@ import android.util.SparseArray;
 
 import org.catrobat.catroid.common.CatroidService;
 import org.catrobat.catroid.devices.mindstorms.MindstormsConnection;
+import org.catrobat.catroid.devices.mindstorms.nxt.Command;
+import org.catrobat.catroid.devices.mindstorms.nxt.CommandByte;
+import org.catrobat.catroid.devices.mindstorms.nxt.CommandType;
 import org.catrobat.catroid.ui.SettingsActivity;
 import org.catrobat.catroid.utils.PausableScheduledThreadPoolExecutor;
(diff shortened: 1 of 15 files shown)


catroid/src/org/catrobat/catroid/devices/mindstorms/nxt/sensors/NXTSensorService.java
```diff
@@ -109,6 +112,18 @@ private NXTSensor createSensor(String sensorTypeName, int port) {
         return sensor;
 	}
 
+	public void deactivateAllSensors(MindstormsConnection connection) {
+
+		for (int port = 0; port < 4; port++) {
+
+			Command command = new Command(CommandType.DIRECT_COMMAND, CommandByte.SET_INPUT_MODE, true);
```
martinbu: if the third parameter is true, it is a reply command, so the nxt will send a reply message. But in line 124 you use send, so the reply is not handled. Please use a non reply command here. So the third parameter should be false.

catroidTest/src/org/catrobat/catroid/uitest/devices/mindstorms/nxt/LegoNXTImplTest.java
```diff
@@ -221,9 +221,9 @@ public void testNXTConnectionDialogGoBack() {
 		commands.add(new int[] { MOTOR_TURN, 2, 515 });
 
 		WaitBrick thirdWaitBrick = new WaitBrick(500);
-		LegoNxtPlayToneBrick legoPlayToneBrick = new LegoNxtPlayToneBrick(5000, 1000);
+		LegoNxtPlayToneBrick legoPlayToneBrick = new LegoNxtPlayToneBrick(5000, 1);
 		//Tone does not return a command
-		commands.add(new int[] { PLAY_TONE, 5000, 1000 });
+		commands.add(new int[] { PLAY_TONE, 5000, 1 });
```
martinbu: Maybe use 1000 here and remove the times 1000 in line 167.

PR #1242, opened 2015-05-04
Gsoc backpack looks

sjaindl:
Fixes backpack implementation for Sounds
- Save and load operations to backpack path corrected
- Remove deleted backpack items also from backpack Directory

Implementation for Looks
- Harmonize methods/functionality for Looks and sounds

Testrun:
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/1711
--> Same tests as in current develop branch are failing


obusher:
Your `git-commit-tree` looks weird. Most commits are already in `develop`. Would be nice if you could normalize the commit history.

sjaindl:
should be ok now

catroid/src/org/catrobat/catroid/common/SoundInfo.java
```diff
@@ -76,23 +78,11 @@ public SoundInfo copySoundInfoForSprite(Sprite sprite) {
...
-			return null;
-		}
-	}
-
-	public String getAbsolutePathBackPackSound() {
-		if (fileName != null) {
-			return Utils.buildPath(getPathToBackPackSoundDirectory(), fileName);
+			if (isBackpackSoundInfo) {
+				return Utils.buildPath(getPathToBackPackSoundDirectory(), fileName);
+			} else{
```
aried3r: Missing whitespace after `else`, I guess checkstyle should pick these up (right? ping @obusher), but please run the formatter on your code either way.

catroid/src/org/catrobat/catroid/ui/ScriptActivity.java
```diff
@@ -259,9 +259,15 @@ public boolean onOptionsItemSelected(MenuItem item) {
 				break;
 
 			case R.id.unpacking:
-				Intent intent = new Intent(currentFragment.getActivity(), BackPackActivity.class);
-				intent.putExtra(BackPackActivity.EXTRA_FRAGMENT_POSITION, FRAGMENT_SOUNDS);
-				startActivity(intent);
+				if (currentFragment == lookFragment) {
```
aried3r: Do you really want to use `==` here instead of `instanceof` or `.equals()` (depending on what you want, since these do different things)? Just checking if this was on purpose.

catroid/src/org/catrobat/catroid/common/SoundInfo.java
obusher: Looks like we do not have this rule in our ruleset, because `Catroid-Build-SourceTests-StaticChecks` was green. (@robertpainsi to the rescue!) :cactus:

robertpainsi: @obusher @aried3r 
There is a jira ticket (https://jira.catrob.at/browse/CAT-1321) waiting to introduce all the neat checkstyle rules for those kind of "errors". But first, the AS formatter rules have to be configured (at least the 'Keep Maximum Blank Lines' and the array style) and also a whole formatter run is required (imo).

catroid/src/org/catrobat/catroid/ui/ScriptActivity.java
sjaindl: I'm not sure what the Intention was because the code is > 1 year old.
But i think this was on purpose because it should really be the same object (fragment).

avihric:
reopen when rebased

PR #1247, opened 2015-05-13
CAT-1360 Added Color Bricks to Look Category

bernadettespieler:
Philipp Ferner a member of my XP Team from the lecture Mobile Application created this branch.


bernadettespieler:
there are mistakes in the implementation - color  range should be between -100 to 100 like in Scratch (not 255)!

PR #1251, opened 2015-05-18
CAT-1347 - fix BroadcastActionTest

cheidenreich:
BroadcastActionTest.testWhenScriptRestartingItself made no sense and could not succeed ever in the former implementation.

I added a RepeatBrick so the restart of the WhenScript is properly tested

JIRA:
https://jira.catrob.at/browse/CAT-1347

Jenkins run:
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/1760/


obusher:
It makes sense for me. In my view this test checks if recursion works properly with `BroadCastBricks` and `BroadCastScript` combinations.
The `loopCounter < 20` condition in the while loop aborts this infinite recursion. 
The assert checks if the `BroadCastScript` got executed more then once.
(Most times it should be executed ~20 times) :mahjong:

cheidenreich:
@obusher Then the assert should be AssertFalse, shouldn't it?

cheidenreich:
created PR to wrong branch -> Closed

obusher:
In this testcase the `ChangeXByNBrick` acts only once, so `x` which is initially `0`, gets changed to `1` but it should be bigger than `1` (~20).

cheidenreich:
@obusher thanks, you're right, I totally misinterpreted this test. 

After some struggle understanding the problem with this test, the solution is quite a simple one:
It turned out that on some devices, the allActionsOfAllSpritesAreFinished() gets executed after the broadcast message is sent but before the when script started, so it returned true and the loop stopped.
I just added a wait brick (only 5ms) right after the send broadcast brick. This closes the gap mentioned above and essentially does not change the test itself.

I opened a new PR (this was opened to the wrong target branch): #1258

robertpainsi:
@cheidenreich: So, there is a problem in the broadcast handling and not in the test itself?

cheidenreich:
@robertpainsi I would not say that, but a cannot prove the opposite 100%.
From what I have seen, Look.getAllActionsAreFinished() to stop script execution is only used in tests. When I manually recreate the script used in this test in PocketCode works as expected with infinite recursion even on the devices where the test fails.

robertpainsi:
@cheidenreich: Ah, thanks for the info and testing! :smile_cat:

obusher:
Yeah! :octocat:

PR #1252, opened 2015-05-18
CAT-1316 Replaced Rot-Vect Incl. computation with plane ACC data Incl…

amix73:
…. computation
A version using this fix can be downloaded and tried out: 
You will not notice any changes, and that's good. On phones where it did not work like the Moto G 1st gen or the Indian Micromax A63 it will work.
http://www.ist.tugraz.at/catrobat/pub/Main/ChristianSchindler/PocketCode_InclSensorFix.apk


amix73:
Should be in the release - therefor forget it! Sorry!
;-)

obusher:
I know the format of this class is not in best shape and it is compelling to use auto-format.
But i want to see a clean `github-diff`, so please do not use this magic. :rabbit:

PR #1253, opened 2015-05-18
CAT-1316 Replaced computation of INCL. with ROT-VECT with plane ACC. …

amix73:
…data computation

Which means that it now works on phones without GYRO like the MotoG 1st gen. and the Indian Phone Micromax A63.

A working PC version with this fix can be obtained here: http://www.ist.tugraz.at/catrobat/pub/Main/ChristianSchindler/PocketCode_InclSensorFix.apk


obusher:
I know the format of this class is not in best shape and it is compelling to use auto-format.
But i want to see a clean `github-diff`, so please do not use this magic. :rabbit:

amix73:
Ok, I will fix this tomorrow and create another request.

obusher:
You do not need to open a new PR. The common way is only to update the branch. 
(f.e. with `git add .`, `git commit --amend` and `git push -f`) :vhs:

PR #1254, opened 2015-05-19
CAT-1316: Replaced RotVectIncl computation with plain AccelrometerInc…

amix73:
…l computation.
Works now on Phones without GYRO like the Micromax A63 or the 1st gen MOTOG.

### DIFF
--- a/catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
+++ b/catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
@@ -34,16 +34,18 @@ public final class SensorHandler implements SensorEventListener, SensorCustomEve
 	private static final String TAG = SensorHandler.class.getSimpleName();
 	private static SensorHandler instance = null;
 	private SensorManagerInterface sensorManager = null;
-	private Sensor accelerometerSensor = null;
+	private Sensor linearAccelerationSensor = null;
 	private Sensor rotationVectorSensor = null;
 	private float[] rotationMatrix = new float[16];
 	private float[] rotationVector = new float[3];
 	public static final float RADIAN_TO_DEGREE_CONST = 180f / (float) Math.PI;
 
-	private float linearAcceleartionX = 0f;
-	private float linearAcceleartionY = 0f;
-	private float linearAcceleartionZ = 0f;
-
+	private float linearAccelerationX = 0f;
+	private float linearAccelerationY = 0f;
+	private float linearAccelerationZ = 0f;
+	private Sensor accelerometerSensor = null;
+	private float[] accelerationXYZ = new float [3];
+	private float signAccelerationZ = 0f;
 	private float loudness = 0f;
 	private float faceDetected = 0f;
 	private float faceSize = 0f;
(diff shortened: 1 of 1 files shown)


catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -177,10 +197,20 @@ public void onAccuracyChanged(Sensor arg0, int arg1) {
 	@Override
 	public void onSensorChanged(SensorEvent event) {
 		switch (event.sensor.getType()) {
+			case Sensor.TYPE_ACCELEROMETER:
+				accelerationXYZ = event.values.clone();
+				double norm_Of_g = Math.sqrt(accelerationXYZ[0] * accelerationXYZ[0]
```
aried3r: We use `camelCase` instead of `snake_case`. Also, is this your own code or did you get it from SO? If so, please leave a reference for the source.

aried3r:
Does this not use the gyro at all anymore, even on devices that have one instead of just falling back? (I don't even see `TYPE_GYROSCOPE` being used), but I'm not familiar with the sensor code. @obusher maybe?

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
amix73: @camelCase, of course.
Shall I close the pr, change it to camelCase, squash the commits and open another pr?
Or is there a better wf?

@code: how shall I reference a partial code snippet from stackoverflow or developer.android.com?

Thanks for info.

amix73:
`TYPE_GYROSCOPE` wasn't used before either but the `TYPE_ROTATION_VECTOR` (which implicitly uses the GYRO). So there is no obvious advantage here to use the rotation vector when it can fail and the simpler method will work since *"all"* (i.e. many more) phones have an accelerometer but gyro is much less likely (even today). Furthermore in practice there are no notable differences between both methods, so there's no reason to use the rotation vector (gyro) reading at all.

amix73:
I will expand the Jira issue CAT-1316 to cover the other sensor issues as well.
Q: is there a way to put the PR temporary on hold? Closing and opening another one is IMO not needed since only the scope of the ticket will change.

robertpainsi:
> Closing and opening another one is IMO not needed since only the scope of the ticket will change.

As long as you don't force push to your branch, a pull request can be reopened :wink:

wintersw:
Hi,
looks great, but you should try to work on the start screen based on 
this mockup:
/_Prototyping/websiteIntegration/index.html

What we want to achive is a style that looks the same on all devices and 
browsers- new and built in controls like the <progress />, <slider />, 
.. do all have there individual styles (behavior) on different platforms 
even if they are ment to be fully customizable using css.

As you can e.g. copy layout and styles within the blue border and add a 
progress bar using existing styles.


the progress bar may look like out startup progress bars:
https://web-test.catrob.at/html5/Client/PocketCodePlayer/index.html
(the blue one as we want to show explicit download progress)

You can find the page and source code (as js and css) in our github:
\Client\PocketCodePlayer

I guess this sounds quite easy because all of the required styles are 
already there.. but the layout should not have any errors when resizing, 
which can be a tricky part.
Anyway.. I already did some tests on that concern so hopefully it's not 
a problem.

So.. whats the best way to get the html/css if there is working example 
based on javascript.
First of all,
- have a look at the generated source of the page using the browsers 
developer tools.
- include the same css files so all current styles are also available in 
your mockup
- copy paste related html tags out of the browsers source and add them 
to your mockup screen
- customize it

If you have time during the weekend, we start together using skype and a 
shared screen.. that may help you getting some useful input how to make 
your life coding html/css much easier.

thanks,
Wolfgang



Am 10.06.2015 um 13:25 schrieb Robert Painsi:

obusher:
@amix73 Did you expand your Jirra issue already ?

@wintersw I think you posted something accidentally in this PR. :package:

avihric:
reopen when finished

PR #1257, opened 2015-05-27
[CAT-1359] Added new Start program.

bernadettespieler:
Change of the Start program - Instead of mole (which is a little bit too difficult for children as a start program) we added the program bird. Michael Fischl a member of my XP Team from the lecture mobile application did it. 

@thmq I will forward you the strings for the new program


bernadettespieler:
problems occured

PR #1241, opened 2015-05-04
CAT-1353 fix UserBrickScriptActivityTest

cheidenreich:
setHasOptionsMenu() triggers a callback (onPrepareOptionsMenu()) which uses the variable formulaEditorEditText. This variable was unassigned when calling setHasOptionsMenu() at the beginning of onCreate().

Now the Test passes:
https://jenkins.catrob.at/job/Catroid-Multi-Job-Custom-Branch-RELOADED/1688/#showFailuresLink

JIRA issue:
https://jira.catrob.at/browse/CAT-1353

PR #1259, opened 2015-06-01
Corrected modification of PR 1241 in onCreate method

avihric:
`setHasOptionsMenu` was inserted at wrong place in  #1241


avihric:
Revert was done in #1261

PR #1268, opened 2015-06-10
CAT-1284 changed gitignore and file tracking

jonny911:
CAT-1284 changed gitignore and file tracking


aried3r:
Think this can go now, too https://github.com/Catrobat/Catroid/blob/develop/gradle/intellij_config_tasks.gradle

SimonStefan:
Have you really tried to clone and import a new project without these files? The following things should work:

1. make fresh clone (or git reset --hard && git clean -xdf)
2. import in Android Studio
3. create a new file (there should be a copyright message)
4. there should be the same code style settings as currently (unfortunately the files in the repo are not formatted according to these styles)
5. run a test method (it should not give you the error "Empty test suite". This happens when the test runner of the run configuration is empty and not the polidea test runner)

@aried3r is right. Without the files in the repo, we don't need the Gradle tasks to hide them for git and you can remove the entire file.

obusher:
#1282

PR #1275, opened 2015-06-12
fixed error in notification statusbar

bernadettespieler:
"fixed" error: everytime when the user downloads a project all previous projects are also loading again in the notification status bar (only loading/not downloading). These lines fix only the display of the additional loading bars, but it needs further debugging why the method onReceiveResult() is called nevertheless for every previous project


bernadettespieler:
"fixed" error: everytime when the user downloads a project all previous projects are also loading again in the notification status bar (only loading/not downloading). This lines fixes only the display of the additional loading bars, but it needs further debugging why the method onReceiveResult() is called nevertheless for every previous project

avihric:
ok to test

PR #1283, opened 2015-06-22
fixed missing application extension @ download standalone

X4mp:
fixes the missing file extension for standalone downloadable apk files from share.catrob.at via the PC inner WebViewActivity


X4mp:
please test this

PR #1284, opened 2015-06-23
CAT-1436 fix the bug in the method onReceiveResult

Manu91:
The method onReceiveResult in DownloadUtil.java will called repeatedly with old values, so i add a if-condition in the method downloadProject in ServerCalls.java and now the method onReceiveResult will called only with the newest value.


avihric:
Please reopen onto `master`

avihric:
ping @Manu91

PR #1287, opened 2015-06-29
CAT-1359 Added new Start program.

bernadettespieler:
Michael Fischl was a member in my XP team (SW 15) and he implemented this user story:
Instead of the default program "mole whacked" (what seems to difficult for young users) we got from the UX team the new default program "bird wing".

As long as there are old strings in all language string.xml - jenkins tests will fail!


thmq:
the problem here is that Jenkins does not in fact run any tests on the branch because it can't get passt the lint warnings. i am aware that removing the excess strings on your branch might be a little inefficient and tedious but there seems to be no other option for the moment.

for the future i am looking into finding a solution but for now you'll have to remove them manually in order to obtain a result from Jenkins.

avihric:
reopen when conflicts are resolved

PR #1289, opened 2015-07-02
CAT-1443 Set/Cange variable bugfix

robertriedl:
Fixed the bug in the clone-method, which caused the wrong brick view in the formula-editor.


robertpainsi:
Consider to write (or adapt) tests for the bugfix.

avihric:
A hotfix should track `master` instead of `develop`

PR #1291, opened 2015-07-07
CAT-1442 hotfix for spinner appearance

jonny911:
disabled spinner. 
Visual change: Spinner content is now grey, but should not affect usability.


avihric:
I would say this is not a hotfix :P

PR #1297, opened 2015-07-14
CAT-1440 Cancelling creating a new variable

StefanBuerscher:
will no longer set the variable to New…

### DIFF
--- a/catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
+++ b/catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
@@ -65,7 +65,7 @@ public class VariableBricksTest extends BaseActivityInstrumentationTestCase<Main
 		dataContainer.deleteUserVariableByName("sprite_var2");
 		super.tearDown();
 	}
-
+	
 	public void testVariableBricks() {
 		Spinner setVariableSpinner = solo.getCurrentViews(Spinner.class).get(0);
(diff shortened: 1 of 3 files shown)


catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
```diff
@@ -93,6 +93,14 @@ public void testVariableBricks() {
...
 
+		solo.goBack();
+		solo.goBack();
+		solo.clickOnText("p2");
+		solo.clickOnText("p1");
+		solo.clickOnText("p1");
+		solo.clickOnText(solo.getString(R.string.brick_variable_spinner_create_new_variable));
+		solo.clickOnText(solo.getString(R.string.cancel_button));
+		solo.clickOnText("p1");
 	}
```
obusher: Would be nice if you could do that in a own method, f.e called `testCreatingNewVariableCancelling()`.
In this way your sourcecode gets commented with the method-name and everyone knows what this code does.
Furthermore use some `assert*` to check the correct behavior of the app.

avihric:
reopen when conflicts are resolved

catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
StefanBuerscher: I think in this case asserts are not needed because when the clickOnText-Methods are succesfull the app's behavior is correct.

StefanBuerscher:
#1328

PR #1303, opened 2015-07-17
CAT-1168 Uncommented Test

loibi93:
Removed Commenting from UserConceptTest class.

### DIFF
--- a/catroidTest/src/org/catrobat/catroid/uitest/web/UserConceptTest.java
+++ b/catroidTest/src/org/catrobat/catroid/uitest/web/UserConceptTest.java
@@ -22,225 +22,226 @@
  */
 package org.catrobat.catroid.uitest.web;
 
-//import android.content.SharedPreferences;
-//import android.content.SharedPreferences.Editor;
-//import android.preference.PreferenceManager;
-//import android.test.UiThreadTest;
-//import android.view.View;
-//import android.widget.EditText;
-//import android.widget.TextView;
-//
-//import com.robotium.solo.Solo;
-//
-//import org.catrobat.catroid.R;
-//import org.catrobat.catroid.common.Constants;
+import android.content.SharedPreferences;
+import android.content.SharedPreferences.Editor;
+import android.preference.PreferenceManager;
+import android.test.UiThreadTest;
+import android.view.View;
+import android.widget.EditText;
+import android.widget.TextView;
 
+import com.robotium.solo.Solo;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.common.Constants;
+import org.catrobat.catroid.test.utils.Reflection;
 import org.catrobat.catroid.ui.MainMenuActivity;
+import org.catrobat.catroid.uitest.annotation.Device;
 import org.catrobat.catroid.uitest.util.BaseActivityInstrumentationTestCase;
 
-//import org.catrobat.catroid.uitest.annotation.Device;
-//import org.catrobat.catroid.uitest.util.Reflection;
-//import org.catrobat.catroid.uitest.util.UiTestUtils;
-//import org.catrobat.catroid.web.ServerCalls;
-//
-//import java.util.ArrayList;
-//import java.util.Locale;
+import org.catrobat.catroid.uitest.util.UiTestUtils;
+import org.catrobat.catroid.web.ServerCalls;
 
+import java.util.ArrayList;
+import java.util.Locale;
+
+//Aborts on emulator
 public class UserConceptTest extends BaseActivityInstrumentationTestCase<MainMenuActivity> {
 
-//	private String saveToken;
-//	private String loginDialogTitle;
-//	private String uploadDialogTitle;
+	private String saveToken;
+	private String loginDialogTitle;
+	private String uploadDialogTitle;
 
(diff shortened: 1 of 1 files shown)


loibi93:
Seems that Jenkins did a restart during the test !? But i think it is sufficient if i just do a singleclass test with the UserConceptTest class since nothing but that class has changed !

obusher:
Console:
`12:01:54 org.catrobat.catroid.uitest.web.UserConceptTest:....INSTRUMENTATION_RESULT: shortMsg=java.lang.NullPointerException
12:01:54 INSTRUMENTATION_RESULT: longMsg=java.lang.NullPointerException
12:01:54 INSTRUMENTATION_CODE: 0`
Logcat:
`java.lang.NullPointerException at android.support.v4.app.DialogFragment.show(DialogFragment.java:125) at org.catrobat.catroid.ui.dialogs.LoginRegisterDialog.onRegistrationComplete(LoginRegisterDialog.java:113)`

loibi93:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/322/

obusher:
The whole testrun gets aborted, so it does not matter much if the single-run is green.

obusher:
ok to test

loibi93:
Sorry, i thought that this log means only one test failed ;)

loibi93:
Failed again for the same reason, but as i said, a Single Class test is sufficient here, because the only thing that changed was the test class, and not actual App Code.

robertpainsi:
> Failed again for the same reason, but as i said, a Single Class test is sufficient here, because the only thing that changed was the test class, and not actual App Code.

Unfortunately not. Jenkins has to be kept stable but at least runnable (especially the PR run) otherwise it can't be used anymore :scream_cat: I'll give the PR a quick rerun but if it fails again, it's not mergeable and has to be fixed to run a full testrun as well.

Since the exception is thrown by an AsyncTask, the root of the problem can be really hard to find. And if you are unlucky, the problem occurs because another test doesn't clean up the resources correctly.

robertpainsi:
ok to test

loibi93:
Ok, thats a good reason ;D
I didn't have a look what exactly went wrong, but since it ran fine on the device, it is possible that the test breaks on emulator ! Is it possible to force jenkins using the device instead of emulator ? I would suggest that i test that out before doing the dig ;D

robertpainsi:
> Is it possible to force jenkins using the device instead of emulator

You can use the ```@Device``` annotation
https://github.com/Catrobat/Catroid/blob/develop/catroidTest/src/org/catrobat/catroid/uitest/content/interaction/BrickEditFormulaTest.java#L51

However, I don't know the device annotation policy. Meaning, I don't know when it's acceptable to use it. We once tried to move as many tests as possible to the emulators because they run so much faster.
ping @avihric @obusher

loibi93:
At least the tests won't crash anymore ! Not sure why the three tests failed.

avihric:
@robertpainsi There is no real policy, so if you argue that this test is only runable on Device otherwise it won*t succeed, thats fine.

@loibi93 you have 2 checkstyle warnings, pls fix them.
**Tip**: Use gradle task checkstyle before commiting, it performs the same checks as on Jenkins

obusher:
Maybe we should comment those tests, which only are runable on Device but technically should be run on emulator, with something fancy like `//Aborts on emulator` .

loibi93:
@avihric I am not sure how to fix these warnings, where do i have to put the imports ?

robertpainsi:
@obusher: :+1: 
@loibi93: Ctrl+Alt+O also known as Optimize imports :wink:

loibi93:
Thanks very much :D

loibi93:
ok to test

loibi93:
ok to test

loibi93:
ok to test

loibi93:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/327/

https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/326/

loibi93:
ok to test

loibi93:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/328/

Ok to merge.

thmq:
fix the commit message according to our policy, furthermore it is not just a few failing test cases, the emulator run crashes entirely and the device run is not even started. as @robertpainsi said: the PR has to be runnable on Jenkins at least.

loibi93:
Commit message has already been fixed ;)

The recent run was because of the new commit message, here is the stale run:
https://jenkins.catrob.at/job/Catroid-PullRequest-TestJob/583/

and the failed test in a single run:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/328/

avihric:
Could you rewrite your commit msg with: 

```
CAT-1168 Uncommented Test

Reintroduced UserConceptTest class
```

loibi93:
Rebased and changed Commit message !

PR #1304, opened 2015-07-21
CAT-1305 Replaced ActionBarSherlock

thmq:
refactored ActionBarSherlock, but there are some issues [1]. the failing test cases are not the main concern but the obscure exception, as described in [2]. I'm not entirely sure this is relevant, therefore I'm hoping on some input here (which is why I opened the PR in the first place). the exception can be found here [3], at the bottom.

[1] full run: https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2087/
[2] http://databasefaq.com/index.php/answer/60793/android-android-layout-android-edittext-android-event-android-edittext-onclick-listener-defined-in-layout-fails-with-obscure-exception
[3] https://jenkins.catrob.at/job/Catroid-PartialEmulatorUITest/14050/artifact/logcat.txt


thmq:
@robertpainsi: could I have added the under discussion label myself?

robertpainsi:
> could I have added the under discussion label myself?

If you have write access (e.g. you can merge and close pull requests), yes.
https://help.github.com/articles/permission-levels-for-an-organization-repository/

obusher:
Does this exception get thrown with a lower version (`'com.android.support:appcompat-v7:22.0.0'`) ?

thmq:
@obusher doesn't work with 'com.android.support:appcompat-v7:22.0.0'.

https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2096/

thmq:
As discussed today with @avihric I have tried to remove the support libraries altogether because we are no longer providing support for any api-levels below 16. There still has to be some work todo (e.g. correct import of the the themes, which are currently still from the support library). I have added a second commit on purpose so that one can still see the changes performed when I still tried to use the support libraries.

If there are any issues with this approach please feel free to tell me.

robertpainsi:
ok to test

aried3r:
> I have tried to remove the support libraries altogether because we are no longer providing support for any api-levels below 16.

I do not recommend doing that. Using the support libraries has advantages, like _one_ set of problems instead of many. The Fragment implementation of the support library is [a standalone one](https://github.com/android/platform_frameworks_support/blob/75573e79af58295cfb3f0ceccb163aa37f88c747/v4/java/android/support/v4/app/Fragment.java#L153-L157), meaning when you update the support library, you get the newest versions with all new features and bug fixes. Or new bugs, but the same ones across all platforms.

And in the future (yesterday, rather), this project will probably switch to using the [Toolbar](https://developer.android.com/reference/android/support/v7/widget/Toolbar.html) from the [appcompat library](https://developer.android.com/tools/support-library/features.html#v7-appcompat) instead, which needs a [AppCompatActivity](https://developer.android.com/reference/android/support/v7/app/AppCompatActivity.html), which extends from [FragmentActivity](https://developer.android.com/reference/android/support/v4/app/FragmentActivity.html) which is in the v4 support library.

build.gradle
```diff
@@ -63,6 +61,8 @@ dependencies {
     natives 'com.badlogicgames.gdx:gdx-platform:' + gdxVersion + ':natives-x86'
     natives 'com.badlogicgames.gdx:gdx-platform:' + gdxVersion + ':natives-armeabi'
     natives 'com.badlogicgames.gdx:gdx-platform:' + gdxVersion + ':natives-armeabi-v7a'
+    compile 'com.android.support:support-v4:19.1.0'
+    compile 'com.google.guava:guava:17.0'
```
obusher: Remove `line 65` because of `line 53`.

PR #1309, opened 2015-07-27
Gsoc backpack looks

sjaindl:
Reopened PR #1242 (rebased)

Jenkins testrun:
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2164/
Seems like there are 2 random fails.


thmq:
testUserBrickVariableScope seems to fail all the time, so there should be no problem there. it should probably be refactored at some point. @loibi93 (i don't seem to be able to link him here) might already be working on that problem.

loibi93:
I just fixed SetLookBrickTest, I didn't know about that test !

thmq:
you should check with the comments in Jira, I have mentioned the test there [1]. I thought you'd include it in #1310 

[1] https://jira.catrob.at/browse/CAT-1438

loibi93:
Sorry, I overlooked that, i will see what i can do !

thmq:
no problem, thank you for looking into it.

catroid/res/values-de/strings.xml
```diff
@@ -36,6 +35,7 @@
   <string name="unpacking">Auspacken</string>
   <string name="unpack">Pack es aus</string>
   <string name="copy_addition">Kopie</string>
+  <!--<string name="unpacking_addition">entpackt</string>-->
```
thmq: please remove strings that are no longer needed instead of commenting. if you remove anything from the values/strings.xml you'll have to remove it within all other language files too (Android Studio provides the functionality). When you add strings you don't need to change anything. When you change strings (their identifiers not their content), be sure to remove them first and add the new string only in values/strings.xml

avihric:
Reopen when conflicts are resolved please

PR #1310, opened 2015-07-28
CAT-1438 Fixed testAddNewLook in SetLookBrickTest:

loibi93:
This test failed a lot, because we were not able to do a real click
on the Paintroid Option. We have to use the MockActivity and initiate the
click with startActivityForResult, which is not the same as clicking on
the Option. Due to this false click, some Code was not triggered, and therefore
the App didn't send the Broadcast for a new LookFile. I fixed this
issue by clicking on the Spinner, where we then can see the new File.


loibi93:
Perfect Testrun ;)
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2181/

loibi93:
It was suggested to fix testUserBrickVariableScope too, so I will have to work on that...

PR #1319, opened 2015-08-03
CAT-910 Screnshot in stage is not a screenshot

LauraBebek:
https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Build-SourceTests-StaticChecks/2926/


bernadettespieler:
ok to test

PR #1322, opened 2015-08-03
CAT-1290 Name alerts fixed

robertriedl:
Linked with CAT-1291, CAT-1292, CAT-1293 - merged to 1 pull-request because these tickets relate to the same issue.

@thmq : please update in crowdin

### DIFF
--- a/catroid/res/values/strings.xml
+++ b/catroid/res/values/strings.xml
@@ -51,6 +51,8 @@
     <string name="settings">Settings</string>
     <string name="redo">Redo</string>
     <string name="undo">Undo</string>
+    <string name="name_exists">Name already exists</string>
+    <string name="no_name">No name</string>
     <!--  -->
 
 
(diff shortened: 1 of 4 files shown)


thmq:
this will be uploaded automatically as soon as the pr is merged, then I'll take care of any further steps.

catroid/res/values/strings.xml
```diff
@@ -285,9 +287,10 @@
     <!-- New sprite dialog -->
     <string name="new_sprite_dialog_default_sprite_name">My Object</string>
     <string name="new_sprite_dialog_title">New Object</string>
-    <string name="spritename_already_exists">This object name already exists</string>
+    <string name="spritename_already_exists">Please choose another object name.</string>
     <string name="rename_sprite_dialog">Rename Object</string>
     <string name="spritename_invalid">This object name is invalid</string>
+    <string name="no_spritename_entered">Please enter a object name.</string>
```
aried3r: This should be "an object". Also, it seems that "object" is used inconsistently. Sometimes it's written capitalized, sometimes not. Is it used as a name or a common noun?

robertpainsi: This applies to other nouns like look, sound, image, brick, script, backpack and even more.
I'll write an UX ticket for that tomorrow.

PR #1323, opened 2015-08-03
CAT-0898 Consistency: Wait 1 second brick

LauraBebek:


catroid/src/org/catrobat/catroid/content/bricks/WaitBrick.java
```diff
@@ -55,7 +55,7 @@ public WaitBrick() {
 	}
 
 	public WaitBrick(int timeToWaitInMillisecondsValue) {
-		initializeBrickFields(new Formula(timeToWaitInMillisecondsValue / 1000.0));
+		initializeBrickFields(new Formula(timeToWaitInMillisecondsValue / 1000));
```
robertpainsi: Not a good idea because of:
https://github.com/Catrobat/Catroid/blob/483fac8c850bb7722b5b6680d48fa4c0054e65ff/catroid/src/org/catrobat/catroid/common/StandardProjectHandler.java#L477

PR #1325, opened 2015-08-03
CAT-1316 Sensor-type to GAMING; LIN.ACC.- and ROT.VECT. (inclin.,…

amix73:
…compass-dir) sensor fallback

### DIFF
--- a/catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
+++ b/catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
@@ -37,32 +37,55 @@ import org.catrobat.catroid.devices.mindstorms.nxt.LegoNXT;
 import org.catrobat.catroid.facedetection.FaceDetectionHandler;
 
 public final class SensorHandler implements SensorEventListener, SensorCustomEventListener {
+	public static final float RADIAN_TO_DEGREE_CONST = 180f / (float) Math.PI;
 	private static final String TAG = SensorHandler.class.getSimpleName();
 	private static SensorHandler instance = null;
+	private static BluetoothDeviceService btService = ServiceProvider.getService(CatroidService.BLUETOOTH_DEVICE_SERVICE);
 	private SensorManagerInterface sensorManager = null;
+	private Sensor linearAccelerationSensor = null;
 	private Sensor accelerometerSensor = null;
+	private Sensor magneticFieldSensor = null;
 	private Sensor rotationVectorSensor = null;
 	private float[] rotationMatrix = new float[16];
 	private float[] rotationVector = new float[3];
-	public static final float RADIAN_TO_DEGREE_CONST = 180f / (float) Math.PI;
-
-	private float linearAcceleartionX = 0f;
-	private float linearAcceleartionY = 0f;
-	private float linearAcceleartionZ = 0f;
-
+	private float[] accelerationXYZ = new float[3];
+	private float signAccelerationZ = 0f;
+	private float[] gravity = new float[] { 0f, 0f, 0f };
+	private boolean useLinearAccelerationFallback = false;
+	private boolean useRotationVectorFallback = false;
+	private float linearAccelerationX = 0f;
+	private float linearAccelerationY = 0f;
+	private float linearAccelerationZ = 0f;
 	private float loudness = 0f;
 	private float faceDetected = 0f;
 	private float faceSize = 0f;
 	private float facePositionX = 0f;
 	private float facePositionY = 0f;
 
-	private static BluetoothDeviceService btService = ServiceProvider.getService(CatroidService.BLUETOOTH_DEVICE_SERVICE);
-
 	private SensorHandler(Context context) {
 		sensorManager = new SensorManager(
 				(android.hardware.SensorManager) context.getSystemService(Context.SENSOR_SERVICE));
-		accelerometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_LINEAR_ACCELERATION);
+		linearAccelerationSensor = sensorManager.getDefaultSensor(Sensor.TYPE_LINEAR_ACCELERATION);
+		if (linearAccelerationSensor == null) {
+			useLinearAccelerationFallback = true;
+		}
+
 		rotationVectorSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ROTATION_VECTOR);
+		if (rotationVectorSensor == null) {
+			useRotationVectorFallback = true;
+		}
+
+		if (useRotationVectorFallback) {
+			accelerometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
+			magneticFieldSensor = sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD);
+		} else if (useLinearAccelerationFallback) {
+			accelerometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
+		}
+
+		Log.d(TAG, "*** LINEAR_ACCELERATION SENSOR: " + linearAccelerationSensor);
+		Log.d(TAG, "*** ACCELEROMETER SENSOR: " + accelerometerSensor);
+		Log.d(TAG, "*** ROTATION_VECTOR SENSOR: " + rotationVectorSensor);
+		Log.d(TAG, "*** MAGNETIC_FIELD SENSOR: " + magneticFieldSensor);
 	}
 
 	public static void startSensorListener(Context context) {
@@ -72,10 +95,9 @@ public final class SensorHandler implements SensorEventListener, SensorCustomEve
 		}
 		instance.sensorManager.unregisterListener((SensorEventListener) instance);
 		instance.sensorManager.unregisterListener((SensorCustomEventListener) instance);
-		instance.sensorManager.registerListener(instance, instance.accelerometerSensor,
-				android.hardware.SensorManager.SENSOR_DELAY_NORMAL);
-		instance.sensorManager.registerListener(instance, instance.rotationVectorSensor,
-				android.hardware.SensorManager.SENSOR_DELAY_NORMAL);
+
+		instance.registerListener(instance);
+
 		instance.sensorManager.registerListener(instance, Sensors.LOUDNESS);
 		FaceDetectionHandler.registerOnFaceDetectedListener(instance);
 		FaceDetectionHandler.registerOnFaceDetectionStatusListener(instance);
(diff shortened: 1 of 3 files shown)


catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -40,15 +40,23 @@
...
+	private Sensor magneticFieldSensor = null;
 	private Sensor rotationVectorSensor = null;
 	private float[] rotationMatrix = new float[16];
 	private float[] rotationVector = new float[3];
 	public static final float RADIAN_TO_DEGREE_CONST = 180f / (float) Math.PI;
 
-	private float linearAcceleartionX = 0f;
-	private float linearAcceleartionY = 0f;
-	private float linearAcceleartionZ = 0f;
+	private float[] accelerationXYZ = new float [3];
```
aried3r: Extraneous space between `float` and `[3]`.

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -40,15 +40,23 @@
...
+	private float signAccelerationZ = 0f;
+	private float[] gravity = new float []{0f,0f,0f};
```
aried3r: `new float[] {0f, 0f, 0f};` is I think the correct formatting. Either way, run the formatter.

aried3r:
Please run the formatter on your files, there's quite a few formatting errors.

martinbu:
Everything works fine now on Moto G1.

robertpainsi:
Please take a look at the [checkstyle warnings](https://jenkins.catrob.at/job/Catroid-Build-SourceTests-StaticChecks/2909/checkstyleResult)

amix73:
@aried3r: formatter run. Didn't think of it - but last time I ran the formatter @obusher  had something against it:
> obusher commented on May 18
> I know the format of this class is not in best shape and it is compelling to use auto-format.
> But i want to see a clean github-diff, so please do not use this magic. :rabbit:

I generally understand that - but it wouldn't be a problem when the formatter would be run automagically in every Android Studio on save - so the format wouldn't deteriorate in the first place. 

@martinbu: Thanks for your help finding my copy/paste error - now the values in the Compute-Dialog in the formula-editor are correctly updated :thumbsup:

aried3r:
While I agree that there should be an option to format it automatically, there _is_ an option to only format the lines you have changed.

reformat

And while I don't personally commit from within Android Studio, there is an option to format before committing.

screen shot 2015-08-04 at 22 58 28

robertpainsi:
edit: Dangle-it, @aried3r was faster! :pensive:

JFI: The comment you quoted is outdated :bowtie: and all java classes  (except two lines [[1]](https://github.com/Catrobat/Catroid/blob/develop/catroid/src/org/catrobat/catroid/ui/fragment/SoundFragment.java#L526) and [[2]](https://github.com/Catrobat/Catroid/blob/develop/catroidTest/src/org/catrobat/catroid/uitest/web/UserConceptTest.java#L41)) are formatted according to the Android Studio formatter.
All new checkstyle rules are doing a great job, however, the approach mentioned by @aried3r  should be used/preferred.

amix73:
@aried3r : that's cool thanks -  haven't tried committing from Android Studio yet.

aried3r:
> All new checkstyle rules are doing a great job

Is there a way to have Jenkins run this and report back within a few minutes? If this isn't already the case. I don't want to comment every PR because of these things.

It seems GitHub can list multiple test runs on PRs. See [here](https://github.com/JakeWharton/ProcessPhoenix/pull/7) for an example.

robertpainsi:
> Is there a way to have Jenkins run this and report back within a few minutes? If this isn't already the case. I don't want to comment every PR because of these things.

There is already an experimental jenkins job waiting to replace the current Catroid-Build-SourceTests-StaticChecks job. The new one will abort the complete PR testrun if there are any pmd, checkstyle or lint warnings. So, if a jenkins testrun takes only a few minutes, which you can check even without a jenkins account, it highly indicates new static code analyzing warnings. I'll ping you if the new job has been activated.

> It seems GitHub can list multiple test runs on PRs. See here for an example.

Thanks for the tip! I like it and I'll write a (low priority) ticket for it. :smile_cat:

robertpainsi:
ping @aried3r
The job has been replace, Sire!
Now you can check the jenkins runtime first before commenting on syntax errors.

If you find something that checkstyle misses, let me know guys!

(this message will now self destruct :wink:)

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -117,50 +155,92 @@ public static Double getSensorValue(Sensors sensor) {
...
 
 			case Z_ACCELERATION:
-				return (double) instance.linearAcceleartionZ;
+				return (double) instance.linearAccelerationZ;
 
 			case COMPASS_DIRECTION:
 				float[] orientations = new float[3];
-				android.hardware.SensorManager.getRotationMatrixFromVector(instance.rotationMatrix,
-						instance.rotationVector);
+				if (!instance.useRotationVectorFallback) {
```
thmq: what happens here if useRotationVectorFallback is true? correct me if I'm wrong, but if there is no getRotationMatrixFromVector then the rotationMatrix is not initialised and hence you won't get any reasonable results anyway.

thmq:
apart from the question regarding the compass direction the PR looks fine. any objections against merging it as soon as this has been cleared up? @robertpainsi @martinbu ?

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -117,50 +155,92 @@ public static Double getSensorValue(Sensors sensor) {
...
-						instance.rotationVector);
-				android.hardware.SensorManager.getOrientation(instance.rotationMatrix, orientations);
-				sensorValue = (double) orientations[2];
-				return sensorValue * RADIAN_TO_DEGREE_CONST * -1f;
-
+				if (instance.useRotationVectorFallback) {
+					float rawInclinationX = RADIAN_TO_DEGREE_CONST * (float) (Math.acos(instance.accelerationXYZ[0]));
+					float correctedInclinationX = 0;
+
+					// X-inclination: turn right direction
```
thmq: I don't think the comments are necessary.

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -117,50 +155,92 @@ public static Double getSensorValue(Sensors sensor) {
...
+					if (rawInclinationX >= 90 && rawInclinationX <= 180) {
+						if (instance.signAccelerationZ > 0) {
+							correctedInclinationX = -(rawInclinationX - 90);
+						} else {
+							correctedInclinationX = -(180 + (90 - rawInclinationX));
+						}
+					} else
+						// turn left
```
thmq: same here.

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
amix73: Hi!
If the useRotatitonVectorFallback is true - then the rotatiton-matrix is NOT obtained from the RotationVectorSensor (because this does not exist in this phone) but determined by means of the getRotationMatrix(float[],float[],float[],float[]) with accelerometer data AND macgnetic field data in line 303 in the onSensorChanged(...) method - when the MagneticField Sensor is present.
If NEITHER rotation-matrix sensor NOR magnetic field sensor are present compass direction does not work on this phone.

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
amix73: true

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
amix73: no its not

amix73:
removed the two comments

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
thmq: all right! got it! obviously I hadn't understood it properly before, thank you for clearing that up.

thmq:
retest this please

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -117,50 +155,90 @@ public static Double getSensorValue(Sensors sensor) {
...
+					float correctedInclinationX = 0;
+
+					if (rawInclinationX >= 90 && rawInclinationX <= 180) {
+						if (instance.signAccelerationZ > 0) {
+							correctedInclinationX = -(rawInclinationX - 90);
+						} else {
+							correctedInclinationX = -(180 + (90 - rawInclinationX));
+						}
+					} else
+						if (rawInclinationX >= 0 && rawInclinationX < 90) {
```
robertpainsi: That's an interesting case of else-if coding style...
Do you read it as ```else { if (...``` (because a else block doesn't require braces) or as ```else if (...```?
However, let's see what the android studio formatter does! :wink:

Also removing comments, as suggested by @thmq, does not always end up by just removing them. Please take a look at Clean Code, "Don't use a comment when you can use a function or variable".

However, I prefer this example
```java
// Bad:
// Check to see if the employee is eligible for full benefits
if ((employee.flags & HOURLY_FLAG) && (employee.age > 65))

// Good:
if (employee.isEligibleForFullBenefits())
```

thmq:
@robertpainsi 's right about this. you should try to keep your code as well legible as possible. besides there seems to be a problem with the sensor test [[1]](https://jenkins.catrob.at/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2315/testReport/).

avihric:
retest this please

amix73:
@thmq damn this test uses "reflection"!
I was sure the test was green but... since the "accelerometerSensor" which was originally used in the SensorHandler was actually the "linearAccelerationSensor" AND
"linearAcceleration[X|Y|Z]" were misspelled as "linearAcceleartion[X|Y|Z]" ;-)

So this happens when correcting variable-names AFTER running the "final" tests which uses "reflection". :scream_cat: Shame on me that I didn't have a look at the test - so I would have known.

avihric:
I will merge this because the Release is urgent, but you will have to refactor `Sensorhandler` as described by @robertpainsi 

Btw: I will cherry-pick this commit into master

PR #1326, opened 2015-08-04
Cat 0898

LauraBebek:


aried3r:
I took a look at your branch, and I might be wrong, but I don't think this fixes CAT-898. From what I gathered, the ticket says to have a consistent value across all different possibilities of the WaitBrick being displayed. Within the brick selection (1.0), while being dragged after first adding it (empty value) and within the script (1).

I *think* they want it to be displayed with a decimal mark and a trailing 0 if the number is an integer.

robertpainsi:
There are many tickets addressing the same problem with floats and integer representation/inconsistency. Closed until all tickets are unified and cleaned up.

PR #1342, opened 2015-08-12
facedetection fix and resource fixes needed also for standalone apk

martinbu:
facedetection works now even if it is disabled in the settings. The facedetection setting affects only the sensor list, i.e. if the facedetection sensors are visible or not


wslany:
Please change as follows: If a programs contains facedetection sensors in some formulas, the facedetection checkbox should automatically be ticket in the settings, and cannot be unticked for that program unless the corresponding sensor is . By this, the facedetection sensors will automatically be visible in the sensor selection list in the formula editor. The rationale for this is that when the users sees these sensors in formulas, e.g., in a downloaded program, he or she should be able to add them in other formulas as well.

wslany:
Sorry, I did not want to post my comment, please ignore it. I also did not want to close this, so I am reopening it. Sorry!

martinbu:
@wslany 
1) If a programs contains facedetection sensors in some formulas, the facedetection checkbox should automatically be ticket in the settings --> already done, also for lego and phiro
2) cannot be unticked for that program --> not yet done, should be a new ticket but also for lego and phiro

wslany:
@martinbu Sorry polease ignore my deleted posting. It is fine as it is. Point #2 "cannot be unticket..." is not important to me after all.

PR #1345, opened 2015-08-12
Fix Checkstyle

PatrickRadkohl:


aried3r:
This added the catrobat project license to code that is already copyrighted by the Android Open Source Project. See every changed file in `libraryProjects/actionbarsherlock`.

Instead of deleting this, I'd recommend just deleting actionbarsherlock. AFAICS it's not used anymore in favor of the gradle dependency.

PR #1351, opened 2015-08-18
Cat 897

LauraBebek:


LauraBebek:
created new string which has to added at crowdin first

robertpainsi:
Please remove all unrelated commits (CAT-898 and merge) from the PR.
Also the first letter of the [commit message](https://github.com/Catrobat/Catroid/wiki/Setup-working-environment#commit-guidelines) should be upper case (CAT-897 **A**dded...) and there is a typo in desel**e**ct.

avihric:
> created new string which has to added at crowdin first

I don't understand this really. You did not change any string.

catroid/src/org/catrobat/catroid/ui/fragment/SpritesListFragment.java
```diff
@@ -327,15 +330,22 @@ public boolean onContextItemSelected(MenuItem item) {
 
 	@Override
 	public void onSpriteChecked() {
+		Log.d("Lausi", "ONSpriteChecked");
```
robertpainsi: @avihric: Surely there is :wink:

avihric: Erm still this does not explain the comment, as @LauraBebek mentioned Crowdin. In order to change a string at Crowdin, the `strings.xml` has to be modified, which in this case did not happen.

Btw: "Lausi" is not a good tagname :stuck_out_tongue: , more precisely this line should be removed

robertpainsi: > Btw: "Lausi" is not a good tagname  , more precisely this line should be removed

And now you know why I commented on this line in the first place. :stuck_out_tongue_winking_eye:

avihric: Well done sir :+1:

avihric:
Reopen when issues have been resolved

PR #1356, opened 2015-08-23
CAT-1396 Show/hide Variable Brick added

robertriedl:


catroid/res/values/strings.xml
```diff
@@ -846,6 +846,9 @@
  	<string name="add_line_break">Add Line Break</string>
  	<string name="edit_variable">Edit Variable</string>
  	<string name="variable_hint">Variable Name</string>
+    <string name="brick_show_text_var">Show variable</string>
+    <string name="brick_show_text_pos">"at  "</string>
```
robertpainsi: ping @thmq: Are leading or trailing spaces a good idea for crowdin?

catroid/src/org/catrobat/catroid/content/actions/HideTextAction.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.List;
+
+public class HideTextAction extends TemporalAction {
+	private String text;
+
+	@Override
+	protected void begin() {
+		DataContainer varContainer = ProjectManager.getInstance().getCurrentProject().getDataContainer();
+		List<UserVariable> varList = varContainer.getProjectVariables();
```
robertpainsi: No [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation)

catroid/src/org/catrobat/catroid/formulaeditor/UserVariable.java
```diff
@@ -55,4 +56,12 @@ public String getName() {
 	public void setName(String name) {
 		this.name = name;
 	}
+
+	public boolean getVisibility() {
```
robertpainsi: `isVisible()`
Consider to rename visibility to visible as well because of: `if (visible)` vs `if (visibility)`

catroid/src/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -0,0 +1,99 @@
...
+		init();
+	}
+
+	@Override
+	public void draw(Batch batch, float parentAlpha) {
+		DataContainer containerWithVar = ProjectManager.getInstance().getCurrentProject().getDataContainer();
+		List<UserVariable> myVarList = containerWithVar.getProjectVariables();
+
+		if (variableName.equals("No variable set")) {
+
```
robertpainsi: Remove empty line

catroid/src/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -0,0 +1,99 @@
...
+			font.draw(batch, variableName, posX, posY);
+		} else {
+
```
robertpainsi: Remove empty line

catroid/src/org/catrobat/catroid/content/bricks/ShowTextBrick.java
```diff
@@ -0,0 +1,283 @@
...
+		TextView textY = (TextView) prototypeView.findViewById(R.id.brick_show_text_prototype_text_view_y);
+		textY.setText(String.valueOf(BrickValues.Y_POSITION));
+		return prototypeView;
+	}
+
+	@Override
+	public View getViewWithAlpha(int alphaValue) {
+
+		if (view != null) {
+
```
robertpainsi: Remove empty lines

Most of the times you keep a compact and readable coding style. However sometimes you don't. Please be consistent.

catroid/src/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	private void init() {
+		font = new BitmapFont();
+		font.setColor(Color.BLACK);
+		font.getData().setScale(scale);
+	}
+
+	public void setPosX(int posX) {
```
robertpainsi: `setX`? However, see my next comment.

catroid/src/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -0,0 +1,99 @@
...
+		this.variableName = text;
+		this.posX = posX;
+		this.posY = posY;
+		this.linkedVarName = variableName;
+		init();
+	}
+
+	@Override
+	public void draw(Batch batch, float parentAlpha) {
+		DataContainer containerWithVar = ProjectManager.getInstance().getCurrentProject().getDataContainer();
```
robertpainsi: You mean "varContainer" :wink:

robertpainsi:
There is a big inconsistency with position variable names.
x, posX, xPosition, endX, etc. Please take a look at the project source code to be consistent. I guess `x` is the most suited variable name.

Also the abbreviation for variable 'var' shouldn't be used, imo. It definitively leads to inconsistency. You are using var and variable inconsistently already as well. :wink:

I only commented on a few lines of code. Please take a look at your whole code again and change it according to the feedback information. Maybe you find other stuff as well. :+1:

robertriedl:
Thank you @robertpainsi for your feedback.
I've now corrected your suggestions and some other inconsistencies that you mentioned - except (1) the rename of the projectVariables in HideTextAction.java and  (2) the "visibilty" in UserVariable.java.
(1) I think the name here is useful, because here we have to distinguish projectVariables and spriteVariables created in the app, not in the program-code.
(2) Yes your're right, "visible" sounds better, but then the name of the getter & setter of this function gets a little bit confusing - e.g. "setVisible()"  sounds like this setter only sets the variable visible, but not invisible.
I've pushed the changes to my repository, are they also refreshed automatically at the pull-request or do i have something to do to refresh the changes?

avihric:
retest this please

robertriedl:
retested, without any fails

https://jenkins.catrob.at/view/All-Categories/view/Catroid-multi-job/job/Catroid-Multi-Job-Custom-Branch-RELOADED/2525/

robertpainsi:
@(1): I see. `getProjectVariables()` isn't a suited method name and we should rename that eventually.

@(2): java (awt/swing/javafx), gtk+, qt, libgdx, .net windows forms and *many* more are using visible. I've seen visibility only in a context when there are more than two states, like in Android, asp.net (css) and others, and therefore isVisible can't be used.

To keep it short (and since I currently don't have the time to discuss things), I always prefer a well known approach to developers. Since this is mainly used in libgdx context, which uses visible, I would prefer it as well. Lets postpone this.

catroid/src/org/catrobat/catroid/content/actions/HideTextAction.java
```diff
@@ -0,0 +1,69 @@
...
+	protected void update(float percent) {
+		DataContainer projectVariableContainer = ProjectManager.getInstance().getCurrentProject().getDataContainer();
+		List<UserVariable> variableList = projectVariableContainer.getProjectVariables();
+
+		for (UserVariable var : variableList) {
+			if (var.getName().equals(text)) {
+				break;
+			}
+		}
+	}
```
robertpainsi: Does this method do anything?

catroid/src/org/catrobat/catroid/content/actions/HideTextAction.java
```diff
@@ -0,0 +1,69 @@
...
+		DataContainer projectVariableContainer = ProjectManager.getInstance().getCurrentProject().getDataContainer();
+		List<UserVariable> variableList = projectVariableContainer.getProjectVariables();
+
+		for (UserVariable var : variableList) {
+			if (var.getName().equals(text)) {
+				var.setVisibility(false);
+				break;
+			}
+		}
+	}
```
robertpainsi: The UserVariable will always be the same, right?
Save it once, then you don't need to search it every time the action is started/executed.

catroid/src/org/catrobat/catroid/content/actions/ShowTextAction.java
```diff
@@ -0,0 +1,115 @@
...
+import org.catrobat.catroid.formulaeditor.UserVariable;
+import org.catrobat.catroid.stage.ShowTextActor;
+import org.catrobat.catroid.stage.StageActivity;
+
+import java.util.List;
+
+public class ShowTextAction extends TemporalAction {
+	private Formula endX;
+	private Formula endY;
+	private String text;
```
robertpainsi: This is called text but is always used in code like ```userVariable.getName().equals(text)```. So text is not the text which is displayed but the name of the user variable.

catroid/src/org/catrobat/catroid/content/actions/ShowTextAction.java
```diff
@@ -0,0 +1,115 @@
...
+			StageActivity.stageListener.addActor(actor);
+
+			for (UserVariable userVariable : variableList) {
+				if (userVariable.getName().equals(text)) {
+					userVariable.setVisibility(true);
+					break;
+				}
+			}
+		} catch (InterpretationException e) {
+			Log.d("ShowTextAction.java", "InterpretationException");
```
robertpainsi: Consider to log the exception e as well.

robertpainsi: No, don't do this yet. We have to discuss the formula logging because logging a whole exception isn't a good idea. It spams the logcat. Have already seen that :smile_cat:

catroid/src/org/catrobat/catroid/content/actions/ShowTextAction.java
```diff
@@ -0,0 +1,115 @@
...
+		try {
+			int x = endX.interpretInteger(sprite);
+			int y = endY.interpretInteger(sprite);
+
+			actor.setX(x);
+			actor.setY(y);
+		} catch (InterpretationException e) {
+			Log.d("ShowTextAction.java", "InterpretationException");
+		}
+	}
```
robertpainsi: afaik, if an action is executed, first it will call begin, then update (many times if necessary like a glideTo action) and at last end. If the action is called multiple times, begin, update, end are called multiple times as well. Please research or test this flow and let me know if it works that way.

PR #1362, opened 2015-08-27
CAT-909 Delete multiselection at rename objects

LauraBebek:


avihric:
:two: checkstyle warnings

LauraBebek:
fixed checkstyle warnings :)

avihric:
retest this please

avihric:
retest this please

wslany:
Question: How will a Rename then be possible via the "three-dots" menu in the action bar? Remember that almost all users do not know about long press. If renaming is not possible via the menu, then we need a hint about the long press to educate users about it.

LauraBebek:
Rename via the "three-dots" menu is still possible, only the multiselection when rename already is chosen is removed
https://jira.catrob.at/browse/CAT-909

Have to wait for jenkins with restesting.

thmq:
please rebase this.

PR #1380, opened 2015-09-02
CAT-904 Design problem at menu bar

LauraBebek:


thmq:
could you please provide a description saying what you have fixed?

LauraBebek:
deleted all arrows (on the right side), because it is not conform with android style.
https://jira.catrob.at/browse/CAT-904

PR #1381, opened 2015-09-04
Ignore missing and extra translation checks

robertpainsi:
Translation strings should only be removed in the default translation.
Crowdin will update the removed strings in all other translation files
automatically later. To pass the Jenkins job if strings have been
removed from the default translation only, following checks have been
disabled.

- Remove StringTest.java
- Ignore Lint warning ExtraTranslation which would warn if there is a
  translated string which isn't in the default translation.

http://tools.android.com/tips/lint-checks


avihric:
I think @thmq needs to approve this. We had serious problems in the past with additional strings not occuring in the default xml

robertpainsi:
Sure! :crossed_flags:

thmq:
I have checked this now and I think you're right about this. At least when translations are downloaded from Crowdin, the files should be overwritten anyway. As soon as the Jenkins is up again we should be good to merge this.

PR #1383, opened 2015-09-06
Remove Windows line ending checks

robertpainsi:
According to our .gitattributes configuration, checking out files will
convert line endings to native ones and committing will convert line
endings back to LF. Therefore, checking out files on Windows will
convert line endings to CRLF locally. Checkstyle doesn't permit CRLF
line endings and will show lots of errors when ran locally. Since git
normalize line endings on commits automatically, there is no need to
check for unified line endings.

- Remove Windows line ending regexp check
- Adapt NewlineAtEndOfFile check to allow any newline style.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -207,13 +207,9 @@
 
   <module name="NewlineAtEndOfFile">
     <property name="fileExtensions" value="java" />
-    <property name="lineSeparator" value="lf" />
+    <property name="lineSeparator" value="lf_cr_crlf" />
   </module>
   <module name="UniqueProperties" />
-  <module name="RegexpMultiline">
-    <property name="format" value="\r" />
-    <property name="message" value="Do not use Windows line endings." />
-  </module>
 
   <module name="RegexpHeader">
     <property name="headerFile" value="catroidSourceTest/res/agpl_license_text.txt" />


aried3r:
> Since git normalize line endings on commits automatically

[citation needed], afaik, that's exactly what `.gitattributes` is for, but I might be wrong.

obusher:
O_O!

robertpainsi:
>> Since git normalize line endings on commits automatically

> [citation needed], afaik, that's exactly what .gitattributes is for, but I might be wrong.

Quoting myself:
"According to our .gitattributes configuration, checking out files will
convert line endings to native ones and committing will convert line
endings back to LF."

So, you are right! :wink: However, I'll reword it double-:wink: Since there is no ticket for that, I tried to explain, why these changes had to be made. But if it needs clarification, it's a bad commit message. :crying_cat_face:

Some time ago, egit ignored the .gitattributes file (and maybe still does). So it was possible to commit Windows line endings which led to big problems. To prevent such a disaster, a simple Checkstyle check has been added. However, using such a Checkstyle check currently leads to problems on Windows machines. Line endings are converted to CRLF on checkouts and running Checkstyle locally will print lots of errors.

So these checks have to be removed.

robertpainsi:
Re o==[]::::::::::::::::> ed

robertpainsi:
retest this please

robertpainsi:
ok to test

PR #1392, opened 2015-10-01
Arduino feature implementation

AdrianSch:
Finally we were able to get a green checkstyle run on Jenkins, but the following lines should be checked (pretty ugly): ArduinoImpl.java line 264 - 268 (Martin Burtscher + Me couldn't implement it in another way)

PR #1398, opened 2015-10-30
Revert "Arduino feature implementation"

avihric:
Reverts Catrobat/Catroid#1392


avihric:
The Arduino implementation will be merged from the feature branch

PR #1384, opened 2015-09-08
CAT-1361 Redo/Undo for Looks, Sounds and Sprites

loibi93:
Last test on Jenkins is a while back, but there should be no major Problem. The current version has been approved by the UX Team. I am aware that the tests i wrote are not covering every part of this, but since this is so complex you can not test everything. At least i tried to cover all critical cases I encountered when testing it by myself.

Big thanks to @robertpainsi  for helping me with some problems I encountered ;)

PR #1408, opened 2015-11-04
Revert the Redo/Undo Implementation

thmq:
Reverts Catrobat/Catroid#1384.

UPDATE: here is the corresponding JIRA ticket: https://jira.catrob.at/browse/CAT-1646

According to @robertpainsi there are quite a few problems with this. In short: it cannot be guaranteed that ids are unique.

It should be reworked after the release.


thmq:
@robertpainsi: is that all right with you?

robertpainsi:
Thanks for creating a PR and Jira ticket :+1:
As discussed per :envelope:, hopefully quick-fix https://github.com/Catrobat/Catroid/pull/1409 will solve the unique-id problem for the next release.

PR #1409, opened 2015-11-04
Fix non unique ids for undo/redo

robertpainsi:
After a project has been loaded, all used object ids (Sprite, LookData,
SoundInfo and some Bricks) have to be voided. So the ProjectManager
will only hand out unique ids.

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/org/catrobat/catroid/content/Sprite.java
@@ -64,7 +64,7 @@ public class Sprite implements Serializable, Cloneable {
 	private ArrayList<UserBrick> userBricks;
 	private transient int newUserBrickNext = 1;
 
-	private Integer id;
+	private int id = ProjectManager.getInstance().getNewId();
 
 	public Sprite(String name) {
 		this.name = name;
@@ -112,9 +112,6 @@ public class Sprite implements Serializable, Cloneable {
 		}
 		if (userBricks == null) {
 			userBricks = new ArrayList<UserBrick>();
-		}
-		if (this.id == null) {
-			this.id = ProjectManager.getInstance().getNewId();
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


catroid/src/org/catrobat/catroid/content/Sprite.java
```diff
@@ -64,7 +65,7 @@
 	private ArrayList<UserBrick> userBricks;
 	private transient int newUserBrickNext = 1;
 
-	private Integer id;
+	private int id = ProjectManager.getInstance().getNewId();
```
robertpainsi: For consistency.
If a new Sprite is created, the Sprite gets a new id.
If the Sprite is loaded from the XML, the newly received id at this line will be replaced by the id from the xml by xstream.

robertpainsi:
@thmq: Should I add CAT-1646 to the commit header?

robertpainsi:
Dang-it... :pouting_cat:

> The solution works so far, because either all ids are newly assigned or
all ids are loaded from the XML. Once new ids are partially assigned
and partially loaded, e.g. other classes get an unique id as well, this
quickfix won't work anymore and the ProjectManager will hand out
non-unique ids.

Since the release has been postponed, we should definitely try to solve this entirely.
A little [gist](https://gist.github.com/robertpainsi/a34c1934024ffe667b41) to see how xstream handles members during loading.

thmq:
would be nice if you could do that. I'm with you on that, we should really try to solve it entirely, as it would be a shame if we lost all the work lukas has put into it.

robertpainsi:
I could solve the previously mention problem by lazy assigning the ids. However, there are still drawbacks.
So, I'm just leaving a link to the current state of the PR here. [link](https://jira.catrob.at/browse/CAT-1646?focusedCommentId=17108&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17108)

PR #1411, opened 2015-11-06
Adapt checkstyle rules

robertpainsi:
- Add gradle file extension to Checkstyle.
- Add tabs for indentation check for gradle files.
- Add unnecessary multiple newline check for XML files. Maximum is 2 to conform with (almost all) current XML files.

@jonny911: Since there is a change in the strings.xml file, is there any problem with crowdin?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,26 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2015 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


robertpainsi:
Oh, I forgot to enable the licence header check for gradle files too. :pouting_cat:
However, what's the convention on licence headers? Should I add them to the *.gradle files too?

jonny911:
@robertpainsi I'm quite sure there will be no problem with crowdin.

robertpainsi:
@jonny911 Thanks

robertpainsi:
Since https://github.com/Catrobat/Catroid/pull/1404 was reverted by https://github.com/Catrobat/Catroid/pull/1413, this PR was closed because it had to be adapted.
Reopened, however https://github.com/Catrobat/Catroid/pull/1411#issuecomment-154413944 is still unanswered.

thmq:
what do you think could be the problem with adding the license file?

robertpainsi:
> what do you think could be the problem with adding the license file?

I don't think there is one. The [(comments) syntax](http://stackoverflow.com/a/32569925) is shared with [Groovy](http://www.groovy-lang.org/syntax.html#_comments) and gradle tasks like clean, build, test and checkstyle work as expected. For this PR, I enabled the licence header check for gradle files as well now and fixed all violations (revertable any time :wink:).

However, I'm interested in how to handle license headers and also if they are mandatory. I'm aware of not adding a license to binaries like exe, png, etc. but do you really have to add it to any text-like file where it's possible? I'm just curios. :cat:

PR #1417, opened 2015-11-11
Cat 1530

AdrianSch:
I had to change the PhiroSeekBarTests, because the Jenkins reads 126.0 and not 127.0 as my SGS4


AdrianSch:
No tests failed local on my SGS4, but Jenkins tells me  "Build finished. No test results found."

PR #1422, opened 2015-11-12
CAT-1651 Fixed en_US locale gradle build issue

Shudouken:
Fixes an issue that prevents gradle.build form running on en_US locale systems because of using the wrong/german quotation marks “” surrounding some statements


robertpainsi:
Nice catch! :wink:

PR #1423, opened 2015-11-12
CAT-1636 Order XStream XML fields by annotation

robertpainsi:
The new XStreamFieldKeyOrder annotation provides an easier way to
define a custom order for serialized fields in the XML file. Only
serialized fields have to be specified now. If not specified, the
CatroidFieldKeySorter will throw an exception when saving or loading a
project.


robertpainsi:
Closed for improvements :smirk_cat: and will be reopened shortly.

PR #1424, opened 2015-11-13
Init Sprite members directly in declaration

robertpainsi:
This removes the init() method entirely. If the Sprite is created
normally, all members are initialized. If it's loaded by XStream, the
default constructor will be called and all members are initialized as
well. However, they will be replaced by the value in the XML afterwards.


avihric:
@thmq 1 less PR you have to look into :stuck_out_tongue:

robertpainsi:
Thank you :kissing_heart:

PR #1427, opened 2015-11-16
Hide backpack in default menu

sjaindl:
Hides backpack in SoundFragment when an item is longpressed (and in default menu).


sjaindl:
ping @bernadettespieler

sjaindl:
Not necessary anymore as backpack is ready to merge (#1442)

PR #1432, opened 2015-11-27
CAT-1655 Fix upload for restricted profile

84n4n4:
Trying to log in for upload with restricted profile crashed app.
Restricted profiles are used in school classes on shared tablets.

Quickfix:
- If email returned by AccountManager is null, replace with empty
string, to prevent the null exception.

Further changes to handle arising problems:
- Automaticall invalidate TOKEN to "log out" users again right after
upload, failed upload, and canceled upload. (since without this there
is no possibility to log out to prepare tablets for next class)

Limitations:
- Not possible to register new users using a restricted profile.
- Maybe not automatically logged out if upload process is disrupted.
- Log in error messages for restricted users propably misleading,
  eg. not existing username triggers "invalid email adress" error.

PR #1433, opened 2015-11-27
CAT-1655 Fix upload for restricted profile

84n4n4:
Trying to log in for upload with restricted profile crashed app.
Restricted profiles are used in school classes on shared tablets.

Quickfix:
- If email returned by AccountManager is null, replace with empty
string, to prevent the null exception.

Further changes to handle arising problems:
- Automaticall invalidate TOKEN to "log out" users again right after
upload, failed upload, and canceled upload. (since without this there
is no possibility to log out to prepare tablets for next class)

Limitations:
- Not possible to register new users using a restricted profile.
- Maybe not automatically logged out if upload process is disrupted.
- Log in error messages for restricted users propably misleading,
  eg. not existing username triggers "invalid email adress" error.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.1'
+        classpath 'com.android.tools.build:gradle:1.5.0'
         classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.+'
     }
 }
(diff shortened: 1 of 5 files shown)


sjaindl:
ok to test

robertpainsi:
Please use the same branch for the develop and master branch for hotfixes (if possible). In this case, just create a PR with the same branch ([84n4n4:hotfix-0.9.19/1655-restricted-user-upload](https://github.com/84n4n4/Catroid/tree/hotfix-0.9.19/1655-restricted-user-upload)) as in https://github.com/Catrobat/Catroid/pull/1432

See https://github.com/Catrobat/Catroid/pull/1432#issuecomment-160113266 first

PR #1436, opened 2015-11-27
CAT-1655 Fix upload for restricted profile

84n4n4:
Trying to log in for upload with restricted profile crashed app.
Restricted profiles are used in school classes on shared tablets.

Quickfix:
- If email returned by AccountManager is null, replace with empty
string, to prevent the null exception.

Further changes to handle arising problems:
- Automaticall invalidate TOKEN to "log out" users again right after
upload, failed upload, and canceled upload. (since without this there
is no possibility to log out to prepare tablets for next class)

Limitations:
- Not possible to register new users using a restricted profile.
- Maybe not automatically logged out if upload process is disrupted.
- Log in error messages for restricted users propably misleading,
  eg. not existing username triggers "invalid email adress" error.


thmq:
I'm assuming this was opened by mistake? there already exists a PR for this #1435. If, however, this is the one you want to check in then please look at the commit history and squash/rebase accordingly.

The best way to go about fixing this would be to reuse the first PR #1432 and just apply the changes to the branch that has been used for it. If you need any help you can come see me on Monday.

robertpainsi:
> I'm assuming this was opened by mistake? there already exists a PR for this #1435. If, however, this is the one you want to check in then please look at the commit history and squash/rebase accordingly.

Both PRs are on purpose. One is for the develop and the other is for the master branch.

Squashing or rebasing isn't an option here. @84n4n4 did everything correctly according to our [git workflow](http://nvie.com/posts/a-successful-git-branching-model/).

@avihric sneaked :smirk_cat: in https://github.com/Catrobat/Catroid/commit/05662dd86a0a3d9c6afede069258b0f210dc95ae only into the master. That's the reason why it's listed here. It has been added with https://github.com/Catrobat/Catroid/commit/204502004b5b167af88c45153ca839a7f3ed989d to the develop branch.

The other commits are merge commits which aren't present in the develop branch. This bugs me a little. According to the git workflow, every hotfix will be branched from the master but merged into both, master and develop branch. This will add all merge-commits which are only present in the master, to the develop branch...

> The best way to go about fixing this would be to reuse the first PR #1432

Yes, he could've reused the PR :wink:

thmq:
I somehow missed that one was onto the master and one onto the develop. So the way to go now is to merge the patch into both master and develop I assume. I haven't had the time to review it thoroughly enough yet but if you're confident about the resolution you may merge it ;)

robertpainsi:
cant-someone-else-just-do-it

I'm not familiar with the uploading stuff, but a quick look at the git blame tells me, @sjaindl is (or has been a long time ago). Maybe he can take a quick peek at the code and test it?

I guess there isn't a ui test which tests this process? If this process can be tested, @84n4n4, please create one.

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.1'
+        classpath 'com.android.tools.build:gradle:1.5.0'
```
robertpainsi: This is necessary, because...?

catroid/src/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -344,6 +344,12 @@ public boolean registerOrCheckToken(String username, String password, String use
 			userEmail = emailForUiTests;
 		}
 
+		if (userEmail == null) { //restricted users
```
robertpainsi: If you have to write a comment which exactly describes the condition, create a method instead.

robertpainsi:
Ah, don't forget to increase the version number as well in it's [own commit](http://nvie.com/posts/a-successful-git-branching-model/#creating-the-hotfix-branch). :wink:
[Example](https://github.com/Catrobat/Catroid/commit/0abe59f)
versionCode 27
versionName "0.9.20"

catroid/src/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -344,6 +344,12 @@ public boolean registerOrCheckToken(String username, String password, String use
...
+			userEmail = "restricted";
```
robertpainsi: You surely tested it, but is there a problem if the userEmail isn't a valid email address?
[when-i-find-a-hard-coded-string-during-code-review](http://thecodinglove.com/post/111849339882/when-i-find-a-hard-coded-string-during-code-review) :wink:

catroid/src/org/catrobat/catroid/ui/dialogs/UploadProjectDialog.java
```diff
@@ -265,6 +265,12 @@ private void handleUploadButtonClick() {
 	}
 
 	private void handleCancelButtonClick() {
+		Context context = getActivity().getApplicationContext();
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		if (sharedPreferences.getBoolean(Constants.RESTRICTED_USER, false)) {
+			sharedPreferences.edit().putString(Constants.TOKEN, Constants.NO_TOKEN).commit();
+			sharedPreferences.edit().putString(Constants.USERNAME, Constants.NO_USERNAME).commit();
+		}
```
robertpainsi: Duplicate code
Consider to use a method and use it in ProjectUploadService.java as well.

84n4n4:
@robertpainsi I adressed all issues above, can someone have a look again if its fine to merge now?

PR #1439, opened 2015-12-02
Merge hotfixed version of master

thmq:


thmq:
@robertpainsi apparently there are some history issues again? master is now 3 commits behind with the identical code.

robertpainsi:
> You’re all set—the master branch can be safely deleted.

...should I? :joy_cat:

> @robertpainsi apparently there are some history issues again? master is now 3 commits behind with the identical code.

`git log master..develop` lists all commits which are in the develop but not in the master branch. Since those three commits are pull request merges with an empty changeset, they can be ignored entirely.

Two unique merge commits have been created for each, the master and develop branch, because hotfix and release branches have to be merged into both branches.
Release 0.9.19: https://github.com/Catrobat/Catroid/pull/1434 vs https://github.com/Catrobat/Catroid/pull/1430
CAT-1655 Fix upload for restricted profile: https://github.com/Catrobat/Catroid/pull/1435 vs https://github.com/Catrobat/Catroid/pull/1436

Each pull request merge commit is a new, unique commit in the log. The third commit has been created by this pull request.

However, our current model (forking and pull requests) combined with the [git flow](http://nvie.com/posts/a-successful-git-branching-model/) definitely needs some adaption.

Imo, release and hotifx branches should be created on the official Catroid repo to ensure that the branches have been branched from the correct branch (develop for release and master for hotfixes). Additionally, increasing the version number can be done by the coordinator :wink: and also merged into the corresponding hotfix/release branch shortly before those are merged into the develop/master branch.

thmq:
all right. so the main question now is: should we do something about the current situation (master being 3 behind, although it contains the same code)? or better: can we even do something about it?

robertpainsi:
> all right. so the main question now is: should we do something about the current situation (master being 3 behind, although it contains the same code)? or better: can we even do something about it?

There is nothing we should do about it. This is the expected output for all taken actions and the master will be even more commits behind, when new PRs are merged.

I'm not sure where exactly you are struggling, but take a look at the git flow for hotfixes
A successful Git branching model

As you can see, merging the hotfix branch into both, develop and master, creates a new merge commit on each branch. These commits are only located in one branch but not in the other. So, there is no problem if the develop and master branch diverge :wink: 

You may ask, why the master is only behind the develop? Because of this PR :wink: You merged the master including all merge commits into the develop. Our graph differs from the git branching model and the develop branch is 3 ahead now. (According to the branching model, it should be ahead by 1-2 and behind by 2, because the master shouldn't be merged directly into the develop branch and also the increasing of the version number should have been into the hotfix branch, but these are different topics...).

So, even if the output and actions of the git branching model are different, nothing has been broken nor further actions are needed.

robertpainsi:
Btw: Don't forget to tag release 0.9.19 and 0.9.20 correctly. https://github.com/Catrobat/Catroid/releases

thmq:
all right all right, thank you very much for the clarification.

avihric:
@thmq I would follow the instructions from [Branching model](http://nvie.com/posts/a-successful-git-branching-model/) for future releases/hotfixes
I used it as well and it covers everything we need.

Just as a reminder: 
* hotfixes branch from `master`
* releases branch from `develop`

but **both** are merged back into master **and** develop

PR #1441, opened 2015-12-04
CAT-1653 Logout button and Tokenlogin for explore

84n4n4:
Automatically log in users in explore webview if they are logged
in in the app.
Offer a logout button in menu to logged in users, that
logs them out in the app and the webview.


84n4n4:
This depends on webteam to deploy the tokenLogin functionality, aswell as the fix for the invalid token bug that happens after a successful upload.

PR #1449, opened 2015-12-28
CAT-CLEANUP Remove workarounds and checks for deprecated API levels

84n4n4:
Remove checks and workarounds that were implemented for older API levels
that are now below minimum supported API level.

### DIFF
--- a/catroid/src/org/catrobat/catroid/ui/MainMenuActivity.java
+++ b/catroid/src/org/catrobat/catroid/ui/MainMenuActivity.java
@@ -23,12 +23,8 @@
 package org.catrobat.catroid.ui;
 
 import android.app.ActionBar;
-import android.app.AlertDialog;
-import android.content.DialogInterface;
 import android.content.Intent;
-import android.content.SharedPreferences;
 import android.net.Uri;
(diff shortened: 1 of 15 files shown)


catroid/src/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -200,21 +199,9 @@ public void handleWebButton(View view) {
 	}
 
 	public void startWebViewActivity(String url) {
-		// TODO just a quick fix for not properly working webview on old devices
-		if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.GINGERBREAD_MR1) {
-			final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(this);
-			boolean showBrowserWarning = preferences.getBoolean(SHARED_PREFERENCES_SHOW_BROWSER_WARNING, true);
-			if (showBrowserWarning) {
-				showWebWarningDialog();
```
sjaindl: The method showWebWarningDialog() can be entirely removed, as it is not used anymore when deleting it here.

PR #1450, opened 2015-12-28
CAT-FIX Fix exception by LedBrickTest when run together with other tests

84n4n4:
When LedBrickTest is run right after InserItemIntoUserMenuListTest
it throws an exception and ends the local testrun. This is annoying
since jenkins isnt working properly and it prevents me from running
all tests locally overnight.

### DIFF
--- a/catroidTest/src/org/catrobat/catroid/uitest/content/brick/LedBrickTest.java
+++ b/catroidTest/src/org/catrobat/catroid/uitest/content/brick/LedBrickTest.java
@@ -45,9 +45,9 @@ public class LedBrickTest extends BaseActivityInstrumentationTestCase<ScriptActi
 		super(ScriptActivity.class);
 	}
 
-/*	@Override
+	@Override
 	protected void setUp() throws Exception {
-		createProject();
+	/*	createProject();
 		if (hasLedSystemFeature()) {
 		super.setUp();
 			SensorTestServerConnection.connectToArduinoServer();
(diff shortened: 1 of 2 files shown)


84n4n4:
- only matters if one wants to batch run all uitests from android studio.
- verified that this isnt just a problem on my machine, checked with david, aswell as with multiple different devices.
- LedBrickTest is commented out /deactivated anyways, i just commented out the test function itself in LedBrickTest, instead of just its content.

sjaindl:
Maybe you should comment the whole test class:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/445/testReport/junit/junit.framework/TestSuite$1/warning/

84n4n4:
ok, tried to comment all file content, which leads lint to throw exception.
then tried to leave the package in the java file but commented out the rest, now lint is fine, but jenkins throws exception.
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-Device-Single-ClassAndPackage-Test/456/
i suppose the best way to go here is either fix the test for good, or delete it.

im open for suggestions, i know how to work around this to get my local testruns, and i dont know if anyone else is actually bothered by this, so im also fine if this is pull request is rejected.

84n4n4:
Same happening because of VibrationBrickTest.

84n4n4:
fixed now in corporation with sjaindl. now lint, android studio, and jenkins should be ok with it...

PR #1453, opened 2016-01-07
CAT-1673 Program crashes on resume if no or only one camera available

84n4n4:
Fix program crash that happens when no or only one camera is available
on device in following cases:
- restart any program
- resume any program
- start program that utilizes facedetection
Also added message shown on program start if facedetection is to be used
but no camera is available on device.

### DIFF
--- a/catroid/res/values-de/strings.xml
+++ b/catroid/res/values-de/strings.xml
@@ -699,4 +699,5 @@
   <string name="warning">Warnung</string>
   <string name="abort">Abbruch</string>
   <string name="variable">Variable</string>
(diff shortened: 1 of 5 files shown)


84n4n4:
Prior failing UI test regarding "resume" now ok:
https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/Catroid-CustomEmulator-Single-ClassAndPackage-Test/671/testReport/org.catrobat.catroid.uitest.ui.dialog/StageDialogTest/testPauseOnBackButton/

PR #1456, opened 2016-01-08
MUS-182 Create interface between note-events and gui

joseffilzmaier:
Add classes which serve as interface between note events and future GUI.


joseffilzmaier:
https://jenkins.catrob.at/view/Debugging%20CatroindUITests/job/debug-Catroid-ParalellEmulatorUITest-CustomBranch/61/testReport/

PR #1387, opened 2015-09-17
CAT-1357 Video brick

KohlThomas:
Don't merge yet.
Only for discussion.

PR #1460, opened 2016-01-11
CAT 1703 Remove Camera Picker

davidmarogy:
Currently we have the problem if the device dont have a camera, we would see in the settings the Select camera setting, but this isnt very nice so we check if the device has some camera, if not the setting for choosing a camera is disabled.

### DIFF
--- a/catroid/src/org/catrobat/catroid/ui/SettingsActivity.java
+++ b/catroid/src/org/catrobat/catroid/ui/SettingsActivity.java
@@ -61,33 +61,40 @@ public class SettingsActivity extends PreferenceActivity {
 
 		addPreferencesFromResource(R.xml.preferences);
 
+		screen = getPreferenceScreen();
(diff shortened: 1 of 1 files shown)


thmq:
I think this will have to be reviewed after the changes made bei @KohlThomas.

avihric:
You should also squash your commits, actually fixup because the second is the same as the first.
By the way, when you rewrite your commit change it to `CAT-1703` in order that Jira tracks it correctly. The `-` is important.

wslany:
@davidmarogy please do not spend time on this branch as the camera selection has been removed and is now only possible via a dedicated "Select front/back camera" brick in the latest release candidate. This brick will always remain visible, independently of the availability of a camera.

thmq:
already implemented in #1387 ([CAT-1566](https://jira.catrob.at/browse/CAT-1566)).

PR #1461, opened 2016-01-11
CAT-1357 functional chooseCameraBrick and videoBrick

KohlThomas:
added some tests

set background image automatically transparent

removed FaceDetectionSettings

FaceDetection, Flash and Preview now work

fixed problem with releasing camera

set transparency to zero after preview

fixed testcase


robertpainsi:
You made a lot of commits locally, that's a good thing and the right way to use git :smirk_cat:. However, when squashing all these commits into a single one, you also have to rewrite your commit message. As a start, I'll just point out to our [commit message guidelines](https://confluence.catrob.at/display/KNOWHOW/Commit+Message+Guidelines), especially the listed sources at the end are definitely worth reading.

Hint: Sometimes you create a problem in one commit, but later fix it in another one. Squashing will obscure the problem in the first place. So there is no need to mention the fix in the squashed/final commit :wink:

robertpainsi:
+5 technical debt :crying_cat_face:

PR #1463, opened 2016-01-12
fixed problem on stage

KohlThomas:
With this changes the nullpointerexception cannot occur.


avihric:
@thmq @KohlThomas the 2 lint warnings were not fixed...

PR #1480, opened 2016-01-20
CAT-1649 Vibration Brick

davidmarogy:
Changed default value in brick from 1000 seconds to 1 second, onclick on the value opens now the formula editor instead of option menu


thmq:
This did not fix the problem, you cannot simply copy and paste a method from the base class into the derived class. You could have just fixed it by setting the OnClickHandler to the correct TextView. As for the constants: I won't even talk about what you did there, because ...

PR #1483, opened 2016-01-22
CAT-1725 Open Authentication with Facebook and Google+

sjaindl:
Adds Facebook and Google+ sign-in functionality to Pocket Code. 
After successful authentication with Facebook or Google+ on the device, authentication is also done with our web server.

Furthermore, login and registration are separated. In the registration dialog there is now the possibility to change the E-Mail address as well.


sjaindl:
@kirshanluhana is currently working on a new Jenkins device Job, that replaces the placeholders in oauth_config.xml in the test assets with real values just before the test run, as we don't want to commit credentials to github.

PR #1484, opened 2016-01-27
CAT-Quickfix Repair NewProjectDialogTest

84n4n4:
Fix Test that became outdated/broken with CAT-1732.


robertpainsi:
Fixes [CAT-1734](https://jira.catrob.at/browse/CAT-1734)

PR #1486, opened 2016-02-02
CAT-999 fixed dialog behaviour

jonny911:
* shows now error message when trying to load an image with special chars in filename
-- new string for better user notification
* Dialog behaviour
-- sprite is not created if loading the image fails


thmq:
please have a look at painsibot's suggestions :smile:

PR #1501, opened 2016-02-05
CAT FIX - Deactivate removeAllCookies function in webview

84n4n4:
Deactivate function that is not needed yet* to fix build problems.
(*will be needed as soon as token login is activated)


84n4n4:
to fix build problems that occur building, and to not have whole logout functionality be pulled from release. (i know this is only a quickfix, and its not the nicest way to just comment out the code).

thmq:
is fixed by #1499.

PR #1522, opened 2016-02-15
CAT-1689 Broadcast not called twice in Landscapemode

StefanBuerscher:


thmq:
This will be fixed by #1526

PR #1519, opened 2016-02-15
CAT-1557 Refactor Mergefunction

thomasmauerhofer:
JIRA Link: https://jira.catrob.at/browse/CAT-1557

PR #1523, opened 2016-02-15
CAT-1557 Refactor Mergefunction

thomasmauerhofer:
replaces: #1519


catroid/src/org/catrobat/catroid/content/Sprite.java
```diff
@@ -470,4 +492,26 @@ public boolean containsLookData(LookData lookData) {
 		}
 		return false;
 	}
+
+	public boolean existSoundInfoByName(SoundInfo sound) {
+		for (SoundInfo soundinfo : soundList) {
```
thmq: soundInfo

PR #1524, opened 2016-02-15
CAT-1440 Cancelling creating a new variable

StefanBuerscher:
will no longer name the variable New...

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/bricks/SetVariableBrick.java
+++ b/catroid/src/org/catrobat/catroid/content/bricks/SetVariableBrick.java
@@ -185,6 +185,7 @@ public class SetVariableBrick extends UserVariableBrick {
 				if (position == 0 && ((UserVariableAdapterWrapper) parent.getAdapter()).isTouchInDropDownView()) {
 					NewDataDialog dialog = new NewDataDialog((Spinner) parent, NewDataDialog.DialogType.USER_VARIABLE);
 					dialog.addVariableDialogListener(SetVariableBrick.this);
+					dialog.setUserVariableIfCancell(userVariable);
 					dialog.show(((Activity) view.getContext()).getFragmentManager(),
 							NewDataDialog.DIALOG_FRAGMENT_TAG);
 				}
--- a/catroid/src/org/catrobat/catroid/ui/dialogs/NewDataDialog.java
+++ b/catroid/src/org/catrobat/catroid/ui/dialogs/NewDataDialog.java
@@ -57,6 +57,7 @@ public class NewDataDialog extends DialogFragment {
 	public static final String DIALOG_FRAGMENT_TAG = "dialog_new_data_catroid";
 	Spinner spinnerToUpdate;
 	DialogType dialogType = DialogType.SHOW_LIST_CHECKBOX;
+	UserVariable userVariableIfCancell;
 
 	public static enum DialogType {
 		SHOW_LIST_CHECKBOX, USER_LIST, USER_VARIABLE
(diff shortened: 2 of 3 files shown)


84n4n4:
doesnt fix the problem described in ticket for "show variable", "change variable" , "hide variable" bricks.
only "set variable" has been modified.

catroid/src/org/catrobat/catroid/content/bricks/SetVariableBrick.java
```diff
@@ -185,6 +185,7 @@ public void onItemSelected(AdapterView<?> parent, View view, int position, long
 				if (position == 0 && ((UserVariableAdapterWrapper) parent.getAdapter()).isTouchInDropDownView()) {
 					NewDataDialog dialog = new NewDataDialog((Spinner) parent, NewDataDialog.DialogType.USER_VARIABLE);
 					dialog.addVariableDialogListener(SetVariableBrick.this);
+					dialog.setUserVariableIfCancell(userVariable);
```
84n4n4: why not pass the spinner selection instead?

catroid/src/org/catrobat/catroid/ui/dialogs/NewDataDialog.java
```diff
@@ -89,6 +90,17 @@ public NewDataDialog(Spinner spinnerToUpdate, DialogType dialogType) {
 	public void onCancel(DialogInterface dialog) {
 		super.onCancel(dialog);
 		userListDialogListenerListFinishNewUserListDialog(null);
+		if (userVariableIfCancell != null) {
+			int spinnerSize = spinnerToUpdate.getCount() - 1;
+			for (int spinnerIndex = 1; spinnerIndex <= spinnerSize; spinnerIndex++) {
+				spinnerToUpdate.setSelection(spinnerIndex);
+				if (spinnerToUpdate.getSelectedItem().equals(userVariableIfCancell)) {
```
84n4n4: UserVariableAdapterWrapper offers method getPositionOfItem() that does exactly this on a spinner. also if spinner id is passed instead of the userVariable this could also be skipped.

catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
```diff
@@ -94,6 +94,15 @@ public void testVariableBricks() {
 				dataContainer.getUserVariable("p2", sprite).getValue());
 	}
 
+	public void testCreatingNewVariableCancelling() {
+		solo.clickOnText("p2");
```
84n4n4: access spinners directly like in testVariableBricks, instead of text search.

catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
```diff
@@ -94,6 +94,15 @@ public void testVariableBricks() {
...
+		solo.clickOnText("p1");
+		solo.clickOnText("p1");
+		solo.clickOnText(solo.getString(R.string.brick_variable_spinner_create_new_variable));
+		solo.clickOnText(solo.getString(R.string.cancel_button));
+		solo.clickOnText("p1");
```
84n4n4: use assertion rather than clickOnText to verify that "p1" is visible.

catroid/src/org/catrobat/catroid/ui/dialogs/NewDataDialog.java
```diff
@@ -57,6 +57,7 @@
 	public static final String DIALOG_FRAGMENT_TAG = "dialog_new_data_catroid";
 	Spinner spinnerToUpdate;
```
fwinkelbauer: class member without a visibility modifier get a default visibility - which means that they can be accessed by any class in the same package. In this example all your member variables (spinnerToUpdate, dialogType and userVariableIfCancell [why cancell instead of cancel?]) are exposed to the package. Please make them private!

catroidTest/src/org/catrobat/catroid/uitest/content/brick/VariableBricksTest.java
StefanBuerscher: When I use "solo.pressSpinnerItem(spinnerIndex, -position);" to click on "New...", the "NewDataDialog" does not appear. That's why I used "clickOnText".
Any suggestions?

thmq:
this is replaced by #1541, apparently.

PR #1531, opened 2016-02-19
Cat 1758

LauraBebek:
Facebook and Google+ buttons, fix in inital If value


thmq:
this is exactly the same thing as #1531. please change it accordingly!!!

PR #1532, opened 2016-02-22
"About Pocket Code" and "(c) Catrobat" links should open browser intent

wslany:
https://jira.catrob.at/projects/CAT/issues/CAT-1793


wslany:
Hmm, just thinking, there actually should be regression tests for such things, right?

thmq:
Yes, you're right, we should be looking into it. Just one thing: could you please open a new PR onto the release-0.9.21 branch in order to incorporate this change into the release?

PR #1553, opened 2016-02-29
CAT-FIX Updated gradle version to 1.2.0 and fixed resulting lint issues

thmq:


robertpainsi:
_Why gradle version 1.2.0?_
There are more [recent stable versions](https://bintray.com/android/android-tools/com.android.tools.build.gradle/view), like 1.5.0

_Why even updating?_
Gradle version 2.0 seems to be around the corner. Teaming up with Android Studio 2.

_Since it's a fix, what does it fix? To be more precise, what was the problem before this commit?_

I'm just curious and also have no knowledge about gradle. :smile_cat:

PR #1556, opened 2016-03-01
CAT-1830 Removed ignores from lint and fixed resulting problems

thmq:
- removed unnecessary lint ignores.
- enabled html lint report.
- updated strings.
- replaced `...` with `&#8230;`


robertpainsi:
You missed [one](https://github.com/thmq/Catroid/blob/dec2ee1f1c6e8fa317a3772d12e6248d7f83b150/catroid/res/values-tr/strings.xml#L446) :scream_cat:... :smile_cat:

> replaced `...` with `&#8230;`

Should have been: replaced `...` with `…`
Always use symbols instead of codes.

PR #1557, opened 2016-03-03
Fix CAT-1831 Donot track gradle.properties

Achilles-96:


thmq:
Thank you for your input, however, we have to review this change internally because I cannot tell you if the file is necessary for internal reasons. You can add a workaround for your own configuration by untracking the files via git (You can find more information [here](http://stackoverflow.com/questions/6964297/untrack-files-from-git)).

PR #1560, opened 2016-03-06
CAT-1317 Use constants instead of hardcoded strings

Achilles-96:


Achilles-96:
Should I add any test that makes sure no hard coded strings are used in future.

catroid/src/org/catrobat/catroid/content/bricks/DroneMoveBrick.java
```diff
@@ -45,6 +45,7 @@
 import java.util.List;
 
 public abstract class DroneMoveBrick extends FormulaBrick {
+	private static final String TAG = ProjectManager.class.getSimpleName();
```
thmq: You should use the TAGs corresponding to the class rather than the ProjectManager's TAG.

thmq:
As I line-commented: You should use the TAGs corresponding to the class rather than the ProjectManager's TAG. But in general this is a very good idea, and adding a test (or probably even incorporating this test into checkstyle/pmd/lint) would be incredibly helpful.

Furthermore: are you sure about when to use Log.e vs Log.w vs Log.d? Because if you have any doubts about it someone can surely help you.

Achilles-96:
About `Log.e`, I followed what is suggested on issue page. I just identified all the Log tags containing error string in it and changed their priority to `e`. Yes, any more suggestions in this regard are helpful. Also, it will be great if you can point me to resources on how to write checkstyle/pmd/lint messages.

Achilles-96:
I updated PR. This way, whenever a class name is changed, we need not bother about changing the TAG as well.

Achilles-96:
Any update on this?

robertpainsi:
Hi @Achilles-96! Thanks for contributing to Catrobat. :smiley_cat:

I quickly tested your code. Here is my logcat
```
03-10 14:38:30.553 21998-21998/org.catrobat.catroid D/java.lang.Thread: loadProject My first program
03-10 14:38:30.553 21998-21998/org.catrobat.catroid D/java.lang.Thread: path: /storage/emulated/0/Pocket Code/My first program/code.xml
03-10 14:38:42.848 21998-22018/org.catrobat.catroid D/java.lang.Thread: saveProject My first program
…
```

As you can see, all tags on my machine are _java.lang.Thread_.
Timber, for example, is using a different approach to gather the log tag. See [here](https://github.com/JakeWharton/timber/blob/a6b2e220a33da939da181de37a5cd68be843f63a/timber/src/main/java/timber/log/Timber.java#L423-L446). It also strips packages and the inner class syntax.

However, please use `ClassName.class.getSimpleName()` in this PR. It'll be much easier to switch to timber or an approach like [here](https://github.com/google/iosched/blob/0a90bf8e6b90e9226f8c15b34eb7b1e4bf6d632e/android/src/main/java/com/google/samples/apps/iosched/util/LogUtils.java#L27-L42) in the future, if and only if necessary.

e.g.
```java
public final class FormulaEditorTestUtil {
    private static final String TAG = FormulaEditorTestUtil.class.getSimpleName();
```

Achilles-96:
I'll update the PR. Actually, I thought of getting class name in without using ClassName so that we don't need to update this statement when we rename class. I'll update the PR.

catroid/src/org/catrobat/catroid/ui/dialogs/SignInDialog.java
```diff
@@ -470,7 +471,7 @@ public void onGoogleExchangeCodeComplete() {
 
 	@Override
 	public void onGetFacebookUserInfoTaskComplete(String id, String name, String locale, String email) {
-		Log.d("FB", "User Info complete");
+		Log.d(TAG, "User Info complete");
```
84n4n4: to not loose information here it would be better to include the "FB" in the message: "FB User Info complete"

catroidTest/src/org/catrobat/catroid/uitest/web/ProjectUpAndDownloadTest.java
```diff
@@ -565,11 +565,11 @@ private void uploadProject(String uploadProjectName, String uploadProjectDescrip
 		String resultString = (String) Reflection.getPrivateField(ServerCalls.getInstance(), "resultString");
 
 		try {
-			Log.v("resultString=", "" + resultString);
+			Log.v(TAG, "" + resultString);
 			JSONObject jsonObject;
 			jsonObject = new JSONObject(resultString);
 			serverProjectId = jsonObject.optInt("projectId");
-			Log.v("serverID=", "" + serverProjectId);
+			Log.v(TAG, "" + serverProjectId);
```
84n4n4: same here, move "serverID=" part over to message.

84n4n4:
@Achilles-96 
>  so that we don't need to update this statement when we rename class. 

this is a non-issue since renaming can be easily done with android studios refactor - rename, which does all the work and renames all occurrences and usages of a class to the newly chosen name. (all other modern IDE will have similar functionality)

Achilles-96:
Sure :) . Updated.

PR #1562, opened 2016-03-07
#CAT-1034 fix for the wifi connection to the drone

Manthan-R-Sheth:
This Pull request is in response to issue #CAT-1034 . 
The user can now connect the android device's wifi to the drone's wifi directly from within the app.


thmq:
please squash this into one single commit and be sure not to integrate any .idea files.

Manthan-R-Sheth:
ya i will assure to not include .idea files henceforth.


Sent with Mixmax

On Mon, Mar 7, 2016 at 4:01 PM, Thomas Schranz < notifications@github.com > wrote:
please squash this into one single commit and be sure not to integrate any .idea
files.

fwinkelbauer:
We currently can't merge these changes because you have added changes to our .xml and .iml files. We have to discuss internally why these files exist in our repository in the first place (do we still need them or should we just add them to .gitignore?).

PR #1581, opened 2016-03-20
Fixed default broadcast message

Achilles-96:
Fixes https://jira.catrob.at/browse/CAT-1708.

### DIFF
--- a/catroid/res/values-en-rAU/strings.xml
+++ b/catroid/res/values-en-rAU/strings.xml
@@ -472,7 +472,7 @@
   <string name="brick_change_variable">Change variable</string>
   <string name="brick_variable_spinner_create_new_variable">New&#8230;</string>
   <string name="brick_note_default_value">"add comment here…"</string>
-  <string name="brick_broadcast_default_value">"message 1"</string>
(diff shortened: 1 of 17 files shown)


thmq:
Have you considered @robertpainsi 's comment on this issue? "message 1" is used as the default value in Scratch and can be misleading for children as the drop down field used for creating a new message also reads "new message". However, if you still want to change it please find a more appropriate message and please consider the following points:

Please add CAT-1708 to your commit message, so that it reads:

`CAT-1708 Fixed default broadcast message`

be sure to squash the commits if you add another one or just use:

`git commit --amend`

to change the message without adding another commit. (Further details on the usage can be found [here](https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend).

Furthermore, you did not have to change the message in all translations, it would've sufficed to change the default strings.xml (en-US: found under res/values), as your translation system automatically generates translations for all other languages. If you're interested in getting involved in the translation process, just let me know.

Achilles-96:
By the time I made this commit, the mentioned comment is not made.

After reading the comment, I also feel 'message1' is more apt as it also in accordance with Scratch.

PR #1584, opened 2016-03-23
CAT-968 Formula Editor wrong variable handling

thomasmauerhofer:


fwinkelbauer:
I get the feeling that `DataProvider.java` is a very nice god class. Do we really want to keep adding stuff to it? I mean I get that this change is necessary, but what are our long term plans? Also we only have **one** test for this class... Any ideas or opinions on this? Or is there already a plan?

thomasmauerhofer:
I don't added anything in DataProvider.java, only two methods in DataContainer.java. I think the changes in DataContainer.java are ok, because I need this methods also for CAT-1746 (Renaming of Lists and Variables).

fwinkelbauer:
Ye I can't read. `DataProvider` != `DataContainer`.

I get that we need these changes, my question is a rather general one and not only directed at you :)

PR #1585, opened 2016-03-24
CAT-1820 background issue in version 0.98

StefanBuerscher:


fwinkelbauer:
So we just had a look at this and are unsure if that solves all problems. Loading a project might give a black background for a very short moment. Is that intended?

Also please take a look at the code at line 304. Is this `whiteBackground` somehow involved or still needed after fixing the black background?

thmq:
@fwinkelbauer is right, the implementation is rather chaotic (not due to your changes of course), however, please have a look at what he suggested with special attention to `whiteBackground` and its role in reloading the project as well as its necessity and its dependencies (i.e.: `if(spriteSize ...`). I will close this PR for now, please re-open when you're done. 

I haven't looked into it in too much detail, so if you find out that it is correctly implemented the way it is now, please let me know.

robertpainsi:
The `whiteBackground` is needed for on tapped events. It defines the clickable area of the background if no background is defined. There may be [other _unrelated_ problems (CAT-1876) with this approach](https://jira.catrob.at/browse/CAT-1876).

StefanBuerscher:
So if the white Background is rotated, tapping at a position that is just white because of the gdx-background (but not part of whiteBackground) does not lead to an event.

robertpainsi:
Off topic:
> I will close this PR for now, please re-open when you're done.

Developer with read access only [cannot re-open PRs](https://help.github.com/articles/repository-permission-levels-for-an-organization) and if they force-pushed to their remote branch, even developer with write access cannot re-open PRs.
The HowTo ['How to reopen a pull-request after a force-push?'](https://confluence.catrob.at/pages/viewpage.action?pageId=9666894) is rather useless since only few developer can do that.

GitHub currently improved protected branches. See https://github.com/blog/2137-protected-branches-improvements `User and team restrictions`. It's definitely worth considering.

robertpainsi:
> Developer with read access only cannot re-open PRs

Ah, that's not stated in https://help.github.com/articles/repository-permission-levels-for-an-organization
Please verify first and let me know :smile_cat:

PR #1590, opened 2016-03-30
CAT-1797 Open backpack context menu right after click

sjaindl:
Context menu in backpack now opens right after it has been clicked (not when it has been released).


sjaindl:
ok to test

sjaindl:
ok to test

thmq:
ok to test

PR #1592, opened 2016-03-31
CAT-1820 background issue in version 0.98

StefanBuerscher:
This does not solve the problems mentioned in CAT-1876.


StefanBuerscher:
ok to test

thmq:
ok to test

PR #1601, opened 2016-04-01
CAT-1884 Fix crash when adding two looks from media library

Achilles-96:
Fixes [CAT-1884](https://jira.catrob.at/browse/CAT-1884)


thmq:
ok to test

catroid/src/org/catrobat/catroid/ui/dialogs/OverwriteRenameMediaDialog.java
```diff
@@ -218,8 +218,6 @@ private boolean handleOkButton() {
 					for (LookData lookData : ProjectManager.getInstance().getCurrentSprite().getLookDataList()) {
 						if (lookData.getLookName().compareTo(mediaName) == 0) {
 							lookToRemove = lookData;
-							ProjectManager.getInstance().getCurrentSprite().getLookDataList().remove(lookData);
```
thmq: Please have another look at this. I think this can be implemented more elegantly. For example: you could just remove the look here rather than in lines 223 - 226, and add a break in order to prevent the loop from iterating through all looks after the correct one has been found (same goes for the sounds: lines 235 - 238).

thmq:
please reopen after you've had a look at my suggestions.

PR #1603, opened 2016-04-03
CAT-1529 Grouped list functions together in FE

Fasusu:
CatroidCheck: https://jenkins.catrob.at/view/All-Categories/view/Catroid/job/CatroidCheck/48/
SingleClassTest: https://jenkins.catrob.at/job/Catroid-CustomEmulator-Single-ClassAndPackage-Test/720/

Second commit contains only auto format changes where spaces are replaced with tabs and a newline is added to the end of the file, so that checkstyle does not complain anymore

### DIFF
--- a/catroid/src/org/catrobat/catroid/ui/fragment/FormulaEditorListFragment.java
+++ b/catroid/src/org/catrobat/catroid/ui/fragment/FormulaEditorListFragment.java
@@ -136,9 +136,6 @@ public class FormulaEditorListFragment extends ListFragment implements Dialog.On
 	private String actionBarTitle;
 	private int[] itemsIds;
 
-	public FormulaEditorListFragment() {
-	}
-
 	private static int[] concatAll(int[] first, int[]... rest) {
 		int totalLength = first.length;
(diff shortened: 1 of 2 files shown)


thmq:
ok to test

thmq:
could you please format your code according to the bot's suggestions?

fwinkelbauer:
Or just use the Android Studio Formatter like it is mentioned in the first bullet point [here](https://github.com/Catrobat/Catroid/wiki/Creating-a-pull-request)

catroid/src/org/catrobat/catroid/ui/fragment/FormulaEditorListFragment.java
```diff
@@ -135,6 +136,23 @@
 	private String actionBarTitle;
 	private int[] itemsIds;
 
+	public FormulaEditorListFragment() {
```
cheidenreich: the default constructor would be autogenerated by the compiler -> can be removed

thmq:
this is replaced by #1660.

PR #1612, opened 2016-04-07
CAT-1862 physics object type default value

Gorphi:
Set default physics object type value depending on the present physics bricks


meiserm:
reviewed 11.04.2016

PR #1621, opened 2016-04-12
CAT-1883 Correctly handled a just downloaded missing program

LauraBebek:


robertpainsi:
Please take a look at [CAT-1883#commentId=17762](https://jira.catrob.at/browse/CAT-1883?focusedCommentId=17762&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17762). Keeping this code the way it is, is extremely dangerous (and doesn't make any sense).

PR #1622, opened 2016-04-12
CAT-1740 Refactored FormulaEditor Layout

StefanBuerscher:


84n4n4:
Hardcoded percentages inside every single button style are bad way to go around this. If a new button was to be introduced in future all existing buttons have to be modified.

PR #1623, opened 2016-04-13
CAT-1716 Delete the dustbin icon in the action bar and put in the Act…

bernadettespieler:
…ion Bar Menu

### DIFF
--- a/catroid/res/menu/menu_current_project.xml
+++ b/catroid/res/menu/menu_current_project.xml
@@ -43,7 +43,7 @@
     <item
         android:id="@+id/delete"
         android:icon="@drawable/ic_menu_delete"
-        android:showAsAction="ifRoom"
+        android:showAsAction="never"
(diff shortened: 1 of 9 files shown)


thmq:
ok to test

fwinkelbauer:
Please fix the format check like mentioned [here](https://github.com/Catrobat/Catroid/wiki/Creating-a-pull-request)

catroid/res/menu/menu_current_project.xml
```diff
@@ -43,7 +43,7 @@
     <item
         android:id="@+id/delete"
         android:icon="@drawable/ic_menu_delete"
```
thmq: It would make more sense to remove the drawable altogether as you won't need an icon for a menu entry (same for the other three occurrences of this line and the drawable itself).

thmq:
please reopen when you're done fixing this.

PR #1628, opened 2016-04-14
CAT-1570 AsyncTask is added while creating a new project

Manthan-R-Sheth:
I have added async task to the OrientationDialog.java class and executed it on the handleOkButtonClick() method. The freezing of the UI is hence replaced by a progress dialog to achieve the goal.
@robertpainsi can you have a look at this.

### DIFF
--- a/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
+++ b/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
@@ -145,7 +145,7 @@ public class OrientationDialog extends DialogFragment {
 		@Override
 		protected void onPreExecute() {
 			super.onPreExecute();
-			this.dialog.setMessage("Please wait");
+			this.dialog.setMessage(getResources().getString(R.string.please_wait));
 			this.dialog.show();
 		}
 
@@ -163,7 +163,6 @@ public class OrientationDialog extends DialogFragment {
 				}
 
 				Intent intent = new Intent(getActivity(), ProjectActivity.class);
-
 				intent.putExtra(Constants.PROJECTNAME_TO_LOAD, projectName);
 
 				if (isOpenedFromProjectList()) {
@@ -174,7 +173,6 @@ public class OrientationDialog extends DialogFragment {
 				dismiss();
 				return true;
 			} catch (Exception e) {
-
 				return false;
 			}
 		}
--- a/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
+++ b/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
@@ -57,7 +57,7 @@ public class OrientationDialog extends DialogFragment {
 	private RadioButton landscapeMode;
 	private boolean createEmptyProject;
 	private boolean createLandscapeProject = false;
-	private Context mContext;
+	private Context context;
 
 
 	private boolean openedFromProjectList = false;
@@ -67,7 +67,7 @@ public class OrientationDialog extends DialogFragment {
 	public Dialog onCreateDialog(Bundle savedInstanceState) {
 		View dialogView = LayoutInflater.from(getActivity()).inflate(R.layout.dialog_orientation_new_project, null);
 
-		mContext = getActivity();
+		context = getActivity();
 
 		orientationDialog = new AlertDialog.Builder(getActivity()).setView(dialogView)
 				.setTitle(R.string.project_orientation_title)
@@ -105,7 +105,7 @@ public class OrientationDialog extends DialogFragment {
 
 	protected void handleOkButtonClick() {
 		createLandscapeProject = landscapeMode.isChecked();
-		LoadingOnCreatingProject loadingOnCreatingProject = new LoadingOnCreatingProject(mContext);
+		LoadingOnCreatingProject loadingOnCreatingProject = new LoadingOnCreatingProject(context);
 		loadingOnCreatingProject.execute();
 	}
 
--- a/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
+++ b/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
@@ -123,17 +123,26 @@ public class OrientationDialog extends DialogFragment {
 		createDroneProject = isChecked;
 	}
 
-	private class LoadingOnCreatingProject extends AsyncTask<String, Void, Boolean> {
-		private ProgressDialog progressDialog;
+    private class LoadingOnCreatingProject extends AsyncTask<String, Void, Void> {
+        private ProgressDialog progressDialog;
 
 		protected LoadingOnCreatingProject() {
 			progressDialog = new ProgressDialog(getActivity());
 		}
 
 		@Override
-		protected void onPostExecute(Boolean showing) {
-			progressDialog.dismiss();
-		}
+        protected void onPostExecute(Void result) {
+            Intent intent = new Intent(getActivity(), ProjectActivity.class);
+            intent.putExtra(Constants.PROJECTNAME_TO_LOAD, projectName);
+
+            if (isOpenedFromProjectList()) {
+                intent.putExtra(Constants.PROJECT_OPENED_FROM_PROJECTS_LIST, true);
+            }
+
+            getActivity().startActivity(intent);
+            dismiss();
+            progressDialog.dismiss();
+        }
 
 		@Override
 		protected void onPreExecute() {
@@ -142,29 +151,18 @@ public class OrientationDialog extends DialogFragment {
 		}
 
 		@Override
-		protected Boolean doInBackground(String... params) {
+        protected Void doInBackground(String... params) {
 
-			try {
-					ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
-				} catch (IllegalArgumentException illegalArgumentException) {
+            try {
+                ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
+            } catch (IllegalArgumentException illegalArgumentException) {
 					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
 				} catch (IOException ioException) {
 					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
 					Log.e(TAG, Log.getStackTraceString(ioException));
 					dismiss();
 				}
-
-				Intent intent = new Intent(getActivity(), ProjectActivity.class);
-				intent.putExtra(Constants.PROJECTNAME_TO_LOAD, projectName);
-
-				if (isOpenedFromProjectList()) {
-					intent.putExtra(Constants.PROJECT_OPENED_FROM_PROJECTS_LIST, true);
-				}
-
-				getActivity().startActivity(intent);
-				dismiss();
-				return true;
-
-		}
-	}
+            return null;
+        }
+    }
 }
--- a/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
+++ b/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
@@ -26,8 +26,11 @@ package org.catrobat.catroid.ui.dialogs;
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.app.DialogFragment;
+import android.app.ProgressDialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.os.AsyncTask;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.LayoutInflater;
@@ -54,6 +57,8 @@ public class OrientationDialog extends DialogFragment {
 	private RadioButton landscapeMode;
 	private boolean createEmptyProject;
 	private boolean createLandscapeProject = false;
+	private Context mContext;
+
 
 	private boolean openedFromProjectList = false;
 	private boolean createDroneProject = false;
@@ -61,6 +66,8 @@ public class OrientationDialog extends DialogFragment {
 	@Override
 	public Dialog onCreateDialog(Bundle savedInstanceState) {
 		View dialogView = LayoutInflater.from(getActivity()).inflate(R.layout.dialog_orientation_new_project, null);
+
+		mContext = getActivity();
 
 		orientationDialog = new AlertDialog.Builder(getActivity()).setView(dialogView)
 				.setTitle(R.string.project_orientation_title)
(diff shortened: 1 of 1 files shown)


catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +128,55 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+		@Override
+		protected void onPostExecute(Boolean showing) {
+			super.onPostExecute(showing);
+			dialog.dismiss();
+		}
+
+		@Override
+		protected void onPreExecute() {
+			super.onPreExecute();
+			this.dialog.setMessage("Please wait");
```
thomasmauerhofer: Please use R.string... from the Resources

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +128,55 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+				} catch (IllegalArgumentException illegalArgumentException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
+				} catch (IOException ioException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
+					Log.e(TAG, Log.getStackTraceString(ioException));
+					dismiss();
+				}
+
+				Intent intent = new Intent(getActivity(), ProjectActivity.class);
+
```
thomasmauerhofer: remove newline

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +128,55 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+				if (isOpenedFromProjectList()) {
+					intent.putExtra(Constants.PROJECT_OPENED_FROM_PROJECTS_LIST, true);
+				}
+
+				getActivity().startActivity(intent);
+				dismiss();
+				return true;
+			} catch (Exception e) {
+
```
thomasmauerhofer: remove newline

Manthan-R-Sheth:
Updated the PR with the changes you told @thomasmauerhofer

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -54,6 +57,8 @@
 	private RadioButton landscapeMode;
 	private boolean createEmptyProject;
 	private boolean createLandscapeProject = false;
+	private Context mContext;
```
robertpainsi: Please rename to `context`.

Manthan-R-Sheth:
Updated the PR @robertpainsi  @thomasmauerhofer

Manthan-R-Sheth:
Can this be merged now?

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -54,6 +57,8 @@
 	private RadioButton landscapeMode;
 	private boolean createEmptyProject;
 	private boolean createLandscapeProject = false;
+	private Context context;
+
```
thomasmauerhofer: Remove newline

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -62,6 +67,8 @@
 	public Dialog onCreateDialog(Bundle savedInstanceState) {
 		View dialogView = LayoutInflater.from(getActivity()).inflate(R.layout.dialog_orientation_new_project, null);
 
+		context = getActivity();
+
```
thomasmauerhofer: Remove newline

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -62,6 +67,8 @@
 	public Dialog onCreateDialog(Bundle savedInstanceState) {
 		View dialogView = LayoutInflater.from(getActivity()).inflate(R.layout.dialog_orientation_new_project, null);
```
thomasmauerhofer: Remove newline

Manthan-R-Sheth:
updated the pr

Manthan-R-Sheth:
Can this be merged now? @robertpainsi  @thomasmauerhofer

robertpainsi:
Between you and me, I hate to review Android specific stuff :pouting_cat:. Especially if I don't know the best practices for this component and have to do some research first. So I'll just drop some new comments and leave the rest to one of our new seniors :wink:.

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+		@Override
+		protected void onPostExecute(Boolean showing) {
+			super.onPostExecute(showing);
+			dialog.dismiss();
+		}
+
+		@Override
+		protected void onPreExecute() {
+			super.onPreExecute();
+			this.dialog.setMessage(getResources().getString(R.string.please_wait));
```
robertpainsi: Remove super call, create ProgressDialog here (Instead of context, use `getActivity()`). Also remove `this.` here and the next line.

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
 	public void setCreateDroneProject(boolean isChecked) {
 		createDroneProject = isChecked;
 	}
+
+	private class LoadingOnCreatingProject extends AsyncTask<String, Void, Boolean> {
+		private ProgressDialog dialog;
+
+		protected LoadingOnCreatingProject(Context context) {
```
robertpainsi: Remove constructor

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -54,14 +57,15 @@
 	private RadioButton landscapeMode;
 	private boolean createEmptyProject;
 	private boolean createLandscapeProject = false;
-
+	private Context context;
```
robertpainsi: You don't need this member

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
 	public void setCreateDroneProject(boolean isChecked) {
 		createDroneProject = isChecked;
 	}
+
+	private class LoadingOnCreatingProject extends AsyncTask<String, Void, Boolean> {
+		private ProgressDialog dialog;
```
robertpainsi: Rename to progressDialog

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+
+		protected LoadingOnCreatingProject(Context context) {
+			dialog = new ProgressDialog(context);
+		}
+
+		@Override
+		protected void onPostExecute(Boolean showing) {
+			super.onPostExecute(showing);
```
robertpainsi: Remove super call

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+			try {
+				try {
+					ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
+				} catch (IllegalArgumentException illegalArgumentException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
+				} catch (IOException ioException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
+					Log.e(TAG, Log.getStackTraceString(ioException));
+					dismiss();
+				}
```
robertpainsi: The original code had a return in every catch block. You need it as well to stop at these points.

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +125,53 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+				intent.putExtra(Constants.PROJECTNAME_TO_LOAD, projectName);
+
+				if (isOpenedFromProjectList()) {
+					intent.putExtra(Constants.PROJECT_OPENED_FROM_PROJECTS_LIST, true);
+				}
+
+				getActivity().startActivity(intent);
+				dismiss();
+				return true;
+			} catch (Exception e) {
```
robertpainsi: Avoid catching generic exceptions such as RuntimeException or Exception. There is no reason to expect any exception it the code above.

robertpainsi:
Hmm... It seems a little odd that the OrientationDialog handles all this stuff... I'll take a look at it again in a few hours :wink:.

Manthan-R-Sheth:
Thanks for reviewing the code.
```java
	try {
				try {
					ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
				} catch (IllegalArgumentException illegalArgumentException) {
					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
				} catch (IOException ioException) {
					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
					Log.e(TAG, Log.getStackTraceString(ioException));
					dismiss();
				}

				Intent intent = new Intent(getActivity(), ProjectActivity.class);
				intent.putExtra(Constants.PROJECTNAME_TO_LOAD, projectName);

				if (isOpenedFromProjectList()) {
					intent.putExtra(Constants.PROJECT_OPENED_FROM_PROJECTS_LIST, true);
				}

				getActivity().startActivity(intent);
				dismiss();
				return true;
			} catch (Exception e) {
				return false;
			}
```
This part of the code is unchanged from the original code. I just shifted it to the *doInBackground(String... params)* function. Hence the try-catch calls are those which have already been included in the main code.

Manthan-R-Sheth:
Super calls are boilerplate codes provided for the functions of AsyncTask . It would be very helpful if you could explain to me the removal of super calls.
 
I made the constructor to pass the context but will update it by using *getActivity()* instead. Passing context as the constructor parameter is a good practice but I will update it according to what you say.

:smile:

robertpainsi:
> This part of the code is unchanged from the original code. I just shifted it to the doInBackground(String... params) function. Hence the try-catch calls are those which have already been included in the main code.

The inner try/catch block is fine. You somehow added the big, outer try/catch block. ([original code](https://github.com/Catrobat/Catroid/blob/71250deb6dbae42201cf734bc190a7a036110c49/catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java#L99-L125))
```java
try {
...
} catch (Exception e) {
    return false;
}
```

> Super calls are boilerplate codes provided for the functions of AsyncTask . It would be very helpful if you could explain to me the removal of super calls.

The [implementation](https://github.com/android/platform_frameworks_base/blob/6f28d02bedb2cf8a0532b2c9b7184e4363f64135/core/java/android/os/AsyncTask.java#L383-L386) of `onPostExecute` is empty. So calling `super.onPostExecute()` has no effect at all. Same for [onPreExecute](https://github.com/android/platform_frameworks_base/blob/6f28d02bedb2cf8a0532b2c9b7184e4363f64135/core/java/android/os/AsyncTask.java#L367-L369). The Android documentation also provides an usage example where super isn't called. http://developer.android.com/reference/android/os/AsyncTask.html

> Passing context as the constructor parameter is a good practice...

Yes, definitely! The reason why I suggest not to pass it, is, because `LoadingOnCreatingProject` is an inner class. It can easily access the context from its outer class (as it already does in the `doInBackground` method) and also doesn't need to be tested (directly).
If you take a look at the implementation of the [Iterator in the ArrayList](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/ArrayList.java#736), you will see that the constructor also doesn't take any parameters (because it can access the elements of the list easily).
So, at the end and _imo_ it's just not necessary to pass the context to the inner class.

Manthan-R-Sheth:
Sorry for the late response, I was busy with the end of academic year formalities :stuck_out_tongue: 
I have made the changes suggested by you @robertpainsi .

1.  Super Calls have been removed.
2.  context variable is removed and now the inner class accesses the context directly. (constructor is removed)
3.  dialog is changed to progressDialog in the iner class.

Any further changes required?

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +122,49 @@ public void setCreateEmptyProject(boolean isChecked) {
 	public void setCreateDroneProject(boolean isChecked) {
 		createDroneProject = isChecked;
 	}
+
+	private class LoadingOnCreatingProject extends AsyncTask<String, Void, Boolean> {
```
cheidenreich: you are not using the return value so this could be `AsyncTask<String, Void, Void>`

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +122,49 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+					ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
+				} catch (IllegalArgumentException illegalArgumentException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
+				} catch (IOException ioException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
+					Log.e(TAG, Log.getStackTraceString(ioException));
+					dismiss();
+				}
+
+				Intent intent = new Intent(getActivity(), ProjectActivity.class);
```
cheidenreich: As `Activity.startActivity(Intent)` runs on the UI Thread anyway, I'd rather put the Intent creation and Activity start into the `onPostExecute()` which also runs o the UI Thread.
The long term task (which `doInBackground` is designed for) here is the `InitializeNewProject()`

cheidenreich:
please have a look at my code comments and also have a look at @robertpainsi 's comment about returning from the catch block.
also make sure the code is formatted.

After this is done, I'll make a quick test run, but in general it looks good

Manthan-R-Sheth:
@cheidenreich I made the changes suggested by you and updated the PR.
I had kept boolean to allow us to check if the dialogbox was dismissed or not. But void also does the work.
Have a look at the changes .

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +122,47 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+        protected Void doInBackground(String... params) {
+
+            try {
+                ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
+            } catch (IllegalArgumentException illegalArgumentException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
+				} catch (IOException ioException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_new_project);
+					Log.e(TAG, Log.getStackTraceString(ioException));
+					dismiss();
```
cheidenreich: is the `onPostExecute` executed after the dismiss call? if yes, i sense problems in case of an exception

catroid/src/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -143,4 +122,47 @@ public void setCreateEmptyProject(boolean isChecked) {
...
+			this.progressDialog.show();
+		}
+
+		@Override
+        protected Void doInBackground(String... params) {
+
+            try {
+                ProjectManager.getInstance().initializeNewProject(projectName, getActivity(), createEmptyProject, createDroneProject, createLandscapeProject);
+            } catch (IllegalArgumentException illegalArgumentException) {
+					Utils.showErrorDialog(getActivity(), R.string.error_project_exists);
```
cheidenreich: `dismiss()` here too?
it would be a strange behaviour when creating a project which already exists and the manager opens the existing one

thmq:
Please have a look at the suggestions by @cheidenreich and reopen when done.

PR #1631, opened 2016-04-14
Replace &#8230; by ellipsis character …

robertpainsi:
Lint suggests to use the ellipsis character … or the numeric character
reference `&#8230;` instead of three dots. The ellipsis character … is
much more readable and therefore all `&#8230;` have been replaced in the
default strings.xml.

### DIFF
--- a/catroid/res/values/strings.xml
+++ b/catroid/res/values/strings.xml
@@ -52,12 +52,12 @@
     <string name="loading_check_facebook_data">Facebook data is retrieved…</string>
     <string name="loading_check_email">E-Mail availability is checked on our server…</string>
     <string name="loading_check_oauth_token">OAuth token is checked on our server…</string>
-    <string name="loading_check_facebook_token_validity">Facebook token validity is checked on our server&#8230;</string>
-    <string name="loading_check_oauth_username">Username availability is checked on our server&#8230;</string>
-    <string name="loading_facebook_exchange_token">Facebook token is exchanged on our server&#8230;</string>
-    <string name="loading_facebook_login">Facebook login is completed on our server&#8230;</string>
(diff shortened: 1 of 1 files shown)


catroid/res/values-ar/strings.xml
```diff
@@ -391,11 +391,11 @@
   <string name="stage_dialog_maximize">زيادة الامتداد\n</string>
   <string name="stage_dialog_screenshot">لقطة للشاشة</string>
   <!-- Broadcast spinner -->
-  <string name="new_broadcast_message">جديد&#8230;</string>
+  <string name="new_broadcast_message">جديد…</string>
```
thmq: Arabic is spelled from left to right, therefore you should've add the character on the left hand side. Furthermore adding the characters to the translations does not really help, as they are overwritten by Crowdin. The best way to go about changing this would be to change the character in the default strings.xml and then add it in Crowdin's system.

robertpainsi: > ...you should've add the character on the left hand side.

Nah :wink:!
The numeric character reference `&#8230;` is also on the right side (open this line in Android Studio if your browser somehow splits it) and changing the position of the ellipsis is beyond the purpose of this PR.

Your doubt is still valid. I just tested both strings and the `…` is automatically moved to the left side by Android. So I don't know why the Crowdin translator added the three dots on the right side... :confused:

Can you take a look directly at the Crowdin translation? Maybe there is a problem during the integration process.

> ...The best way to go about changing this would be to change the character in the default strings.xml and then add it in Crowdin's system.

Sure! Someone will add them to the Crowdin system for me, right?

thmq: Sure! I'll have a look at it :wink, as soon as I find the time. Thank you for your input.

PR #1634, opened 2016-04-19
Cat 1393 brick value consistency

cheidenreich:
default values of bricks are now integers without comma,
user input in the formula is trimmed if a number has a trailing ".0[...]"


cheidenreich:
I will take care of code formatting tomorrow

catroid/src/org/catrobat/catroid/utils/Utils.java
```diff
@@ -680,4 +680,8 @@ public static boolean isUserLoggedIn(Context context) {
 					|| token.equals(ServerCalls.TOKEN_CODE_INVALID));
 		return tokenValid;
 	}
+
+	public static String getNumberStringForBricks(float value) {
+		return (int) value == value ? "" + (int) value : "" + value;
```
thomasmauerhofer: Please use somthing like value.intValue(); Why do you cast into an integer, when return is a String?

cheidenreich: This method returns the display string for brick values. It gets rid of the comma and everything behind, if the float value is an integer. This is a requirement by the UX team for this issue

e.g. 15.0f > "15"
4.00000000f > "4"
4.31f > "4.31"

thmq:
y u no squash?

thmq:
please reopen when done

PR #1635, opened 2016-04-19
Cat 1877 - Prevent double backpacking of sounds, looks & objects

sjaindl:
Prevents backpacking of sounds, looks & objects that are already in the backpack (same name). The users can decide whether they want to replace the existing items.

I've also updated to the newest gradle version and fixed the resulting exception of Jake Wharton's SDK Manager as well as the CatroidFieldKeySorter, as @robertpainsi described in CAT-1658.

For multidexing javaMaxHeapSize should be increased for faster apk builds.

Works like a charm with the new AS Instant Run! :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,12 +22,18 @@
  */
 
 buildscript {
-    repositories {
-        mavenCentral()
-    }
-    dependencies {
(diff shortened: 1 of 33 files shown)


sjaindl:
ok to test

thmq:
ok to test

PR #1637, opened 2016-04-20
CAT-1824 Fixed Detail links on Marketing activity

Chypsylon:
For newer standalone apps the detail-link in the final marketing
activity stopped working as the url-tag in code.xml doesn't contain the
full url anymore.

### DIFF
--- a/buildAPK.gradle
+++ b/buildAPK.gradle
@@ -191,7 +191,7 @@ task buildStandalone(dependsOn: ':downloadProject') << {
 }
 
 task downloadProject() << {
-    def id = 824;//3143; //debugging purpose only
+    def id = 6760;//3143; //debugging purpose only
     def url = "";
     if (project.hasProperty("download")) {
         url = project["download"]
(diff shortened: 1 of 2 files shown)


buildAPK.gradle
```diff
@@ -191,7 +191,7 @@ task buildStandalone(dependsOn: ':downloadProject') << {
 }
 
 task downloadProject() << {
-    def id = 824;//3143; //debugging purpose only
+    def id = 6760;//3143; //debugging purpose only
```
thmq: but why?

catroid/src/org/catrobat/catroid/ui/MarketingActivity.java
```diff
@@ -84,6 +84,9 @@ public void onClick(View v) {
 			@Override
 			public void onClick(View v) {
 				String url = ProjectManager.getInstance().getCurrentProject().getXmlHeader().getUrl();
+				if (!url.trim().startsWith("http://")) {
+					url = "https://share.catrob.at" + url;
```
thmq: you should always use constants instead of hardcoded values (see Constants.java).

thmq:
please rethink this PR and reopen when done.

PR #1639, opened 2016-04-20
CAT-1868 Fix the camera front, flashlight, vibration bug

Manu91:
If you want to turn on the flashlight on the front camera, but you doesn't have a flashlight there and you also have an vibration brick in your program, you will get an error, the stage will be destroyed, but the vibrator doesn't stop. So i added an check to pause the vibration.


sjaindl:
ok to test

thmq:
ok to test

PR #1643, opened 2016-04-21
CAT-1931 Add Locale.getDefault to toUpperCase

JoachimLesser:
Add Locale.getDefault in StorageHandler .toUpperCase to distuingish
languages

### DIFF
--- a/catroid/src/org/catrobat/catroid/io/StorageHandler.java
+++ b/catroid/src/org/catrobat/catroid/io/StorageHandler.java
@@ -685,7 +685,7 @@ public final class StorageHandler {
 			Log.e(TAG, "file " + inputFilePath + " doesn`t exist or can`t be read");
 			return null;
 		}
-		String inputFileChecksum = Utils.md5Checksum(inputFile).toUpperCase();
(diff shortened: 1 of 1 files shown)


catroid/src/org/catrobat/catroid/io/StorageHandler.java
```diff
@@ -685,7 +685,7 @@ private File copyFileBackPack(String programSubDirectory, String backpackSubDire
 			Log.e(TAG, "file " + inputFilePath + " doesn`t exist or can`t be read");
 			return null;
 		}
-		String inputFileChecksum = Utils.md5Checksum(inputFile).toUpperCase();
+		String inputFileChecksum = Utils.md5Checksum(inputFile).toUpperCase(Locale.getDefault());
```
robertpainsi: `If you want the methods to just perform ASCII replacement, for example to convert an enum name, call String#toUpperCase(Locale.US) instead.`

However, I think upper casing the checksum isn't necessary (or shouldn't be done) in the first place. ping @sjaindl

robertpainsi:
:pouting_cat: As referred above, `Locale.US` should be used instead of `Locale.getDefault()`.

However, I still doubt that upper case should be used at all...

PR #1652, opened 2016-04-27
CAT-1160 Formula Editor: Fix Wrong value of pi

bernadettespieler:


catroid/src/org/catrobat/catroid/formulaeditor/Formula.java
```diff
@@ -272,8 +272,6 @@ public String getResultForComputeDialog(Context context) {
 			} catch (InterpretationException interpretationException) {
 				return "ERROR";
 			}
-			interpretationResult *= 100;
-			interpretationResult = (double) (Math.round(interpretationResult) / 100f);
```
robertpainsi: Removing these lines changes the default compute output behavior. Such a change has to be discussed with the UX team first.

Current output: The result rounded up to two decimal places*
Your output: The result :smirk_cat:

\* It currently doesn't because of floating-point inaccuracy and the next line...

bernadettespieler: what do you mean with current output: rounded up to two decimals?
Currently: https://www.dropbox.com/s/ko2sskxlsv5zd6a/Screenshot_2016-04-27-17-18-41.png?dl=0 
all decimal places but the wrong value - we only changed the value to the right one: 3.141592653589793 - why do you need input from the ux?

robertpainsi: You're right and this fixes the problem, of course.
Back then we limited the number of decimals to two to prevent a weird floating point error output (e.g. `0.1+0.1+0.1=0.30000000000000004`).

PR #1653, opened 2016-04-27
CAT-1924 Made strings editable in formula editor

bernadettespieler:


catroid/src/org/catrobat/catroid/formulaeditor/InternFormula.java
```diff
@@ -912,6 +912,40 @@ public int getExternSelectionEndIndex() {
 		return externSelectionEndIndex;
 	}
 
+	private InternToken getSelectedToken() {
+		if (internFormulaTokenSelection == null || internFormulaTokenSelection.getTokenSelectionType() != TokenSelectionType.USER_SELECTION) {
```
thomasmauerhofer: Please use an boolean value in the if.. it's better readable

catroid/src/org/catrobat/catroid/formulaeditor/InternFormula.java
```diff
@@ -912,6 +912,40 @@ public int getExternSelectionEndIndex() {
...
+			return null;
+		}
+		int currentIndex = 0;
+		for (InternToken token : internTokenFormulaList) {
+			if (token.getInternTokenType() == InternTokenType.STRING
```
thomasmauerhofer: Please use an boolean value in the if.. it's better readable

thmq:
please reopen when done.

PR #1658, opened 2016-04-30
 CAT-1595 Implemented Rating Dialog

jonny911:
If a user uploads two projects, a dialog will ask to rate Pocket Code on Google Play Store.


thmq:
ok to test

PR #1664, opened 2016-05-03
CAT-1904 Added description field expansion in programm overview

bernadettespieler:


catroid/res/layout/dialog_text_multiline_dialog.xml
```diff
@@ -21,30 +21,37 @@
   ~ You should have received a copy of the GNU Affero General Public License
   ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   -->
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
thmq: This will not work. Hint: Layout has no root. Check even suggests that. Furthermore you should always reopen old PRs instead of just creating a new one.

PR #1686, opened 2016-05-11
CAT-1740 Refactored FormulaEditor Layout

StefanBuerscher:


thmq:
ok to test

cheidenreich:
not sure about the exact goals of this ticket --> see JIRA for details

bernadettespieler:
The children had problems to tap on the buttons because they are too small on tablets. This solves the problem!
The font size is another step which will be coverd with https://jira.catrob.at/browse/CAT-1984

cheidenreich:
okay, thanks for clearing that ;)

PR #1690, opened 2016-05-12
CAT-1884 Fix crash when adding two looks or sounds from media library

Achilles-96:
Fixes [CAT-1884](https://jira.catrob.at/browse/CAT-1884)


cheidenreich:
ok to test

PR #1699, opened 2016-05-14
CAT-1970 VideoBrick did not resume

Manu91:
If you want to resume your project, the VideoBrick did not resume. So i added a new brick resources "VIDEO" and resume it correctly.

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/bricks/CameraBrick.java
+++ b/catroid/src/org/catrobat/catroid/content/bricks/CameraBrick.java
@@ -165,7 +165,7 @@ public class CameraBrick extends BrickBaseType {
 
 	@Override
 	public int getRequiredResources() {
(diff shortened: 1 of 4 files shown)


thmq:
ok to test

catroid/src/org/catrobat/catroid/content/bricks/CameraBrick.java
```diff
@@ -165,7 +165,7 @@ public View getViewWithAlpha(int alphaValue) {
 
 	@Override
 	public int getRequiredResources() {
-		return Brick.NO_RESOURCES;
+		return Brick.VIDEO;
```
thmq: Why do you need a separate bit mask for video? wouldn't it work with CAMERA_BACK || CAMERA_FRONT?

PR #1703, opened 2016-05-18
CAT FIX - remove broken testInvalidProject ui test

84n4n4:
Test is broken, test brakes whole test run, and this shouldnt be a
ui test in the first place.


84n4n4:
ready to test

PR #1711, opened 2016-05-20

jonny911:


thmq:
please re-open onto develop

PR #1720, opened 2016-05-29
HOT - change collision identifier from \0 to \t for XML v1.0

Gorphi:
changing the Collision Identifier from the zero-byte to the byte with value 9 (\t) in order to be usable with xml 1.0

locally tested with https://share.catrob.at/pocketcode/program/9546 and https://share.catrob.at/pocketcode/program/9547.


Gorphi:
ok to test

thmq:
ok to test

robertpainsi:
Ahm, this Pull Request didn't get it in the play store release 0.9.22. But it will be in the next hotfix :smile_cat:

`<receivedMessage>Object&lt;&#x0;-&#x0;&gt;&#x0;ANYTHING&#x0;</receivedMessage>`

PR #1725, opened 2016-06-01
CAT-1922 Formular Editor: Modify the way of deleting

bernadettespieler:


Bioxar:
Longclick should be implemented for back button in input field as well. reopen when done.

bernadettespieler:
No! The button in the input field will be deleted by another ticket see: https://github.com/Catrobat/Catroid/pull/1727

PR #1732, opened 2016-06-04
Cat 1949 - Comment Out Bricks

p4p4:
The “comment out” feature allows to disable a Brick and make it visually distinguishable from the active ones. This means that the users no longer have to delete a script if they want to temporarily disable it.

In the context menu of a Brick, you can choose "Enable Brick" or "Disable Brick"


thmq:
ok to test

p4p4:
i have to fix stuff..

PR #1735, opened 2016-06-08

jonny911:


thmq:
When you return from a script, look or sound list you arrive at the project activity. What should happen is that you should be returned to the sprite the list belongs to.

Furthermore: When you step into a deeper level (i.e. from the sprite list to the script list) the new screen moves in from the right and when using the back button the previous screen moves in from the left. When you use the top bar to navigate back the new screen moves in from the right.

Could you change that please? If there is any reason (e.g. Android's material design guidelines please let me know).

PR #1739, opened 2016-06-08
CAT-955 File chooser

jonny911:
File chooser opens instead of brick categories screen
Refactored Activity-Fragment interaction


Bioxar:
ok to test

Bioxar:
ok to test

PR #1740, opened 2016-06-08
CAT-1925 Look/Background Number/Name

bernadettespieler:


thmq:
ok to test

catroid/src/org/catrobat/catroid/ui/fragment/FormulaEditorListFragment.java
```diff
@@ -182,6 +189,13 @@ public void onStart() {
 
 		if (tag.equals(OBJECT_TAG)) {
 			itemsIds = OBJECT_ITEMS;
+			Context context = this.getActivity().getApplicationContext();
+			if (ProjectManager.getInstance().getCurrentSprite().getName().equals(context.getString(R.string
+					.background))) {
```
robertpainsi: Currently there are two ways to check if a sprite is the background sprite.
`sprite.getName().equals(context.getString(R.string.background))`
or
`project.isBackgroundObject(sprite)` which internally uses `spriteList.indexOf(sprite) == 0`

Personally I prefer the second one. The first element in the sprite list is always the background. You can't delete the background. `R.string.background` depends on the language and is not unique. A project created from an English version of PC may have the background called `Background` and an Object called `Hintergrund`. If you upload this project and download it from a German version of PC, the background is called `Hintergrund` and also the Object is called `Hintergrund`. (Haven't tested it, just assuming it). So the background string doesn't verify if the object is the background.

Feedback is still appreciated on this topic. :smirk_cat:

Bioxar:
ok to test

bernadettespieler:
errors with compute

PR #1753, opened 2016-06-15
CAT-1295 Miscalculation with decimal places

bernadettespieler:

PR #1754, opened 2016-06-16
CAT-1066: Add math operation for exponentiations(power)

bernadettespieler:
Changed code like suggested from @robertpainsi but for special cases may the try and catch block is needed (class FormulaElement.java):

try {
    return java.lang.Math.pow(doubleValueOfLeftChild, doubleValueOfRightChild);
    } catch (Exception e) {
    return 0d;
    }


catroid/src/org/catrobat/catroid/formulaeditor/Formula.java
```diff
@@ -279,7 +279,61 @@ public String getResultForComputeDialog(Context context) {
...
+		}
+
+		private static String normalCut(String doubleString, int decimals) {
+			int decimalIndex = doubleString.indexOf(COMMA);
+			int endIndex = decimalIndex + decimals + 1;
+
+			if (endIndex >= doubleString.length()) {
+				endIndex = doubleString.length();
+			}
+			return doubleString.substring(0, endIndex);
```
robertpainsi: These changes are part of another Pull Request/Jira ticket. (https://github.com/Catrobat/Catroid/pull/1753 / [CAT-1295](https://jira.catrob.at/browse/CAT-1295)). Please remove them.

robertpainsi:
> but for special cases may the try and catch block is needed (class FormulaElement.java):

Which cases do you mean?

bernadettespieler:
ok according to the programmer there are no cases left that need a try and catch block. sorry please ignore

catroid/src/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -354,6 +354,10 @@ private Object interpretFunction(Functions function, Sprite sprite) {
 				return doubleValueOfLeftChild == null ? 0d : java.lang.Math.toDegrees(Math.atan(doubleValueOfLeftChild));
 			case EXP:
 				return doubleValueOfLeftChild == null ? 0d : java.lang.Math.exp(doubleValueOfLeftChild);
+			case POWER:
+				return ((doubleValueOfLeftChild == null || doubleValueOfRightChild == null)
+						|| (doubleValueOfLeftChild < 0 && doubleValueOfRightChild < 1 && doubleValueOfRightChild > -1))
```
robertpainsi: There are just too many input/output combinations listed in the [Math.pow](https://developer.android.com/reference/java/lang/Math.html#pow(double, double)) documentation... :smile_cat:
I just tested two cases you catch here (only in java):
```java
Math.pow(-1, 0.5); // == NaN
Math.pow(-1, 0) // == 1.0
```
So, why should both cases return `0` instead?

HCrane: Many apologies! There has been a twist in our thinking process :scream_cat: Of course 
```Java
Math.pow(-1,0)
```

should be 1! Mhm since there are a truckload of special cases wouldnt it be better to use try and catch blocks? I know they are a bit "ugly" because they disrupt the current layout of the cases. On the other hand it would cover all special cases as listet in the [documentation](https://developer.android.com/reference/java/lang/Math.html#pow(double,%20double)). Ill try some tests but I have to restart my PC for it so this is it for now :smiley_cat:

84n4n4: what happens if the method would return NaN? (like it currently does with SQRT when a negative number is passed)

does it make sense to replicate all special cases/behaviours of all used methods of math library, or wouldnt it be better to refactor the whole switch case part of the method to instead of doing special case handling and right away returning method call result to write them some place and check for NaN after the switch case and then returning value?

HCrane: @robertpainsi I was able to determine the special case (NaN) we had and where the try and catch came in handy and not at the same time.
```
power(-5,(1/5))
```
will return  ```Error``` and I believe that is the Try/Catch way of handling with NaN but I dont understand why because:
```java
try {
	return java.lang.Math.pow(doubleValueOfLeftChild, doubleValueOfRightChild);
} catch (Exception e) {
	return 0d;
}
```
Should evaluate to 0 whatever exception is thrown..

robertpainsi: NaN is a valid return value of the `Formula.interpretFunction` function.

The compute button will show ERROR because children have no idea what NaN is.
In the stage, if a formula evaluates to NaN, the action is completely ignored. At least if the formula should be a number like in the SetXAction.

So, using following code snipped is fine
```java
case POWER:
    return (doubleValueOfLeftChild == null || doubleValueOfRightChild == null) ? 0d : java.lang.Math.pow(doubleValueOfLeftChild, doubleValueOfRightChild);
```

Btw, `Math.pow` doesn't throw any exception :wink:

PR #1755, opened 2016-06-19
CAT-2015 HOTFIX 0.9.21 - Exception Handling

84n4n4:
Deal with unhandled exceptions and properly finish and restart app.


84n4n4:
ok to test

PR #1757, opened 2016-06-20
CAT-1863 Enumerating newly created backgrounds

Achilles-96:
Fixes [CAT-1863](https://jira.catrob.at/browse/CAT-1863)
@thmq , Sorry for late reply. But I was asked to make it as two commits - one for refactoring and other for the rest by @robertpainsi . Initially, this was one commit.


thmq:
retest this please

PR #1759, opened 2016-06-21
CAT-1330 Added visible distinction for logical operators

Achilles-96:
Fixes [CAT-1330](https://jira.catrob.at/browse/CAT-1330)
Though mentioned change is to make it bold, I felt changing color appeared better. I will change if you insist on making it bold. If we use color, we can extend this to math operators as well with different colors.


thmq:
ok to test

thmq:
retest this please

thmq:
test this please

PR #1766, opened 2016-06-23
CAT-1954 IF-Then brick without ELSE

p4p4:
New Brick:
- IF-then without ELSE

reuses a lot of the standard if-else brick via inheritance. uses the same Action as the existing brick

This PR includes a TestCase

reviewed & refactored

Checkstyle OK:
- [jenkins CatroidCheck results](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/229/)

Tests:
- [jenkins package 'test' results](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/2170/)
- this pull does not break any tests
-- the two failing tests already fail in the current develop branch (June 23rd): [results for 'develop' branch](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/2170/)

PR #1770, opened 2016-06-24
CAT-1954 IF-Then brick without ELSE

p4p4:
**this reopens PR #1766 after rebasing** as requested by @thmq 

New Brick:
- IF-then without ELSE

reuses a lot of the standard if-else brick via inheritance. uses the same Action as the existing brick

This PR includes a TestCase

reviewed & refactored

Tests:
- [jenkins package 'test' results](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/2170/)
- this pull does not break any tests
-- the two failing tests already fail in the current develop branch (June 23rd): [results for 'develop' branch](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/2170/)

### DIFF
--- a/catroid/src/org/catrobat/catroid/io/StorageHandler.java
+++ b/catroid/src/org/catrobat/catroid/io/StorageHandler.java
@@ -103,6 +103,8 @@ import org.catrobat.catroid.content.bricks.IfLogicBeginBrick;
 import org.catrobat.catroid.content.bricks.IfLogicElseBrick;
 import org.catrobat.catroid.content.bricks.IfLogicEndBrick;
 import org.catrobat.catroid.content.bricks.IfOnEdgeBounceBrick;
+import org.catrobat.catroid.content.bricks.IfThenLogicBeginBrick;
+import org.catrobat.catroid.content.bricks.IfThenLogicEndBrick;
 import org.catrobat.catroid.content.bricks.InsertItemIntoUserListBrick;
 import org.catrobat.catroid.content.bricks.LegoNxtMotorMoveBrick;
 import org.catrobat.catroid.content.bricks.LegoNxtMotorStopBrick;
@@ -330,6 +332,8 @@ public final class StorageHandler {
 		xstream.alias("brick", IfLogicBeginBrick.class);
 		xstream.alias("brick", IfLogicElseBrick.class);
 		xstream.alias("brick", IfLogicEndBrick.class);
+		xstream.alias("brick", IfThenLogicBeginBrick.class);
+		xstream.alias("brick", IfThenLogicEndBrick .class);
 		xstream.alias("brick", IfOnEdgeBounceBrick.class);
 		xstream.alias("brick", InsertItemIntoUserListBrick.class);
 		xstream.alias("brick", FlashBrick.class);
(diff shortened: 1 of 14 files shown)


Bioxar:
ok to test

Bioxar:
ok to test

Bioxar:
Please check this issue: [checkstyle](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidTriggerPullRequestTest/317/)

p4p4:
@Bioxar this seems to be an inconsistency between Checkstyle and the Formatter

Checkstyle does not allow to import with an asterisk:
    Using the '.*' form of import should be avoided - org.catrobat.catroid.content.bricks.*.

But that's exacty what our code formatter suggests! (before there were import statements for each and every Brick subclass, which were reported as unformatted.) 

Maybe @robertpainsi  has an idea what should be used? I don't see how to resolve this problem because either code is reported as unformatted or a checktyle warning is raised, no matter what I do.

catroid/src/org/catrobat/catroid/io/StorageHandler.java
```diff
@@ -57,103 +57,7 @@
...
-import org.catrobat.catroid.content.bricks.UserListBrick;
-import org.catrobat.catroid.content.bricks.UserScriptDefinitionBrick;
-import org.catrobat.catroid.content.bricks.UserScriptDefinitionBrickElement;
-import org.catrobat.catroid.content.bricks.UserVariableBrick;
-import org.catrobat.catroid.content.bricks.VibrationBrick;
-import org.catrobat.catroid.content.bricks.WaitBrick;
-import org.catrobat.catroid.content.bricks.WhenBrick;
-import org.catrobat.catroid.content.bricks.WhenNfcBrick;
-import org.catrobat.catroid.content.bricks.WhenStartedBrick;
+import org.catrobat.catroid.content.bricks.*;
```
p4p4: how to do the import?
formatter says 
    import org.catrobat.catroid.content.bricks.*;
checkstyle says
   don't do this

thmq:
There are several reasons why you should not use the import on demand (i.e. *). Although there is no difference in runtime performance the first and foremost, in my opinion, is that you cannot guarantee a clear dependency structure. Which means that if you refer to a class A which might be present in two or more sub packages your code will simply not compile. I am aware that in our structure this might not be the case at the moment, however, I think that it is best practice to prevent future bugs by thinking ahead and not taking these kinds of risks.

I have also read reasons for using *, such as the fact that it apparently simplifies refactoring. Anyway, I'd suggest you use the full list rather than the import on demand, but if anyone else has a good reason why we should use it please feel free to add your input. @robertpainsi @fwinkelbauer

p4p4:
Ok I understand your reasons. -> changed imports  back & rebased. 
Let's see if the FormatBot complains ;)

robertpainsi:
> I'd suggest you use the full list rather than the import on demand

I will create a pull request to disable wildcard imports in Android Studio asap.

thmq:
ok to test

robertpainsi:
@p4p4 please rebase your branch.

p4p4:
@robertpainsi  done

wslany:
Close https://jira.catrob.at/browse/CAT-1954

PR #1774, opened 2016-06-24
CAT-2015 HOTFIX 0.9.22 - Exception Handling

84n4n4:
Deal with unhandled exceptions and properly finish and restart app.


catroid/src/org/catrobat/catroid/utils/Utils.java
```diff
@@ -114,6 +116,30 @@ public void onClick(DialogInterface dialog, int which) {
 		return true;
 	}
 
+	public static boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
+		SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+		if (preferences.getBoolean(BaseExceptionHandler.RECOVERED_FROM_CRASH, false)) {
+
+			if (!(context instanceof MainMenuActivity)) {
+				context.finish();
```
robertpainsi: @84n4n4, why do we have to call finish? Does killing the app and it's process not restart the app in the MainMenuActivity?

cc @nishantt95

PR #1778, opened 2016-06-28
Disable import on demand (.*) in Android Studio

robertpainsi:
Disable import on demand (.*) in Android Studio

The Android Studio Formatter converts imports to a star import if the
configured limit has been exceeded. Since it has in the StorageHandler,
the limit has been highly increased and can be considered disabled now.

https://github.com/Catrobat/Catroid/pull/1770/#issuecomment-228997145


robertpainsi:
Thank you :heart:

PR #1780, opened 2016-06-29

jonny911:


thmq:
have you tried clicking on the pocket code icon in the main menu? if not please do. 😉

PR #1784, opened 2016-06-29
CAT-2034 Merge ApkGenerator in develop

Chypsylon:


thmq:
You should never commit the .gitignore file unless you have got a good reason for it. In your case you should remove it. Furthermore you should have a look at activity_main_menu_splashscreen.xml as you apparently renamed a resource file instead of adding a new .xml. And finally you should not have changed the lint and pmd config (you should not be adding random new lines in random files). If you have any questions or need assistance fixing these issues, just drop by so we can have a look at it together.

thmq:
This will be replaced by a version with a cleaner history.

PR #1786, opened 2016-06-30
CAT-2013 loading circle in webview

jonny911:


catroid/res/values/styles.xml
```diff
@@ -606,66 +606,89 @@
...
+    <item name="android:bottomDark">@color/transparent</item>
+    <item name="android:bottomMedium">@color/transparent</item>
+    <item name="android:centerBright">@color/transparent</item>
+    <item name="android:centerDark">@color/transparent</item>
+    <item name="android:centerMedium">@color/transparent</item>
+    <item name="android:fullBright">@color/transparent</item>
+    <item name="android:fullDark">@color/transparent</item>
+    <item name="android:topBright">@color/transparent</item>
+    <item name="android:topDark">@color/transparent</item>
+</style>
```
robertpainsi: Both style blocks aren't formatted correctly.

robertpainsi:
> Things to avoid when creating commits: Mixing whitespace changes with functional code changes.
> https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits

Please remove all unrelated changes like reformatting. A clean commit in this branch would look like `git diff @~ --ignore-all-space`.
You can easily fix this with git. Give it a try. You can find my solution [here](https://gist.github.com/robertpainsi/2137ce6ee16792aa3da621b3d8dfe728#gistcomment-1814968)
You still have to format the two style blocks mentioned above.

PR #1791, opened 2016-07-09
Cat-1949

p4p4:
allows the user to disable ("comment out") or enable a Brick
A disabled brick gets grayed out in order to make it distinguishable to enabled bricks.
Additionally,  disabled bricks are intended and separated to enabled bricks by a padding.
The enabled/disabled status of a brick gets written into the code.xml file

This PR includes a unit test.

[Static checks ok](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/252/)

*waiting for code-review* 
feedback appreciated


catroid/src/org/catrobat/catroid/content/Script.java
```diff
@@ -59,8 +61,12 @@ public Script() {
 
 	public void doCopy(Sprite copySprite, Script cloneScript) {
 		ArrayList<Brick> cloneBrickList = cloneScript.getBrickList();
+		cloneScript.commentedOut = commentedOut;
 		for (Brick brick : getBrickList()) {
 			Brick copiedBrick = brick.copyBrickForSprite(copySprite);
+			if (!(copiedBrick instanceof ScriptBrick)) {
+				copiedBrick.setCommentedOut(brick.isCommentedOut());
+			}
```
loibi93: Why is there no else case needed ?

catroid/src/org/catrobat/catroid/content/bricks/BrickBaseType.java
```diff
@@ -47,6 +51,89 @@
...
+		if (commentedOut) {
+			grayscaleBackground();
+		} else {
+			undoGrayscaleBackground();
+		}
+		doPadding();
+	}
+
+	private void grayscaleBackground() {
+
```
loibi93: remove newline

catroid/src/org/catrobat/catroid/content/bricks/BrickBaseType.java
```diff
@@ -47,6 +51,89 @@
...
+				ColorMatrixColorFilter filter = new ColorMatrixColorFilter(matrix);
+
+				background.setColorFilter(filter);
+			}
+		}
+	}
+
+	private void undoGrayscaleBackground() {
+
```
loibi93: remove newline

catroid/src/org/catrobat/catroid/content/Script.java
p4p4: Because the comment out status of a script Brick is already set in the copy of the Script (line 64). The brick then is created using that script.

ScriptBricks store the status in the script, whereas other (Standard-)bricks store it internally, die to the code.xml structure

catroid/src/org/catrobat/catroid/content/bricks/BrickBaseType.java
```diff
@@ -47,6 +51,89 @@
...
+			return;
+		}
+
+		int next = adapter.getBrickList().indexOf(this) + 1;
+		boolean hasNext = next < adapter.getBrickList().size();
+		boolean nextCommentStateDifferent = hasNext && (this.isCommentedOut() != adapter.getBrickList().get(next).isCommentedOut());
+		boolean nextIsScriptbrick = hasNext && adapter.getBrickList().get(next) instanceof ScriptBrick;
+
+		int paddingLeft = isCommentedOut() ? 75 : 0;
+		if (nextCommentStateDifferent && !nextIsScriptbrick) {
```
loibi93: maybe do the same as above like
`int paddingBottom = nextCommentStateDifferent && !nextIsScriptbrick ? 50 : 0;
view.setPadding(paddingLeft, view.getPaddingTop(), 0, paddingBottom);`

would save some lines ;)

catroid/src/org/catrobat/catroid/content/bricks/WhenNfcBrick.java
```diff
@@ -80,6 +80,10 @@ public WhenNfcBrick(WhenNfcScript script) {
 		this.oldSelectedNfcTag = null;
 		this.nfcTag = script.getNfcTag();
 		this.whenNfcScript = script;
+
+		if (script != null && script.isCommentedOut()) {
+			setCommentedOut(true);
+		}
```
loibi93: Why is the clone method not altered in order to set the cloned brick comment state ?

catroid/src/org/catrobat/catroid/content/bricks/WhenRaspiPinChangedBrick.java
```diff
@@ -55,6 +55,10 @@ public WhenRaspiPinChangedBrick(RaspiInterruptScript script) {
 		if (script != null) {
 			pinString = script.getPin();
 			eventString = script.getEventValue();
+
+			if (script.isCommentedOut()) {
+				setCommentedOut(true);
+			}
 		}
```
loibi93: Why is the clone method not altered in order to set the cloned brick comment state ?

catroid/src/org/catrobat/catroid/content/bricks/WhenStartedBrick.java
```diff
@@ -43,6 +43,10 @@
 
 	public WhenStartedBrick(Script script) {
 		this.script = script;
+
+		if (script != null && script.isCommentedOut()) {
+			setCommentedOut(true);
+		}
 	}
```
loibi93: Why is the clone method not altered in order to set the cloned brick comment state ?

catroid/src/org/catrobat/catroid/content/bricks/WhenTouchDownBrick.java
```diff
@@ -52,6 +52,10 @@ public WhenTouchDownBrick() {
 
 	public WhenTouchDownBrick(WhenTouchDownScript script) {
 		this.whenTouchDownScript = script;
+
+		if (script != null && script.isCommentedOut()) {
+			setCommentedOut(true);
+		}
 	}
```
loibi93: Why is the clone method not altered in order to set the cloned brick comment state ?

catroid/src/org/catrobat/catroid/physics/content/bricks/CollisionReceiverBrick.java
```diff
@@ -63,6 +63,10 @@ public CollisionReceiverBrick(String spriteName) {
 	public CollisionReceiverBrick(CollisionScript collisionScript) {
 		this.collisionScript = collisionScript;
 		this.selectedMessage = "";
+
+		if (collisionScript != null && collisionScript.isCommentedOut()) {
+			setCommentedOut(true);
+		}
 	}
```
loibi93: Why is the clone method not altered in order to set the cloned brick comment state ?

catroidTest/src/org/catrobat/catroid/test/content/bricks/CommentOutTest.java
```diff
@@ -0,0 +1,63 @@
...
+		Brick disabledBrick = new WaitBrick(1);
+		disabledBrick.setCommentedOut(true);
+		script.addBrick(disabledBrick);
+
+		script.run(sprite, sequence);
+
+		assertEquals("action of disable brick should not be in sequence:", sequence.getActions().size, 3);
+	}
+}
+
```
loibi93: Maybe you can write some more tests, for example, when you comment out nesting bricks, do all Bricks get commented out ? Or the same for a Script Brick.

catroid/src/org/catrobat/catroid/content/Script.java
loibi93: Ok, perfect ;)

catroid/src/org/catrobat/catroid/content/bricks/WhenNfcBrick.java
p4p4: instead of copying the comment out status in the clone() method of each and every Brick I only copied it at places where clone() was called (...) and only when copying the comment-out status made sense. This way I had to change less files. Still, I am a bit dissatisfied because if someone adds a new feature where brick-cloning is used has to remember to copy the value as well. On the other hand, if someone adds a new Brick, she also has to think of copying the boolean in the clone() method.
Unfortunately I can't think of a good solution using inheritance either because clone() doesn't modify any member variables, therefore copying via super.clone() doesn't work either.
What do you think, should I leave it as it is, should I copy the boolean in every brick?
That's one thing I've already been pondering about while developing the feature.

catroid/src/org/catrobat/catroid/content/bricks/BrickBaseType.java
p4p4: great idea :+1:

catroidTest/src/org/catrobat/catroid/test/content/bricks/CommentOutTest.java
p4p4: that's a good idea! more test cases coming up! :clap: 

greetings from :it:

catroid/src/org/catrobat/catroid/content/bricks/WhenNfcBrick.java
loibi93: You could copy it in the BrickBaseType. Would there be any reason, why you shouldn't copy it for every Brick ? Even if the Brick does not use it ? I think this should not be a problem, since every Brick has the Member and if it never gets used it does not matter what the value is set to anyway. But please correct me if I am overseeing something ;)

p4p4: Agreed, it can always be copied, that's no problem. But still I'm not sure how this can be achieved with the least effort. 

in a class SpecialBrick, we override the method
Brick clone() 
which returns a SpecialBrick that is derived from BrickBasetype.
If i want to do the copying of the variable, I'd have to create the following member in BrickBaseType:
Brick copy = null; 
and the clone() method in BrickBaseType looks as follows:
copy.isCommentedOut = isCommentedOut

and in the  SpecialBrick clone() method looks as follows:
{
    copy = new SpecialBrick();
    // or: copyBaseValues()
    return super.clone(); // now we can copy the value (and maybe other values in the future)
} 

1) I'm not sure if that's a good design
2) I still have to modify every Brick to call a method from BrickBaseType. That would only pay off if multiple values have to be copied (maybe in future)

Or how would you suggest to solve this? Thanks for any hint! :+1:

loibi93: Ahhh, now I see your Problem ;) I think you are right that this cannot be done without editing every Brick class. This is really a tough one ;D Maybe you could have a method in BrickBaseType called `setMembers(BrickBaseType copyBrick)` where you set all Members the same as in copyBrick. Then in every "SpecialBrick" you call super.clone() and after that clonedBrick.setMembers(this). You still need to change every Brick class, but whenever someone adds a new Member in BrickBaseType, he/she just needs to add one line in the setMember method.
I am not perfectly sure, but I think this should be the easiest way with the smallest amount of work, especially for the future while preserving a quite easy to read and clean code. But it would be very nice if someone has a better idea ;)

catroid/src/org/catrobat/catroid/content/bricks/WhenRaspiPinChangedBrick.java
p4p4: because the comment out status of a Brick is copied only when needed, e.g. when packing a Script to the Backpack. With that I don't have to modify the clone() method of each brick.

PR #1793, opened 2016-07-12
CAT-2042 Select/deselect not shown properly

Achilles-96:
Fixes [CAT-2042](https://jira.catrob.at/browse/CAT-2042). This fixes only for programs and objects. For other items in issue, this is implemented in a slightly different way i.e `select all` is hidden after all items are selected and made visible if any of it is unselected. I just wanted to make sure if we want to override that.


thmq:
ok to test

thmq:
This has already been implemented by one of our team members and will be merged soon. I am sorry about you having done all the work in vain due to this misunderstanding.

Achilles-96:
I think it will be better if the person working on it assigns it to himself/herself in jira.

PR #1799, opened 2016-07-13
CAT-REF Refactor tagging code from PR#1797

robertpainsi:
* Use List<String> for tags instead of a single String or String[].
* Parse tags from a JSONObject.
* Fix endless loop in UploadDialogTest.
* Change ServerCalls.useTestUrl value back to false.

Additional but unimportant cleanup
* Remove unnecessary logs.
* Add log TAGs.
* Use primitives instead of wrappers.

https://github.com/Catrobat/Catroid/pull/1797
$

If a single commit isn't k, let me know. :smile_cat:
I have _not_ tested the code.

@LauraBebek, please review the changes.

### DIFF
--- a/catroid/src/org/catrobat/catroid/web/ServerCalls.java
+++ b/catroid/src/org/catrobat/catroid/web/ServerCalls.java
@@ -63,7 +63,6 @@ import okio.Okio;
 public final class ServerCalls {
 
 	public static final String BASE_URL_TEST_HTTPS = "https://catroid-test.catrob.at/pocketcode/";
-	//"https://web-test.catrob.at/pocketcode/";
 	public static final String TEST_FILE_UPLOAD_URL_HTTP = BASE_URL_TEST_HTTPS + "api/upload/upload.json";
 	public static final String TEST_FILE_TAG_URL_HTTP = BASE_URL_TEST_HTTPS + "api/tags/getTags.json";
 	public static final int TOKEN_LENGTH = 32;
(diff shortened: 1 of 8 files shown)


catroid/src/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -63,7 +63,6 @@
 public final class ServerCalls {
 
 	public static final String BASE_URL_TEST_HTTPS = "https://catroid-test.catrob.at/pocketcode/";
-	//"https://web-test.catrob.at/pocketcode/";
```
robertpainsi: Any reason (jenkins?) this comment has been added previously?

catroid/src/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -152,7 +151,7 @@
 
 	private static final ServerCalls INSTANCE = new ServerCalls();
 
-	public static boolean useTestUrl = true;
+	public static boolean useTestUrl = false;
```
robertpainsi: https://github.com/Catrobat/Catroid/pull/1797/files?diff=unified#diff-8e10fb43a2b53e349b11d6c768b1a65fL153
Other code also relies on this flag. I don't know if false or true is the correct value, so I changed it to highlight it. :wink:
@LauraBebek, please take a look at this.

robertpainsi:
ok to test

PR #1800, opened 2016-07-15
CAT-1717 Integrate Scenes

loibi93:
With Scenes you are able to have "Programs in Programs". It is another layer between the project layer and the object layer. If there is only one Scene (by default), the project is presented to the user just like before.
Features:
- Start Scene and Continue Scene Bricks
- Copy, Delete, Rename and Backpack Scenes (soon also DragNDrop)
- global Variables are available for all Scenes


loibi93:
ok to test

thmq:
ok to test

catroid/src/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -268,20 +278,61 @@ void menuPause() {
...
+		scene.firstStart = true;
+		create();
+	}
+
 	public void reloadProject(StageDialog stageDialog) {
 		if (reloadProject) {
 			return;
 		}
 		this.stageDialog = stageDialog;
+		if (!ProjectManager.getInstance().getCurrentProject().getDefaultScene().getName().equals(scene.getName())) {
```
loibi93: use project not ProjectManager.....

catroid/src/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -713,4 +765,83 @@ public Stage getStage() {
...
+		backup.sprites = sprites;
+		backup.axesOn = axesOn;
+		backup.deltaActionTimeDivisor = deltaActionTimeDivisor;
+		backup.cameraRunning = CameraManager.getInstance().isCameraActive();
+		if (backup.cameraRunning) {
+			CameraManager.getInstance().pauseForScene();
+			//CameraManager.getInstance().releaseCamera();
+		}
+		return backup;
+	}
```
loibi93: Remove this line

catroid/src/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -713,4 +765,83 @@ public Stage getStage() {
...
+		camera = backup.camera;
+		batch = backup.batch;
+		font = backup.font;
+		passepartout = backup.passepartout;
+		viewPort = backup.viewPort;
+		sprites = backup.sprites;
+		axesOn = backup.axesOn;
+		deltaActionTimeDivisor = backup.deltaActionTimeDivisor;
+		if (backup.cameraRunning) {
+			//CameraManager.getInstance().startCamera();
```
loibi93: Remove line

loibi93:
I ran all tests locally with two emulators and two devices for both the develop and this branch. The results are saved in two folders. I then compared the results, and wrote down the tests, which failed on this branch, but not on the develop branch. I reran each of these tests. Some did just randomly fail, and worked fine in the new run, and some were actually broken. Now those are fixed too. Some test classes crashed PC on this branch, but also on the develop for the same test. I know this is not perfect, but it is better than nothing.
[Scene Tests.zip](https://github.com/Catrobat/Catroid/files/383272/Scene.Tests.zip)

thmq:
ok to test

catroid/src/org/catrobat/catroid/content/Scene.java
```diff
@@ -0,0 +1,444 @@
...
+
+	public int getOriginalWidth() {
+		return originalWidth;
+	}
+
+	public int getOriginalHeight() {
+		return originalHeight;
+	}
+
+	public List<Sprite> getSpriteList() {
```
p4p4: we should not forget to adapt this for my clone feature ( PR #1823 ) after this or my feature has been merged ;)

loibi93: For sure, but I think there will be a lot of PRs that need to change something ;)

PR #1804, opened 2016-07-19
CAT-1936 CurrentTime in FormulaEditor added

robertriedl:


rsamer:
@robertriedl no further complaints from my side

fwinkelbauer:
Most of the code in `catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java` is pure copy/paste. There's already a `SensorDate` abstract class, which would be a perfect place for all that duplication!

Also: what's the point of all those singletons? Please have a look at the Confluence blog to read more about avoiding them :smiley:

catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java
```diff
@@ -0,0 +1,410 @@
...
+package org.catrobat.catroid.formulaeditor;
+
+import android.os.Handler;
+
+import java.util.ArrayList;
+import java.util.Calendar;
+
+public abstract class SensorDate {
+	public static final int UPDATE_INTERVAL = 50;
+	public float lastValue = 0f;
```
fwinkelbauer: You should use the `protected` keyword instead of `public` or no keyword (`default`) in all these variables:

- lastValue
- handler
- calendar

catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java
```diff
@@ -0,0 +1,410 @@
...
+	public SensorDate() {
+	}
+}
+
+class SensorYear extends SensorDate {
+	private static SensorYear instance = null;
+	private ArrayList<SensorCustomEventListener> listenerList = new ArrayList<SensorCustomEventListener>();
+
+	private SensorYear() {
+		super();
```
fwinkelbauer: There's already an implicit call to `super()` in a constructor. You can simply remove it in all child classes

catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java
```diff
@@ -0,0 +1,410 @@
...
+	private SensorMonth() {
+		super();
+		handler = new Handler();
+	}
+
+	Runnable statusChecker = new Runnable() {
+		@Override
+		public void run() {
+			calendar = Calendar.getInstance();
+			float[] date = new float[1];
```
fwinkelbauer: Why do you need an array of size one here?

catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java
```diff
@@ -0,0 +1,410 @@
...
+	public Handler handler;
+	Calendar calendar = null;
+
+	public SensorDate() {
+	}
+}
+
+class SensorYear extends SensorDate {
+	private static SensorYear instance = null;
+	private ArrayList<SensorCustomEventListener> listenerList = new ArrayList<SensorCustomEventListener>();
```
fwinkelbauer: You could use a [thread safe collection](https://developer.android.com/reference/java/util/concurrent/package-summary.html). This way your methods don't have to be `synchronized`.

catroid/src/org/catrobat/catroid/formulaeditor/SensorDate.java
```diff
@@ -0,0 +1,410 @@
...
+	public synchronized boolean registerListener(SensorCustomEventListener listener) {
+		if (listenerList.contains(listener)) {
+			return true;
+		}
+		listenerList.add(listener);
+		statusChecker.run();
+		return true;
+	}
+
+	public synchronized void unregisterListener(SensorCustomEventListener listener) {
```
fwinkelbauer: Guard clauses are in general more readable:

```java
if (!listenerList.contains(listener)) {
  return;
}

listenerList.remove(listener);
// ...
```

fwinkelbauer:
Some meaningful tests would be great too. The current one only checks a few lines from `FormulaElement`.

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -337,6 +356,20 @@ public static Double getSensorValue(Sensors sensor) {
 				}
 			case LOUDNESS:
 				return Double.valueOf(instance.loudness);
+			case DATE_YEAR:
+				return Double.valueOf(instance.dateYear);
```
p4p4:
```java
return Double.valueOf(Calendar.getInstance().get(Calendar.YEAR));
```
should suffice

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -337,6 +356,20 @@ public static Double getSensorValue(Sensors sensor) {
...
+			case DATE_MONTH:
+				return Double.valueOf(instance.dateMonth);
```
p4p4:
```java
return Double.valueOf(Calendar.getInstance().get(Calendar.MONTH));
```
should suffice

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -337,6 +356,20 @@ public static Double getSensorValue(Sensors sensor) {
...
+			case DATE_DAY:
+				return Double.valueOf(instance.dateDay);
```
p4p4:
```java
return Double.valueOf(Calendar.getInstance().get(Calendar.DAY_OF_MONTH));
```
should suffice

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -337,6 +356,20 @@ public static Double getSensorValue(Sensors sensor) {
...
+			case DATE_WEEKDAY:
+				return Double.valueOf(instance.dateWeekday);
```
p4p4: here you could use 
```java
Double.valueOf(Calendar.getInstance().get(Calendar.DAY_OF_WEEK ));
```
returns 1 for sunday, 2 for monday, ...
you might want to  adapt this if you want to start the week with a monday (monday = 1, sunday = 7)

catroid/src/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -337,6 +356,20 @@ public static Double getSensorValue(Sensors sensor) {
...
+			case TIME_HOUR:
+				return Double.valueOf(instance.timeHour);
```
p4p4: etc. see definitions in `Calendar.java`

catroid/src/org/catrobat/catroid/content/bricks/Brick.java
```diff
@@ -83,6 +83,7 @@
 	int SENSOR_COMPASS = 0x8000;
 	int NFC_ADAPTER = 0x10000;
 	int VIDEO = 0x20000;
+	int DATE = 0x40000;
```
p4p4: this should be able to implement without an additional resource

p4p4:
There has already been some lower-level feedback (formatting, code style etc.), so I tried to find out how to simplify the implementation:


- I am not sure, but do you really need a Brick-Resource for Date and Time? both should be available at any time without having to do something in the Pre-Stage activity.
- in the `Sensorhandler.java` class, you could directly use the `Calendar` class, [as I wrote in other in-line-comments](https://github.com/Catrobat/Catroid/pull/1804#discussion-diff-73348569).
     - with that you should not need any: Listeners, Resources, Helper-Classes etc

Please correct me if I am wrong, but as far as I can tell you could simplify a lot and make it less complicated.
If you have further questions feel free to ask. :)

PR #1810, opened 2016-07-23
CAT-1937 Ask Brick

p4p4:
- ask a user defined question
- answer is entered in a dialog by the user
- program is paused while user enters answer
- program is resumed as soon as an answer is provided by the user


catroid/src/org/catrobat/catroid/content/bricks/AskBrick.java
```diff
@@ -0,0 +1,266 @@
...
+	@Override
+	public int getRequiredResources() {
+		return getFormulaWithBrickField(BrickField.ASK_QUESTION).getRequiredResources();
+	}
+
+	@Override
+	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
+		sequence.addAction(sprite.getActionFactory().createAskAction(sprite,
+				getFormulaWithBrickField(BrickField.ASK_QUESTION), userVariable));
+		return null;
```
fwinkelbauer: Just curious: would returning an empty list make sense too?

p4p4: Yes, the result would be the same. This would actually be better, according to the [How To Do A Code Review](https://confluence.catrob.at/display/KNOWHOW/How+to+do+Code+Reviews) article ;) (and [others](http://www.javapractices.com/topic/TopicAction.do?Id=59) ).

This method is called in Script::run(), where a null check is performed, but returning an empty list also yields the same result (namely: nothing happens). 

If all of us would return an empty list instead of  null, this check wouldn't be necessary

`List<SequenceAction> actions = brickList.get(i).addActionToSequence(sprite,
					sequenceList.get(sequenceList.size() - 1));
if (actions != null) {
	for (SequenceAction action : actions) { ... }
}`

Because of that I'll change that to an empty list! Thanks for the feedback

p4p4: I suggest returning `Collections.emptyList()` as the preferred way if the returned list is not modified

catroid/src/org/catrobat/catroid/content/bricks/AskBrick.java
```diff
@@ -0,0 +1,267 @@
...
+	@Override
+	public int getRequiredResources() {
+		return getFormulaWithBrickField(BrickField.ASK_QUESTION).getRequiredResources();
+	}
+
+	@Override
+	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
+		sequence.addAction(sprite.getActionFactory().createAskAction(sprite,
+				getFormulaWithBrickField(BrickField.ASK_QUESTION), userVariable));
+		return Collections.emptyList();
```
p4p4: @fwinkelbauer asked: 
> Just curious: would returning an empty list make sense too?

Yes, the result would be the same. This would actually be better, according to the [How To Do A Code Review](https://confluence.catrob.at/display/KNOWHOW/How+to+do+Code+Reviews) article ;) (and [others](http://www.javapractices.com/topic/TopicAction.do?Id=59) ).

This method is called in Script::run(), where a null check is performed, but returning an empty list also yields the same result (namely: nothing happens). 

```java
List<SequenceAction> actions = brickList.get(i).addActionToSequence(sprite,
					sequenceList.get(sequenceList.size() - 1));
if (actions != null) {
	for (SequenceAction action : actions) { ... }
}
```

If all of us would return an empty list instead of  null, this check wouldn't be necessary
Because of that I'll change that to an empty list! Thanks for the feedback

I suggest returning `Collections.emptyList()` as the preferred way if the returned list is not modified

that's the implementation of `emptyList()`:
```java
public static final <T> List<T> emptyList() {
    return (List<T>) EMPTY_LIST;
}
```
because of the immutability it should have a slightly better performance (CPU & memory).

Similarily `Collections.emptySet()` and `Collections.emptyMap()` should be used for sets and maps respectively.

fwinkelbauer: Thanks!

catroid/src/org/catrobat/catroid/content/actions/AskAction.java
```diff
@@ -0,0 +1,111 @@
...
+	private Formula questionFormula;
+	private UserVariable answerVariable;
+
+	private boolean questionAsked = false;
+	private boolean answerReceived = false;
+
+	private void askQuestion() {
+		if (questionFormula == null) {
+			Log.e(getClass().getSimpleName(), "formula interpretation in ask brick failed");
+			return;
```
loibi93: maybe do the same as below with the empty string ? Just for consistency ;)

catroid/src/org/catrobat/catroid/content/bricks/AskBrick.java
```diff
@@ -0,0 +1,267 @@
...
+
+		DataAdapter dataAdapter = ProjectManager.getInstance().getCurrentProject().getDataContainer()
+				.createDataAdapter(context, currentBrick, ProjectManager.getInstance().getCurrentSprite());
+		UserVariableAdapterWrapper userVariableAdapterWrapper = new UserVariableAdapterWrapper(context,
+				dataAdapter);
+		userVariableAdapterWrapper.setItemLayout(android.R.layout.simple_spinner_item, android.R.id.text1);
+
+		variableSpinner.setAdapter(userVariableAdapterWrapper);
+
+		if (!(checkbox.getVisibility() == View.VISIBLE)) {
```
loibi93: More readable:
`if (checkbox.getVisibility() == View.VISIBLE) {.......`

catroid/src/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -115,6 +130,55 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+				if (message.what == ASK_MESSAGE) {
+					showDialog((String) params.get(1), (AskAction) params.get(0));
+				}
+			}
+		};
+	}
+
+	private void showDialog(String question, final AskAction askAction) {
+		pause();
```
loibi93: Unfortunately this pause does not pause everything (for example vibration is not stopped. For whatever reason only StageActivity.pause() stops everything. I am not sure if this is intended behavior, so changing the StageListeners pause method may break other functionality.

catroidTest/src/org/catrobat/catroid/test/content/actions/AskActionTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+	public void testAskAndCheckAnswer() {
+		AskAction action = (AskAction) testSprite.getActionFactory().createAskAction(testSprite, new Formula(ASK_QUESTION),
+				userVariableForAnswer);
+		action.act(1f);
+		action.setAnswerText(ASK_ANSWER);
+
+		assertEquals("answer incorrect", ASK_ANSWER, userVariableForAnswer.getValue().toString());
+	}
+}
```
loibi93: Maybe you can also test, if the stage is paused when a question is asked ;)

catroid/src/org/catrobat/catroid/content/actions/AskAction.java
p4p4: thank you! changed. :+1:

catroid/src/org/catrobat/catroid/content/bricks/AskBrick.java
p4p4: thank you! changed. :+1:

catroid/src/org/catrobat/catroid/stage/StageActivity.java
p4p4: you may have confused StageActivity and StageListener. This is the StageActivity class, so StageActivity.pause() is called, where vibration is stopped as well ;)
I tested it as well: Both Vibration and flash are paused during the dialogue is shown and resumed afterwards. Thank you though for reviewing my code in such a detail! :)

loibi93: Ah, stupid me ;)
You are right, its perfectly fine !

catroidTest/src/org/catrobat/catroid/test/content/actions/AskActionTest.java
loibi93: Ok, seems legit ;)

PR #1816, opened 2016-07-27
CAT-2059 Pocket Code crashes when Looks are changed

stefanreichenauer:


catroid/src/org/catrobat/catroid/ui/adapter/LookAdapter.java
```diff
@@ -59,7 +60,14 @@ public long getItemId(int position) {
...
 		LookData item = getItem(position);
-		return idMap.get(item);
+
+		for (Map.Entry<LookData, Integer> e : idMap.entrySet()) {
+			if (item.equals(e.getKey())) {
+				return e.getValue();
+			}
+		}
+
+		return INVALID_ID;
```
robertpainsi: Please explain (to me :wink:)  what exactly was the problem in the code and how this change fixes it?

robertpainsi: Hmm, the implementation of LookData.hashCode is wrong, right?
https://github.com/Catrobat/Catroid/blob/dba80632feb54a4ef9dada08c46e94df57d518dd/catroid/src/org/catrobat/catroid/common/LookData.java#L84-L87

stefanreichenauer: It seems like the HashMap throws a NullPointerException, even if the element is in the map (so it is equal by the definition of LookData.equal() ). So I check the entries myself, and it seems to work.

robertpainsi: Please fix `LookData.hashCode` and `LookData.equals` instead. Both methods should also work if name or fileName are null.

---

The hashCode calculation is also wrong (`+ super.hashCode()`). But I'm not sure if we should fix that as well before the release. There may be side effects I'm not aware of. @thmq, what do you think?

robertpainsi: Hmm, maybe there shouldn't be a LookData with a null name or fileName in the first place?

stefanreichenauer: LookDatas Name or FileName does not become null. HashMap.get() returns just null, because the hash values are different.

stefanreichenauer: I've changed it, so that it works with the HashMap methods. It appears that if a file was changed, the LookAdapter.idMap was just not updated.

robertpainsi: Ah, that's the problem. Thanks for the update! :smile_cat:
I created a ticket for the hashCode. https://jira.catrob.at/browse/CAT-2068

PR #1822, opened 2016-08-03
CAT-1968 "When Background switches to", "Switch background and wait"

p4p4:
- brick "When Background switches to": Runs a script when background
  switches to a certain look (see Scratch)
- brick: "Switch background to _"
    - now available in all Sprites
- brick: "Switch background to _ and wait"
  waits until all "When Background _" scripts are completed

Checkstyle OK:
- [jenkins CatroidCheck results](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/297/)

Tests OK:
- [jenkins package 'test' results](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/3637/)

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/org/catrobat/catroid/content/Look.java
@@ -254,6 +254,11 @@ public class Look extends Image {
 	public void setLookData(LookData lookData) {
 		this.lookData = lookData;
 		imageChanged = true;
+
+		boolean isBackgroundLook = getZIndex() == 0;
+		if (isBackgroundLook) {
+			BackgroundWaitHandler.fireBackgroundChangedEvent(lookData);
+		}
 	}
 
 	public boolean getAllActionsAreFinished() {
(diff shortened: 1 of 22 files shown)


catroid/src/org/catrobat/catroid/content/Look.java
```diff
@@ -254,6 +255,10 @@ public LookData getLookData() {
 	public void setLookData(LookData lookData) {
 		this.lookData = lookData;
 		imageChanged = true;
+
+		if (this.equals(ProjectManager.getInstance().getCurrentProject().getSpriteList().get(0).look)) {
```
loibi93: I am not hundred Percent sure, but maybe you can check this with `if (this.getZindex() == 0)` (not sure what method, but I think you get the point). This should normally work, but you should first check if the MoveBackNSteps Brick would break that functionality. Please ask if unclear ;)

p4p4: you are right! I changed it. That way, the functionality is not susceptible to changes in the ``Project`` class. (Which are likely to happen when the Scenes feature has been merged)

loibi93: Awesome, cool that this is working :)

PR #1823, opened 2016-08-03
CAT-1938 Clone Sprites Feature

p4p4:
- create a clone of a Sprite (= Object) via a special Brick
- create a special "Clone" brick that clones the Object in which it is
    - all Scripts, Sounds, Variables etc are copied
- add a "When i start as a clone" brick
    - is a script-brick that is only executed when Object is cloned
    - "When started" ScriptBricks are NOT executed for the clone
- add "Delete this clone" brick
    - when executed, the cloned object containing this brick
      is deleted
- global variables are shared, local (=per object) variables are copied
  for clones.

[tests ok](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/3646/) and [checkstyle ok](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/300/)

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/Project.java
+++ b/catroid/src/org/catrobat/catroid/content/Project.java
@@ -24,6 +24,7 @@ package org.catrobat.catroid.content;
 
 import android.content.Context;
 import android.os.Build;
+import android.util.Log;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 
@@ -41,6 +42,7 @@ import org.catrobat.catroid.formulaeditor.UserList;
 import org.catrobat.catroid.formulaeditor.UserVariable;
 import org.catrobat.catroid.physics.PhysicsWorld;
 import org.catrobat.catroid.physics.content.ActionPhysicsFactory;
+import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.ui.SettingsActivity;
 import org.catrobat.catroid.utils.Utils;
 
@@ -116,6 +118,10 @@ public class Project implements Serializable {
 	}
 
 	public synchronized void addSprite(Sprite sprite) {
+		if (sprite.isClone) {
+			Log.e("Project", "Do NOT add clones to the projectmanager!");
+			return;
+		}
 		if (spriteList.contains(sprite)) {
 			return;
 		}
@@ -128,6 +134,14 @@ public class Project implements Serializable {
 
 	public List<Sprite> getSpriteList() {
 		return spriteList;
+	}
+
+	public List<Sprite> getSpriteListWithClones() {
+		if (StageActivity.stageListener != null) {
+			return StageActivity.stageListener.getSpritesFromStage();
+		} else { // e.g. for ActionTests there is no Stage, only use sprites from Project
+			return spriteList;
+		}
 	}
 
 	public void setName(String name) {
--- a/catroid/src/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/org/catrobat/catroid/content/Sprite.java
@@ -45,6 +45,7 @@ import org.catrobat.catroid.formulaeditor.DataContainer;
 import org.catrobat.catroid.formulaeditor.UserVariable;
 import org.catrobat.catroid.physics.PhysicsLook;
 import org.catrobat.catroid.physics.PhysicsWorld;
+import org.catrobat.catroid.stage.StageActivity;
 
 import java.io.Serializable;
 import java.util.ArrayList;
@@ -66,6 +67,7 @@ public class Sprite implements Serializable, Cloneable {
 	private List<UserBrick> userBricks = new ArrayList<>();
 	private List<NfcTagData> nfcTagList = new ArrayList<>();
 	private transient ActionFactory actionFactory = new ActionFactory();
+	public transient boolean isClone = false;
 
 	public Sprite(String name) {
 		this.name = name;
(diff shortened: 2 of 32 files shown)


catroid/src/org/catrobat/catroid/content/Project.java
```diff
@@ -130,6 +136,14 @@ public synchronized boolean removeSprite(Sprite sprite) {
 		return spriteList;
 	}
 
+	public List<Sprite> getSpriteListWithClones() {
+		if (StageActivity.stageListener != null) {
+			return StageActivity.stageListener.getSpritesFromStage();
+		} else { // e.g. for ActionTests there is no Stage, only use sprites from Project
```
loibi93: Also if you never started the stage I assume ?

catroid/src/org/catrobat/catroid/content/Project.java
```diff
@@ -405,6 +419,17 @@ public Sprite getSpriteBySpriteName(Sprite searchedSprite) {
 		return spriteBySpriteName;
 	}
 
+	public Sprite getSpriteBySpriteName(String searchedSpriteName) {
```
loibi93: Is it not possible to use `getSpriteBySpriteName(Sprite sprite)` ?

catroid/src/org/catrobat/catroid/content/Sprite.java
```diff
@@ -302,6 +304,32 @@ public Sprite clone() {
 		return cloneSprite;
 	}
 
+	public Sprite cloneForCloneBrick() {
+		final Sprite cloneSprite = new Sprite();
+
+		cloneSprite.setName(this.getName() + "-c" + StageActivity.getAndIncrementNumberOfClonedSprites());
+		cloneSprite.isClone = true;
+		cloneSprite.isBackpackObject = false;
```
loibi93: delete this line, this one is false by default ;)

catroid/src/org/catrobat/catroid/content/Sprite.java
```diff
@@ -302,6 +304,32 @@ public Sprite clone() {
...
+
+		//cloneSprite.lookList = this.lookList;
+		cloneSprite.soundList = this.soundList;
```
loibi93: Why do you clone the looks but not sounds ?

catroid/src/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -216,6 +219,49 @@ public void create() {
 		}
 	}
 
+	public void cloneSpriteAndAddToStage(Sprite cloneMe) {
```
loibi93: Not sure, but maybe synchronize this method, since the stage is also accessing the sprite list !

catroid/src/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -575,6 +621,7 @@ public void dispose() {
 		font.dispose();
 		axes.dispose();
 		disposeTextures();
+		disposeClonedSprites();
```
loibi93: Is this called when you leave the stage ? Just curious, because the cloned Sprites need to be deleted when we return from stage !?

catroid/src/org/catrobat/catroid/content/Project.java
p4p4: correct. This method should never be called if there is no stage. If somebody calls is anyway (for some reason or by mistake) he gets the standard sprite-list, since there can't be any clones without a stage.

catroid/src/org/catrobat/catroid/content/Project.java
p4p4: Thank you for the hint. I see that I don't even use the method, so I removed it. :smiley_cat:

catroid/src/org/catrobat/catroid/content/Sprite.java
p4p4: done :+1:

catroid/src/org/catrobat/catroid/content/Sprite.java
p4p4: I thought that I can reuse the sounds since the should never be changed, whereas Looks might be changed by some Bricks (color effects like brightness, alpha etc). Additionally, looks are added to the stage as actors, and I was wondering if adding the identical look multiple times might lead to problems, especially when I call ``sprite.look.remove()`` I only want to remove the copy.

But I did not find a reason why I should copy sounds yet, as far as I can see we would only waste memory by loading the same soundFile into memory for each clone.  (Cloning Look objects is not that costly because there is no deep copy of the LookData, only the reference is copied).

But please correct me if I'm wrong. :smile:

catroid/src/org/catrobat/catroid/stage/StageListener.java
p4p4: correct, it's called when the stage is destroyed.

catroid/src/org/catrobat/catroid/content/Sprite.java
loibi93: Ahh I see. You are partially right, the look needs to be cloned, but cloning lookDatas is not needed. But as you said, this is not costly, so I would say you can leave it that way since otherwise you would need either another method or more code which is no good ;)

catroid/src/org/catrobat/catroid/stage/StageListener.java
loibi93: Ok perfect :)

PR #1824, opened 2016-08-03
CAT-1226 fixed upload bug

jonny911:
Uploading a project where the project name contains a restricted character is now prohibited.


robertpainsi:
https://share.catrob.at/pocketcode/program/12231
Project name: `XXX:YYY`
Magic!!!

@Gorphi already fixed the problem for project names with reserved characters of the android file system. [UtilFile.java#L312-L344](https://github.com/Catrobat/Catroid/blob/dba80632feb54a4ef9dada08c46e94df57d518dd/catroid/src/org/catrobat/catroid/utils/UtilFile.java#L312-L344)

Here is the code I used to fix [CAT-1226](https://jira.catrob.at/browse/CAT-1226). Uploading works and also the size is displayed correctly.
https://github.com/Catrobat/Catroid/compare/develop...robertpainsi:upload-project-with-fs-reserved-characters

@jonny911 Please test it and let me know if there is still a problem :grinning:

robertpainsi:
Thanks for updating the PR! :bow:

PR #1825, opened 2016-08-04
Fixes DragNDrop Multitouch and Rename Bug

loibi93:
Multitouch bug was a one line fix, which just prevents multitouching for the ListView.
The rename option was maybe lost during a rebase or something and is now back.


thmq:
ok to test

PR #1827, opened 2016-08-05
CAT-1969 When condition becomes true brick

p4p4:
- script is executed when condition becomes true
- to be executed again, the condition has first to be false
  and then become true again
- did some refactoring: ScriptBrick is now an interface instead of an
  abstract class

this includes some tests.

[checkstyle ok](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/303/) and [tests ok](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/3715/)

### DIFF
--- a/catroid/src/org/catrobat/catroid/content/bricks/WhenConditionBrick.java
+++ b/catroid/src/org/catrobat/catroid/content/bricks/WhenConditionBrick.java
@@ -1,0 +1,168 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2016 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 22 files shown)


catroid/src/org/catrobat/catroid/content/bricks/WhenConditionBrick.java
```diff
@@ -0,0 +1,165 @@
...
+			script = new WhenConditionScript(this);
+			formulaMap = script.getFormulaMap();
+		}
+		return script;
+	}
+
+	@Override
+	public Brick clone() {
+		// TODO clone script
+		return new WhenConditionBrick(script);
```
sjaindl: Not absolutely sure, but I think it would be better to take null as parameter here and to override copyScriptForSprite(..) and set a reference/clone there (same as in the other ScriptBricks). 
Please check if copying of scripts/sprites, backpacking/unpacking works as expected

catroid/src/org/catrobat/catroid/content/bricks/WhenConditionBrick.java
```diff
@@ -0,0 +1,168 @@
...
+		if (script == null) {
+			script = new WhenConditionScript(this);
+			formulaMap = script.getFormulaMap();
+		}
+		return script;
+	}
+
+	@Override
+	public Brick clone() {
+		return new WhenConditionBrick();
```
p4p4: in the old diff, @sjaindl commented:

> Not absolutely sure, but I think it would be better to take null as parameter here and to override copyBrickForSprite(..) and set a reference/clone there (same as in the other ScriptBricks). Please check if copying of scripts/sprites, backpacking/unpacking works as expected

thank you, you are right. 

Since this is a ``ScriptBrick`` i do the copying in ``copyScriptForSprite(..)`` now. (copy*Brick*ForSprite seems to be useless for ScriptBricks, the ScriptBricks are typically generated out of the Script).

I tested copying and backpacking, both work :smiley:

PR #1831, opened 2016-08-08
CAT-1838 Separate events from other control bricks

robertriedl:


catroid/res/values/styles.xml
```diff
@@ -443,6 +443,23 @@
...
     </style >
 
+    <style name="BrickContainer.Event" ></style >
+
+    <style name="BrickContainer.Event.Small" >
+        <item name="android:minHeight" >@dimen/brick_height_small</item >
+        <item name="android:background" >@drawable/brick_brown_1h_when_9</item >
+    </style >
+
+    <style name="BrickContainer.Event.MediumWhen9" >
```
loibi93: Why MediumWhen**9** ?

robertriedl: because of the file-names ;)

loibi93: The 9 stands for 9 Patch Graphik, no need to include it in the style name, this is just confusing ;D

robertriedl: just skimmed over it and didn't see that this isn't the file name xD

PR #1832, opened 2016-08-08
CAT-1925 Look/Background Number/Name added to Formula-Editor

robertriedl:


p4p4:
looks good :+1:

catroid/src/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -78,6 +80,12 @@
 			R.string.formula_editor_object_x_velocity, R.string.formula_editor_object_y_velocity,
 			R.string.formula_editor_object_angular_velocity };
 
+	private static final int[] OBJECT_ITEMS_LOOK = { R.string.formula_editor_object_look_number, R.string
+			.formula_editor_object_look_name };
+
+	private static final int[] OBJECT_ITEMS_BACKGROUND = { R.string.formula_editor_object_background_number, R.string
+			.formula_editor_object_background_name };
```
robertpainsi: Please seperate both items in the array at a reasonable position :wink:

catroid/src/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -739,6 +745,18 @@ private Object interpretObjectSensor(Sensors sensor, Sprite sprite) {
...
+				returnValue = 1.0d + ((lookData != null) ? lookDataList.indexOf(lookData) : 0);
+				break;
+			case OBJECT_LOOK_NAME:
+				returnValue = (lookData != null) ? lookData.getLookName() : null;
+				break;
+			case OBJECT_BACKGROUND_NUMBER:
+				returnValue = 1.0d + ((lookData != null) ? lookDataList.indexOf(lookData) : 0);
+				break;
+			case OBJECT_BACKGROUND_NAME:
+				returnValue = (lookData != null) ? lookData.getLookName() : null;
```
robertpainsi: Will there be a problem when returning `null`?
e.g. concating two strings, `join(Hello, !background.name which will return null!)` = ?

robertriedl: I've read when concating a string with another string, which is null, the reference null is converted to the string "null".
e.g. String str = null; str + "hi";   will return "nullhi".

robertpainsi: `null` is a valid look name. A look can't be named to an empty string `""`.
So I would suggest to return an empty string instead of null. Also children who have never used a programming language have no idea what null means and are definetily confused by it.

If there is no lookdata, look_number will also return 1 which is the same as the first look is set.

Please discuss this with the person responsible for the catrobat language first. Once it's released, it can't be changed.

PR #1835, opened 2016-08-09
CAT-1962 Play sound and wait Brick added

robertriedl:


p4p4:
org.catrobat.catroid.test.ui.CategoryBricksFactoryTest.testSoundBricks()
please increase number of expected bricks from 5 to 6.

catroid/src/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,372 @@
...
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.io.SoundManager;
+import org.catrobat.catroid.ui.ScriptActivity;
+import org.catrobat.catroid.ui.controller.SoundController;
+import org.catrobat.catroid.ui.fragment.SoundFragment;
+import org.catrobat.catroid.ui.fragment.SoundFragment.OnSoundInfoListChangedAfterNewListener;
+
+import java.util.List;
+
+public class PlaySoundAndWaitBrick extends BrickBaseType implements OnItemSelectedListener,
```
p4p4: add XStreamAlias for Brick in ``StorageHandler.setProgramXstreamAliases()``
and add brickIndo in ``XStreamToSupportCatrobatLanguageVersion099AndBefore.initializeBrickInfoMap()``

I used to forget this myself in the past, but now I try to be very attentive :smile:

catroid/src/org/catrobat/catroid/content/actions/PlaySoundAndWaitAction.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.catrobat.catroid.content.actions;
+
+import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
+
+import org.catrobat.catroid.common.SoundInfo;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.io.SoundManager;
+
+public class PlaySoundAndWaitAction extends TemporalAction {
```
p4p4: What's the difference of this class to ``PlaySoundAction``?
As far as I can see this class is identical to ``PlaySoundAction``.

The name ``PlaySoundAndWaitAction`` is misleading as well, because this action does not perform a *wait*. (Instead, you do a wait in a separate class ``WaitSoundAction``

catroid/src/org/catrobat/catroid/content/actions/WaitSoundAction.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.content.actions;
+
+import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
+
+public class WaitSoundAction extends TemporalAction {
```
p4p4: couldn't you use the existing``WaitAction`` instead?

just use the method from the ``ActionFactory`` like this:
``actionfactory.createWaitAction(sprite, new Formula(delayAsIntegerNumber))``

catroid/src/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,372 @@
...
+		adapterView = parent;
+	}
+
+	@Override
+	public void onNothingSelected(AdapterView<?> arg0) {
+	}
+
+	@Override
+	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
+		sequence.addAction(sprite.getActionFactory().createPlaySoundAndWaitAction(sprite, sound));
```
p4p4: the standard ``PlaySoundAction`` should suffice. use 
``sprite.getActionFactory().createPlaySoundAction()`` instead

catroid/src/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -61,6 +61,35 @@ public synchronized void playSoundFile(String pathToSoundfile) {
 		}
 	}
 
+	public synchronized void playSoundFileAndWait(String pathToSoundfile) {
```
p4p4: this method is unnecessary, since it's the same as the existing ``public synchronized void playSoundFile(String pathToSoundfile)``

catroid/src/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
robertriedl: Thanks :)

catroid/src/org/catrobat/catroid/content/actions/PlaySoundAndWaitAction.java
robertriedl: That happened at the beginning, when a different method playSoundFile had to be called, where the scripts were stopped - but now i've solved it with a wait action and there you're right, the two methods do the same

catroid/src/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	private ArrayAdapter<SoundInfo> createSoundAdapter(Context context) {
+		ArrayAdapter<SoundInfo> arrayAdapter = new ArrayAdapter<SoundInfo>(context,
+				android.R.layout.simple_spinner_item);
+		arrayAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+		SoundInfo dummySoundInfo = new SoundInfo();
+		dummySoundInfo.setTitle(context.getString(R.string.new_broadcast_message));
+		arrayAdapter.add(dummySoundInfo);
+		for (SoundInfo soundInfo : ProjectManager.getInstance().getCurrentSprite().getSoundList()) {
```
loibi93: Not sure if available for ArrayAdapter but if possible use addAll ;)

catroid/src/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,374 @@
...
+	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
+		sequence.addAction(sprite.getActionFactory().createPlaySoundAction(sprite, sound));
+		sequence.addAction(sprite.getActionFactory().createWaitAction(sprite,
+				new Formula(getDurationOfSoundFile(sprite, sound))));
+		return null;
+	}
+
+	private int getDurationOfSoundFile(Sprite sprite, SoundInfo sound) {
+		int durationInMilliseconds = 0;
+		int durationInSeconds = 0;
```
loibi93: This is not needed, just `return durationInMilliseconds / 1000;`

PR #1836, opened 2016-08-09
CAT-1714 Object Grouping

sjaindl:
- Adding BaseExpandableListView for automatic grouping
- Views and logic for groups and children
- Adding of groups via actionbar and longPress on add button
- Current group selections
- Refactoring
- Adding SpriteFactory
- Handle view recycling (convert view) appropriately in SpriteListFragment
- Group deletion via actionbar and longPress
- Backpack support
- Renaming of groups
- Drag&Drop ListView
- Add converter for loading old projects
- Update select all view in SpritesListFragment
- Fix rename object in ProgramMenuActivity
- Fix drag & drop listview
- Add group state indicator drawables
- Update id after renaming sprites
- Fix and add tests
- UX changes
- Make shallow copy of convertedSprites instead of deep copy (convert old project + drag&drop)
- New group layout


robertpainsi:
Are your gradle changes necessary? `Xmx5120M`

The Formatter Bot is nagging like crazy :smile_cat: I guess it's because the Raspberry Pi only has 1GB of RAM. However, my laptop also has only 4GB.

catroid/res/values-de/strings.xml
```diff
@@ -159,6 +160,10 @@
   <string name="deselect_all">Nichts auswählen</string>
   <string name="backpack_replace_object">Das Objekt \"%1$s\" befindet sich bereits im Rucksack. Möchtest du es ersetzen?</string>
   <string name="backpack_replace_object_multiple">Im Rucksack sind bereits Objekte vorhanden. Möchtest du diese ersetzen?</string>
+  <string name="groups_create">Gruppe hinzufügen</string>
+  <string name="ungroup">Ausgruppieren</string>
```
p4p4: vielleicht bessser: "Gruppierung aufheben" ?

catroid/src/org/catrobat/catroid/common/Constants.java
```diff
@@ -29,7 +29,7 @@
 	// Reflection in testcases needed
 	// http://stackoverflow.com/questions/1615163/modifying-final-fields-in-java?answertab=votes#tab-top
 
-	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = Float.valueOf(0.99f);
+	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = Float.valueOf(1.0f);
```
p4p4: it's only internal, but are we really *ready* for ``1.0``, or more for something like ``0.991``.  In other words: do we have a convention for this number, or are we happy as long as the value is increasing?

catroid/src/org/catrobat/catroid/ui/fragment/SpriteFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class SpriteFactory {
+
+	public static final String SPRITE_BASE = Sprite.class.getSimpleName();
+	public static final String SPRITE_SINGLE = SingleSprite.class.getSimpleName();
+	public static final String SPRITE_GROUP = GroupSprite.class.getSimpleName();
+	public static final String SPRITE_GROUP_ITEM = GroupItemSprite.class.getSimpleName();
+
+	public Sprite newInstance(String type) {
+		Sprite sprite = null;
```
p4p4: just a minor thing, but could you replace the body of this method with
```java
return newInstance(type, null);
```
in order to get rid of duplicate code?

catroid/src/org/catrobat/catroid/content/Sprite.java
```diff
@@ -274,20 +282,21 @@ public void setActionFactory(ActionFactory actionFactory) {
 
 	@Override
 	public Sprite clone() {
```
p4p4: Not a suggestion, just a note:

I'd be glad if we could start merging our features as soon as possible. I suppose that my Cloning feature (PR #1823) has to be  adapted to the grouping feature as well (and a lot of the PRs might have conflicts with other PRs). I am affraid that 10 hours won't suffice for that. :warning:

catroid/src/org/catrobat/catroid/content/Project.java
```diff
@@ -424,4 +421,33 @@ public boolean containsSprite(Sprite selectedSprite) {
...
+			if (!(brick instanceof PointToBrick)) {
+				continue;
+			}
+			PointToBrick pointToBrick = (PointToBrick) brick;
+			Sprite pointedSprite = pointToBrick.getPointedObject();
+			if (pointedSprite == null) {
+				continue;
+			}
+
+			Sprite newSprite = getSpriteBySpriteName(pointToBrick.getPointedObject());
```
p4p4: I'm curious: where does the PointTo brick point to when a group is selected? :)

catroid/src/org/catrobat/catroid/content/GroupItemSprite.java
```diff
@@ -0,0 +1,35 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.content;
+
+public class GroupItemSprite extends Sprite {
```
p4p4: I don't know what's best practice here
a) extend a class and check with ``instanceof`` operator if an object is an instance of a certain subclass
b) add an ``enum`` spriteTypes in ``Sprite`` and check ``if (sprite.getSpriteType() == SpriteType.GROUP_ITEM)

I guess the answer to which method is better depends on the likelihood that a subclass could override or add functionality in the future. I don't expect you to change anything here, since it might involve changing lots of "instanceof" checks to type-checks.

gradle.properties
```diff
@@ -23,4 +23,4 @@
 
 org.gradle.daemon=true
 org.gradle.parallel=true
-org.gradle.jvmargs=-XX:MaxPermSize=1024m
+org.gradle.jvmargs=-Xmx5120M
```
p4p4: I'm not sure if this should be in the repository since a meaningful value depends on each developer's own hardware

p4p4:
I tested your PR locally and didn't encounter any bugs yet :)

gradle.properties
sjaindl: @robertpainsi @p4p4 
Well, it has nothing to do with the object grouping - it's just one of the things you encounter during developing :)
Actually it should speed up development builds, especially with Instant Run - it should be set to a minimum of 2GB that faster builds can be done with DexInProcess. 

There's a good article: 
https://medium.com/google-developers/faster-android-studio-builds-with-dex-in-process-5988ed8aa37e#.jjwhs8idz

It also states that it should correspond to the javaMaxHeapSize option in build.gradle, which currently is 4GB. So maybe we should change it to -Xmx4096M. 
But I can also remove it here and create a separate ticket, since it has nothing to do with object grouping ;) What do you think?

catroid/src/org/catrobat/catroid/common/Constants.java
sjaindl: @thmq : Are we ready for 1.0? ;)

catroid/src/org/catrobat/catroid/content/Project.java
sjaindl: The PointToBrick can only point to other objects, not to a group at all (the group is just a layer around the Sprites).
And the method refreshSpriteReferences(..) is needed because all PointTo-Bricks have a reference to the old (ready for GC) sprites - that need to be set to the newly created Sprites ;)

p4p4: Ok I understand. The groups should not appear in the 

The group should not appear in the dropdown list of sprites in the PointToBrick, i guess?
https://www.dropbox.com/s/ddhtq0gw292xf57/2016-08-10%2017.26.07.png?dl=0

catroid/src/org/catrobat/catroid/content/GroupItemSprite.java
sjaindl: Hmm..  maybe the cleanest approach would be to use some sort of visitor pattern. Not sure though.

catroid/src/org/catrobat/catroid/content/Project.java
sjaindl: Yes, thx for the hint! Should be fixed now (I've added a check in PointToBrick.java)! ;)

gradle.properties
robertpainsi: > But I can also remove it here and create a separate ticket, since it has nothing to do with object grouping ;) What do you think?

Sounds great :+1:

sjaindl: Perfect, I've created a ticket for it: https://jira.catrob.at/browse/CAT-2089
and removed it here.

catroid/src/org/catrobat/catroid/common/Constants.java
sjaindl: I've set it to 0.992 now.

PR #1838, opened 2016-08-10
CAT-1998 Physics collision not working in downloaded programs

stefanreichenauer:


catroid/src/org/catrobat/catroid/io/StorageHandler.java
```diff
@@ -1120,4 +1120,9 @@ public static boolean deleteDirectory(File path) {
 		}
 		return path.delete();
 	}
+
```
thmq: This should be called within XStreamToSupportCatrobatLanguageVersion0991AndBefore.java at line 584, so that it is called for all programs that contain the old version of the message, regardless if they are downloaded or locally opened.

catroid/src/org/catrobat/catroid/transfers/ProjectDownloadService.java
```diff
@@ -88,6 +88,7 @@ protected void onHandleIntent(Intent intent) {
 					StorageHandler.getInstance().saveProject(projectTBRenamed);
 				}
 			}
+			StorageHandler.getInstance().updateCodefileOnDownload(projectName);
```
thmq: this is no longer needed then.

catroid/src/org/catrobat/catroid/io/StorageHandler.java
stefanreichenauer: That won't work, because the wrong BroadcastMessage won't cause the super.fromXML(file)-method (line 171) to throw a ConversionException. Not the tags are wrong, just the value. It should work right under line 170, but then it's called every time the program will be opened after switching programs.

thmq: all right then!

thmq:
here is a green run: https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/573/

PR #1841, opened 2016-08-10
CAT-1629 fix landscape mode for devices without HW-buttons

p4p4:
The error occurred for devices without hardware buttons (menu, back, home).
Previously in a landscape program, a white area was visible that did not
belong to the stage.
Recomputing the screen dimensions in the StageActivity fixes that problem.


thmq:
could you please open another PR onto release-0.9.24?

PR #1845, opened 2016-08-11
Feature/chrome cast

jochenflachhuber:


p4p4:
at first sight this pull-request looks a bit scary because it contains lots of commits which are not ChromeCast-related. But when looking more carefully and [comparing the Pull-Request against develop](https://github.com/Catrobat/Catroid/compare/develop...jochenflachhuber:feature/ChromeCast), one can see that the PR should only consist of ChromeCast commits.
After merging, the feature/ChromeCast branch should be up with the develop branch :)

PR #1863, opened 2016-08-16
CAT-1925 Look/Background Number/Name added to Formula-Editor

robertriedl:
reopens [#1832](https://github.com/Catrobat/Catroid/pull/1832)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -739,6 +745,18 @@ private Object interpretObjectSensor(Sensors sensor, Sprite sprite) {
...
+				break;
+			case OBJECT_LOOK_NAME:
+				returnValue = (lookData != null) ? lookData.getLookName() : "";
+				break;
+			case OBJECT_BACKGROUND_NUMBER:
+				returnValue = 1.0d + ((lookData != null) ? lookDataList.indexOf(lookData) : 0);
+				break;
+			case OBJECT_BACKGROUND_NAME:
+				returnValue = (lookData != null) ? lookData.getLookName() : "";
+				break;
```
sjaindl: It would be shorter to write it like:
case OBJECT_LOOK_NUMBER:
case OBJECT_BACKGROUND_NUMBER:
    returnValue = 1.0d + ((lookData != null) ? lookDataList.indexOf(lookData) : 0);
    break;

The same for OBJECT_LOOK_NAME and OBJECT_BACKGROUND_NAME, as they have the same logic.

PR #1864, opened 2016-08-16
CAT-1934 Go to brick

robertriedl:


catroid/src/main/java/org/catrobat/catroid/content/actions/GoToOtherSpritePositionAction.java
```diff
@@ -0,0 +1,54 @@
...
+import org.catrobat.catroid.content.Sprite;
+
+public class GoToOtherSpritePositionAction extends TemporalAction {
+	private Sprite sprite;
+	private Sprite destinationSprite;
+
+	@Override
+	protected void update(float percent) {
+		if (destinationSprite == null) {
+			destinationSprite = this.sprite;
```
loibi93: Please just return, saves the app from executing 4 lines of code that basically do nothing (i know this will never hit performance, but if its not needed you don't need to do it ;D)

catroid/src/main/java/org/catrobat/catroid/content/actions/GoToRandomPositionAction.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.stage.StageActivity;
+
+public class GoToRandomPositionAction extends TemporalAction {
+	private Sprite sprite;
+
+	@Override
+	protected void update(float percent) {
+		sprite.look.setXInUserInterfaceDimensionUnit((int) (Math.random()
```
loibi93: Should it be possible that the random position is off screen ? Or did you limit the random generator to some values, I can't see that anywhere !

robertriedl: the following 2 lines limit it to screen size ;)

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+		spinnerAdapterWrapper = new SpinnerAdapterWrapper(context, goToSpinner, spinnerAdapter);
+
+		goToSpinner.setAdapter(spinnerAdapterWrapper);
+
+		goToSpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+			@Override
+			public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
+				String itemSelected = parent.getSelectedItem().toString();
+
+				if (itemSelected.equals(context.getString(R.string.brick_go_to_touch_position))) {
```
loibi93: `.equals(touchPositionLabel)`

also use this two lines further down ;)

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+		}
+	}
+
+	private ArrayAdapter<String> createArrayAdapter(Context context) {
+		final ArrayAdapter<String> arrayAdapter = new ArrayAdapter<String>(context,
+				android.R.layout.simple_spinner_item);
+
+		arrayAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+
+		arrayAdapter.add(context.getString(R.string.brick_go_to_touch_position));
```
loibi93: you can use the global strings here too.

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+		arrayAdapter.add(context.getString(R.string.brick_go_to_random_position));
+
+		final ArrayList<Sprite> spriteList = (ArrayList<Sprite>) ProjectManager.getInstance().getCurrentProject()
+				.getSpriteList();
+
+		for (Sprite sprite : spriteList) {
+			String spriteName = sprite.getName();
+			String currentSprite = ProjectManager.getInstance().getCurrentSprite().getName();
+			if (!spriteName.equals(currentSprite) && !spriteName.equals(context.getString(R.string.background))) {
```
loibi93: and here too

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+
+		@Override
+		public Object getItem(int paramInt) {
+			return spinnerAdapter.getItem(paramInt);
+		}
+
+		@Override
+		public long getItemId(int paramInt) {
+			String currentItemName = spinnerAdapter.getItem(paramInt);
+			if (!(currentItemName.equals(context.getString(R.string.brick_go_to_touch_position))
```
loibi93: also here

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GoToBrick extends BrickBaseType {
+	private static final long serialVersionUID = 1L;
+
+	private Sprite destinationSprite;
+	private transient String oldSelectedObject;
+	private String touchPositionLabel;
```
loibi93: If you use those in all the places I marked out below, you need to ensure that they are never null (I am not sure if getView is called first amongst the other methods where you should use the strings)

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -713,4 +713,20 @@ public Stage getStage() {
 	public void removeActor(Look look) {
 		look.remove();
 	}
+
+	public float getVirtualWidth() {
```
loibi93: You can get those values from projectManager !

catroid/src/main/java/org/catrobat/catroid/content/actions/GoToRandomPositionAction.java
loibi93: Does random give back numbers between 0 and 1 ?

robertriedl: yes

loibi93: Ahhh ok, then it is perfectly fine ;)

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
robertriedl: i need this labels only for the oldSelectedObject in the setSpinnerSelection-Method, where i have no context to get the string of R.string.... For the prototypeView it's not relevant if they are null and in getView the assignments are called before setSpinnerSelection

loibi93: then it is fine ;)

catroid/src/androidTest/java/org/catrobat/catroid/test/content/actions/GoToActionsTest.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	public void testNullActor() {
+		ActionFactory factory = new ActionFactory();
+		Action action = factory.createGoToAction(null, null, 0);
+		try {
+			action.act(1.0f);
+			fail("Execution of GlideToBrick with null Sprite did not cause a " + "NullPointerException to be thrown");
+		} catch (NullPointerException expected) {
+		}
```
sjaindl: There should be another 2 tests for random position (just check if position is different after the act()-method call) and, if possible, also for the last touched position

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -0,0 +1,381 @@
...
+
+	@Override
+	public int getRequiredResources() {
+		return NO_RESOURCES;
+	}
+
+	@Override
+	public Brick copyBrickForSprite(Sprite sprite) {
+		GoToBrick copyBrick = (GoToBrick) clone();
+
```
sjaindl: You could do that in one line:
return clone();

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
robertriedl: i don't want to use the labels here, because they're only needed in the setSpinnerSelection method because there's no Context and if i use it everywhere else they're not set in time for some cases

catroid/src/androidTest/java/org/catrobat/catroid/test/content/actions/GoToActionsTest.java
robertriedl: yes it would be fine, but it's not possible here to simulate a touch position since there are no ui-methods here and also the stage isn't loaded here at the unit-tests

loibi93: maybe put dummy values in by hand ?

robertriedl: tried this out, but didn't work with my variant - how would you get the stage here?

loibi93: this is your code for the go to touch action:
`int touchIndex = TouchUtil.getLastTouchIndex();
 sprite.look.setXInUserInterfaceDimensionUnit(TouchUtil.getX(touchIndex));
 sprite.look.setYInUserInterfaceDimensionUnit(TouchUtil.getY(touchIndex));`

Now just add a method in TouchUtil `public void setDummyTouchForTest(int x, int y)`
and set the values in there.

robertriedl: this didn't solve the problem because the stage still is not loaded - so you have a null-pointer-reference and this leads to test fail

loibi93: You do not need the stage, just add a point to the `touches` member in TouchUtil, and then call goToTouchPositionAction.act() ! Then check if the sprite is at the specified position.

loibi93: And the random Position you can also test without a Stage, because you can refactor the code in your Action to not use the stage, which is better, because then you do not need the new methods in stage Listener !

robertriedl: thanks, without using the stage my unit-tests would work :)

PR #1865, opened 2016-08-16
CAT-1935 Distance to

robertriedl:
reopens [#1853](https://github.com/Catrobat/Catroid/pull/1853)


robertriedl:
reopens [#1853](https://github.com/Catrobat/Catroid/pull/1853)

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -290,6 +291,17 @@ public void setYInUserInterfaceDimensionUnit(float y) {
 		setY(y - getHeight() / 2f);
 	}
 
+	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
+		float distanceToTouchPosition;
+		int touchIndex = TouchUtil.getLastTouchIndex();
+
+		distanceToTouchPosition = (float)
```
loibi93: just return here, then you don't need the extra variable

PR #1868, opened 2016-08-16
CAT-1962 Play sound and wait Brick added

robertriedl:
reopens [#1835](https://github.com/Catrobat/Catroid/pull/1835)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
@@ -1,0 +1,381 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2016 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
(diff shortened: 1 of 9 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,376 @@
...
+		return prototypeView;
+	}
+
+	private void setSpinnerSelection(Spinner spinner) {
+		if (ProjectManager.getInstance().getCurrentSprite().getSoundList().contains(sound)) {
+			oldSelectedSound = sound;
+			spinner.setSelection(ProjectManager.getInstance().getCurrentSprite().getSoundList().indexOf(sound) + 1, true);
+		} else {
+			if (spinner.getAdapter() != null && spinner.getAdapter().getCount() > 1) {
+				if (ProjectManager.getInstance().getCurrentSprite().getSoundList().indexOf(oldSelectedSound) >= 0) {
```
sjaindl: Please add a List<SoundInfo>-variable to assign ProjectManager.getInstance().getCurrentSprite().getSoundList() and use that variable - 
So you don't need to call the entire function all the time.

catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundAndWaitBrick.java
```diff
@@ -0,0 +1,376 @@
...
+
+	@Override
+	public void onSoundInfoListChangedAfterNew(SoundInfo soundInfo) {
+		sound = soundInfo;
+		oldSelectedSound = soundInfo;
+	}
+
+	@Override
+	public void storeDataForBackPack(Sprite sprite) {
+		SoundInfo backPackedSoundInfo = SoundController.getInstance().backPackHiddenSound(this.getSound());
```
sjaindl: Please just use     sound    instead of this.getSound() - this is a bit more compact

PR #1877, opened 2016-08-16
CAT-1969 When condition becomes true brick

p4p4:
**reopens #1827**

- script is executed when condition becomes true
- to be executed again, the condition has first to be false
  and then become true again
- did some refactoring: ScriptBrick is now an interface instead of an
  abstract class


p4p4:
Checkstyle: [checkstyle](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/336/)
Tests: [tests](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/4494/)

PR #1878, opened 2016-08-16
CAT-1938 Clone Sprites Feature

p4p4:
**reopens  #1823**

- create a clone of a Sprite (= Object) via a special Brick
- create a special "Clone" brick that clones the Object in which it is
    - all Scripts, Sounds, Variables etc are copied
- add a "When i start as a clone" brick
    - is a script-brick that is only executed when Object is cloned
    - "When started" ScriptBricks are NOT executed for the clone
- add "Delete this clone" brick
    - when executed, the cloned object containing this brick
      is deleted
- global variables are shared, local (=per object) variables are copied
  for clones.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Project.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Project.java
@@ -39,6 +39,7 @@ import org.catrobat.catroid.formulaeditor.DataContainer;
 import org.catrobat.catroid.formulaeditor.UserList;
 import org.catrobat.catroid.formulaeditor.UserVariable;
 import org.catrobat.catroid.physics.content.ActionPhysicsFactory;
+import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.ui.SettingsActivity;
 import org.catrobat.catroid.utils.Utils;
 
@@ -198,6 +199,14 @@ public class Project implements Serializable {
 
 	public void setName(String name) {
 		xmlHeader.setProgramName(name);
+	}
+
+	public List<Sprite> getSpriteListWithClones() {
+		if (StageActivity.stageListener != null) {
+			return StageActivity.stageListener.getSpritesFromStage();
+		} else { // e.g. for ActionTests there is no Stage, only use sprites from Project
+			return getDefaultScene().getSpriteList();
+		}
 	}
 
 	public String getName() {
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/DeleteThisCloneBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/DeleteThisCloneBrick.java
@@ -1,0 +1,109 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2016 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.content.bricks;
+
+import android.content.Context;
+import android.graphics.drawable.Drawable;
+import android.view.View;
+import android.widget.BaseAdapter;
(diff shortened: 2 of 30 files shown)


catroid/src/main/java/org/catrobat/catroid/content/Project.java
```diff
@@ -116,6 +118,10 @@ private void ifPortraitSwitchWidthAndHeight() {
 	}
 
 	public synchronized void addSprite(Sprite sprite) {
+		if (sprite.isClone) {
+			Log.e("Project", "Do NOT add clones to the projectmanager!");
```
sjaindl: Please create a log tag instead of "Project", as suggested by @robertpainsi in #1881 :

private static final String TAG = Project.class.getSimpleName();

This would be consistent with all other source files, it looks cleaner and the next time someone wants to log something, s|he can use the TAG right away 😉.

catroid/src/main/java/org/catrobat/catroid/content/bricks/DeleteThisCloneBrick.java
```diff
@@ -0,0 +1,112 @@
...
+	@Override
+	public Brick copyBrickForSprite(Sprite sprite) {
+		DeleteThisCloneBrick copyBrick = (DeleteThisCloneBrick) clone();
+		return copyBrick;
+	}
+
+	@Override
+	public View getViewWithAlpha(int alphaValue) {
+		if (view != null) {
+			Log.d(TAG, "VIEW != NULL");
```
sjaindl: this log shouldn't be necessary

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -216,6 +219,49 @@ public void create() {
...
+	public void removeClonedSpriteFromStage(Sprite sprite) {
+		if (!sprite.isClone) {
+			return;
+		}
+
+		Project currentProject = ProjectManager.getInstance().getCurrentProject();
+		DataContainer userVariables = currentProject.getDataContainer();
+		userVariables.removeVariableListForSprite(sprite);
+
+		BroadcastHandler.getScriptSpriteMapMap().remove(sprite);
```
sjaindl: There's a typo in the existing code: please change getScriptSpriteMapMap --> getScriptSpriteMap

catroid/src/main/res/layout/brick_clone.xml
```diff
@@ -0,0 +1,62 @@
...
+
+        <Spinner
+            android:id="@+id/brick_clone_spinner"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:clickable="false"
+            android:focusable="false" />
+    </org.catrobat.catroid.ui.BrickLayout>
+
+</LinearLayout>
```
sjaindl: add newline

catroid/src/main/res/layout/brick_delete_this_clone.xml
```diff
@@ -0,0 +1,53 @@
...
+        app:horizontalSpacing="@dimen/brick_flow_layout_horizontal_spacing"
+        app:verticalSpacing="@dimen/brick_flow_layout_vertical_spacing">
+
+        <TextView
+            android:id="@+id/brick_delete_clone_label"
+            style="@style/BrickText.SingleLine"
+            android:text="@string/brick_delete_this_clone" />
+    </org.catrobat.catroid.ui.BrickLayout>
+
+</LinearLayout>
```
sjaindl: add newline

catroid/src/main/res/layout/brick_when_cloned.xml
```diff
@@ -0,0 +1,55 @@
...
+        app:verticalSpacing="@dimen/brick_flow_layout_vertical_spacing">
+
+        <TextView
+            style="@style/BrickText.SingleLine"
+            android:text="@string/brick_when_cloned">
+        </TextView>
+
+    </org.catrobat.catroid.ui.BrickLayout>
+
+</LinearLayout>
```
sjaindl: add newline

catroid/src/main/res/layout/brick_clone.xml
p4p4: since I often forget the newline at the end of an XML file: is it possible to add a static check for this?

robertpainsi: The Checkstyle check NewlineAtEndOfFile is currently enabled for java and gradle files only ([see here](
https://github.com/Catrobat/Catroid/blob/98f506ed6e6ad1d19a83d0d6c42744daca1fb309/catroid/config/checkstyle.xml#L208-L211)).

Android Studio is the reason why this check is not enabled for xml files. Creating a new xml file/resource through AS will create a file _not_ ending with a newline. However newly created java files will always end with a newline. I haven't found a setting (which is also shared) yet to automatically add newlines to xml files as well. :crying_cat_face: 

I would prefer to keep the default line endings for created files with AS, but since it's not possible to check **No**NewlineAtEndOfFile with checkstyle afaik, I'm fine to force newlines in xmls as well. It also seems to be the preferred way when working with git.

> Rationale: Any source files and text files in general should end with a line separator to let other easily add new content at the end of file and "diff" command does not show previous lines as changed.

p4p4:
Tests ok : [tests](https://jenkins.catrob.at/job/CatroidSinglePackageEmulatorTest/4488/)
checkstyle ok: [checkstyle](https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidCheck/331/)

PR #1879, opened 2016-08-16
CAT-2056 Bugfixes and Refactoring for Show variable Brick

robertriedl:


catroid/src/main/java/org/catrobat/catroid/content/actions/ShowVariableAction.java
```diff
@@ -40,26 +40,28 @@
...
+	private Formula xPosition;
+	private Formula yPosition;
+	private UserVariable variableToShow;
+
 	private Sprite sprite;
 	private UserBrick userBrick;
-	private ShowTextActor actor;
+	private ShowVariableActor actor;
+
+	public static final String TAG = ShowVariableAction.class.getSimpleName();
```
sjaindl: the TAG should be above all other member variables

catroid/src/main/java/org/catrobat/catroid/stage/ShowVariableActor.java
```diff
@@ -69,26 +72,31 @@ public void draw(Batch batch, float parentAlpha) {
...
 		if (variableList == null) {
 			return;
 		}
-		for (UserVariable variable : variableList) {
-			if (variable.getName().equals(variableName)) {
-				variableValue = variable.getValue().toString();
-				if (variable.getVisible()) {
-					font.draw(batch, variableValue, posX, posY);
+
+		if (variableToShow.getDummy()) {
```
sjaindl: Maybe a bit clearer: getDummy() --> isDummy()

PR #1880, opened 2016-08-17
CAT-1636 Order xstream fields by annotation

robertpainsi:
See commit messages for more details.

> ...If there is an alias defined for the field, the alias has to be used in the annotation.

No real reason behind that. I just think that renaming an alias is less likely then renaming a member variable. So it would be less error prone... But if you guys prefer the member name, let me know.

I created two commits, because the CatroidFieldKeySorter wasn't tested at all. So I tested the general case and then **extended** it to work with annotation as well. I can create two separate Pull Requests if preferred.

@p4p4 :wink: 

Reopens https://github.com/Catrobat/Catroid/pull/1423


p4p4:
nice :)

loibi93:
Looks like we do not have to change the KeySorter anymore even if we add transient members, right ? :D

robertpainsi:
> Looks like we do not have to change the KeySorter anymore even if we add transient members, right ? :D

Exactly!

PR #1881, opened 2016-08-18
Fixed bug using the clone method of LookData

loibi93:
- clone method fixes the bug
- had to add Checksum before clone because otherwise there is no checksum found for the LookData
- added Log Message to increment Usage such that we know which file is missing

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/common/FileChecksumContainer.java
+++ b/catroid/src/main/java/org/catrobat/catroid/common/FileChecksumContainer.java
@@ -22,6 +22,8 @@
  */
 package org.catrobat.catroid.common;
 
+import android.util.Log;
+
 import java.io.FileNotFoundException;
 import java.io.Serializable;
 import java.util.HashMap;
@@ -30,6 +32,7 @@ import java.util.Map;
 public class FileChecksumContainer implements Serializable {
 
 	private static final long serialVersionUID = 1L;
+	private static final String TAG = FileChecksumContainer.class.getSimpleName();
 
 	private class FileInfo {
 		private int usageCounter;
@@ -91,13 +94,13 @@ public class FileChecksumContainer implements Serializable {
 	public void incrementUsage(String filepath) throws FileNotFoundException {
 		String checksum = null;
 		for (Map.Entry<String, FileInfo> entry : checksumFileInfoMap.entrySet()) {
-
 			if (entry.getValue().path.equalsIgnoreCase(filepath)) {
 				checksum = entry.getKey();
 				break;
 			}
 		}
 		if (checksum == null) {
+			Log.e(TAG, "incrementUsage: FileNotFound: " + filepath);
 			throw new FileNotFoundException();
 		}
 		FileInfo fileInfo = checksumFileInfoMap.get(checksum);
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/common/FileChecksumContainer.java
```diff
@@ -91,13 +93,13 @@ public int getUsage(String checksum) {
...
 		String checksum = null;
 		for (Map.Entry<String, FileInfo> entry : checksumFileInfoMap.entrySet()) {
-
 			if (entry.getValue().path.equalsIgnoreCase(filepath)) {
 				checksum = entry.getKey();
 				break;
 			}
 		}
 		if (checksum == null) {
+			Log.e(this.getClass().getSimpleName(), "incrementUsage of " + filepath);
```
sjaindl: getClass().getSimpleName() should be sufficient

catroid/src/main/java/org/catrobat/catroid/common/defaultprojectcreators/DefaultProjectCreatorDefault.java
```diff
@@ -163,11 +163,14 @@ public Project createDefaultProject(String projectName, Context context, boolean
 
 			ProjectManager.getInstance().getFileChecksumContainer().addChecksum(soundInfo1.getChecksum(), soundInfo1.getAbsolutePath());
 			ProjectManager.getInstance().getFileChecksumContainer().addChecksum(soundInfo2.getChecksum(), soundInfo2.getAbsolutePath());
+			ProjectManager.getInstance().getFileChecksumContainer().addChecksum(birdWingUpLookData.getChecksum(), birdWingUpLookData.getAbsolutePath());
+			ProjectManager.getInstance().getFileChecksumContainer().addChecksum(birdWingDownLookData.getChecksum(), birdWingDownLookData.getAbsolutePath());
+			ProjectManager.getInstance().getFileChecksumContainer().addChecksum(cloudLookData.getChecksum(), cloudLookData.getAbsolutePath());
```
sjaindl: I think you should add the checksum for cloudLookData two times. The first time to create the checksum key and the second time to increment the usage (for the cloned look below).

catroid/src/main/java/org/catrobat/catroid/common/FileChecksumContainer.java
robertpainsi: Create a log tag instead
```java
private static final String TAG = FileChecksumContainer.class.getSimpleName();
```
This would be consistent with all other source files, it looks cleaner and the next time someone wants to log something, s|he can use the TAG right away :wink:.

Maybe you can write a more meaningful log message to hint to the problem?

sjaindl: even better +1

catroid/src/main/java/org/catrobat/catroid/common/defaultprojectcreators/DefaultProjectCreatorDefault.java
loibi93: This is done in the clone method!

sjaindl: Ok, then it's perfectly fine!

loibi93:
Log call should be pretty now ;)

michael-fischl:
Please also check [this](https://github.com/Catrobat/Catroid/pull/1884) when merging since it's addressing the same problem.

loibi93:
Maybe we should link those two Tickets together then, and only merge one. @thmq

PR #1882, opened 2016-08-18
Cat 2054

OliverPapst:


robertpainsi:
Please clean up your log. There shouldn't be any merge commits in this PR.

```
git checkout CAT-2054
git reset --soft dfd3b3c
git commit -m 'CAT-2054 Updating XML Files'
git diff f39ac85
```
No warranty!
Always start with a clean working tree and know how to undo things (git reflog and reset --hard). The diff at the end should be empty and there shouldn't be any further commits addressing `CAT-2054`. However, by executing `git log --grep=2054` you will find two commits. The currently committed one and d4cd37b. The second commit is from another PR https://github.com/Catrobat/Catroid/pull/1857. So, in this case it's fine if there are two commits.

After that, rebase onto the current develop branch and force-push your changes.

Side note: Don't merge your current branch in your... current one :hushed:. If you are working on a short living branch or create a single commit PR, rebase onto the current develop branch instead.

PR #1884, opened 2016-08-19
[CAT-2071] Fix for sprite clone not working consistent

michael-fischl:


catroid/src/main/java/org/catrobat/catroid/common/SoundInfo.java
```diff
@@ -99,7 +84,7 @@ public SoundInfo copySoundInfoForSprite(Sprite sprite) {
 	public String getAbsolutePath() {
 		if (fileName != null) {
 			if (isBackpackSoundInfo) {
```
loibi93: this if does not make any sense ! Did you also test, if backpack is still working ? I really don't think it is with this if case !

PR #1885, opened 2016-08-20
CAT-2053 Added Bubble Bricks

loibi93:
Added four Bricks from Scratch:
- Think __
- Say __
- Think __ for __
- Say __ for __

Was implemented using an Actor for every Sprite. Like in Scratch every Sprite can always only have one Bubble. This was achieved by a Map in the StageListener. I also created a new Wait Action which overrides the end() method, to remove the Bubble actor for the specified Sprite. The code for drawing the Bubbles in ShowBubbleActor is commented, in order to help people work on the code. It would have been a bit messy otherwise.
UX Feedback can be found here: https://confluence.catrob.at/display/USABILITY/GSOC+APKs+Feedback
Unfortunately only in german.
For the points in there:
- cropping can be done by Pocket Paint, unfortunately I can't do that
- I found out, that the Bubble may get pixelated, if PC crashes, but I could not find any reason for that. If you change the text, then everything is fine again. Maybe someone has a clue, but in my opinion this is a minor flaw and can be ignored.
- Cloudy Bubbles are not possible to draw dynamically (only with a huge monster of code)
- Crashes: This is not my fault, there is already a Ticket for that
- Maybe show a Error message, which would need a Context in the action (good idea ?). I could not reproduce the error, because the FormelEditor does not allow to put something in other than text or numbers
- you can not remove the bubbles (same as in scratch) 
- not sure what "does not work in one object" means, I tried it in one object, and it works
- fixed the different look of bricks in the overview and when actually inserted (during insert all bricks look different)


catroid/src/main/res/layout/brick_say_bubble.xml
```diff
@@ -0,0 +1,73 @@
...
+            android:layout_gravity="center_vertical"
+            app:layout_inputType="text"
+            app:layout_textField="true"
+            android:gravity="left"
+            android:maxLines="3"
+            android:minLines="1"
+            android:singleLine="false" />
+    </org.catrobat.catroid.ui.BrickLayout >
+
+</LinearLayout>
```
sjaindl: Please add a newline

catroid/src/main/res/layout/brick_say_for_bubble.xml
```diff
@@ -0,0 +1,94 @@
...
+        </TextView >
+
+        <TextView
+            android:id="@+id/brick_say_for_bubble_seconds_label"
+            style="@style/BrickText.SingleLine"
+            android:text="@string/phiro_seconds" >
+        </TextView >
+    </org.catrobat.catroid.ui.BrickLayout >
+
+</LinearLayout>
```
sjaindl: Add newline

catroid/src/main/res/layout/brick_think_bubble.xml
```diff
@@ -0,0 +1,73 @@
...
+            android:layout_gravity="center_vertical"
+            app:layout_inputType="text"
+            app:layout_textField="true"
+            android:gravity="left"
+            android:maxLines="3"
+            android:minLines="1"
+            android:singleLine="false" />
+    </org.catrobat.catroid.ui.BrickLayout >
+
+</LinearLayout>
```
sjaindl: Add newline

catroid/src/main/res/layout/brick_think_for_bubble.xml
```diff
@@ -0,0 +1,94 @@
...
+        </TextView >
+
+        <TextView
+            android:id="@+id/brick_think_for_bubble_seconds_label"
+            style="@style/BrickText.SingleLine"
+            android:text="@string/phiro_seconds" >
+        </TextView >
+    </org.catrobat.catroid.ui.BrickLayout >
+
+</LinearLayout>
```
sjaindl: same here ;)

PR #1886, opened 2016-08-20
Pen bricks

loibi93:
Added new Pen Category with following Bricks:
- Pen down
- Pen up
- Set Pen Color
- Set Pen Size
- Stamp
- Clear Background

UX Feedback can be found here (only German):
https://confluence.catrob.at/display/USABILITY/GSOC+APKs+Feedback

- Color Changer does give always a message because I reused the one Phiro uses. (Maybe a ticket for that ?)
- The visual problems seem to happen after a crash, I tested it thoroughly and had no issues
- default color is now blue (this is also shown in the overview and when brick is added)
- all bricks that accept numbers also accept strings. I think a fix for that does not belong to my ticket and should be fixed by another ticket (if even it needs to be fixed, no one can expect that this should work, so no one will even use a brick like that)
- I think the brick color is gorgeous ;)


catroid/src/main/res/layout/brick_category_pen.xml
```diff
@@ -0,0 +1,32 @@
...
+  -->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    style="@style/BrickCategory" >
+
+    <TextView
+        style="@style/BrickCategoryText.Pen"
+        android:text="@string/category_pen" >
+    </TextView>
+
+</LinearLayout>
```
sjaindl: Please add a newline in all xml files

PR #1887, opened 2016-08-20
Add values for Longitude, Latitude, Location accuracy and altitude

loibi93:
The implementation always tries to get GPS Signal, if not available after 3 Seconds, it tries Network GPS. When the GPS Signal is available again, it is also used.
UX Feedback can be found here (only german):
https://confluence.catrob.at/display/USABILITY/GSOC+APKs+Feedback


catroid/src/main/java/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -67,10 +75,19 @@
 	private float faceSize = 0f;
 	private float facePositionX = 0f;
 	private float facePositionY = 0f;
+	private double latitude = 0d;
+	private double longitude = 0d;
+	private float location_accuracy = 0f;
```
sjaindl: should be named: locationAccuracy

thmq:
ok to test

PR #1889, opened 2016-08-21
CAT-1859 Integrate Scratch2Catrobat Converter Client

rsamer:
This first beta (!) version includes the following (basic) features:
* Searching for Scratch programs (accesses official Scratch API)
* Users can see details of Scratch programs and navigate through remixed programs
* The integrated WebSocket client gets notified with progress updates by the server while the conversion is in progress
* Automatic download of converted programs (users don't have to care about that)
* Client automatically restores current progress-state after the app is restarted or network connection lost (i.e. in order to convert a Scratch program the server requires no continuous network connection to the client)

Limitations:
* Users can convert up to 3 programs simultaneously (a server limit also exists for this)
* The converter only supports Scratch version 2.0, i.e. programs created before 2013-05-09 cannot be converted (for more details, please follow this link: https://wiki.scratch.mit.edu/wiki/Scratch_2.0)

Remaining issue:
* The authentication is anonymous (i.e. works without using a password):
Users could easily manipulate the client ID and see converted programs of other unknown/anonymous persons.
Therefore we make sure that no private data (IPs, etc.) about our users is stored on the server at the moment.
Thus, this should be no real privacy concern since the Scratch Converter only accepts and converts published Scratch programs.
But once we store private data about our users, this aspect immediately becomes crucial of course.

Planned for first release:
* Switch to TLS once a server-proxy has been set up and configured (in Catroid only the URL has to be changed -> wss:// instead of ws://)
* Support Single-Sign-On authentication (has to be coordinated with Catroweb-team)
* Maybe get rid of anonymous authentication (not discussed with Bernadette and Wolfgang yet)
* Further performance improvements for the server

Looking forward to your feedback! :)


build.gradle
```diff
@@ -1,3 +1,26 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
```
sjaindl: Please add maven { url "https://jitpack.io" } to the buildscript also in the top-level gradle-file (Jenkins complains: https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/705/console)

catroid/build.gradle
```diff
@@ -103,6 +106,8 @@ dependencies {
     compile 'com.google.guava:guava:18.0'
     compile 'com.squareup.okhttp:okhttp:2.3.0'
     compile 'com.google.code.gson:gson:2.4'
+    compile 'com.koushikdutta.async:androidasync:2.+' // for websockets (Scratch Converter)
```
sjaindl: we don't need a comment here

catroid/src/main/java/org/catrobat/catroid/scratchconverter/WebSocketClient.java
```diff
@@ -0,0 +1,304 @@
...
+	@Override
+	public void onUserCanceledConversion(long jobID) {
+		Preconditions.checkState(state == State.CONNECTED_AUTHENTICATED);
+		Preconditions.checkState(clientID != INVALID_CLIENT_ID);
+		messageListener.onUserCanceledConversion(jobID);
+	}
+
+	@Override
+	public void onBaseMessage(BaseMessage baseMessage) {
+		// case: InfoMessage
```
sjaindl: We don't need the comment // case: InfoMessage
This is obvious here (the same applies to some other comments so far)

catroid/src/main/java/org/catrobat/catroid/scratchconverter/protocol/message/job/JobMessage.java
```diff
@@ -0,0 +1,145 @@
...
+
+			case JOB_PROGRESS:
+				final short progress = (short) jsonData.getInt(JsonDataKeys.PROGRESS.toString());
+				return (T) new JobProgressMessage(jobID, progress);
+
+			case JOB_FINISHED:
+				final String downloadURL = jsonData.getString(JsonDataKeys.URL.toString());
+				final String cachedUTCDateKey = JsonDataKeys.CACHED_UTC_DATE.toString();
+				final String dateUTC = jsonData.isNull(cachedUTCDateKey) ? null : jsonData.getString(cachedUTCDateKey);
+				final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.US);
```
sjaindl: Please create a constant for the DateFormat   (DATE_FORMAT_ISO_8601 doesn't match exactly)

catroid/src/main/java/org/catrobat/catroid/scratchconverter/WebSocketClient.java
rsamer: fixed

catroid/src/main/java/org/catrobat/catroid/scratchconverter/protocol/message/job/JobMessage.java
rsamer: fixed

catroid/build.gradle
rsamer: ok, removed

build.gradle
rsamer: I have added it to main/top-level build.gradle file. This fixes the problem, but then another issue occurred.

The Jenkins server seems to have problems with fetching the FlowText-library from the given HTTPS-url. It says that the peer is not authenticated. I think, that this might be a TLS-related problem, since when opening the URL in my browser (or manually download the xml-file via curl/wget) I do not encounter any problems: https://jitpack.io/com/github/deano2390/FlowTextView/2.0.4/FlowTextView-2.0.4.pom

`FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':catroid'.
> Could not resolve all dependencies for configuration ':catroid:_standaloneDebugCompile'.
   > Could not resolve com.github.deano2390:FlowTextView:2.0.4.
     Required by:
         CatroidTriggerPullRequestTest:catroid:unspecified
      > Could not resolve com.github.deano2390:FlowTextView:2.0.4.
         > Could not get resource 'https://jitpack.io/com/github/deano2390/FlowTextView/2.0.4/FlowTextView-2.0.4.pom'.
            > Could not GET 'https://jitpack.io/com/github/deano2390/FlowTextView/2.0.4/FlowTextView-2.0.4.pom'.
               > peer not authenticated
      > Could not resolve com.github.deano2390:FlowTextView:2.0.4.
         > Could not get resource 'https://jitpack.io/com/github/deano2390/FlowTextView/2.0.4/FlowTextView-2.0.4.pom'.
            > Could not GET 'https://jitpack.io/com/github/deano2390/FlowTextView/2.0.4/FlowTextView-2.0.4.pom'.
               > peer not authenticated
`

PR #1902, opened 2016-08-24
CAT-FIX Grouping/Backpack Issues

sjaindl:


catroid/src/main/java/org/catrobat/catroid/ui/dynamiclistview/UtilDynamicListView.java
```diff
@@ -608,7 +616,9 @@ public void notifyListItemTouchActionUp() {
 	private void touchEventsCancelled() {
 		if (cellIsMobile) {
 			View mobileView = getViewForID(mobileItemId);
-			unsetSpriteMobileState();
+			if (forSpriteList) {
+				unsetSpriteMobileState();
+			}
 			aboveItemId = INVALID_ID;
 			mobileItemId = INVALID_ID;
 			belowItemId = INVALID_ID;
```
loibi93: Could you please override the getAdapter method for the ListView with the following method:
@Override
	public ListAdapter getAdapter() {
		ListAdapter adapter = super.getAdapter();
		if (adapter instanceof HeaderViewListAdapter) {
			return ((HeaderViewListAdapter) adapter).getWrappedAdapter();
		}

		return adapter;
	}

This will fix the crash for DragNDrop in Scenes after copying (please verify)

sjaindl: Thx for the hint, it fixes the crashes for copying, renaming Scenes/Sounds/Looks etc. 👍

sjaindl:
I've added a second commit that also fixes CAT-2099

catroid/src/main/res/menu/menu_script_activity.xml
```diff
@@ -81,6 +81,7 @@
     <item
         android:id="@+id/comment_in_out"
         android:showAsAction="never"
-        android:title="@string/comment_in_out"/>
+        android:title="@string/comment_in_out"
+        android:visible="false"/>
```
p4p4: @sjaindl , it's very kind of you to fix a problem that I introduced. :smile: 
However, I already fixed that problem in PR #1915 which is why you can remove these lines again from this PR.
Thanks to @bernadettespieler for pointing this out.

p4p4: or wait, I close my PR so that it is just one PR that needs to be accepted ;)

PR #1903, opened 2016-08-25
Restore old functions for compatibility-reasons

robertriedl:
Restore some old functions of the show variable brick to be compatible to old pocket-code-programs.


catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextBrick.java
```diff
@@ -268,6 +270,19 @@ public void onClick(View view) {
 		return null;
 	}
 
+	void setUserVariableName(UserVariable userVariable) {
+		userVariableName = Constants.NO_VARIABLE_SELECTED;
+		try {
+			userVariableName = userVariable.getName();
+		} catch (NullPointerException e) {
+			Log.d(TAG, "Nothing selected yet.");
+		}
```
robertpainsi: Use if/else

catroid/src/main/java/org/catrobat/catroid/content/bricks/HideTextBrick.java
```diff
@@ -183,6 +187,19 @@ public View getPrototypeView(Context context) {
 		return null;
 	}
 
+	void setUserVariableName(UserVariable userVariable) {
+		userVariableName = Constants.NO_VARIABLE_SELECTED;
+		try {
+			userVariableName = userVariable.getName();
+		} catch (NullPointerException e) {
+			Log.d(TAG, "Nothing selected yet.");
+		}
```
robertpainsi: Use if/else here as well

robertpainsi:
To be compatible with older programs, it's not necessary to keep the class names.
e.g. previously `transparency` was called `ghost effect`. See [here](https://github.com/Catrobat/Catroid/blob/aabbecb5aa3484e874a88eeb82593b6f7f9dce7b/catroid/src/main/java/org/catrobat/catroid/io/XStreamToSupportCatrobatLanguageVersion0991AndBefore.java#L925-L927).

catroid/src/main/java/org/catrobat/catroid/content/bricks/HideTextBrick.java
robertpainsi: Ah, these new methods aren't used at all. Remove them.

About `public String userVariableName;`... Please do some research how to handle removed members with XStream.

catroid/src/main/java/org/catrobat/catroid/content/bricks/HideTextBrick.java
```diff
@@ -183,6 +184,18 @@ public View getPrototypeView(Context context) {
...
+			userVariableName = userVariable.getName();
+		} else {
+			userVariableName = Constants.NO_VARIABLE_SELECTED;
+		}
+	}
+
+	void setUserVariableName(String userVariableName) {
+		this.userVariableName = userVariableName;
+	}
+
```
robertpainsi: >Ah, these new methods aren't used at all. Remove them.
>
>About public String userVariableName;... Please do some research how to handle removed members with XStream.

PR #1906, opened 2016-08-26
CAT-FIX Quick fix for CAT-2098 until further insight (PR for develop)

michael-fischl:


thmq:
Apparently there are more problems with the new folder structure (i.e. master/develop differences) than we'd previously expected. I'm closing this PR for now until further notice.

PR #1907, opened 2016-08-26
CAT-FIX Quick fix for CAT-2098 until further insight (PR for master)

michael-fischl:


thmq:
The same here. (For details see #1906)

PR #1913, opened 2016-09-02
CAT-FIX change Order of Raspberry Bricks

p4p4:
first control bricks, then RaspberryPi bricks


p4p4:
closed & included to PR #1919

PR #1920, opened 2016-09-08
CAT-FIX Fix FormulaEditor layout and crash on device button (PR master)

michael-fischl:


catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -369,7 +372,14 @@ public void onStart() {
...
 
 		for (int i = 0; i < length; i++) {
-			noParametersList[i] = R.string.formula_editor_function_no_parameter;
+			//Dirty hack until further insight is gained
+			try {
+				Log.i(TAG, "Trying to get string resource: " + getString(R.string
+						.formula_editor_function_no_parameter));
+				noParametersList[i] = R.string.formula_editor_function_no_parameter;
+			} catch (Resources.NotFoundException exception) {
+				Log.e(TAG, "formula_editor_function_no_parameter not found!" + Log.getStackTraceString(exception));
```
robertpainsi: There is also [`Log.e(String, String, Throwable)`](https://developer.android.com/reference/android/util/Log.html#e(java.lang.String, java.lang.String, java.lang.Throwable)). :wink:

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -369,7 +372,14 @@ public void onStart() {
 		int[] noParametersList = new int[length];
 
 		for (int i = 0; i < length; i++) {
-			noParametersList[i] = R.string.formula_editor_function_no_parameter;
+			//Dirty hack until further insight is gained
+			try {
+				Log.i(TAG, "Trying to get string resource: " + getString(R.string
+						.formula_editor_function_no_parameter));
+				noParametersList[i] = R.string.formula_editor_function_no_parameter;
```
robertpainsi: This works without any problems?
When the exception is thrown, noParametersList[i] stays 0.
→ `getString(0) == ""`?

PR #1874, opened 2016-08-16
CAT-1968 "When Background switches to", "Switch background and wait"

p4p4:
**reopens #1822**

- brick "When Background switches to": Runs a script when background
  switches to a certain look (see Scratch)
- brick: "Switch background to _"
    - now available in all Sprites
- brick: "Switch background to _ and wait"
  waits until all "When Background _" scripts are completed

PR #1923, opened 2016-09-08
CAT-2109 Downloaded project not converting properly

michael-fischl:


catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
```diff
@@ -231,7 +231,7 @@ public Brick clone() {
 
 	@Override
 	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
-		sequence.addAction(sprite.getActionFactory().createSetLookAction(getSprite(), look, wait));
+		sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
```
p4p4: @michael-fischl 
could you check if the [SetBackgroundBrick](https://github.com/michael-fischl/Catroid/blob/852e9cc42ffeb64fcb4f63f5ef0c078a17846db6/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java) and [SetBackgroundAndWaitBrick](https://github.com/michael-fischl/Catroid/blob/852e9cc42ffeb64fcb4f63f5ef0c078a17846db6/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundAndWaitBrick.java) are still working? Thank you!
See PR #1874 and CAT-1968

robertpainsi: Checking something is always a great idea. :smirk_cat:

Since you commented on this line. The difference between `getSprite` and `sprite` is
* `getSprite` returns the current opened sprite in the IDE. There is always only _one_ current sprite. It's just a helper method. `getCurrentSprite` may be a better name because it's similar to `ProjectManager.getInstance().getCurrentSprite()`. Or somehow mention that's the current Sprite of the Ide :wink:.
* `sprite` is the Sprite where the brick is in.

The method [SetBackgroundBrick.java#L36](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java#L36) violates this. If there is only a single Scene, there won't be a problem. However, if there are multiple Scenes (`ProjectManager.getInstance().__getCurrentScene()__.getSpriteList().get(0);`), and the method is misused, it's harder to detect the error. I would suggest not to override it.

p4p4: I'm not sure if I understand you correctly.

`ProjectManager.getInstance().getCurrentScene().getSpriteList().get(0);` should always return the Background Sprite of the current scene, right? Can't I use that to access/change the background from *any other* Sprite?

robertpainsi:
As @p4p4 stated, this PR breaks the background functionality. :boom:

PR #1924, opened 2016-09-09
MUS-195 different color for black and white key rows

joseffilzmaier:


catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/NoteView.java
```diff
@@ -48,10 +52,22 @@ public void onClick(View v) {
...
 
 	public boolean isToggled() {
 		return toggled;
 	}
+
+	public int getUntoggledColor() {
+		return untoggledColor;
+	}
+
+	public void setUntoggledColor(int untoggledColor) {
```
fwinkelbauer: Are you going to use this set method in the future? If that's the case - nevermind! If not: please remove it :)

joseffilzmaier: No, this method gets removed with MUS-196, as the color is then replaced with an image.
I'll fix this now. :+1:

PR #1930, opened 2016-09-21
Add checkstyle check to use ellipsis character … instead of &#8230;

robertpainsi:
The ellipsis character … is much more readable and therefore should be
used in java and xml files instead of `&#8230;`. Translated strings.xml
files are ignored because they are handled by Crowdin.

### DIFF
--- a/catroid/config/checkstyle_suppressions.xml
+++ b/catroid/config/checkstyle_suppressions.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Catroid: An on-device visual programming system for Android devices
+  ~ Copyright (C) 2010-2016 The Catrobat Team
+  ~ (<http://developer.catrobat.org/credits>)
+  ~
+  ~ This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


catroid/config/checkstyle_suppressions.xml
```diff
@@ -0,0 +1,28 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  ~ GNU Affero General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<!DOCTYPE suppressions PUBLIC "-//Puppy Crawl//DTD Suppressions 1.1//EN" "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+  <suppress id="EllipsesCheck" files=".*/res/values-.*/strings.xml"/>
```
robertpainsi: Can someone on a Windows machine run Checkstyle locally as well. Maybe I have to use `[\\/]` for slashes.

PR #1932, opened 2016-09-22
CAT-FIX Correct value of month

robertriedl:


robertpainsi:
:+1:, except that tests are missing overall. [CAT-2139](https://jira.catrob.at/browse/CAT-2139)

Btw,
```java
return Double.valueOf(Calendar.getInstance().get(Calendar.DAY_OF_WEEK) - 1);
```
is incompatible with Scratch. The DAY_OF_WEEK in Scratch starts with Sunday, value 1, as it does in Java.

robertriedl:
Are there some tests needed? It even doesn't exist a test class for the SensorHandler and in all other classes are just some additions where also are no test for each element.

We discussed that with the UX-team and wolfgang, it was decided that the week should start with monday.

robertpainsi:
> Are there some tests needed?

Yes, always. Feel free to create tickets if you find code which isn't (well) tested.

> We discussed that with the UX-team and wolfgang, it was decided that the week should start with monday.

Just out of curiosity, what arguments lead to this decision?

* Currently Sunday  will be 0 but should be 7.
* I don't know the progress of the Scratch to Catrobat converter. If it already supports datetime stuff, this needs some special handling.

robertriedl:
> Just out of curiosity, what arguments lead to this decision?
I think it was that it could be irritating to the kids.

> Currently Sunday will be 0 but should be 7.
Thanks, i didn't think about that case.

> I don't know the progress of the Scratch to Catrobat converter. If it already supports datetime 
> stuff, this needs some special handling. Let them know.
done at GSoC-Meetings

robertpainsi:
>> Just out of curiosity, what arguments lead to this decision?

> I think it was that it could be irritating to the kids.

k. Luckily I don't have to decide such things. Considering that for **b**illions of people the week starts on Sunday and Scratch starts on Sunday as well, this is a really tough one.

>> I don't know the progress of the Scratch to Catrobat converter. If it already supports datetime stuff, this needs some special handling. Let them know.

> done at GSoC-Meetings

Awesome! :smiley_cat:

catroid/src/main/java/org/catrobat/catroid/formulaeditor/SensorHandler.java
```diff
@@ -397,11 +397,13 @@ public static Double getSensorValue(Sensors sensor) {
...
 			case DATE_MONTH:
-				return Double.valueOf(Calendar.getInstance().get(Calendar.MONTH));
+				return Double.valueOf(Calendar.getInstance().get(Calendar.MONTH) + 1);
 			case DATE_DAY:
 				return Double.valueOf(Calendar.getInstance().get(Calendar.DAY_OF_MONTH));
 			case DATE_WEEKDAY:
-				return Double.valueOf(Calendar.getInstance().get(Calendar.DAY_OF_WEEK) - 1);
+				return Double.valueOf(
+						(Calendar.getInstance().get(Calendar.DAY_OF_WEEK) == Calendar.SUNDAY) ? Calendar.SATURDAY :
+								Calendar.getInstance().get(Calendar.DAY_OF_WEEK) - 1);
```
robertpainsi: Please create a method for that. Preferable, but no must, with a parameter to pass in the day of week starting on Sunday.

robertpainsi: Ignore the review summary. Note to me:
* The review summary **can** be empty
* The review summary can't be edited :crying_cat_face:

fwinkelbauer: @robertpainsi If the review should be ignored you should click the "Approve changes" button :smile:

robertpainsi: @fwinkelbauer The review itself shouldn't be ignored. Just the review summary:
> Leave a comment... :smiley_cat:

There are still two things missing for review summaries
* Edit review summaries
* Directly comment on review summaries

I'm sure it will be available sooner or later. \*fingercrossed\*

fwinkelbauer: Oh I misunderstood ^^"

robertpainsi: > The review summary can't be edited 

The review summary __can__ be edited (now or always has been :smile_cat: ).

PR #1934, opened 2016-09-23
Refactored BrickAdapter and Brick handling

thmq:


robertpainsi:
`+445 −5,240` BAM! :metal:

sjaindl:
Cool!
+1 for this PR 👍

thmq:
Thanks. 😏 And this is only the beginning. A lot of things can be optimised further but in the spirit of continuous integration I decided to open a PR with what I've done so far.

Bioxar:
Nice :)

catroid/src/main/java/org/catrobat/catroid/content/bricks/BrickBaseType.java
```diff
@@ -23,30 +23,24 @@
...
-import org.catrobat.catroid.ui.BrickLayout;
 import org.catrobat.catroid.ui.adapter.BrickAdapter;
 
 import java.util.List;
 
 public abstract class BrickBaseType implements Brick {
 	private static final long serialVersionUID = 1L;
 	private static final String TAG = BrickBaseType.class.getSimpleName();
-	protected transient View view;
+	public transient View view;
```
fwinkelbauer: Does this have to be public?

catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
```diff
@@ -93,20 +91,17 @@
 	private int toEndDrag;
 	private boolean retryScriptDragging;
 	private boolean showDetails = false;
+	public boolean isDragging = false;
```
fwinkelbauer: Please turn this into a private member and add setter/getter methods

catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
```diff
@@ -1324,245 +1181,176 @@ public int getAmountOfCheckedItems() {
...
-				brickList.get(index).setCheckboxVisibility(visibility);
-			} else {
-				if (currentBrick instanceof ScriptBrick) {
-					currentBrick.setCheckboxVisibility(View.VISIBLE);
-				} else {
-					currentBrick.setCheckboxVisibility(View.INVISIBLE);
+	}
+
+	public void setCheckboxVisibility() {
+		switch (actionMode) {
```
fwinkelbauer: You could swap the order of the switch and the for loop statements:

```java
for (Brick brick : bricklist) {
  switch (actionMode) {
    // ...
  }
}
```

catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
thmq: this is only a temporary fix and should be changed asap. I'll have another look at it today so it might no longer be necessary anyway.

catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
thmq: thank you for pointing that out, I'll change it immediately.

PR #1941, opened 2016-09-28
CAT-FIX crash report dialog

jonny911:


catroid/build.gradle
```diff
@@ -59,6 +60,7 @@ android {
             buildConfigField "boolean", "FEATURE_ARDUINO_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_RASPI_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_POCKETMUSIC_ENABLED", "true"
+            buildConfigField "boolean", "FIREBASE_CRASH_REPORT", "false"
```
fwinkelbauer: Is there a good reason that this flag should be named different than the others? If not, please change it to something like `FEATURE_FIREBASE_CRASH_REPORT`

jonny911: It's not considered to be a feature...

thmq: In the spirit of consistency you should consider naming it `FIREBASE_CRASH_REPORT_ENABLED`. 😼

fwinkelbauer: :+1:

PR #1946, opened 2016-09-29
CAT-FIX Fix build errors in tests

robertpainsi:
This will not fix failing tests in BrickValueParameterTest.


catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/CollisionDetectionAdvancedTest.java
```diff
@@ -80,7 +80,7 @@ protected void initializeSprite(Sprite sprite, int resourceId, String filename)
 		collisionInformation.loadOrCreateCollisionPolygon();
 
 		sprite.look.setLookData(lookData);
-		sprite.addLookData(lookData);
+		sprite.getLookDataList().add(lookData);
```
robertpainsi: Normally I wouldn't do that :joy_cat:
but errors...!

PR #1949, opened 2016-10-03
CAT-2031 Dont display for bricks without description

FlorianWeissensteiner:
CAT-2031

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
@@ -49,6 +49,11 @@ import org.catrobat.catroid.content.bricks.BrickBaseType;
 import org.catrobat.catroid.content.bricks.BrickViewProvider;
 import org.catrobat.catroid.content.bricks.DeadEndBrick;
 import org.catrobat.catroid.content.bricks.FormulaBrick;
+import org.catrobat.catroid.content.bricks.IfLogicElseBrick;
+import org.catrobat.catroid.content.bricks.IfLogicEndBrick;
+import org.catrobat.catroid.content.bricks.IfThenLogicEndBrick;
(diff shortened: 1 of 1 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/adapter/BrickAdapter.java
```diff
@@ -1138,6 +1144,22 @@ private boolean brickHasAFormula(Brick brick) {
...
+		} else if (name.equals(IfThenLogicEndBrick.class.getSimpleName())) {
+			return true;
+		} else if (name.equals(LoopEndlessBrick.class.getSimpleName())) {
+			return true;
+		} else if (name.equals(LoopEndBrick.class.getSimpleName())) {
+			return true;
+		}
+		return false;
+	}
+
```
robertpainsi: Apply all the Android Studio inspection warnings (Yellow light bulbs).

Why do you compare the classes by `getSimpleName` instead of the `class` (`==`) itself or `instanceof`?

The return value of `isBrickWithoutDescription` is always used inverted. So invert the method name and adapt the code accordingly. `!isBrickWithoutDescription` -> `isBrickWithDescription`
Personally I prefer a name like `hasDescription(Brick brick)`.

FlorianWeissensteiner: Why do I compare the classes like that, good question. That is redundant. I was still in the mentality from the previous commit where I had to use the simple name for the URL.

I also agree with your suggestion according the rename of the function. I had not thought of that. 

If Thomas deems those changes worthy enough to un-merge the commit, I will be sure to adjust the code accordingly. Thanks!

robertpainsi: Thanks for the update!
No need to revert the commit. Whenever you find the time, just create a PR with the suggested changes. :smile_cat:

PR #1951, opened 2016-10-05
CAT-1741 Prevent standalone crash on first start

Chypsylon:


catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -430,6 +433,7 @@ public void setSignInDialog(SignInDialog signInDialog) {
 	}
 
 	private void unzipProgram() {
+		StorageHandler.getInstance();
```
sjaindl: For what do you need that line? Please remove.

Chypsylon: It's necessary to call the StorageHandler constructor so that the Pocketcode root directory is created in case it doesn't exist yet. I agree that it's not really obvious why that's done here.

PR #1935, opened 2016-09-23
fix CAT-1945 (set digital Arduino pins)

jobermue:

PR #1953, opened 2016-10-06
CAT-2142 Fix overwriting of whole port when setting single pin

jobermue:


robertpainsi:
ok to test

robertpainsi:
Jenkins doesn't seem to work at the moment.

Please run the Gradle task `check` and fix all reported warnings/violations introduced by your code only. There are only three Checkstyle warnings to fix.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -225,27 +229,27 @@ public void setAnalogArduinoPin(int pin, int value) {
...
-			}
+			PinNumberOfPort = digitalPinNumber - 8;
+		}
+
+		PortValue = arduinoListener.getuCPortValue(digitalPort);
+		if (pinValue > 0) { // set pin
+			PortValue = PortValue | (1 << PinNumberOfPort);
+			arduinoListener.setPortValue(digitalPinNumber, 1);
+		} else { // clear pin
+			PortValue = PortValue & ~(1 << PinNumberOfPort);
```
robertpainsi: If you want, you could create a util method to set a bit. Like
```java
public static int setBit(int number, int index, int value) {
  if (value == 0) {
    return number & ~(1 << index);
  } else {
    return number | (1 << index);
  }
}
```
When using the method somewhere in the code, the reader immediately knows what it does. Then you can also remove the comments here.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -171,4 +174,12 @@ public int getPortValue(int pin) {
...
 		this.portValue[pin] = value;
 	}
+
+	public int getuCPortValue(int port) {
+		return uCPortValue[port];
+	}
+
+	public void setuCPortValue(int port, int value) {
+		this.uCPortValue[port] = value;
+	}
```
robertpainsi: Is this method really necessary? I also doubt there should be a setter for `setPortValue`.
The port values should only be set in the `onAnalogMessageReceived`, right?

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -187,6 +187,10 @@ private void tryInitialize() throws IOException, SerialException {
 			sendFirmataMessage(new SetPinModeMessage(pin, SetPinModeMessage.PIN_MODE.PWM.getMode()));
 		}
 		reportSensorData(true);
+
+		// get status of digital ports
+		sendFirmataMessage(new ReportDigitalPortMessage(0, true));
+		sendFirmataMessage(new ReportDigitalPortMessage(1, true));
```
robertpainsi: Why do you send this firmata messages here?
Before reading a pin, the same firmata message is send anyway. (see [here](https://github.com/Catrobat/Catroid/blob/509c87e5041a5d173e439183c01b109b5e824726/catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java#L262))

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
robertpainsi: Stupid me, forget that ^^

robertpainsi:
@jobermue, I need your feedback on following scenario.

* Set pin#2 to 0.
  * If pin#2 is not connected to anything, the pin may be floating (or maybe not?). At least I can change the read value to 1 by adding a wire between pin#2 and 5V[1]. So reading pin#2 is either 0 or 1.
* Now read from and then write to an unrelated pin on the same port, e.g. pin#3.
  * [When reading a pin, the whole port is stored](https://github.com/jobermue/Catroid/blob/23d6b90259821a9d2a88c1cbe1a3eebe267d317a/catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java#L96). As stated above, the value for pin#2 is 0 or 1.
  * Writing to pin#3 will overwrite the whole port. 
* pin#2 may be changed to 0 or 1.

Is this a scenario which may happen?

[1] This may be a short curcuit. So add a resistor as well.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -45,6 +45,7 @@
 	private int analogPin5 = 0;
 
 	int[] portValue = new int[]{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+	int[] uCPortValue = new int[] { 0, 0 };
```
jobermue: I renamed portValue (and associated getter&setter) to pinValue as this reflects its actual usage. I then renamed uCPortValue to portValue (the name that probably fits best and thatI wanted to use initially, but was already taken).

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -171,4 +174,12 @@ public int getPortValue(int pin) {
 	public void setPortValue(int pin, int value) {
 		this.portValue[pin] = value;
 	}
+
+	public int getuCPortValue(int port) {
```
jobermue: renamed & adapted to camel case

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -225,27 +229,27 @@ public void setAnalogArduinoPin(int pin, int value) {
...
-				sendDigitalFirmataMessage(digitalPort, digitalPinNumber, (int) value);
-				arduinoListener.setPortValue(digitalPinNumber, 1);
-			} else {
-				sendDigitalFirmataMessage(digitalPort, digitalPinNumber, 0);
-				arduinoListener.setPortValue(digitalPinNumber, 0);
-			}
+			PinNumberOfPort = digitalPinNumber - 8;
+		}
+
+		PortValue = arduinoListener.getuCPortValue(digitalPort);
```
jobermue: ok

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
jobermue: implemented this as suggested

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
jobermue: After some testing I removed it in my most recent commit, as it is unnecessary.
I initially thought this was necessary to get the current state of the outputs (e.g. after restarting a program) - but this actually works without these two lines. I'm not sure why, but I suppose this is because AsuroImpl keeps running in the background even if  the user's program is restarted?

jobermue:
Thanks for your feedback!
I tried to fix the problems pointed out in your comments.
I still have some troubles with formatting though, as I'm not able to run gradle check (tried from commandline in Catroid directory, but it failed). Is it possbile to run "gradle check" from inside Android Studio (apart from terminal window) and if so, how?
I added lots of commits with the fixes now - shall I squash them all into one commit?

Regarding your question: only input pins (without pull-up or pull-down resistor) may float, an output pin will not do so - unless you create a situation like the one described, where multiple output pins are connected together (which should normally not be the case). Then whoever is stronger defines the signal on the common connection. 
That's actually the reason why input pins can be in a High-z state (causing the floating if not connected to anything), so that they don't influence the signal level on a common bus.

robertpainsi:
> Is it possbile to run "gradle check" from inside Android Studio (apart from terminal window) and if so, how?

Yes, see [here](https://alincode.files.wordpress.com/2014/06/screen-shot-2014-06-30-at-12-20-50-am.png). In the current AS version it may be on the bottom of the right panel.
Catroid > Catroid > Tasks > verification > check
The task `check` will run Checkstyle, PMD, Lint,… and has to run without errors, warnings or reports. You have to manually inspect the output which can be found in the Gradle Console, lower panel on the right. See [here](https://alincode.files.wordpress.com/2014/06/screen-shot-2014-06-30-at-12-20-50-am.png). You can also run each check individually. e.g. Checkstyle and PMD are in the folder `others`.

> I added lots of commits with the fixes now - shall I squash them all into one commit?

:+1: 

> Regarding your question: only input pins…

Thanks for the explanation. :smile_cat: So let us drop this scenario.

jobermue:
I've run check (lint aborts on error not related to my code) as well as checkstyle and lint individually - none reports any problems in my code. Thus I've now squashed all my commits into a single commit and pushed that to this branch. IMHO it should be ready for merging now, please check if there are any problems left.

jobermue:
I noticed that because I renamed 'portValue' to 'pinValue' in this pull request (which IMHO makes sense as the array actually stores pin values - and helps getting rid of the clumsy 'ucPortValue' name), I can't provide the fix for pin7 now (without causing a merge conflict), as this would access that variable, but develop still uses 'portValue'. Any suggestions on how to proceed?

robertpainsi:
LGTM!

> Any suggestions on how to proceed?

Unfortunately there is no `Rebase and merge --no-ff` option on Github yet :crying_cat_face:

You could branch of from this PR, add a fix commit and open a PR now. Then the changes are already reviewable. When the other PRs are merged, I'll drop a comment on the new PR, so you just have to rebase the branch or, if you make changes in this PR, cherry-pick the commit. Overall, it'll be simple and smooth.

Another option is, to create a Jira ticket which is blocked by the other two Jira tickets. Also add a link to the fixing commit.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -222,30 +222,36 @@ public void setAnalogArduinoPin(int pin, int value) {
...
 	}
 
+	public static int setBit(int number, int index, int value) {
+		if (value == 0) {
+			return number & ~(1 << index);
+		} else {
+			return number | (1 << index);
+		}
+	}
+
```
robertpainsi: Please create a test for this method.

There is one ui test class for the Arduino feature. I've created a ticket because the Arduino feature can and should be tested by unit tests. [CAT-2196](https://jira.catrob.at/browse/CAT-2196).

@jobermue @thmq, is there someone who can review the Arduino changes? Someone who is familiar with the Arduino code and also can test it with an Arduino.

jobermue:
I rebased this pull request on [#1935](https://github.com/Catrobat/Catroid/pull/1935), did the refactoring of the two arrays and added some tests. PMD and checkstyle run without any problems, lint reports quite a few problems but none of them is related to the code I added/modified.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -164,11 +155,28 @@ public int getAnalogPin5() {
...
+		}
+		return 0;
+	}
+
+	public void setPinValue(int pin, int value) {
+		if ((pin >= 0) && (pin < ArduinoImpl.NUMBER_OF_DIGITAL_PINS)) {
+			int port = (pin + 7) / 8;
+			int index = pin % 8;
+			this.portValue[port] = ArduinoImpl.setBit(this.portValue[port], index, value);
+		}
```
robertpainsi: `getPinValue` and `setPinValue` share a lot of code.
I would suggest you create 3 methods. A method...
* for the guard clause if the pin is in range.
* to get the port. You can reuse it to create the size of the port array and if still needed, in `ArduinoImpl.setDigitalArduinoPin`
* to get the pin on the port. Can be reused in `ArduinoImpl.setDigitalArduinoPin` as well, if still needed.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -164,11 +155,28 @@ public int getAnalogPin5() {
...
+	}
+
+	public int getPortValue(int port) {
+		return portValue[port];
 	}
 
-	public void setPortValue(int pin, int value) {
-		this.portValue[pin] = value;
+	public void setPortValue(int port, int value) {
+		this.portValue[port] = value;
```
robertpainsi: Remove setter. The port pin values are either set through `setPinValue` or updated in `onDigitalMessageReceived`.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -164,11 +155,28 @@ public int getAnalogPin5() {
...
+	public void setPinValue(int pin, int value) {
+		if ((pin >= 0) && (pin < ArduinoImpl.NUMBER_OF_DIGITAL_PINS)) {
+			int port = (pin + 7) / 8;
+			int index = pin % 8;
+			this.portValue[port] = ArduinoImpl.setBit(this.portValue[port], index, value);
+		}
+	}
+
+	public int getPortValue(int port) {
+		return portValue[port];
```
robertpainsi: I'm not sure how the method `setDigitalArduinoPin` will change.
Maybe it makes sense to replace this method by `getPortValueByPin`.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -87,24 +87,15 @@ public void onDigitalMessageReceived(DigitalMessage message) {
...
+				portValue[0] = message.getValue();
 				break;
 			case ArduinoImpl.PORT_DIGITAL_1:
-				portValue [8] = (message.getValue() & 0x1) == 0 ? 0 : 1;
-				portValue [9] = (message.getValue() & 0x2) == 0 ? 0 : 1;
-				portValue [10] = (message.getValue() & 0x4) == 0 ? 0 : 1;
-				portValue [11] = (message.getValue() & 0x8) == 0 ? 0 : 1;
-				portValue [12] = (message.getValue() & 0x16) == 0 ? 0 : 1;
-				portValue [13] = (message.getValue() & 0x32) == 0 ? 0 : 1;
+				portValue[1] = message.getValue();
```
robertpainsi: `portValue[0]` is aka `portValue[PORT_DIGITAL_0]` and `portValue[1]` is aka `portValue[PORT_DIGITAL_1]`.
So you could remove the switch statement and use `portValue[message.getPort()]` instead.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
```diff
@@ -164,11 +155,28 @@ public int getAnalogPin5() {
 		return analogPin5;
 	}
 
-	public int getPortValue(int pin) {
-		return portValue[pin];
+	public int getPinValue(int pin) {
+		if ((pin >= 0) && (pin < ArduinoImpl.NUMBER_OF_DIGITAL_PINS)) {
+			int port = (pin + 7) / 8;
```
robertpainsi: Is this correct? Why you're adding 7 to the pin?

Please create a variable for the magic number :8ball:.

jobermue: no, it's wrong - thanks for pointing that out.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
jobermue: In principle a good idea, but in `setDigitalArduinoPin` I need to get the port anyway (it is needed as parameter of `sendDigitalFirmataMessage`) - thus this would result in calculating the port from the pin twice instead of once.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -222,30 +224,46 @@ public void setAnalogArduinoPin(int pin, int value) {
...
-				arduinoListener.setPortValue(digitalPinNumber, 1);
-			} else {
-				sendDigitalFirmataMessage(digitalPort, digitalPinNumber, 0);
-				arduinoListener.setPortValue(digitalPinNumber, 0);
-			}
+			pinNumberOfPort = digitalPinNumber - 8;
+		}
+
+		int portValue = arduinoListener.getPortValue(digitalPort);
+		portValue = setBit(portValue, pinNumberOfPort, pinValue);
```
jobermue: ok, changed that.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
jobermue: ok.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
jobermue: ok. Which of the following styles do you prefer:
`public int getPinValue(int pin) {
		if (ArduinoImpl.isValidPin(pin)) {
			return Utils.getBit(portValue[ArduinoImpl.getPortFromPin(pin)], ArduinoImpl.getIndexOfPinOnPort(pin));
		}
		return 0;
	}

	public void setPinValue(int pin, int value) {
		if (ArduinoImpl.isValidPin(pin)) {
			int port = ArduinoImpl.getPortFromPin(pin);
			int index = ArduinoImpl.getIndexOfPinOnPort(pin);
			this.portValue[port] = Utils.setBit(portValue[port], index, value);
		}
	}`

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -222,30 +224,46 @@ public void setAnalogArduinoPin(int pin, int value) {
...
+		}
+		return number;
+	}
+
+	public static int getBit(int number, int index) {
+		if ((index >= 0) && (index < 32)) {
+			return (number >> index) & 0x1;
+		}
+		return 0;
+	}
```
jobermue: ok.

jobermue:
I just changed the stuff you pointed out. I added it as a separate commit for now, so you can see the diff and will squash everything into a single commit as soon as the code is in its final state. Please see also the question regarding code style in one of my comments to your requests.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoListener.java
robertpainsi: > Which of the following styles do you prefer?

It's completely up to you :wink:

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -293,6 +273,22 @@ public double getAnalogArduinoPin(int analogPinNumber) {
 		return 0;
 	}
 
+	public static boolean isValidPin(int pin) {
+		if ((pin >= 0) && (pin < NUMBER_OF_DIGITAL_PINS)) {
+			return true;
+		} else {
+			return false;
+		}
+	}
```
robertpainsi: Please simplify (remove if/else)

jobermue:
Is there anything else you want me to change (as you haven't changed the "changes requested" status of this PR)?
Otherwise I would squash everything into a single commit so it can be merged.

robertpainsi:
LGTM!

As already mentioned, I can't test the new code because I don't have the equipment to do so and we should also create unit tests for the Arduino feature [CAT-2196](https://jira.catrob.at/browse/CAT-2196).

jobermue:
smashed everything into a single commit.

robertpainsi:
Testrun https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/1648/ (Checkstyle warnings not related to PR)

PR #1961, opened 2016-10-13
Fix default scene name string in all languages and first scene string

loibi93:
Resolve issues with NewSceneDialog and First Scene exclamation marks.


thmq:
this does not really make sense. you do not need a variable placeholder in the scenes string if you replace it with a whitespace anyway. you should have removed the placeholder altogether. 

therefore this is replaced by #1962 .

PR #1964, opened 2016-10-16
Spanish translation for create@school-Strings

sjaindl:


catroid/src/createatschool/res/values-es/strings.xml
```diff
@@ -0,0 +1,276 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<!--Generated by crowdin.com-->
```
thmq: please remove this line 😉

sjaindl: Removed ;)

PR #1965, opened 2016-10-16
CAT-FIX Hotfix variables/lists for release

sjaindl:
- Hotfix variables/lists (remove null and double entries when loading old programs)
- Convert sprites of old programs to SingleSprites in XStreamSpriteConverter to prevent double entries in dataContainer

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/io/XStreamSpriteConverter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/io/XStreamSpriteConverter.java
@@ -32,6 +32,7 @@ import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
 import com.thoughtworks.xstream.mapper.Mapper;
 
+import org.catrobat.catroid.content.SingleSprite;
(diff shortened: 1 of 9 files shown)


catroid/src/main/java/org/catrobat/catroid/io/XStreamSpriteConverter.java
```diff
@@ -66,6 +67,14 @@ public Object doUnmarshal(Object result, HierarchicalStreamReader reader, Unmars
 			} catch (ClassNotFoundException exception) {
 				Log.e(TAG, "Sprite class not found : " + result.toString(), exception);
 			}
+		} else {
+			try {
+				Class cls = Class.forName(SPRITES_PACKAGE_NAME + "." + SpriteFactory.SPRITE_SINGLE);
+				Sprite sprite = (Sprite) reflectionProvider.newInstance(cls);
```
robertpainsi: I havent tested it, but would something like
```java
Sprite sprite = (Sprite) reflectionProvider.newInstance(SingleSprite.class);
```
work as well?

sjaindl: Good point! Yes, that works as well.

PR #1969, opened 2016-10-17
HOTFIX GoTo brick crash with clone - spinner value not copied

p4p4:
The GoTo brick did not copy the spinner-selection, which caused crashes when
appearing in a WhenCloned script (and potentially also in other features
like backpack, copy brick etc). This should fix the issue.

This is a very HOT HotFix ;)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
@@ -43,6 +43,7 @@ import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.ui.controller.BackPackSpriteController;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class GoToBrick extends BrickBaseType {
@@ -51,8 +52,6 @@ public class GoToBrick extends BrickBaseType {
 
 	private Sprite destinationSprite;
 	private transient String oldSelectedObject;
-	private String touchPositionLabel;
-	private String randomPositionLabel;
 
 	private transient SpinnerAdapterWrapper spinnerAdapterWrapper;
 	private int spinnerSelection;
@@ -86,9 +85,6 @@ public class GoToBrick extends BrickBaseType {
 
 		view = View.inflate(context, R.layout.brick_go_to, null);
 		view = BrickViewProvider.setAlphaOnView(view, alphaValue);
-
-		this.touchPositionLabel = context.getString(R.string.brick_go_to_touch_position);
-		this.randomPositionLabel = context.getString(R.string.brick_go_to_random_position);
 
 		setCheckboxView(R.id.brick_go_to_checkbox);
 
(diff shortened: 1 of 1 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -305,7 +306,9 @@ public boolean onTouch(View paramView, MotionEvent paramMotionEvent) {
 
 	@Override
 	public Brick clone() {
-		return new GoToBrick(destinationSprite);
+		GoToBrick copy = new GoToBrick(destinationSprite);
+		copy.spinnerSelection = spinnerSelection;
+		return copy;
```
robertpainsi: Should the other members be cloned as well, especially the non-transient?

p4p4: probably.  I did not (yet) have time to check if this brick contains other bugs as well *let me have a look*

p4p4: copied 
``` java
	private String touchPositionLabel;
	private String randomPositionLabel;
```
as well.

but IMHO this should be refactored when there is more time (WHY not use something static instead).

p4p4: thx for the hint, made the labels transient for now. However, this should still be refactored in a future PR.

robertpainsi: Great idea!
The `setSpinnerSelection` method is called from methods where there is a context. Instead of getting the context from the `CatroidApplication`, you could pass in a parameter context instead.

p4p4: thanks for the hint, I passed the context as a parameter.

catroid/src/main/java/org/catrobat/catroid/content/bricks/GoToBrick.java
```diff
@@ -152,19 +148,19 @@ public View getPrototypeView(Context context) {
 	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
 		sequence.addAction(sprite.getActionFactory().createGoToAction(sprite, destinationSprite, spinnerSelection));
 
-		return null;
+		return Collections.emptyList();
```
p4p4: Not it is not necessary, it's just suggested to not return or pass around null and use empty collections instead. And since the file was already open, I "left it a bit more nice than I found it" ;)

robertpainsi: Hehe, yeah! However

>  "Ensure no-op code refactoring is done separately from functional changes"

or better, create atomic commits. The improvement should be in it's own commit. It's just easier to review. Otherwise if someone looks at it later, he will think this change is necessary for the fix.

PR #1970, opened 2016-10-17
HOTFIX conversion problems special chars in project name

p4p4:
Programs of an older catrobat language version are converted to the new one.
For this purpose, it is necessary to create a DefaultScene.
The migration did not copy images and sounds due to double-escaping the path.
This led to situations where e.g. "Foo:bar" was first escaped to "Foo%3ABar" and then in the second
step, the '%' was (erroneously) escaped again, resulting in "Foo%253ABar", which is NOT a valid file path anymore. This commit fixes that issue.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
+++ b/catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
@@ -613,8 +613,6 @@ public final class StorageHandler {
 	}
 
 	private void fixFolderStructureForSupportProject(String projectName, String sceneName) throws IOException {
-		projectName = UtilFile.encodeSpecialCharsForFileSystem(projectName);
-		sceneName = UtilFile.encodeSpecialCharsForFileSystem(sceneName);
 		String projectPath = buildProjectPath(projectName);
 		String scenePath = buildScenePath(projectName, sceneName);
 		File looksDirectory = new File(buildPath(projectPath, IMAGE_DIRECTORY));


catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
```diff
@@ -613,9 +613,8 @@ public static void copyDirectory(File destinationFile, File sourceFile) throws I
 	}
 
 	private void fixFolderStructureForSupportProject(String projectName, String sceneName) throws IOException {
-		projectName = UtilFile.encodeSpecialCharsForFileSystem(projectName);
-		sceneName = UtilFile.encodeSpecialCharsForFileSystem(sceneName);
-		String projectPath = buildProjectPath(projectName);
+		String projectNameEscaped = UtilFile.encodeSpecialCharsForFileSystem(projectName);
+		String projectPath = buildPath(Constants.DEFAULT_ROOT, projectNameEscaped);
```
robertpainsi: There is a helper method to create an escaped path for a project.
[`Utils.buildProjectPath(String projectName)`](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java#L349-L351)

Thanks for the awesome commit message! It was easy peasy lemon squeezy to review! :bowing_man:

p4p4: thanks for pointing this out, this makes the fix even smaller (only 2 deletions) ;)

PR #1971, opened 2016-10-17
CAT-2149 Fixes remix-data in header of merged programs

rsamer:
For a detailed description about the issue, please see the following URL: https://jira.catrob.at/browse/CAT-2149

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/UtilsTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/UtilsTest.java
@@ -35,6 +35,7 @@ import org.catrobat.catroid.content.Script;
 import org.catrobat.catroid.content.SingleSprite;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.content.WhenScript;
+import org.catrobat.catroid.content.XmlHeader;
 import org.catrobat.catroid.content.bricks.Brick;
 import org.catrobat.catroid.content.bricks.HideBrick;
 import org.catrobat.catroid.content.bricks.SetLookBrick;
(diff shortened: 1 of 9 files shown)


rsamer:
Fixed merge conflict. Please review again.

catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/UtilsTest.java
```diff
@@ -212,6 +213,188 @@ public void testLoadProjectIfNeeded() {
...
+	}
+
+	public void testExtractRemixUrlsOfRemergedProgramHeaderUrlFieldContainingMixedUrls() {
+		final String expectedFirstProgramRemixUrl = "https://scratch.mit.edu/projects/117697631/";
+		final String expectedSecondProgramRemixUrl = "/pocketcode/program/3570";
+		final String expectedThirdProgramRemixUrl = "https://scratch.mit.edu/projects/121648946/";
+		final String expectedFourthProgramRemixUrl = "https://share.catrob.at/pocketcode/program/16267";
+
+		final XmlHeader headerOfFirstProgram = new XmlHeader();
+		headerOfFirstProgram.setProgramName("やねうら部屋(びっくりハウス) remix お化け屋敷");
```
robertpainsi: Please use Latin characters and English words.

catroid/src/main/java/org/catrobat/catroid/content/XmlHeader.java
```diff
@@ -64,12 +64,26 @@
...
+	//  Keep in mind that the remixOf-field is used and supposed to be modified
+	//  by Catroweb's web application *only*!
+	//  Once new Catrobat programs get uploaded, Catroweb automatically
+	//  updates the remixOf-field and sets the program as being remixed.
+	//  In order to do so, Catroweb takes the value from the remixUrlsString-field and
+	//  assigns it to the remixOf-field.
+	//
+	//  With that said, the only correct way to set a new remix-URL (e.g. when
+	//  two programs get merged locally) is to assign it to the remixUrlsString-field.
+	//---------------------------------------------------------------------------
```
robertpainsi: Yes, but why?

robertpainsi: ^^ I liked your comment and in this case it's definitely necessary. It isn't obvious how the field is handled at all.

My question is still _why?_
Why you reuse the url to temporary save the remix info? Why you set the remixOf information on the server?

rsamer:
Ok, this is because the program ID of the uploaded program is generated on the server. The program URL then contains the program ID.
You don't know the program's ID before the program is uploaded.

Let's say when you download a program e.g. Galaxy War from the server then you actually downloaded a copy of Galaxy War and not the original Galaxy War program itself.
This copy is nothing else then a remix of the original Galaxy War program.

With that said, the url in the url-XML-header field does not refer to your copy but refers to the original Galaxy War program (that still lies on the server).
That means, no other user can possess the original program, only the server and the creator of the program own the original program.

But there is a minor and essential difference between the original program the creator has and the original program on the server:
the creator's program's XML file contains the URLs of its parents in the url-field and the server's original program's XML file contains the true URL of this original program in the url-field.
Also the remixOf fields urls differ and the username-fields.
The key problem is you don't know the program ID until the user uploads the program.

In order to work around this issue, one could generate UUIDs locally in Catroid instead, but they won't be compatible with Catroweb's ID "management" unless Catroweb plans to support this.

Hope this answers your open questions.
Sorry, for being so verbose.
I know that you are smart enough to easily figure that out yourself.
But I just wanted to save you some time, because I think you cannot be aware of all our ideas concerning remixing and probably don't have an in-depth view of the Catroweb project since you are not member of our team.

robertpainsi:
Thanks for the detailed explanation! :+1:

rsamer:
Ready to merge?

PR #1972, opened 2016-10-17
CAT-1384 HOTFIX cloned object look: copy all properties 

p4p4:
Previously a cloned object always appeared at the origin of the coordinate
system. Now the position is copied as well.
Also copied: transparancy, rotation, color
Last minute request by @wslany for more scratch compatibility in clone brick

Edit: this seems to be an old ticket: [CAT-1384](https://jira.catrob.at/browse/CAT-1384)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -251,7 +251,7 @@ public class StageListener implements ApplicationListener {
 
 	public void cloneSpriteAndAddToStage(Sprite cloneMe) {
 		Sprite copy = cloneMe.cloneForCloneBrick();
-		copy.resetSprite();
+		//copy.resetSprite(); // can I safely remove this?
 		copy.look.createBrightnessContrastHueShader();
 		stage.addActor(copy.look);
 		copy.resume();
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -138,6 +138,25 @@ public class Look extends Image {
 		cloneLook.whenParallelAction = null;
 		cloneLook.allActionsAreFinished = this.allActionsAreFinished;
 
+		cloneLook.setPositionInUserInterfaceDimensionUnit(this.getXInUserInterfaceDimensionUnit(),
+				this.getYInUserInterfaceDimensionUnit());
+		cloneLook.setTransparencyInUserInterfaceDimensionUnit(this.getTransparencyInUserInterfaceDimensionUnit());
+		cloneLook.setColorInUserInterfaceDimensionUnit(this.getColorInUserInterfaceDimensionUnit());
+
+		int rotationMode = this.getRotationMode();
+		cloneLook.setRotationMode(rotationMode);
+		if (rotationMode != Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && this.isFlipped()) {
+			cloneLook.getLookData().getTextureRegion().flip(true, false);
+			cloneLook.setFlipped(false);
+		}
+		boolean orientedLeft = this.getDirectionInUserInterfaceDimensionUnit() < 0;
+		if (rotationMode == Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && orientedLeft) {
+			cloneLook.getLookData().getTextureRegion().flip(true, false);
+			cloneLook.setFlipped(true);
+		}
+		cloneLook.setDirectionInUserInterfaceDimensionUnit(this.getDirectionInUserInterfaceDimensionUnit());
+		cloneLook.setBrightnessInUserInterfaceDimensionUnit(this.getBrightnessInUserInterfaceDimensionUnit());
+
 		return cloneLook;
 	}
 


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -265,6 +265,24 @@ public void cloneSpriteAndAddToStage(Sprite cloneMe) {
...
+			copy.look.getLookData().getTextureRegion().flip(true, false);
+			copy.look.setFlipped(false);
+		}
+		boolean orientedLeft = copy.look.getDirectionInUserInterfaceDimensionUnit() < 0;
+		if (rotationMode == Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && orientedLeft) {
+			copy.look.getLookData().getTextureRegion().flip(true, false);
+			copy.look.setFlipped(true);
+		}
+		copy.look.setDirectionInUserInterfaceDimensionUnit(cloneMe.look.getDirectionInUserInterfaceDimensionUnit());
+
```
robertpainsi: Both, cloning a look through [the IDE](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L335) or [the clone brick](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L362), will call [Sprite.cloneLooks(Sprite)](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L490) which will call [Look.copyLookForSprite(Sprite)](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L132) which already sets some of the look members. For consistency and also to fix [CAT-1384](https://jira.catrob.at/browse/CAT-1384), it would be awesome if you adapt the [Look.copyLookForSprite(Sprite)](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L132) method instead.

p4p4: I like that idea! :)  Unfortunately, the clone calls [Sprite.resetSprite()](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L172) later on, which overwrites the look again. Do you see any reason why this method should be called at all after a sprite is cloned (I'm not sure about the physics stuff...)?

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -251,7 +251,7 @@ public void create() {
 
 	public void cloneSpriteAndAddToStage(Sprite cloneMe) {
 		Sprite copy = cloneMe.cloneForCloneBrick();
-		copy.resetSprite();
+		//copy.resetSprite(); // can I safely remove this?
```
p4p4: can I safely remove the call to [Sprite.resetSprite()](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L172)?

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
robertpainsi: Good catch!
The biggest flaw in the current implementation is, that the Look needs a reference to the Sprite, which isn't necessary at all. Currently it's used for

* `sprite.isPaused` → Either use [Actor.setTouchable](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/scenes/scene2d/Actor.html#setTouchable-com.badlogic.gdx.scenes.scene2d.Touchable-) or move the isPaused member to the Look. It would be even better to disable touch events on the whole stage.
* `sprite.createWhenScriptActionSequence` → The Sprite should register a listener on the Look, which for simplicity just calls `createWhenScriptActionSequence`

Unfortunately a simple `Look.clone()` method still can't be created. The PhysicsLook needs a fresh PhysicsObject as parameter when created... If you come up with a solution for this, let me know. :smile_cat:

> In other words, is it OK to remove this line?

The method is used to create a Look or PhysicsLook depending on the used Bricks. Is it possible to move the method to `cloneForCloneBrick`? It's important that `resetSprite` is called __after__ the bricks have been cloned. Otherwise the wrong Look will be used.

1. clone bricks
2. call `resetSprite`
3. call `Look.copyLookForSprite` (also override this method in the PhysicsLook to clone friction, bounce factor, etc).

Hopefully this will work!? :hushed:

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
robertpainsi: I guess not, see https://github.com/Catrobat/Catroid/pull/1972#discussion_r83835195

But please test the clone feature with a Sprite containing a physics brick, e.g. SetBounceBrick.
It's important that the cloned sprite also executes such a brick _after_ it has been cloned. I guess a [RuntimeException("No physics action available in non-physics sprite!");](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java#L764) will be thrown.

robertpainsi: My guess was wrong. The Problem is not a [RuntimeException("No physics action available in non-physics sprite!");](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java#L764).
The problem won't be an exception but strange behavior. The created Look would be a normal Look which is not connected to the Physics Engine. So it will behave just like a normal Look. (No collision or forces like gravity).

p4p4: No, your guess was quite good. At least with certain bricks a ClassCastException was thrown (while trying to cast a `Look` (base-type) as a `PhysicsLook`).

robertpainsi: Ah, of course, thanks!

PR #1973, opened 2016-10-19
HOTFIX clone looks with physics bricks

p4p4:
after a Sprite is cloned (using the clone brick), it is necessary to reset the sprite because it might contain Physics-bricks. This creates a PhysicsLook instead of a Look (otherwise Pocket Code would crash). The look properties have to be copied *after* the instantiation of the look.

After this PR, please also merge #1980 

suggested by @robertpainsi in #1972

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -251,7 +251,6 @@ public class StageListener implements ApplicationListener {
 
 	public void cloneSpriteAndAddToStage(Sprite cloneMe) {
 		Sprite copy = cloneMe.cloneForCloneBrick();
-		//copy.resetSprite(); // can I safely remove this?
 		copy.look.createBrightnessContrastHueShader();
 		stage.addActor(copy.look);
 		copy.resume();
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -251,7 +251,13 @@ public void create() {
...
 		Sprite copy = cloneMe.cloneForCloneBrick();
-		//copy.resetSprite(); // can I safely remove this?
+
+		copy.resetSprite();
+		Sprite originalSprite = ProjectManager.getInstance().getCurrentSprite();
+		ProjectManager.getInstance().setCurrentSprite(copy);
+		cloneMe.cloneLooks(copy);
+		ProjectManager.getInstance().setCurrentSprite(originalSprite);
+
 		copy.look.createBrightnessContrastHueShader();
```
robertpainsi: I __cannot test it__, but it LGTM for a hot-fix!

You could also move that code in the `Sprite.cloneForCloneBrick` itself so you don't have to care about the `getCurrentSprite` and `setCurrentSprite` but that's up to you.

p4p4: done :+1:

robertpainsi:
@thmq 
This PR wasn't ready to merge yet. It'll fix the problem by creating the correct Look (Look or PhysicsLook) but creates another problem. The LookData references in e.g. SetLookBrick of the clone point to a origin LookData and not to a cloned one. See detailed description and proposed solution above.

thmq:
Well, that's a problem. I was told to merge it by @bernadettespieler. But I was expecting problems. Anyway, what should we do then? revert it?

PR #1974, opened 2016-10-19
HOTFIX set Background brick did not work anymore

p4p4:
1. commit introduced in PR #1923 broke the SetBackgroundBrick
    -> fixed
2. commit 6c25a4e448e8f0dc165a05b688e6271d87d5784f replaced wrong strings on brick: ´
      should be `Switch to Background [...] and wait`, not `Switch to look [...] Background`

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java
@@ -23,8 +23,13 @@
 
 package org.catrobat.catroid.content.bricks;
 
+import com.badlogic.gdx.scenes.scene2d.actions.SequenceAction;
+
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.content.Sprite;
+
+import java.util.Collections;
+import java.util.List;
 
 public class SetBackgroundBrick extends SetLookBrick {
 
@@ -34,6 +39,14 @@ public class SetBackgroundBrick extends SetLookBrick {
 	@Override
 	protected Sprite getSprite() {
 		return ProjectManager.getInstance().getCurrentScene().getSpriteList().get(0);
+	}
+
+	@Override
+	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
+		Sprite backgroundSprite = ProjectManager.getInstance().getSceneToPlay().getSpriteList().get(0);
+		sequence.addAction(sprite.getActionFactory().createSetLookAction(backgroundSprite, look, wait));
+
+		return Collections.emptyList();
 	}
 
 	@Override
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
@@ -131,7 +131,7 @@ public class SetLookBrick extends BrickBaseType implements OnLookDataListChanged
 		setSpinnerSelection(lookBrickSpinner);
 
 		if (getSprite().getName().equals(context.getString(R.string.background))) {
-			TextView textField = (TextView) view.findViewById(R.id.brick_set_look_and_wait);
+			TextView textField = (TextView) view.findViewById(R.id.brick_set_look_prototype_text_view);
 			textField.setText(R.string.brick_set_background);
 		}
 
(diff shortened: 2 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java
```diff
@@ -32,7 +32,7 @@ public SetBackgroundBrick() {
 	}
 
 	@Override
-	protected Sprite getSprite() {
+	protected Sprite getSpriteToChange() {
```
robertpainsi: Please revert this change. It's not important for the current release. I'll explain to you what I meant at https://github.com/Catrobat/Catroid/pull/1923#discussion_r83840129 later. :wink:

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
```diff
@@ -182,18 +182,23 @@ public Brick clone() {
 
 	@Override
 	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
-		sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
+		if (this instanceof SetBackgroundBrick) {
+			Sprite backgroundSprite = getSpriteToChange();
+			sequence.addAction(sprite.getActionFactory().createSetLookAction(backgroundSprite, look, wait));
+		} else {
+			sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
+		}
```
robertpainsi: Is this change really necessary?
The sprite should always be the correct sprite, either a normal sprite or a background.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
```diff
@@ -182,7 +182,8 @@ public Brick clone() {
 
 	@Override
 	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
-		sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
+		sequence.addAction(sprite.getActionFactory().createSetLookAction(getSprite(), look, wait));
```
robertpainsi: Please revert

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
p4p4: ok then back to you initial question:
 - yes I think it should be necessary to do this:

if this is a `SetBackgroundBrick`, then use `getSprite()`, which [is overridden](https://github.com/Catrobat/Catroid/blob/b5a9fc8323373031b4d561cba3d69de77d094319/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java#L36) for `SetBackgroundBrick` and always returns the background sprite.
in all other cases, the passed `sprite` is used.

Apparantly it seems to be time to stop working for today... :smiley:

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
```diff
@@ -182,7 +182,12 @@ public Brick clone() {
 
 	@Override
 	public List<SequenceAction> addActionToSequence(Sprite sprite, SequenceAction sequence) {
-		sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
+		if (this instanceof SetBackgroundBrick) {
+			Sprite backgroundSprite = getSprite();
+			sequence.addAction(sprite.getActionFactory().createSetLookAction(backgroundSprite, look, wait));
+		} else {
+			sequence.addAction(sprite.getActionFactory().createSetLookAction(sprite, look, wait));
+		}
```
robertpainsi: You could also override the method in the SetBackgroundBrick :wink:

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookBrick.java
robertpainsi: We definitely have to add some tests for the BackgroundBricks

p4p4: you are right, for the action it is necessary to use the "SceneToPlay", whereas the Spinners in the IDE need the "CurrentScene".

If anybody has time for writing a test she is welcome to do so.
My previous hotfixes were only on a voluntary basis to help out for the next release :grinning:

robertpainsi:
@p4p4 thanks for fixing it!

PR #1980, opened 2016-10-24
Set look data when cloning

robertpainsi:
This Pull Requests builds on https://github.com/Catrobat/Catroid/pull/1973 and is just a info that it should be merged before the next release. After https://github.com/Catrobat/Catroid/pull/1973 has been merged, I'll rebase and update the changes accordingly. I'll add a comment when everything is ready and also will remove this text.

###
When cloning a Sprite in the Stage or IDE [1], all current values of the
Sprite should be copied as well. This commit only addresses the current
LookData because it will fix a NullPointerException.

`E/BaseExceptionHandler: unhandled exception
  java.lang.NullPointerException: Attempt to invoke virtual method 'com.badlogic.gdx.graphics.g2d.TextureRegion org.catrobat.catroid.common.LookData.getTextureRegion()' on a null object reference
    at org.catrobat.catroid.content.Look.copyLookForSprite(`[Look.java:154](https://github.com/Catrobat/Catroid/blob/4de61c5d6e17c447abfbe18e5f76ac9377830e67/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L154)`)`

[1] https://jira.catrob.at/browse/CAT-1384

PR #1983, opened 2016-10-25
CAT-FIX Fix LookData reference in Bricks for clones

robertpainsi:
It is important that the LookDatas are cloned before the Bricks are
cloned. Otherwise the SetLookBrick LookData would refer to an origin
LookData instead of a cloned one.

It is also important that the Look is created and cloned after the
Bricks are cloned. Otherwise always a Look instance is created instead
of a Look or PhysicsLook.

To fix this problem, first, the LookDatas are cloned, then the Bricks,
then the Look.

https://github.com/Catrobat/Catroid/pull/1973#issuecomment-256002440

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
@@ -337,6 +337,8 @@ public class Sprite implements Serializable, Cloneable {
 		cloneUserBricks(cloneSprite);
 		cloneNfcTags(cloneSprite);
 		cloneScripts(cloneSprite);
+		cloneSprite.resetSprite();
+		cloneLook(cloneSprite);
 
 		setUserAndVariableBrickReferences(cloneSprite, userBricks);
 
@@ -359,11 +361,12 @@ public class Sprite implements Serializable, Cloneable {
 		Sprite originalSprite = ProjectManager.getInstance().getCurrentSprite();
 		ProjectManager.getInstance().setCurrentSprite(cloneSprite);
 
+		cloneLooks(cloneSprite);
 		cloneUserBricks(cloneSprite);
 		cloneSpriteVariables(ProjectManager.getInstance().getCurrentScene(), cloneSprite);
 		cloneScripts(cloneSprite);
 		cloneSprite.resetSprite();
-		cloneLooks(cloneSprite);
+		cloneLook(cloneSprite);
 		setUserAndVariableBrickReferences(cloneSprite, userBricks);
 
 		ProjectManager.getInstance().setCurrentSprite(originalSprite);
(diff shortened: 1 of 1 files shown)


robertpainsi:
@thmq, @p4p4 please review and test.
Overall, I split the cloneLooks method into two separate methods, so I can first clone the LookDatas, then clone the bricks (the cloned LookData references are used) and after that, clone the Look itself.

I did it for both, cloning through the IDE and the Stage.
Cloning through the IDE perfectly shows what this PR does because it doesn't contain the changes made in https://github.com/Catrobat/Catroid/pull/1973

robertpainsi:
For this change, it's mandatory that the following methods do NOT operate on/with the Look. I checked it and couldn't find any code which does. However, it's still possible that I missed something.
```java
cloneUserBricks(cloneSprite);
cloneSpriteVariables(ProjectManager.getInstance().getCurrentScene(), cloneSprite);
cloneScripts(cloneSprite);
```

thmq:
looks alright to me. any objections to merging it? @p4p4 @robertpainsi ?

p4p4:
Yes the change makes sense to me. Unfortunately I am not able to test it right now since Android Studio does not yet seem to work (seems to build forever..) on my new hardware-setup, so no objections from my side.

robertpainsi:
> objections to merging it? @p4p4 @robertpainsi ?

No, LGTM! :joy_cat:
I'll immediately update https://github.com/Catrobat/Catroid/pull/1980 when merged.

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -488,17 +491,19 @@ private void cloneSpriteVariables(Scene currentScene, Sprite cloneSprite) {
 		}
 	}
 
+	private void cloneLook(Sprite cloneSprite) {
+		cloneSprite.look = this.look.copyLookForSprite(cloneSprite);
+		if (cloneSprite.getLookDataList().size() > 0) {
+			cloneSprite.look.setLookData(cloneSprite.getLookDataList().get(0));
```
p4p4: This sets the look always to the first one of the `lookDataList`, regardless of the original clone's look.
would it make sense to replace it with something like
``` java
cloneSprite.look.setLookData(cloneSprite.lookList.get(this.lookList.indexOf(this.look.getLookData())));
```
(not tested...)

robertpainsi: Yes, definitely!
It's already done in https://github.com/Catrobat/Catroid/pull/1980 :wink:

p4p4: ok thank you, in that case I do not have any complaints. :+1:

PR #1987, opened 2016-10-27
CAT-2087 fix for crash related to TRUE and FALSE in conditional brick

michael-fischl:
`internFormulaTokenSelection = null` wasn't executed where needed, so FormulaEditor still thought something is highlighted although it wasn't. Added this line and refactored `setCursorPositionAndSelectionAfterInput` to make it a little bit more slim and understandable.


cheidenreich:
If formatting and jenkins are passed, this can be merged

PR #1988, opened 2016-10-27
CAT-2126 Fix for PC crashes when open media library without an intern…

michael-fischl:
…et connection

Added checking of network connection before accessing media library in Sounds, Looks and Sprites with the help of `Utils.isNetworkAvailable(Context context, boolean createDialog)`


cheidenreich:
If formatting and jenkins are passed, this can be merged

PR #1991, opened 2016-10-28
CAT-FIX Don't serialize spinner values of StopScriptBrick

robertpainsi:
The language version also has to be increased in the next release, I guess.

Old XML

``` xml
<brick type="StopScriptBrick">
  <commentedOut>false</commentedOut>
  <spinnerSelection>1</spinnerSelection>
  <spinnerValue>
    <string>Stop this script</string>
    <string>Stop all scripts</string>
    <string>Stop other scripts of this object</string>
  </spinnerValue>
</brick>
```

New XML

``` xml
<brick type="StopScriptBrick">
  <commentedOut>false</commentedOut>
  <spinnerSelection>1</spinnerSelection>
</brick>
```

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
@@ -42,15 +42,12 @@ public class StopScriptBrick extends BrickBaseType {
 
 	private static final long serialVersionUID = 1L;
 
-	private String[] spinnerValue;
 	private int spinnerSelection;
 
 	public StopScriptBrick() {
-		this.spinnerValue = new String[3];
 	}
 
 	public StopScriptBrick(int spinnerSelection) {
-		this.spinnerValue = new String[3];
 		this.spinnerSelection = spinnerSelection;
 	}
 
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
```diff
@@ -110,6 +107,7 @@ public View getPrototypeView(Context context) {
 	}
 
 	private ArrayAdapter<String> createArrayAdapter(Context context) {
+		String[] spinnerValue = new String[3];
 		spinnerValue[BrickValues.STOP_THIS_SCRIPT] = context.getString(R.string.brick_stop_this_script);
 		spinnerValue[BrickValues.STOP_ALL_SCRIPTS] = context.getString(R.string.brick_stop_all_scripts);
 		spinnerValue[BrickValues.STOP_OTHER_SCRIPTS] = context.getString(R.string.brick_stop_other_scripts);
```
robertpainsi: Even tough I'm not a fan of arrays and hardcoded indices, I'm leaving it as is.

Some bricks are using spinners with hard coded values like here. We should create a __single__ and future proof solution for this once and for all. What I have in mind, is, that the values may be expanded, the values are reordered or new values may be inserted in between.

robertpainsi:
There are other bricks having the same problem as well. I'll create a new PR.

PR #1997, opened 2016-10-31
CAT-2178 Wrong icon for standalone apps

Chypsylon:


thmq:
ok to test

catroid/gradle/standalone_apk_tasks.gradle
```diff
@@ -182,23 +182,40 @@ def removeUnecessaryPermissions() {
...
+            println "Found manual screenshot: " + it.getPath()
         }
     }
 
-    copy {
-        from screenshot.getPath();
-        into project.ext.resPath
-        rename { String fileName ->
-            fileName.replace(screenshot.getName(), 'splash_screen.png')
+    if (screenshot != null && !automaticScreenshots.isEmpty() && automaticScreenshots.get(0).exists()) {
```
robertpainsi: This is just a guess. Should it be `screenshot == null`?
When there is no manual screenshot (`screenshot == null`), an automatic one will be chosen. That's what this if clause should do, right?

Chypsylon: You're right, that's a leftover from an earlier check.

PR #1998, opened 2016-11-02
CAT-FIX Don't serialize members which shouldn't be in the XML

robertpainsi:
Also see https://github.com/Catrobat/Catroid/pull/1991

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
@@ -42,15 +42,12 @@ public class StopScriptBrick extends BrickBaseType {
 
 	private static final long serialVersionUID = 1L;
 
-	private String[] spinnerValue;
 	private int spinnerSelection;
 
 	public StopScriptBrick() {
-		this.spinnerValue = new String[3];
 	}
 
 	public StopScriptBrick(int spinnerSelection) {
-		this.spinnerValue = new String[3];
 		this.spinnerSelection = spinnerSelection;
 	}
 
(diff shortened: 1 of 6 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/StopScriptBrick.java
```diff
@@ -110,6 +107,7 @@ public View getPrototypeView(Context context) {
 	}
 
 	private ArrayAdapter<String> createArrayAdapter(Context context) {
+		String[] spinnerValue = new String[3];
 		spinnerValue[BrickValues.STOP_THIS_SCRIPT] = context.getString(R.string.brick_stop_this_script);
 		spinnerValue[BrickValues.STOP_ALL_SCRIPTS] = context.getString(R.string.brick_stop_all_scripts);
 		spinnerValue[BrickValues.STOP_OTHER_SCRIPTS] = context.getString(R.string.brick_stop_other_scripts);
```
robertpainsi: As you can see, it doesn't make sense to serialize __localized__ spinner values.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/UserVariable.java
```diff
@@ -29,8 +29,8 @@
 
 	private String name;
 	private transient Object value;
-	private boolean visible = true;
-	private boolean dummy = false;
+	private transient boolean visible = true;
+	private transient boolean dummy = false;
```
robertpainsi: `visible` and `dummy` are used for the ShowTextActor.

`visible` is set in ShowTextBrick and HideTextBrick. So it's just used for the current stage.
`dummy` is only used to draw the the NO_VARIABLE_SELECTED thingy.

There is no use for them in the XML.

PR #2000, opened 2016-11-05
CAT-REF Remove paused member from Script and Sprite

robertpainsi:
The paused member was used in the old implementation when we handled the
action execution by threads. A long time ago we switched to libGDX
actions. So there is no need for the member anymore.

The only case where the member is used, is in Look.doTouchDown [1].
However, when pausing the stage, a grey overlay covers the whole stage.
All touch events are captured by this overlay and no stage touch events
are fired.

It's important that no touch events are fired when the stage is paused
for Sprites and TouchUtil.

[1] [Look.java#L164-L166](https://github.com/Catrobat/Catroid/blob/2ce3242fda5f8c020757a9e31fd6d78f111822c5/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L164-L166)

###end-of-commit-message###

Do not ask me what all the pause tests did because the pause/resume logic in the Script and Sprite did nothing.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -161,9 +161,6 @@ public class Look extends Image {
 	}
 
 	public boolean doTouchDown(float x, float y, int pointer) {
-		if (sprite.isPaused) {
-			return true;
-		}
 		if (!isLookVisible()) {
 			return false;
 		}
--- a/catroid/src/main/java/org/catrobat/catroid/content/Script.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Script.java
@@ -50,11 +50,8 @@ public abstract class Script implements Serializable {
 
 	protected boolean commentedOut = false;
 
-	private transient volatile boolean paused;
-
 	public Script() {
 		brickList = new ArrayList<>();
-		init();
 	}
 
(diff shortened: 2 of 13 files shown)


catroid/src/main/java/org/catrobat/catroid/content/Script.java
```diff
@@ -162,14 +154,6 @@ public void removeBrick(Brick brick) {
 		return brickList;
 	}
 
-	public void setPaused(boolean paused) {
-		this.paused = paused;
-	}
-
-	public boolean isPaused() {
-		return paused;
-	}
```
robertpainsi: `isPaused` is only used in tests.
The paused member itself is private and nowhere else read.

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -161,9 +161,6 @@ public Look copyLookForSprite(final Sprite cloneSprite) {
 	}
 
 	public boolean doTouchDown(float x, float y, int pointer) {
-		if (sprite.isPaused) {
-			return true;
-		}
```
robertpainsi: This is the only real use-case of a paused member.
However, as explained in the commit message, touch events are only fired when the stage is running.
When the stage is paused, a grey overlay covers the stage and captures all touch events.

robertpainsi:
Pull-Request :hash::two::zero::zero::zero: :balloon::tada:

PR #2005, opened 2016-11-13
[CAT-2158] Remove duplicate code for flipping a Look (SetRotationStyle)

michael-fischl:


catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -377,6 +368,25 @@ public float getDirectionInUserInterfaceDimensionUnit() {
...
+		}
+		boolean orientedLeft = sprite.look.getDirectionInUserInterfaceDimensionUnit() < 0;
+		if (mode == Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && orientedLeft) {
+			flipAllLookDatas();
+			sprite.look.setFlipped(true);
+		}
+	}
+
+	public void flipAllLookDatas() {
+		for (LookData lookData : sprite.getLookDataList()) {
```
robertpainsi: The sprite is only used in [Look.java#L177](https://github.com/Catrobat/Catroid/blob/310dc99d43f993f1f58adc4d997fce554e0f0754/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L177) which can be removed by applying Inversion of Control (creating a callback). This will remove the constructor parameter and member Sprite entirely.

However, I'm still not sure if we should change it... What do you think?

michael-fischl: I think this is a nice idea since simplification is always a good thing to do. Consequently, one could transfer `flipAllLookDatas()` and `rotateLookDatasIfNeeded(int mode)` into the `Sprite` classwhere it logically also belongs in my opinion. So `protected Sprite sprite;` would be obsolete then.

robertpainsi: I'm currently in a struggle. On the one hand I like your idea of flipping all LookDatas. You don't have to care about the flip state when changing the LookData. Flipping is also a really cheap operation. On the other hand it will immediately load *all* images into memory. This may block the UI thread for a while. Also moving the `flipAllLookDatas` and `rotateLookDatasIfNeeded` to the Sprite will create another dependency between Look and Sprite (e.g. by setDirectionInUserInterfaceDimensionUnit). :crying_cat_face:

Maybe I'm just overrating the loading problem because I really want to remove the Sprite reference from the Look so bad. I haven't even evaluated nor tested the result. Maybe it isn't a good idea to remove the Sprite anyway.

However, I would suggest to not flip all Looks at once. Just the current one. You also have to update the flip state when the [image is changed](https://github.com/Catrobat/Catroid/blob/310dc99d43f993f1f58adc4d997fce554e0f0754/catroid/src/main/java/org/catrobat/catroid/content/Look.java#L243). Since `Look.isFlipped` won't tell you reliably if the image is flipped, you've to use `TextureRegion.isFlipX` which is the better solution anyway.
What do you think?

michael-fischl: I tested it, you are right, on the first time when the need to flip occurs the UI Thread blocks for a very short but noticeable amount of time. I didn't know about `TextureRegion.isFlipX` which seems like a really good thing to do. I will look into that.

robertpainsi: Awesome! Thanks for testing it :smiley_cat:

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -377,6 +368,17 @@ public float getDirectionInUserInterfaceDimensionUnit() {
...
+	}
+
+	private void rotateLookDataIfNeeded(int mode) {
+		if (mode != Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && sprite.look.isFlipped()) {
+			getLookData().getTextureRegion().flip(true, false);
+		}
+		boolean orientedLeft = sprite.look.getDirectionInUserInterfaceDimensionUnit() < 0;
+		if (mode == Look.ROTATION_STYLE_LEFT_RIGHT_ONLY && orientedLeft) {
+			getLookData().getTextureRegion().flip(true, false);
+		}
```
cheidenreich: I'm not sure about this code.. wouldn't it flip twice if both sprite.look.isFlipped() and orientedLeft are true?
if the one excludes the other, the two ifs could be simplified to one

michael-fischl: The first if in line 375 is responsible for the case, if a sprite had LR rotation before and got flipped at some point and while being flipped, the rotation style changed to something different than LR so it needs to be flipped to normal again.
The second if in 379 flips the lookdata if it's oriented left but not flipped yet.
One does exclude the other, if I get correctly what you meant to say, because of the "mode" part, only one if will ever be true so I could combine the ifs to make it smaller.

michael-fischl:
Please review again :)

catroid/src/main/java/org/catrobat/catroid/content/Look.java
cheidenreich: ah, i see, thanks for clarification :)
also, it is more readable with your last change ;)

cheidenreich:
One more Question:
Why are those lines in the test classes removed?

michael-fischl:
There is no isFlipped() method anymore since I am using the more reliable method from the textureRegion itself. Those lines were basically a test if my isFlipped() method is behaving correctly. This is now obsolete since I'm using the other built-in one, and so what is left should be sufficient.

cheidenreich:
Thanks ;)
Looks good now 👍

bernadettespieler:
ok to test

thmq:
test this please

thmq:
green run: https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidTriggerPullRequestTest/1502/

PR #2006, opened 2016-11-14
CAT-1900 Amount of Bricks (Show details) is wrong

sheryl31:


catroid/src/main/res/values-ar/strings.xml
```diff
@@ -326,12 +326,12 @@
...
+  <string name="number_of_objects">Objects: %1$d</string>
   <string name="start_scene_name">%1$s (First Scene)</string>
-  <string name="number_of_scripts">النص:</string>
-  <string name="number_of_bricks">اللّبنات:</string>
-  <string name="number_of_looks">المشهد:</string>
-  <string name="number_of_sounds">الأصوات:</string>
+  <string name="number_of_scripts">النص: %1$d</string>
+  <string name="number_of_bricks">اللّبنات: %1$d</string>
+  <string name="number_of_looks">المشهد: %1$d</string>
+  <string name="number_of_sounds">الأصوات: %1$d</string>
```
robertpainsi: Maybe Chrome is just tricking me or that's how RTL languages work but this change looks odd.
To test the Pull Request, [CAT-2199](https://jira.catrob.at/browse/CAT-2199) needs to be fixed first.

sheryl31: I don't know why but at my local files it all looks okay and in the right order. 
capture1

robertpainsi: In AS 2.2.2 on Ubuntu it also looks different image :smile_cat:

I quickly tested only the number_of_bricks text and it looks good  image

Google translate: `Building blocks: 12` → `اللبنات: 12`*
The number ١٢ represents 12 and seems to be written LTR ([source](https://www.quora.com/Why-are-numerals-written-left-to-right-in-Arabic)).

\* JFI: The translated text for building blocks depends on the quantity. `Building blocks: 12` → `اللبنات: 12` but `Building blocks: 123` → `كتل البناء: 123`. Not important for this Pull Request but something we should consider with internationalization in the future.

robertpainsi: Oh, I completely forgot... afaik these changes will be overwritten by Crowdin anyway.
You always only need to change the English strings.

@thmq, how should we handle such "translations"?
The string itself hasn't changed. However it has to be retranslated again in every language :disappointed:.

bernadettespieler:
please consider comments and reopen

PR #2014, opened 2016-11-18
CAT-REF Add parameterized constructor to LookData and SoundInfo

robertpainsi:
The name and file name are nearly always set immediately after creating
a LookData or SoundInfo. A parameterized constructor comes in handy to
remove these set calls.


84n4n4:
please rebase

robertpainsi:
Done!
As always, those 8 web tests (org.catrobat.catroid.test.web.ServerCallsTest.*) fail :worried: 

Thank you guys for reviewing!!! :smile_cat:

PR #2016, opened 2016-11-20
Cat nolb v3

sjaindl:
Rebased to develop
Fix large Templates-Icon-Bug
Fix accessibility settings with refactored UI
Fix translation of project variables
Add missing strings


catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -391,6 +391,7 @@ protected void onDestroy() {
 		CameraManager.getInstance().stopPreviewAsync();
 		CameraManager.getInstance().releaseCamera();
 		CameraManager.getInstance().setToDefaultCamera();
+		ProjectManager.getInstance().setSceneToPlay(ProjectManager.getInstance().getCurrentScene());
```
loibi93: Why is this done here ? The sceneToPlay is used while the Stage is loaded!

sjaindl: I don't know.
I've just rebased to the current develop, where this code is already merged. 
This PR goes to the feature/NOLB-branch, so only the last commit should be worthwhile to review.

robertpainsi: > Why is this done?

https://github.com/Catrobat/Catroid/pull/2016/commits/73c1f8f8572d09033977b0fcaead5b3f8ea6eabf
Much "bla bla" in the commit message. Summarized, in the IDE, the `currentScene` and `sceneToPlay` always have to be the same so formulas in the IDE are evaluated correctly.. When exiting the Stage, the `sceneToPlay` may have changed and has to be reset to the `currentScene`.

> Why is this done **here**?

`StageActivity.onDestroy` is the point where many resources are released and reset. Just right before going back to the IDE.

PR #2017, opened 2016-11-20
MUS-191 GUI to structure

joseffilzmaier:
Implement the button press handling so that the data structure
is updated upon the onClick callback.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/NoteView.java
+++ b/catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/NoteView.java
@@ -30,25 +30,33 @@ import android.view.View;
 import android.widget.ImageView;
 
 import org.catrobat.catroid.R;
+import org.catrobat.catroid.pocketmusic.note.NoteLength;
+import org.catrobat.catroid.pocketmusic.note.trackgrid.GridRowPosition;
+
+import java.util.List;
 
 public class NoteView extends ImageView implements View.OnClickListener {
 
 	private static final int HIDDEN = 0;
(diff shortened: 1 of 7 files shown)


catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/TrackRowView.java
```diff
@@ -86,7 +88,11 @@ private void initializeRow() {
 			noteColor = ContextCompat.getColor(getContext(), R.color.white);
 		}
 		for (int i = 0; i < QUARTER_COUNT; i++) {
-			noteViews.add(new NoteView(getContext(), noteColor));
+			List<GridRowPosition> gridRowPositions = new ArrayList<>();
+			if (gridRow != null) {
+				gridRowPositions = gridRow.getGridRowPositions().get(0);
```
rohmab06: get(0) is not very intuitive

catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/NoteView.java
```diff
@@ -30,25 +30,33 @@
...
+import org.catrobat.catroid.pocketmusic.note.trackgrid.GridRowPosition;
+
+import java.util.List;
 
 public class NoteView extends ImageView implements View.OnClickListener {
 
 	private static final int HIDDEN = 0;
 	private static final int FULL_VISIBLE = 255;
+	private final List<GridRowPosition> gridRowPositions;
+	private final int index;
```
rohmab06: please rename the two new member variables - it is not clear what they are doing here

PR #2018, opened 2016-11-20
CAT-FIX XStream sort by annotation with derived classes

robertpainsi:
CAT-FIX XStream sort by annotation with derived classes

Derived classes do not inherit annotations and therefore will always be
sorted alphabetically. The CatroidFieldKeySorter fixes this problem by
searching for the XStreamFieldKeyOrder annotation in the class hierarchy
now. If a derived class adds new members, which aren't specified in a
base class, a XStreamMissingSerializableFieldException is thrown.

This fix currently addresses only the Sprite and its subclasses. The
Sprite specifies a custom field order but the subclasses don't and
therefore the fields were always sorted alphabetically. Now, each
Subclass looks for the XStreamFieldKeyOrder annotation in a base class,
the Sprite, and sorts it accordingly.

The @Inherited would've also solved this problem but is easily forgotten
and more error-prone.

###END-OF-COMMIT

I didn't add a `XML change` label to this Pull Request because there is no need to increase the language version. XStream doesn't care about the order of XML tags and attributes*. However, since every currently uploaded project wrongly sorts the Sprite tags alphabetically, every team (ios, html5, windows phone, …) should be informed to test their XML parser.

The XML looks good now but please recheck.
```xml
<object type="SingleSprite" name="Background">
  <lookList>
    <look name="Background">
      <fileName>9eb91177d3dffb4acec24ecbdbb61c3e_Background.png</fileName>
    </look>
  </lookList>
  <soundList/>
  <scriptList/>
  <userBricks/>
  <nfcTagList/>
</object>
```

\* It just cares about not using [forward referencing](https://github.com/x-stream/xstream/issues/78).

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/io/CatroidFieldKeySorterTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/io/CatroidFieldKeySorterTest.java
@@ -197,4 +197,39 @@ public class CatroidFieldKeySorterTest extends AndroidTestCase {
 		private int a;
 		private int b;
 	}
+
+	public void testSortByAnnotationIsInBaseClass() {
+		xstream.toXML(new SubClassWithoutAnnotation());
+
+		MoreAsserts.assertEquals("Sorted fields differ",
+				new String[] { "b", "a" }, fieldKeySorter.getFieldNames(SubClassWithoutAnnotation.class));
+	}
+
+	public void testMissingFieldInSubClassWithoutAnnotationThrowsException() {
+		try {
+			xstream.toXML(new SubClassWithNewMemberButWithoutAnnotation());
+			fail("XStream didn't throw an exception for missing field c in annotation");
+		} catch (XStreamMissingSerializableFieldException expected) {
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


robertpainsi:
We should merge this PR in the current release. So the Sprite members are serialized in the correct order.

Current order:
```xml
<object type="SingleSprite" name="Background">
  <lookList/>
  <nfcTagList/>
  <scriptList/>
  <soundList/>
  <userBricks/>
</object>
```

After merging the PR into the release branch:
```xml
<object type="SingleSprite" name="Background">
  <lookList/>
  <soundList/>
  <scriptList/>
  <userBricks/>
  <nfcTagList/>
</object>
```

That's the order specified in [Sprite.java#L68-L75](https://github.com/Catrobat/Catroid/blob/e66d9a4b481347c51754047ab05bcf3e57df4e03/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java#L68-L75)
```java
@XStreamFieldKeyOrder({
		"name",
		"lookList",
		"soundList",
		"scriptList",
		"userBricks",
		"nfcTagList"
})
```

I changed the base to `origin/release-0.9.28`.

robertpainsi:
Ah, forgot to ping @84n4n4 @thmq

PR #2021, opened 2016-11-22
Fix Scene-Reference in VariableActions

sjaindl:
The Show & HideVariableActions need a reference to the correct Scene that holds the relevant sprite variables. The currentScene was definitely wrong as it only works when the program is  started coincidentally from that Scene.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/HideTextAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/HideTextAction.java
@@ -37,20 +37,21 @@ import java.util.Map;
 public class HideTextAction extends TemporalAction {
 
 	private UserVariable variableToHide;
+
+	private Sprite sprite;
 	private UserBrick userBrick;
 
 	@Override
 	protected void begin() {
-		DataContainer dataContainer = ProjectManager.getInstance().getCurrentScene().getDataContainer();
-		Sprite currentSprite = ProjectManager.getInstance().getCurrentSprite();
+		DataContainer dataContainer = ProjectManager.getInstance().getSceneToPlay().getDataContainer();
 
 		List<UserVariable> variableList = dataContainer.getProjectVariables();
 		Map<Sprite, List<UserVariable>> spriteVariableMap = dataContainer.getSpriteVariableMap();
 
 		setVariablesVisible(variableList);
 
-		if (currentSprite != null) {
-			List<UserVariable> spriteVariableList = spriteVariableMap.get(currentSprite);
+		if (sprite != null) {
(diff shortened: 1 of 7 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/HideTextAction.java
```diff
@@ -37,20 +37,22 @@
...
+
+	private Scene scene;
+	private Sprite sprite;
 	private UserBrick userBrick;
 
 	@Override
 	protected void begin() {
-		DataContainer dataContainer = ProjectManager.getInstance().getCurrentScene().getDataContainer();
-		Sprite currentSprite = ProjectManager.getInstance().getCurrentSprite();
+		DataContainer dataContainer = scene.getDataContainer();
```
loibi93: You can use getSceneToPlay() here! With that you do not need the getSceneOfSprite Method.

robertpainsi: I agree with @loibi93, `getSceneOfSprite` seems to be unnecessary.
Since it should be merged to the NOLP branch asap, I'm also fine to use `getSceneToPlay()` here.

However, DI as you already use it [here](https://github.com/Catrobat/Catroid/pull/2021/files#diff-a353972b5dc62a8ccef1151508102651R922) is a cleaner solution. You should also be able use `getSceneToPlay` there. The best solution would be to pass the parameter Scene all from the start but this'll be a bigger refactoring. Also I can't guarantee if it'll work as expected. :wink:

sjaindl: Good hint, works fine!

loibi93: Actually we never need to pass the scene as a parameter, since there is always just one running scene. Action of Sprite is executed -> Sprite is part of running Scene -> Sprite is part of sceneToPlay. There should never ever be another possibility (only if at some point concurrent scene execution is introduced). This behaviour is defined [here](https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java#L349).

robertpainsi: > Dependency injection is about decoupling code.
>
> When you avoid the use of globals by passing arguments you are decoupling code. You are removing the dependency the code has on the globals. [[1]](http://stackoverflow.com/questions/5440800/how-does-dependency-injection-differ-from-just-not-using-global-variables)

thmq:
test this please

sjaindl:
@thmq test added & rebased

PR #2022, opened 2016-11-22
Fix Scene-Reference in VariableActions

sjaindl:
The Show & HideVariableActions need a reference to the correct Scene that holds the relevant sprite variables. The currentScene was definitely wrong as it only works when the program is  started coincidentally from that Scene.


robertpainsi:
Do you really want to merge the Musicdroid stuff?

sjaindl:
Actually, the MusicDroid stuff are just the commits that the feature/NOLB branch is behind develop (as I branched away from the develop).
I could make a separate branch, but I think we should keep the NOLB-branch up-to-date to develop anyway..

robertpainsi:
Got it!

In the future, I still would suggest to create separate Pull Requests to update the NOLB branch. It's much cleaner. You can easily create PRs either from the develop branch or any commit you like through the GitHub GUI. No extra branch needed.

PR #2023, opened 2016-11-23
CAT-REF Remove duplicate range checks for friction

robertpainsi:
The PhysicsObject.setFriction trims the friction value to the correct
range [1]. All other places are just duplicate code, will lead to
inconsitancy and have been removed.

[1] [PhysicsObject.java#L268-L273](https://github.com/Catrobat/Catroid/blob/b10f22e2bd81e6f7ce70a7a2c68d0add6901c7ab/catroid/src/main/java/org/catrobat/catroid/physics/PhysicsObject.java#L268-L273)


84n4n4:
the 3 failing tests from https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/1596/ are the ones calling scratch webservice, everything else is fine.

PR #2024, opened 2016-11-23
CAT-FIX Remove empty updateReferenceAfterMerge method overrides

robertpainsi:
In most cases the updateReferenceAfterMerge is an empty implementation.
So doing something in this method is an exception, not the normal case.
Therefore I added a default empty implementation in the base class and
removed all empty overrides.


84n4n4:
the 3 failing tests from https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidTriggerPullRequestTest/1596/ are the ones calling scratch webservice, everything else is fine.

PR #2032, opened 2016-12-03
[CAT-2095] Fix formula editor keys gradually disappear when formula becomes longer

michael-fischl:


cheidenreich:
working well, please fix the conflict in catroid/src/main/res/layout/fragment_formula_editor.xml

michael-fischl:
done!

PR #2033, opened 2016-12-08
MUS-207 Integrating the sound library

paggerbua:
Integrate the C++ midi sound library. The compiled C++ files are provided too.


catroid/src/main/java/org/catrobat/catroid/pocketmusic/mididriver/MidiInstruments.java
```diff
@@ -0,0 +1,158 @@
...
+	public static final byte SYNTH_DRUM                     = 118;
+	public static final byte REVERSE_CYMBAL                 = 119;
+	public static final byte GUITAR_FRET_NOISE              = 120;
+	public static final byte BREATH_NOISE                   = 121;
+	public static final byte SEASHORE                       = 122;
+	public static final byte BIRD_TWEET                     = 123;
+	public static final byte TELEPHONE_RING                 = 124;
+	public static final byte HELICOPTER                     = 125;
+	public static final byte APPLAUSE                       = 126;
+	public static final byte GUNSHOT                        = 127;
```
robertpainsi: The formatter doesn't like horizontal aligning :scream_cat:. You can use
```
// @formatter:off
...
// @formatter:on
```
to disable the formatter and Checkstyle (https://github.com/Catrobat/Catroid/commit/d599f93b8282119017ca5483a0001f5bc7001626). Please use it sparely!

Enums are really powerful. They can store constants and also have methods. See [here (Planets)](http://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).
If you still prefer numbers, take a look at [enum-annotation](https://developer.android.com/studio/write/annotations.html#enum-annotations). I'm not sure if it works with bytes though.

oliewa92:
The PullRequest will be made at a later date.

PR #2035, opened 2016-12-08
CAT-FIX Various fixes from feature/NOLB-branch

sjaindl:
These are some commits that are in feature/NOLB and should also be merged into develop


catroid/src/main/java/org/catrobat/catroid/common/LookData.java
```diff
@@ -190,16 +197,16 @@ public String getChecksum() {
 		return fileName.substring(0, 32);
 	}
 
-	protected String getPathToImageDirectory() {
+	String getPathToImageDirectory() {
 		return Utils.buildPath(Utils.buildProjectPath(ProjectManager.getInstance().getCurrentProject().getName()),
 				getSceneNameByLookData(), Constants.IMAGE_DIRECTORY);
```
robertpainsi: Use `Utils.buildScenePath`. It already internally encodes the project and scene names.

Eventually we should create a handful of helper methods for many scenarios. e.g. in this case `buildImagePath(Project, Scene)` and [Replace String paths by File](https://jira.catrob.at/browse/CAT-1304).

catroid/src/main/java/org/catrobat/catroid/content/bricks/BackPackedData.java
```diff
@@ -0,0 +1,50 @@
...
+import org.catrobat.catroid.formulaeditor.UserVariable;
+
+import java.io.Serializable;
+
+public class BackPackedData implements Serializable {
+	public UserVariable userVariable;
+	public Integer userVariableType;
+
+	public UserList userList;
+	public Integer userListType;
```
robertpainsi: Stupid question: Why does a BackPackedData store two things, a UserVariable and a UserList?

catroid/src/main/java/org/catrobat/catroid/common/LookData.java
sjaindl: Yes, I would vote for that - the helper methods should also include sounds and the backpack.

catroid/src/main/java/org/catrobat/catroid/content/bricks/BackPackedData.java
sjaindl: Actually, I've moved it to a central class out of the Variable- and List-Bricks.
But I admit that it makes sense to split it into 2 classes - one for variables and one for lists. I've changed that accordingly.

sjaindl:
rebased
@robertpainsi should be fine now ;)

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -746,20 +747,40 @@ private void correctAllNestedReferences(Script script) {
 			} else if (currentBrick instanceof LoopBeginBrick) {
 				loopBeginList.add((LoopBeginBrick) currentBrick);
 			} else if (currentBrick instanceof LoopEndBrick) {
+				if (!hasItems(loopBeginList)) {
```
robertpainsi: `!hasItems(loopBeginList)` === `loopBeginList.isEmpty()`

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -769,6 +790,18 @@ private void correctAllNestedReferences(Script script) {
...
 		}
+
+		removeBricksWithInvalidReferences(bricksWithInvalidReferences, script);
+	}
+
+	private void removeBricksWithInvalidReferences(ArrayList<Brick> bricksWithInvalidReferences, Script script) {
+		for (Brick brick : bricksWithInvalidReferences) {
+			script.removeBrick(brick);
+		}
+	}
```
robertpainsi: I think the script itself is a better place for this method. `script.removeBricks(List bricks)`. Then you can use `script.removeBricks(bricksWithInvalidReferences)`.

sjaindl: Yes, that should be clearer

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
sjaindl: changed

robertpainsi: thanks!

catroid/src/main/java/org/catrobat/catroid/content/Script.java
```diff
@@ -146,6 +146,12 @@ public void removeInstancesOfUserBrick(UserBrick userBrickToRemove) {
 		}
 	}
 
+	public void removeBricksWithInvalidReferences(List<Brick> bricksWithInvalidReferences) {
```
robertpainsi: Please rename to `removeBricks(List<Brick> bricks)`. So it's a more generic method.

sjaindl: true true

PR #2025, opened 2016-11-24
CAT-1658 Upgrade gradle version to 2.2.3 to enable instant run

robertpainsi:

PR #2039, opened 2016-12-18
MUS-207 Integrating the sound library

joseffilzmaier:
Integrate the C++ midi sound library. The compiled C++ files are provided too.


joseffilzmaier:
The sound library is written in native code. We need Gradle 2.2.3 in order to automatically compile native code when building android apps. For this reason the library has been precompiled and added to the project (*.so files). When Gradle 2.2.3 is integrated those *.so files can be removed.

Can a core team member please verify this PR as there is a lot of external code?

robertpainsi:
> We need Gradle 2.2.3

https://github.com/Catrobat/Catroid/pull/2025

joseffilzmaier:
For now we only included the precompiled binaries as automatic building of ndk's is not supported with the current gradle version.
Once the gradle version is updated we can still use ndk builds if necessary.
This way we can merge this code and continue development independent of the gradle version.

PR #2040, opened 2016-12-19
Changing wording of Collision Formula. Fixes CAT-2179, CAT-2180, CAT-2181.

comeclarity:
The wording of the collision formula changed from "object_a touches object_b" to "touches_object(object_b)"
To do this, the CatrobatLanguageVersion is now 0.993f (was 0.992f).
Old projects are converted to the new version when loaded.

This also fixes problems with downloaded programs, which where created by an android device with a different language.

Collision now also works with groups. (returns true if one of the sprites in the group collides)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
@@ -228,7 +228,7 @@ public class FormulaEditorCategoryListFragment extends ListFragment implements D
 					.findFragmentByTag(FormulaEditorFragment.FORMULA_EDITOR_FRAGMENT_TAG);
 			if (formulaEditor != null) {
 				if (itemsIds[position] == R.string.formula_editor_function_collision) {
-					showChooseSpriteDialog(formulaEditor, position);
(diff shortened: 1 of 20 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -277,10 +277,9 @@ public void onDismiss(DialogInterface dialogInterface) {
 						}
 					}
 					if (secondSprite != null) {
-						String formula = firstSprite.getName() + " "
-								+ getActivity().getString(itemsIds[pos]) + " " + dialog.getSprite();
+						//String formula = getActivity().getString(itemsIds[pos]) + "(" + dialog.getSprite() +")";
 
-						formulaEditor.addCollideFormulaToActiveFormula(formula);
+						formulaEditor.addCollideFormulaToActiveFormula(secondSprite.getName());
```
cheidenreich: if this takes a sprite name, then the String parameter name should also be changed.. right now its addCollideFormulaToActiveFormula(String formula)

cheidenreich:
ok to test

thmq:
no, not really, but I have no idea why.

comeclarity:
fixed requested changes and added tests

PR #2041, opened 2016-12-20
Changing wording of Collision Formula. Fixes CAT-2179, CAT-2180, CAT-2181.

comeclarity:
The wording of the collision formula changed from "object_a touches object_b" to "touches_object(object_b)"
To do this, the CatrobatLanguageVersion is now 0.993f (was 0.992f).
Old projects are converted to the new version when loaded.

This also fixes problems with downloaded programs, which where created by an android device with a different language.

Collision now also works with groups. (returns true if one of the sprites in the group collides)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -243,6 +243,10 @@ public final class ProjectManager implements OnLoadProjectCompleteListener, OnCh
 				project.setCatrobatLanguageVersion(0.992f);
 			}
 			if (project.getCatrobatLanguageVersion() == 0.992f) {
+				project.updateCollisionFormulasToNewVersion();
+				project.setCatrobatLanguageVersion(0.993f);
+			}
+			if (project.getCatrobatLanguageVersion() == 0.993f) {
 				project.setCatrobatLanguageVersion(Constants.CURRENT_CATROBAT_LANGUAGE_VERSION);
 			}
 //			insert further conversions here
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -245,6 +245,10 @@ public final class ProjectManager implements OnLoadProjectCompleteListener, OnCh
 			if (project.getCatrobatLanguageVersion() == 0.992f) {
 				project.updateCollisionFormulasToNewVersion();
 				project.setCatrobatLanguageVersion(0.993f);
+				asynchronousTask = false;
+				saveProject(context);
+				loadProject(projectName, context);
+				asynchronousTask = true;
 			}
 			if (project.getCatrobatLanguageVersion() == 0.993f) {
 				project.setCatrobatLanguageVersion(Constants.CURRENT_CATROBAT_LANGUAGE_VERSION);
--- a/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
@@ -32,6 +32,7 @@ import com.badlogic.gdx.scenes.scene2d.actions.ParallelAction;
 import com.badlogic.gdx.scenes.scene2d.actions.SequenceAction;
 import com.thoughtworks.xstream.annotations.XStreamAsAttribute;
 
+import org.catrobat.catroid.CatroidApplication;
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.common.BrickValues;
 import org.catrobat.catroid.common.BroadcastSequenceMap;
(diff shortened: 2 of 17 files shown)


catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -880,6 +878,66 @@ public boolean hasToCollideWith(Sprite other) {
...
+				}
+			}
+			for (Brick brick : script.getBrickList()) {
+				if (brick instanceof UserBrick) {
+					UserBrick formulaBrick = (UserBrick) brick;
+					for (Formula formula : formulaBrick.getFormulas()) {
+						formula.updateCollisionFormulasToNewVersion();
+					}
+				}
+				if (brick instanceof FormulaBrick) {
```
sjaindl: This should be an "else if"

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -214,16 +208,36 @@ public void updateCollisionFormula(String oldName, String newName) {
...
+
+	public void updateCollisionFormulaToNewVersion() {
+		if (leftChild != null) {
+			leftChild.updateCollisionFormulaToNewVersion();
+		}
+		if (rightChild != null) {
+			rightChild.updateCollisionFormulaToNewVersion();
+		}
+		if (type == ElementType.COLLISION_FORMULA) {
+			int indexOfSpriteInFormlua = -1;
```
sjaindl: indexOfSpriteInForm**lua**  --> indexOfSpriteInForm**ula** ;)

catroid/src/main/java/org/catrobat/catroid/ui/fragment/SpritesListFragment.java
```diff
@@ -710,7 +704,7 @@ public void onReceive(Context context, Intent intent) {
 			if (intent.getAction().equals(ScriptActivity.ACTION_SPRITE_RENAMED)) {
 				String newSpriteName = intent.getExtras().getString(RenameSpriteDialog.EXTRA_NEW_SPRITE_NAME);
 				String oldSpriteName = spriteToEdit.getName();
-				renameSpritesInCollisionFormulas(oldSpriteName, newSpriteName, getActivity());
+				//renameSpritesInCollisionFormulas(oldSpriteName, newSpriteName, getActivity());
```
sjaindl: Please remove, if not needed

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -243,6 +243,14 @@ public void loadProject(String projectName, Context context) throws LoadingProje
 				project.setCatrobatLanguageVersion(0.992f);
 			}
 			if (project.getCatrobatLanguageVersion() == 0.992f) {
+				project.updateCollisionFormulasToNewVersion();
+				project.setCatrobatLanguageVersion(0.993f);
+				asynchronousTask = false;
+				saveProject(context);
+				loadProject(projectName, context);
```
sjaindl: As this is bad for the performance (especially for big programs): Do you really need to save and load the program again?

comeclarity: This way, one can be absolutely sure that everthing was changed properly. Furthermore, the new program has to be saved anyways. As this has to be done only once, in my mind the performance loss is bearable

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -214,16 +208,36 @@ public void updateCollisionFormula(String oldName, String newName) {
...
+		if (type == ElementType.COLLISION_FORMULA) {
+			int indexOfSpriteInFormula = -1;
+			for (Scene scene : ProjectManager.getInstance().getCurrentProject().getSceneList()) {
+				for (Sprite sprite : scene.getSpriteList()) {
+					int index = value.indexOf(sprite.getName());
+					if (index > indexOfSpriteInFormula) {
+						indexOfSpriteInFormula = index;
+					}
+				}
+			}
```
robertpainsi: Consider a value of `a touches ab` and three sprites (`a`, `b`, `ab`). The parsed secondSpriteName will be `b` instead of `ab`.
If you pass the firstSprite as parameter, you are able to parse the secondSpriteName reliably (at least if a translation of `touches` won't screw up your parsing :wink:).

robertpainsi:
Please add tests

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
robertpainsi: > This way, one can be absolutely sure that everthing was changed properly.

Please be more specific. When/Where do you think there will be a problem?

bernadettespieler:
only in develop

PR #2043, opened 2016-12-20
[CAT-2192] adds collision with finger and edge

comeclarity:
Adds the possibility to check if a sprite is colliding with a finger (is touched) or with the edge of the screen.
This works with the collision polygons similar to the normal collision between objects.


cheidenreich:
Just a general heads-up: the commit message should not have square brackets around the ticket number, just CAT-####

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionDetection.java
```diff
@@ -119,4 +126,86 @@ public static boolean checkCollisionForPolygonsInPolygons(Polygon[] first, Polyg
...
+		If that's the case, check if any line segment of the polygon intersects with the circle, to evaluate
+		a collision. If there is no intersection, but the circle is contained in an uneven number of polygons
+		it means that there still is a collision
+		Example:
+		   _______
+		  |  ___ x|  point "o" is not colliding, while point "x" is colliding
+		  | | o | |
+		  | |___| |
+		  |_______|
+		*/
```
cheidenreich: ❤️

PR #2046, opened 2017-01-09
CAT-2223 Refactored TextDialog, Fixed Codestyle Warnings ...

thmq:
This is only an early version. All RenameDialogs should be simplified and replaced by just one RenameItemDialog, furthermore, the callbacks from the dialogs should be changed to Interfaces instead of BroadCasts.

I will either add another commit on this or, what is more likely, include the other changes into the refactoring process, as I am going to have to change the fragment accordingly.

I have also refactored a few strings, so please just have a look at the RenameDialogs and the TextDialog.


catroid/src/main/java/org/catrobat/catroid/content/bricks/BroadcastReceiverBrick.java
```diff
@@ -158,23 +157,19 @@ private void setSpinnerSelection(Spinner spinner) {
 	// TODO: BroadcastBrick and BroadcastReceiverBrick contain this identical method.
 	private void showNewMessageDialog(final Spinner spinner) {
 		final Context context = spinner.getContext();
-		BrickTextDialog editDialog = new BrickTextDialog() {
+		BrickTextDialog editDialog = new BrickTextDialog(R.string.dialog_new_broadcast_message_title, R.string
+				.dialog_new_broadcast_message_name, context.getString(R.string.new_broadcast_message)) {
```
thomasmauerhofer: Maybe there should be a second constructor at the TextDialog which takes three integer, for a better readability.

PR #2048, opened 2017-01-11
Reset gradle version

thomasmauerhofer:
Reset gradle version, because Jenkins didn't work with 2.2.3


robertpainsi:
What exactly was the problem with Jenkins?

PR #2052, opened 2017-01-12
CAT-2242 Add clone behind origin

robertpainsi:
To be compatible with Scratch, the clone is added behind the original
Object. Previously a clone was added to the top. I didn't add a test for
this behavior.

Beware, cloning will change the z-index of the original Look.


robertpainsi:
I've created a [Jira ticket](https://jira.catrob.at/browse/CAT-2242) and also added an `Under discussion` label because this behavior change hasn't been discussed nor approved.

PR #2054, opened 2017-01-15
CAT-1588 AskBrick and NoNetworkDialog

krellsebastian:
Adds 3 features:
* Brick to get a Speechrecognition-dialog in stage and stores recognized word in user-variable.
* Intents can now be registered easily in the StageActivity. Will be executed automatically and the result returned to the implemented interface.
* No-Network-Ressource for bricks. The resource is checked in PreStage, when no network available, bricks with this resource will be rendered in a dialog to show the user, which bricks will use the connection.

screenshot_2017-01-15-23-21-42


catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
```diff
@@ -240,6 +245,33 @@ public void onClick(DialogInterface dialog, int id) {
 			task.execute();
 		}
 
+		if ((requiredResources & Brick.NETWORK_CONNECTION) > 0) {
```
cheidenreich: Please change this condition to (requiredResources & Brick.NETWORK_CONNECTION) != 0
As this is a bitwise AND between two integers, it could get negative as we may reach the 32nd bit with future changes

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -142,13 +150,25 @@ public void onCreate(Bundle savedInstanceState) {
...
+						showDialog((String) params.get(1), (AskAction) params.get(0));
+						break;
+					case REGISTER_INTENT:
+						currentStage.queueIntent((IntentListener) params.get(0));
+						break;
+					case PERFORM_INTENT:
+						currentStage.startQueuedIntent((Integer) params.get(0));
+						break;
+					default:
+						Log.e(TAG,"Unhandled message in messagehandler, case "+message.what);
```
cheidenreich: eventhough not necessary, please add a break also at the default case.. this prevents errors if these case statements are reordered

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -438,4 +458,50 @@ public void run() {
 	public static int getAndIncrementNumberOfClonedSprites() {
 		return ++numberOfSpritesCloned;
 	}
+
+
+
+	public synchronized void queueIntent(IntentListener asker) {
+		int newIdentId;
+		do {
+			newIdentId = StageActivity.randomGenerator.nextInt(100000);
```
cheidenreich: is there a reason for the bound of 100000? why not full integer range? is 0 a valid number here?

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -438,4 +458,50 @@ public void run() {
...
+		} while (intentListeners.containsKey(newIdentId));
+
+		intentListeners.put(newIdentId, asker);
+
+		ArrayList<Object> params = new ArrayList<>();
+		params.add(newIdentId);
+		if (StageActivity.messageHandler == null) {
```
cheidenreich: this can be done before creating the ArrayList.

catroid/src/main/java/org/catrobat/catroid/content/actions/AskSpeechAction.java
```diff
@@ -0,0 +1,149 @@
...
+	private Formula questionFormula;
+	private UserVariable answerVariable;
+
+	private boolean questionAsked = false;
+	private boolean answerReceived = false;
+
+	private void askQuestion() {
+		ArrayList<Object> params = new ArrayList<>();
+		params.add(this);
+		if (StageActivity.messageHandler == null) {
```
cheidenreich: This can be done before creating the ArrayList (also in AskAction)

krellsebastian:
Changed the bitwise condition to check for resource in all cases. Good argument about exceeding the posInt space.

About testing the dialog: the only functionality is to stop execution of the project when no network is available. So for unit-testing, it would only be the resource-mechanism and the isOnline-function. About UI, don't know if you want to test a dialog that way.

Also changed the random-integer space (max was just a made up value). 0-key is no problem, it will be used as a lookup-identifier in the list and as request, where it's checked for >=0.

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
```diff
@@ -563,6 +595,31 @@ private void nfcInitialize() {
 		resourceInitialized();
 	}
 
+	private static ArrayList<Brick> getBricksRequieringResource(int resource) {
```
robertpainsi: The method name is misleading. All other implementations with the same method name return **all** bricks whereas this also filters duplicate classes. As mentioned a few lines below, `Project.getBricksRequiringResource` will return all bricks. So you just need to create a filter method passing in this list.

I would love to see a generic filter version :wink:

Btw: It's [requ**i**ring](http://www.dictionary.com/browse/requiring)

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
```diff
@@ -563,6 +595,31 @@ private void nfcInitialize() {
...
+			Brick c = i.next();
+			Iterator<Brick> secIt = brickList.iterator();
+			while (secIt.hasNext()) {
+				Brick proofBrick = secIt.next();
+				if (proofBrick.getClass() == c.getClass() && proofBrick != c) {
+					i.remove();
+					break;
+				}
+			}
+		}
```
robertpainsi: Please take a look at the android utils stuff and guava. Maybe they provide something useful.
Let me know if the result has to be sorted or not.

catroid/src/main/java/org/catrobat/catroid/content/Script.java
```diff
@@ -258,4 +258,15 @@ public void setCommentedOut(boolean commentedOut) {
 			}
 		}
 	}
+
+	public ArrayList<Brick> getBricksRequiringResources(int resource) {
+		ArrayList<Brick> resourceBrickList = new ArrayList<Brick>();
```
cheidenreich: note: only on the left side of the "=". When creating it with "new" you have to specify the Implementation.

see: http://stackoverflow.com/questions/2279030/type-list-vs-type-arraylist-in-java

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
krellsebastian: Got that covered. 
Filtering could be done generic, but here we want to distinct the list by comparing one of their properties (the class) **to each other**. So not really filtering. Tried to implement it with `Lists.transform`, but does not seem right. 
Added a function to our own Utils-class, see in next push.

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
```diff
@@ -563,6 +595,31 @@ private void nfcInitialize() {
 		resourceInitialized();
 	}
 
+	private static ArrayList<Brick> getBricksRequieringResource(int resource) {
+		ArrayList<Brick> brickList = new ArrayList<Brick>();
+		ArrayList<Sprite> spriteList = (ArrayList<Sprite>) ProjectManager.getInstance().getCurrentProject()
+				.getSpriteListWithClones();
```
krellsebastian: Not really necessary to implement it in every container. Saw that the function `getRequiredResources` just uses the `sceneList` and iterates through their sprites. 
Did the same now.

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
krellsebastian: No need for a specific order.

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -87,6 +93,8 @@
 	private static int numberOfSpritesCloned;
 
 	public static Handler messageHandler;
+	public static NavigableMap<Integer, IntentListener> intentListeners = new TreeMap<>();
```
krellsebastian: changed

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/NoNetworkDialog.java
```diff
@@ -0,0 +1,95 @@
...
+		super(context);
+	}
+
+	protected NoNetworkDialog(Context context, int theme) {
+		super(context, theme);
+	}
+
+	protected NoNetworkDialog(Context context, boolean cancelable, OnCancelListener cancelListener) {
+		super(context, cancelable, cancelListener);
+	}
```
krellsebastian: done

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -1104,4 +1104,14 @@ public static boolean isUserLoggedIn(Context context) {
 	public static String getNumberStringForBricks(float value) {
 		return (int) value == value ? "" + (int) value : "" + value;
 	}
+
+	public static <T extends Object> List<T> distinctListByClassOfObjects(List<T> listToDistinct) {
```
thmq: 👍

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/NoNetworkDialog.java
```diff
@@ -0,0 +1,87 @@
...
+		this.brickList = brickList;
+	}
+
+	@Override
+	public void onCreate(Bundle savedInstanceState) {
+		View dialogLayout = View.inflate(this.getContext(), R.layout.dialog_error_networkconnection, null);
+		LinearLayout imageLayout = (LinearLayout) dialogLayout
+				.findViewById(R.id.dialog_error_network_brickimages_layout);
+
+		for (Brick networkBrick : brickList) {
```
thmq: if you're adding the bricks in `onCreate(...)` couldn't you have implemented a constructor that directly takes this brick list as a parameter? (I'm not insisting on you doing it, it's just a question)

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -1104,4 +1104,14 @@ public static boolean isUserLoggedIn(Context context) {
...
+		List<T> distinctList = new ArrayList<>();
+		HashMap<Class, T> uniqueMap = new HashMap<>();
+		for (T objectInstance : listToDistinct) {
+			uniqueMap.put(objectInstance.getClass(), objectInstance);
+		}
+		distinctList.addAll(uniqueMap.values());
+		return distinctList;
```
robertpainsi: Perfect! :+1: 


Just a few things. I'm a little picky here :grimacing:


No need to `extends Object`.
Please use `Map` as declaration type.
```java
return new ArrayList<>(uniqueMap.values());
```
The method is a perfect text-book example for tests. :wink:

krellsebastian: `extends Object` is necessary (IMHO) because i call `Object.getClass()`. Would not be able to do that (accourding to [this question](http://stackoverflow.com/questions/3437897/how-to-get-a-class-instance-of-generics-type-t).
Map as declaration for what exactly?

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/NoNetworkDialog.java
krellsebastian: Initially i design the dialog for generic "you need network". But since the chosen string in the description (following bricks will need it), it seems the dialog without a brick list will never be used.

So yeah, into constructor would be better 👍

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
robertpainsi: > extends Object is necessary (IMHO) because i call Object.getClass(). Would not be able to do that (accourding to this question ).

Just give it a try by writing a few tests. :wink:
<details>
<summary>After that, expand me!</summary>
As you already know, for <a href="http://stackoverflow.com/questions/339699/java-generics-type-erasure-when-and-what-happens">some reason</a> T.class isn't valid. However, <code>T.class != Object.getClass()</code>. You can always call getClass() on any Java Object. Even on <code>String.class.getClass()</code> (hopefully this won't be confusing :D String.class is just the class object of the String). You just can't use .class with generics.<br>
If not specified otherwise, every Class extends from Object. So <code>class A {…}</code> == <code>class A extends Object {…}</code>. In java <code>extends Object</code> is (almost) always redundant.
The <a href="https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/util/Collections.java#L215">jdk</a> and <a href="https://github.com/google/guava/blob/8a991157845af123f701f223c8f2ecc245e82977/guava/src/com/google/common/collect/Lists.java#L111">guava</a> also do not use extends Object.
</details>

</br>

> Map as declaration for what exactly?
```java
Map<Class, T> uniqueMap = new HashMap<>();
````

Comment (krellsebastian) [inline catroid/src/main/java/org/catrobat/catroid/utils/Utils.java]: Map instead of Hashmap on a **local** variable? thats not picky, it's useless. But here you go.

File: catroid/src/main/java/org/catrobat/catroid/ui/dialogs/NoNetworkDialog.java (hunk 13 above)
Comment (krellsebastian) [inline catroid/src/main/java/org/catrobat/catroid/ui/dialogs/NoNetworkDialog.java]: done ;)

**Comment by thmq** [comment]
there is a AskBrick as well as a AskSpeechBrick, can anyone explain why? @krellsebastian

**Comment by bernadettespieler** [comment]
There are two kinds of ask bricks! One is the same like that one in Scratch (from GSoC) - you type in the answer as a text. The second ask brick records the answer - that's especially if you use robots (e.g. Phiro).

**Comment by robertpainsi** [comment]
Please add tests in a following up PR. Especially unit tests and tests for the IDE stuff.

File: catroid/src/androidTest/java/org/catrobat/catroid/uitest/content/brick/AskBrickTest.java [hunk 15]
```diff
@@ -0,0 +1,162 @@
...
+import org.catrobat.catroid.content.bricks.AskSpeechBrick;
+import org.catrobat.catroid.content.bricks.WaitBrick;
+import org.catrobat.catroid.formulaeditor.DataContainer;
+import org.catrobat.catroid.stage.StageActivity;
+import org.catrobat.catroid.uitest.annotation.Device;
+import org.catrobat.catroid.uitest.util.UiTestUtils;
+
+import java.util.ArrayList;
+
+public class AskBrickTest extends ActivityInstrumentationTestCase2<StageActivity> {
```
robertpainsi: Rename to AskSpeechBrickTest (credits go to @thmq)
Also rename every occurrences of askBrick... to askSpeechBrick... in code, strings, etc.

catroid/src/main/java/org/catrobat/catroid/stage/PreStageActivity.java
robertpainsi: The current implementation of [Project.getRequiredResources](https://github.com/Catrobat/Catroid/blob/8203e3adabc5583264b30b1f822282befee03092/catroid/src/main/java/org/catrobat/catroid/content/Project.java#L281-L299) is rubbish. It unnecessarily handles the ActionFactory and that's the reason why it's implemented that way. I'm not sure if I already wrote a ticket for that but I definitely have it on my private Catroid Cache list.

We definitely have to take a look at your implementation later (since I'm short on time) but currently it's just split up.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
robertpainsi: > Map instead of Hashmap on a local variable? thats not picky, it's useless. But here you go.

* Program to interfaces, not implementations
* `Vector v = new Vector()` implies that the functionality of Vector is needed (e.g. a function which the Vector has only or thread-safety).
* Dependency inversion principle
* Liskov substitution principle
* Effective Java Item 52
  * Refer to objects by their interfaces
  * > If you get into the **habit** of using interfaces as types, …

I know your original code and that you were using implementations as [return types](https://github.com/krellsebastian/Catroid/commit/44742bd64d10586e8b2a8d0a35d61d565cc74012#diff-addd92b36328ca27d9d2ede2ca3ab9c5R915), [parameters](https://github.com/krellsebastian/Catroid/commit/44742bd64d10586e8b2a8d0a35d61d565cc74012#diff-0b1215285c675bebd81342f589acb64eR57), [members](https://github.com/krellsebastian/Catroid/commit/44742bd64d10586e8b2a8d0a35d61d565cc74012#diff-0b1215285c675bebd81342f589acb64eR43) and I don't even know about [this cast](https://github.com/krellsebastian/Catroid/commit/44742bd64d10586e8b2a8d0a35d61d565cc74012#diff-69b6ec2fdde36c910b08ab1eafb094abR600). Maybe you're coming from a dynamic typed language or you just aren't used to interfaces. That's no problem. Just start to break your bad habit by using interfaces all the way down to local variables.

And always feel free to ask. (But don't expect an answer too sone because currently I'm really short on time.)

robertpainsi: I may have to exclude the Liskov substitution principle from the list for _local_ variables. If I know it for sure, I'll let you know.

PR #2056, opened 2017-01-16
CAT-2243 Dismiss actionMode after copying a scene

sjaindl:


catroid/src/main/java/org/catrobat/catroid/ui/fragment/ListActivityFragment.java
```diff
@@ -102,6 +102,7 @@ public void onDestroyActionMode(ActionMode mode) {
 				clearCheckedItems();
 			} else {
 				copyCheckedItems();
+				clearCheckedItems();
```
cheidenreich: this is called in both branches of the if.. it can be done once below the if/else, and by altering the condition, the else branch would be obsolete

sjaindl:
good hint, fixed ;)

thmq:
replaced by #2066.

PR #2065, opened 2017-01-19
CAT-2173 fixed Layer order

cheidenreich:
This is more like a quick fix, with the current implementation it works.

The StageListener sneaks in a PenActor on top of the background layer which caused the offset.
Other bricks which adds actors (display text, etc) put the actor over the Passepartout, which does not cause further Problems.

If we want to have this fix in the release, this is the easiest solution and the way to go.

I'd work on my first approach right after the release to provide a clean solution for that problem.

Since Actor is a LibGDX class and we are using that directly in Z order readout, a refactoring of all types of actors with introduction of an intermediate layer would be needed for a clean solution.

~~//EDIT:~~
~~please put label "Under Discussion" as i still have to adapt the Layer Unit Tests~~


cheidenreich:
can someone please trigger the test, seems like i don't have permission to trigger PR tests

thmq:
ok to test

thmq:
test this please

cheidenreich:
https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidTriggerPullRequestTest/1503/

PR #2067, opened 2017-01-20
MUS-213 Back button

joseffilzmaier:
Override the onOptionsMenuSelected function within PocketMusicActivity and
implement the home button to finish the activity.


paggerbua:
ok to test

PR #2068, opened 2017-01-23
CAT-2256 Templates API

sjaindl:


catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	public void setOnFetchTemplatesCompleteListener(OnFetchTemplatesCompleteListener listener) {
+		onFetchTemplatesCompleteListener = listener;
+	}
+
+	@Override
+	protected void onPreExecute() {
+		super.onPreExecute();
+		if (context == null) {
```
thomasmauerhofer: Is it possible, that the context is null?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+		String message = context.getString(R.string.fetching_templates);
+		progressDialog = ProgressDialog.show(context, title, message);
+	}
+
+	@Override
+	protected String doInBackground(Void... arg0) {
+		try {
+			if (!Utils.isNetworkAvailable(context)) {
+				exception = new WebconnectionException(WebconnectionException.ERROR_NETWORK, "Network not available!");
+				return null;
```
thomasmauerhofer: Why you return null, and don't throw the exception, to handle it outside?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+			}
+
+			return ServerCalls.getInstance().fetchTemplatesList();
+		} catch (WebconnectionException webconnectionException) {
+			Log.e(TAG, Log.getStackTraceString(webconnectionException));
+			message = webconnectionException.getMessage();
+		}
+		return null;
```
thomasmauerhofer: Do something in the catch, or remove it if you handle it outside.. Then you don't need this line.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Override
+	protected void onPostExecute(String templatesList) {
+		super.onPostExecute(templatesList);
+
+		if (progressDialog != null && progressDialog.isShowing() && !((Activity) context).isFinishing()) {
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1 && !((Activity) context).isDestroyed()) {
+				progressDialog.dismiss();
+			} else {
+				progressDialog.dismiss();
```
thomasmauerhofer: Your if else construct is redundant ;)

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+
+		if (Utils.checkForNetworkError(exception)) {
```
thomasmauerhofer: Why you have to check an Exception?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+			showDialog(R.string.error_internet_connection);
+			return;
+		}
+
+		if (onFetchTemplatesCompleteListener != null) {
```
thomasmauerhofer: Is it possible that the listener is null, or is it only to avoid an exception?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+			onFetchTemplatesCompleteListener.onFetchTemplatesComplete(templatesList);
+		}
+	}
+
+	private void showDialog(int messageId) {
```
thomasmauerhofer: Please use Toasts to show errors, or create a TextDialog. Don't build them in a method of a class.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
```diff
@@ -33,37 +33,49 @@
...
+	private void fetchTemplates() {
+		FetchTemplatesTask task = new FetchTemplatesTask(context);
+		task.setOnFetchTemplatesCompleteListener(this);
+		task.execute();
+	}
 
-		add(new TemplateData(TemplateConstants.TEMPLATE_PUZZLE_NAME, TemplateConstants.TEMPLATE_PUZZLE_IMAGE_NAME,
-				TemplateConstants.TEMPLATE_PUZZLE_LANDSCAPE_FILENAME, TemplateConstants.TEMPLATE_PUZZLE_PORTRAIT_FILENAME));
+	private void initTemplates() {
+		if (templateContainer == null) {
```
thomasmauerhofer: Is it possilbe that the templateContainer is null?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,121 @@
...
+
+	private static final String TAG = FetchTemplatesTask.class.getSimpleName();
+
+	private Context context;
+	private ProgressDialog progressDialog;
+	private OnFetchTemplatesCompleteListener onFetchTemplatesCompleteListener;
+	private WebconnectionException exception;
+	private String message;
+
+	public FetchTemplatesTask(Context activity) {
```
thomasmauerhofer: Plase name the context context, and not activity

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
```diff
@@ -33,37 +33,49 @@
...
+			return;
+		}
 
-		add(new TemplateData(TemplateConstants.TEMPLATE_QUIZ_NAME, TemplateConstants.TEMPLATE_QUIZ_IMAGE_NAME,
-				TemplateConstants.TEMPLATE_QUIZ_LANDSCAPE_FILENAME, TemplateConstants.TEMPLATE_QUIZ_PORTRAIT_FILENAME));
+		for (TemplateData template : templateContainer.getTemplateData()) {
+			add(template);
```
thomasmauerhofer: Why not addAll? Should be possible in the ArrayAdapter.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
```diff
@@ -110,14 +122,31 @@ public void onClick(View view) {
...
+
+		initTemplates();
 	}
 
 	public interface OnTemplateEditListener {
 		void onTemplateEdit(TemplateData templateData);
 	}
 
-	public static class ViewHolder {
+	private static class ViewHolder {
```
thomasmauerhofer: This class is completely unnecessary...

catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
```diff
@@ -1417,14 +1417,11 @@ private void copyFile(InputStream in, OutputStream out) throws IOException {
 	}
 
 	public void unzipTemplate(String projectName, String templateName, String zipFileName, Activity activity) {
```
thomasmauerhofer: There is also a UtilZip. All Zipfiles should be handled in the UtilZip or the StorageHandler. Not a few here, and a few there...

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
```diff
@@ -0,0 +1,124 @@
...
+
+	private Context context;
+	private final TemplateData templateData;
+	private final String zipFileUrl;
+	private ProgressDialog progressDialog;
+	private OnDownloadTemplateCompleteListener onDownloadTemplateCompleteListener;
+	private WebconnectionException exception;
+	private String message;
+
+	public DownloadTemplateTask(Context activity, TemplateData templateData, String zipFileUrl) {
```
thomasmauerhofer: Context should be named context...

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	public void setOnDownloadTemplateCompleteListener(OnDownloadTemplateCompleteListener listener) {
+		onDownloadTemplateCompleteListener = listener;
+	}
+
+	@Override
+	protected void onPreExecute() {
+		super.onPreExecute();
+		if (context == null) {
```
thomasmauerhofer: This is a member of the class... This never should be happened...

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
```diff
@@ -0,0 +1,124 @@
...
+		}
+		String title = context.getString(R.string.please_wait);
+		String message = context.getString(R.string.status_downloading);
+		progressDialog = ProgressDialog.show(context, title, message);
+	}
+
+	@Override
+	protected Void doInBackground(Void... arg0) {
+		try {
+			if (!Utils.isNetworkAvailable(context)) {
```
thomasmauerhofer: Think this is very similar to the above code... Isn't there a better solution, so you don't have the same code twice...

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			String zipFileString = Utils.buildPathForTemplatesZip(templateData.getName());
+
+			ServerCalls.getInstance().downloadProject(zipFileUrl, zipFileString, null, null, 0);
+		} catch (WebconnectionException | IOException webconnectionException) {
+			Log.e(TAG, Log.getStackTraceString(webconnectionException));
+			message = webconnectionException.getMessage();
+		}
+		return null;
```
thomasmauerhofer: Same as above...

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
```diff
@@ -0,0 +1,124 @@
...
+			showDialog(R.string.error_internet_connection);
+			return;
+		}
+
+		if (onDownloadTemplateCompleteListener != null) {
+			onDownloadTemplateCompleteListener.onDownloadTemplateComplete();
+		}
+	}
+
+	private void showDialog(int messageId) {
```
thomasmauerhofer: Don't create a Dialog in a method of a class... Same as above...

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -108,16 +109,16 @@ public void onClick(View view) {
 	}
 
 	@Override
-	public void onAttach(Activity context) {
```
thomasmauerhofer: Please name a activity activity.....

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -135,12 +136,19 @@ protected void handleOkButtonClick() {
...
 	}
 
-	private void loadStageProject() {
-		String templateZipFileName =
-				createLandscapeProject ? templateData.landscapeZipFileName : templateData.portraitZipFileName;
+	private void downloadTemplateFile() {
+		String zipFileUrl = baseUrlForTemplates;
+		zipFileUrl += createLandscapeProject ? templateData.getLandscape() : templateData.getPortrait();
+
+		DownloadTemplateTask downloadTemplateTask = new DownloadTemplateTask(getActivity(), templateData, zipFileUrl);
```
thomasmauerhofer: There is also a getActivity()? Then why you have to put the activity as a member?

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
```diff
@@ -135,12 +136,19 @@ protected void handleOkButtonClick() {
...
+		downloadTemplateTask.setOnDownloadTemplateCompleteListener(this);
+		downloadTemplateTask.execute();
+	}
 
-		LoadProjectTask loadProjectTask = new LoadProjectTask(getActivity(), projectName, templateData.templateName,
-				templateZipFileName, false, false);
+	private void loadStageProject() {
```
thomasmauerhofer: Shouldn't be done in a Dialog. Handle the OK-Button is not path of the dialog...

catroid/src/main/java/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -510,6 +513,24 @@ public void downloadProject(final String url, final String filePath, final Strin
 			httpClient.setConnectionSpecs(Arrays.asList(ConnectionSpec.CLEARTEXT));
 		}
 
+		if (receiver != null) {
```
thomasmauerhofer: Can this be happened?

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
thmq: have a look at [this](https://developer.android.com/guide/topics/ui/dialogs.html).

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: Well, it could be if the TemplatesFragment isn't attached to its activity.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: True, thx for the hint, I was doing something different here before

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: Your suggestion is simply not possible. 
First, the doInBackground function overrides from AsyncTask. So you can't simply add a throws someException to the method signature. This doesn't even compile.
Second, the doInBackground function runs on a background thread. This means the error dialog can't be shown at that stage (it's not the UI thread). This can be done onPostExecute(..). The exception is stored as member variable to get the error message. 

Actually, this is the same implementation as in most other Tasks in the org.catrobat.catroid.transfers package.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: I do save a possible error message from the server response here, so that it can be displayed later on the UI thread. 
It can't be handled outside as written in the comment above.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: To check whether to display an error message or not. 
Maybe it can be refactored somehow, but, again, this is the same implementation as in all other Tasks.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: Shouldn't be the case, as the listener is set before executing the task. However, IMO it makes sense to have that additional check.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: Yes, I'd say according to the Android Design Guidelines it should be a Toast.
Toast: A toast provides simple feedback about an operation in a small popup (https://developer.android.com/guide/topics/ui/notifiers/toasts.html) vs.
Dialog: A dialog is a small window that prompts the user to make a decision or enter additional information (https://developer.android.com/guide/topics/ui/dialogs.html).
The user doesn't need to make a decision, the error occurs anyway. However, I don't know what the UX team thinks about that. I remember that @thmq wanted to discuss that with the UX team.
I would vote for replacing the showDialog() methods with Toasts in all Tasks (the same method is also in most other Tasks).

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
sjaindl: It could be e.g. when there is an invalid server response and TemplateContainer class could not be deserialized.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
sjaindl: This class isn't "completely unncessary". The ViewHolder is actually an important design pattern for performance optimization in Android --> see https://developer.android.com/training/improving-layouts/smooth-scrolling.html

catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
sjaindl: Well, there are more zip-methods in the StorageHandler. I've moved them to UtilZip.

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
sjaindl: Well, it could be - again - if TemplatesFragment isn't attached to its Activity for any reason (it is originally passed from there).

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
sjaindl: I can't think of a good way to encapsulate that code for any AsyncTask (a base class isn't reasonable as there are different class parameters for different AsyncTasks), but I'm open for suggestions.

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
sjaindl: also same answer as above..

catroid/src/main/java/org/catrobat/catroid/transfers/DownloadTemplateTask.java
sjaindl: also same answer as above.. As said, I'd vote for Toasts here

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/OrientationDialog.java
sjaindl: When getActivity() is called, it is safe to do so. The fragment is attached to its activity.
However, due to the LoadProject-AsyncTask the fragment may be detached from its activity. This would lead to getActivity() returning null. That's why the activity is saved in onAttach.

catroid/src/main/java/org/catrobat/catroid/web/ServerCalls.java
sjaindl: Yes, this happens intentionally when no progress updates in percent should be given to the user that is the case when download a template.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
cheidenreich: This "forged" exception is a bit useless IMO.
You only use it to call Utils.checkForNetworkError(exception) in onPostExecute() and you know that this will return true with your created exception.
Why not directly store the boolean !Utils.isNetworkAvailable(context) and check the boolean in onPostExecute?

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,116 @@
...
+
+	@Override
+	protected void onPostExecute(String templatesList) {
+		super.onPostExecute(templatesList);
+
+		if (progressDialog != null && progressDialog.isShowing() && !((Activity) context).isFinishing()) {
+			progressDialog.dismiss();
+		}
+
+		if (Utils.checkForNetworkError(exception)) {
```
cheidenreich: see my comment about this exception above

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
cheidenreich: If this should always be set, you can do it in the constructor, thus forsing the developer to implement your interface.

In general:
I would not check it for null. If someone used this task incorrectly he/she gets a NullPointerException which is immediate feedback.
With the check its kept silent, no hint at all

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
```diff
@@ -0,0 +1,116 @@
...
+		if (onFetchTemplatesCompleteListener != null) {
+			onFetchTemplatesCompleteListener.onFetchTemplatesComplete(templatesList);
+		}
+	}
+
+	private void showDialog(int messageId) {
+		if (context == null) {
+			return;
+		}
+		if (message == null) {
```
cheidenreich: I don't like that messageId is totally ignored, if message is not null. This behaviour is not logical if you think through the program flow (behaviour based on wheter an event occoured in the AsyncTask). You should always pass a string here and keep any logic outside of this method.

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
```diff
@@ -110,14 +120,31 @@ public void onClick(View view) {
 	}
 
 	private void setImage(TemplateData templateData, ViewHolder holder) {
-		Picasso.with(context).load(templateData.templateResourceId).into(holder.image);
+		Picasso.with(context).load(getBaseUrl() + templateData.getThumbnail()).into(holder.image);
```
cheidenreich: this loads directly via HTTP? could there be problems if there is no internet connection?

catroid/src/main/java/org/catrobat/catroid/utils/UtilZip.java
```diff
@@ -173,4 +180,52 @@ public static void deleteZipFile(String fileName, String destinationFolder) {
...
 	}
+
+	public static void unzipTemplate(String projectName, String templateName, String zipFileName) {
+		Log.d(TAG, "zip file name:" + zipFileName);
+		Archiver archiver = ArchiverFactory.createArchiver("zip");
+		File unpackedDirectory = new File(Constants.DEFAULT_ROOT + "/" + templateName);
+		try {
+			archiver.extract(new File(zipFileName), unpackedDirectory);
+		} catch (IOException e) {
+			Log.d(TAG, "Can't extract program", e);
```
cheidenreich: why is this error not handled anywhere? Will the app go on as if nothing happened?

catroid/src/main/java/org/catrobat/catroid/utils/UtilZip.java
```diff
@@ -173,4 +180,52 @@ public static void deleteZipFile(String fileName, String destinationFolder) {
...
+		AssetManager assetManager = resources.getAssets();
+		String[] files = null;
+		try {
+			files = assetManager.list("");
+		} catch (IOException e) {
+			Log.e("STANDALONE", "Failed to get asset file list.", e);
+		}
+		for (String filename : files) {
+			if (filename.contains(zipFileName)) {
+				InputStream in;
```
cheidenreich: file copy on android can be done with Files.copy(original, copy);

catroid/src/main/java/org/catrobat/catroid/utils/UtilZip.java
```diff
@@ -173,4 +180,52 @@ public static void deleteZipFile(String fileName, String destinationFolder) {
...
+				try {
+					in = assetManager.open(filename);
+					File outFile = new File(Constants.DEFAULT_ROOT, filename);
+					out = new FileOutputStream(outFile);
+					StorageHandler.getInstance().copyFile(in, out);
+					out.flush();
+					out.close();
+					in.close();
+				} catch (IOException e) {
+					Log.e("STANDALONE", "Failed to copy asset file: " + filename, e);
```
cheidenreich: again, no error handling

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/transfers/FetchTemplatesTask.java
sjaindl: Yes, that's a good argument. I've changed it accordingly.
Remark: This should be refactored in most other tasks, too! There is the same problem!

sjaindl:
@cheidenreich I've changed the code according to your suggestions.
2 remarks: 
1. The same problems that you commented (error handling, dialog methods, listener != null check) are in most other Tasks, too. This should also be refactored by somebody.
2. Dialog vs. Toast: This is now inconsistent with other tasks, where dialogs are still shown in case of errors (--> UX team).

catroid/src/createatschool/java/org/catrobat/catroid/createatschool/ui/adapter/TemplateAdapter.java
sjaindl: Yes, it directly loads via http/https. If there's no Internet connection, this is handled by Picasso, so that no image is loaded at all (empty ImageView). However, this shouldn't happen as the Internet connection is already checked before templates are downloaded at all.

PR #2073, opened 2017-01-25
Bugfix of merging scenes

thomasmauerhofer:
It wasn't possible to merge projects with more than one scene.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/merge/MergeTask.java
+++ b/catroid/src/main/java/org/catrobat/catroid/merge/MergeTask.java
@@ -23,7 +23,7 @@
 
 package org.catrobat.catroid.merge;
 
-import android.app.Activity;
+import android.content.Context;
 import android.util.Log;
 
 import org.catrobat.catroid.ProjectManager;
@@ -52,7 +52,7 @@ import java.util.List;
 
 public class MergeTask {
 
-	private Activity activity;
+	private Context context;
 	private Scene mergeResult = null;
 	private Scene firstScene = null;
 	private Scene secondScene = null;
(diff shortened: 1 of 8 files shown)


cheidenreich:
when i long press a project in the project list, why can i only merge with the first one in the list?

catroid/src/main/java/org/catrobat/catroid/merge/MergeTask.java
```diff
@@ -132,34 +134,41 @@ public boolean mergeProjects(String mergeProjectName) {
...
+		File projectCodeFile = new File(Utils.buildPath(Utils.buildProjectPath(mergeProjectName), Constants.PROJECTCODE_NAME));
+		adapter.insert(new ProjectData(mergeProjectName, projectCodeFile.lastModified()), 0);
+		adapter.notifyDataSetChanged();
 
-		if (adapter != null) {
-			File projectCodeFile = new File(Utils.buildPath(Utils.buildProjectPath(mergeProjectName),
-					Constants.PROJECTCODE_NAME));
-			adapter.insert(new ProjectData(mergeProjectName, projectCodeFile.lastModified()), 0);
-			adapter.notifyDataSetChanged();
+		String msg = firstProject.getName() + " " + activity.getString(R.string.merge_info) + " " + secondProject.getName() + "!";
```
cheidenreich: is this correct for localisation?

e.g. Arabic is right to left.. so probably add placeholder like %FIRST_PROJECT% in the string resource and string replace it here?

thmq:
@thomasmauerhofer please have a look at the failing test (i.e. tests should be adapted as well).

PR #2075, opened 2017-01-26
[CAT-2064] Editing text field in formula editor should be larger than…

singhalok641:
… one line


bernadettespieler:
Please exclude these files from your branch:
 .idea/encodings.xml 
 .idea/misc.xml 
 .idea/vcs.xml 
 Catroid.iml 
catroid/catroid.iml

singhalok641:
I have excluded the files you have mentioned

bernadettespieler:
are you sure? I still see them in the "Files changed" tap.
And now you also have 2 commits - please squash them to one.

singhalok641:
I checked it again and deleted the files and squashed the commits into a single commit - which is commit number 4 in the commits tap.

thmq:
@singhalok641: You should not have deleted the .idea/.iml files because merging your changes would also delete them from the repository. What you should rather do is to execute the "gitSkipWorktreeForIntellijConfigFiles" task from the "other" folder in our gradle configuration. This task will not remove the files but simple mark them as assume unchanged for git.

Furthermore, a small implementation like this one does not need to consist of multiple commit, as @bernadettespieler has told you: please squash your commits into one. If you are not sure how to do that have a look at the git documentation ([here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)) or simple delete the branch, create a new one and copy your changes to it. Merging the develop branch into your feature branch is also not necessary.

Finally, creating a new style and overwriting all properties is strongly discouraged, it only results in unnecessary code duplicates. You should have extended the style from the existing and only add necessary changes.

singhalok641:
Thanks @thmq I will do the following changes and create a new pull request.

PR #2077, opened 2017-01-29
CAT-REF Add catroid/catroid.iml to gradle task gitSkipWorktreeForIntellijConfigFiles

robertpainsi:
With the new folder structure, a new IntelliJ config file has been added
which now will also be ignored when running the gradle task
gitSkipWorktreeForIntellijConfigFiles.


robertpainsi:
Closed, there is [more to do](https://jira.catrob.at/browse/CAT-2258).

PR #2078, opened 2017-01-30
CAT-2157 Extension for Distance to

robertriedl:


catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaParser.java
```diff
@@ -298,6 +306,49 @@ private FormulaElement collision() throws InternFormulaParserException {
...
+
+		for (Sprite sprite : ProjectManager.getInstance().getSceneToPlay().getSpriteList()) {
+			if (sprite.getName().compareTo(currentSpriteName) == 0) {
+				spriteCount++;
+			}
+		}
+
+		if (!(spriteCount == 1)) {
+			throw new InternFormulaParserException("Parse Error, Sprite was not found");
+		}
```
robertpainsi: `Scene.containsSpriteBySpriteName` or `Scene.getSpriteBySpriteName` (Those method names though… :worried:)
Use these methods everywhere possible in this PR.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternToExternGenerator.java
```diff
@@ -300,6 +298,11 @@ private String generateExternStringFromToken(InternToken internToken, boolean tr
 				String collisionTag = CatroidApplication.getAppContext().getString(R.string
 						.formula_editor_function_collision);
 				return collisionTag + "(" + internToken.getTokenStringValue() + ")";
+			case DISTANCE_TO_TOUCH_POSITION:
+			case DISTANCE_TO_SPRITE_POSITION:
+				String distanceTo = CatroidApplication.getAppContext().getString(R.string
+						.formula_editor_object_distance_to);
+				return distanceTo + "(" + internToken.getTokenStringValue() + ")";
```
robertpainsi: RTL languages

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternToken.java
```diff
@@ -78,6 +78,18 @@ public void updateCollisionFormulaToVersion(float catroidLanguageVersion) {
...
+			tokenStringValue = newName;
+		}
+	}
+
+	public void updateDistanceToSpritePositionFormula(String oldName, String newName) {
+		if (internTokenType == InternTokenType.DISTANCE_TO_SPRITE_POSITION && tokenStringValue.equals(oldName)) {
+			tokenStringValue = newName;
+		}
+	}
+
```
robertpainsi: Used?

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -319,14 +319,20 @@ public void setYInUserInterfaceDimensionUnit(float y) {
 		setY(y - getHeight() / 2f);
 	}
 
-	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
+	public double getDistanceToTouchPositionInUserInterfaceDimensions() {
```
robertriedl: `TouchUtil` is needed to get the last touch position

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaParser.java
robertriedl: These methods in `Scene` stop, when the first sprite is found, but in this case the currentSprite and the secondSprite have to be found

robertpainsi: Always keep [premature optimization](https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize) in mind.

However, you hinted me to something else.
IMO it doesn't make sense to search for the `CurrentSprite` in the `SceneToPlay`* anyway. A lot of code builds on the premise that the opened Scene,Sprite in the UI is the `getCurrentScene`,`getCurrentSprite`.
Searching for the second sprite however makes sense. If you've deleted the second sprite, the formula becomes invalid.**
What do you think?

\* `SceneToPlay == CurrentScene`. I'll create a PR for that.
\** I'm not 100% sure if *a parser* should check this, but it's not important for this PR.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -270,10 +277,34 @@ public Object interpretRecursive(Sprite sprite) {
...
 
+	private Object interpretDistanceToSpritePosition(Sprite currentSprite, String value) {
+		String secondSpriteName = value;
+		Sprite secondSprite;
+
+		try {
+			secondSprite = ProjectManager.getInstance().getSceneToPlay().getSpriteBySpriteName(secondSpriteName);
+		} catch (Resources.NotFoundException exception) {
+			return 0d;
+		}
```
robertpainsi: I should've tested it in Scratch and also do more research… returning `0` is fine (I guess). Many calculations in the FormulaElement return `0` if there is some kind of undefined intermediate result or state. The `InterpretationException` is just used if the **result** of the Formula is NaN.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -270,10 +277,43 @@ public Object interpretRecursive(Sprite sprite) {
...
+	}
+
+	private Object interpretDistanceToSpritePosition(Sprite currentSprite, String secondSpriteName) {
+		Sprite secondSprite;
+
+		try {
+			secondSprite = ProjectManager.getInstance().getSceneToPlay().getSpriteBySpriteName(secondSpriteName);
+		} catch (Resources.NotFoundException exception) {
+			return 0d;
+		}
```
robertpainsi: None of the methods in the try block throw a `NotFoundException`.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -76,16 +76,15 @@
 	private CategoryListAdapter adapter;
 
 	private static final int[] OBJECT_GENERAL_PROPERTIES_ITEMS = { R.string.formula_editor_object_transparency,
-			R.string.formula_editor_object_brightness, R.string.formula_editor_object_color/*,
-			R.string.formula_editor_object_distance_to*/ };
```
robertpainsi: Just to be sure, the `distance to` feature hasn't been enabled in any release yet, right?
So we don't have to cover backward compatibility.

robertpainsi:
The value `touch position` will be (unnecessarily) stored in the XML.
The value stored in the XML will be shown to the user, independently of the users language.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
robertriedl: no, distance to hasn't been in any release

PR #2079, opened 2017-01-31
CAT-RELEASE Increased Version Code, Version Name and Language Version

thmq:


catroid/src/main/java/org/catrobat/catroid/io/XStreamToSupportCatrobatLanguageVersion0993AndBefore.java
```diff
@@ -183,14 +183,14 @@
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
 
-public class XStreamToSupportCatrobatLanguageVersion0992AndBefore extends XStream {
+public class XStreamToSupportCatrobatLanguageVersion0993AndBefore extends XStream {
```
robertpainsi: This unnecessary change always bugged me. The class and test class has always been updated to the newest version for years now. Always the same way. It also destroys the git history feature.
I suggest to rename it. `SupportCatrobatLanguageXStream`, `BackwardCompatibleCatrobatLanguageXStream`, `XStreamToSupportCurrentCatrobatLanguageVersionAndBefore` (or `…AndOlder`) or someone can come up with a better name.

PR #2081, opened 2017-02-01
EV3 Sensor String in formula editor

MightyMendor:
Adapted the EV3 strings to match the naming of all other devices in the formula editor, and also to match the NXT sensors


MightyMendor:
will be redone with one commit only

PR #2085, opened 2017-02-02
CAT-RELEASE String Update

comeclarity:


thmq:
Apparently there is something wrong with the strings, in the Jenkins build you can see that some strings are missing. Furthermore, there is one regional identifier that should not be there: values-gl-rEs. I.e. it should be just values-gl. Please also make sure the correct Android language code is used for any new languages. And finally: please reword the commit message to something like: CAT-RELEASE String Update.

Thank you.

comeclarity:
anyone has an idea why this doesn't pass on jenkins?

robertpainsi:
> anyone has an idea why this doesn't pass on jenkins?

> … in the Jenkins build you can see that some strings are missing …

e.g. https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/1561/console

```
…/CatroidTriggerPullRequestTest/catroid/src/androidTest/java/org/catrobat/catroid/uitest/util/UiTestUtils.java:1831: error: cannot find symbol
					|| overflowMenuItemName.equals(solo.getString(R.string.unpack_keep))) {
```

robertpainsi:
Oh, you were asking `why`. That's an really interesting question. :smile_cat:

You removed all occurrences of `unpack_keep`. So there is no string called `unpack_keep` in any language causing a compile error.

For Android and during compile time, it's fine to refer to a string if it's defined at least in one language (e.g. `values-de/strings.xml`). This is the reason why it compiled just fine previously. However, if your phone is set to English, an exception will be thrown during runtime when you try to access this resource. (Resources.NotFoundException).

Shouldn't lint warn about such things anyways?

robertpainsi:
> Shouldn't lint warn about such things anyways?

Ah, I remember the discussion about it. We decided to ignore MissingTranslation because Crowdin handles all the translations for us. So when removing a string from the default translation, it will be automatically removed from all translations in the next Crowdin PR. So no need to manually remove it from every file.

catroid/src/main/res/values/strings.xml
```diff
@@ -41,6 +41,7 @@
     <string name="backpack">Backpack</string>
     <string name="packing">Pack</string>
     <string name="unpack">Unpack</string>
+    <string name="unpack_keep">Unpack and keep</string>
```
robertpainsi: Please remove `unpack_keep` and also [UiTestUtils.java#L1831](https://github.com/Catrobat/Catroid/blob/7f9dc2fcd5239be748a851a78210c385d22078af/catroid/src/androidTest/java/org/catrobat/catroid/uitest/util/UiTestUtils.java#L1831). It's not used anymore.

thmq:
@comeclarity: Thank you for fixing the problem with the "lost" string.

PR #2094, opened 2017-02-08
CAT-2178

Chypsylon:


robertpainsi:
Rebasing will :sparkles:magically:sparkles: remove all merge commits. :smile_cat:

thmq:
Rebasing will also ⭐️ effectively ⭐️ destroy our git history as this is a hot-fix, that has been merged into the master branch but not the develop. We would have to have a look at the commit hash here, if it is the same as the one in the master we should be able to fix this somehow, if not, well ... I got a bad feeling about this.

robertpainsi:
Ah, yes. For hot-fixes no rebasing of course. :smile_cat:
It's :dizzy:quite obvious:dizzy: that these merge commits are from the master branch and therefore this is a hot-fix. My fault. :flushed:

thmq:
Alright then! Still good to have someone ask questions like these 😉

PR #2097, opened 2017-02-09
CAT-2267 EV3 Motor Move Brick changed to match NXT Brick

MightyMendor:
Now the EV3 Motor Move brick is the same as the NXT Brick. No longer a "Duration in Seconds" needs to be set, but only the percentage of speed the motor should spin. 

This should make it easier to port NXT-Programs to EV3 Programs and should prevent future problems if we decide to remove the EV3 Motor Move brick with the duration-input. 
We still have the option to later add the old Brick with additional Input-field as an additional option, which will not result in broken programs (removing would).

Please be advised that programs that where created in the meantime, that use the EV3 Motor Move brick with duration-field will no longer work after this change and may result in a crash, but because EV3 is not yet released this should only apply to our developers.


catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoEv3MotorMoveBrick.java
```diff
@@ -81,26 +77,21 @@ protected Object readResolve() {
...
 
 	@Override
 	public View getPrototypeView(Context context) {
 		prototypeView = View.inflate(context, R.layout.brick_ev3_motor_move, null);
-
-		TextView textPower = (TextView) prototypeView.findViewById(R.id.ev3_motor_move_power_edit_text);
-
-		textPower.setText(String.valueOf(BrickValues.LEGO_POWER));
+		TextView textSpeed = (TextView) prototypeView.findViewById(R.id.ev3_motor_move_speed_edit_text);
+		textSpeed.setText(String.valueOf(BrickValues.LEGO_SPEED));
```
cheidenreich: why is it, that sometimes constants related to this brick are called "LEGO_EV3_SPEED" and sometimes "LEGO_SPEED"?

catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -480,7 +480,7 @@
 		legoEV3BrickList.add(new LegoEv3MotorTurnAngleBrick(LegoEv3MotorTurnAngleBrick.Motor.MOTOR_A,
 				BrickValues.LEGO_ANGLE));
 		legoEV3BrickList.add(new LegoEv3MotorMoveBrick(LegoEv3MotorMoveBrick.Motor.MOTOR_A,
-				BrickValues.LEGO_POWER, BrickValues.LEGO_DURATION));
+				BrickValues.LEGO_POWER));
```
cheidenreich: shouldn't this be LEGO_SPEED now?
If yes, please remove LEGO_POWER from BrickValues.

catroid/src/main/java/org/catrobat/catroid/devices/mindstorms/ev3/LegoEV3.java
```diff
@@ -44,6 +44,7 @@
 	void moveMotorStepsSpeed(byte outputField, int chainLayer, int speed, int step1Tacho, int step2Tacho,
 			int step3Tacho, boolean brake);
 	void moveMotorTime(byte outputField, int chainLayer, int power, int step1TimeInMs, int step2TimeInMs, int step3TimeInMs, boolean brake);
```
cheidenreich: moveMotorTime should be unused now, correct? if yes, please remove dead code

cheidenreich:
please have a look at my comments and squash into one single commit

catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoEv3MotorMoveBrick.java
MightyMendor: LEGO_SPEED is from BrickValues (default values for Brick Input fields). These are used both for EV3 and NXT Bricks (same dafault values for same behaviour bricks) that is why it is named different here.

MightyMendor:
Done. Please note that the test-case was altered too (used the dead function and was forgotten to be updated)

Sidenote: I dislike the squashing when there are different authors, now there are changes under Andre's commit that he has no idea I made.

thmq:
You did not need to squash in this case, actually it is discouraged. Multiple commits are fine as long as every single commit can be built and run. (Just a short remark on PRs containing multiple commits: do not rebase them! merge the develop branch into them instead. This prevents problems with commits that cannot be built. If you need further information regarding this issue feel free to contact me via mail.)

PR #2098, opened 2017-02-09
Extended merge function and refactoring

thomasmauerhofer:
Implement the new functionality of the merge function. If there exist an conflict between global and local userobjects, the global one will be renamed automatically.

Also refactored the MergeTask, and the associated Dialogs. All UI-stuff is made outside of the MergeTask, and the MergeManager was removed completely.
I have started to create the Dialogs in a style like @thmq use it in his refactoring.


catroid/src/main/java/org/catrobat/catroid/merge/MergeManager.java
```diff
@@ -1,133 +0,0 @@
...
-			adapter) {
-		Project firstProject = StorageHandler.getInstance().loadProject(firstProjectName, activity);
-		Project secondProject = StorageHandler.getInstance().loadProject(secondProjectName, activity);
-
-		if (firstProject == null || secondProject == null) {
-			Utils.showErrorDialog(activity, R.string.error_load_project);
-			return;
-		}
-
-		boolean justAddAsScene = firstProject.getSceneList().size() == 1 ^ secondProject.getSceneList().size() == 1;
```
robertpainsi: Out of curiosity, did you figure it out what this line was for?

robertpainsi: Ah, thanks!

catroid/src/main/java/org/catrobat/catroid/merge/ConflictHelper.java
```diff
@@ -20,92 +20,86 @@
...
-				return false;
 			}
 		}
-		return true;
 	}
 
-	private static String checkVariableNames(List<UserVariable> globalValues, List<UserVariable> localValues) {
-		for (UserVariable global : globalValues) {
-			for (UserVariable local : localValues) {
+	private static List<String> checkObjectNames(List<? extends UserObject> globalObjects, List<? extends UserObject> localObjects) {
```
cheidenreich: I would name this method findNameConflicts to be more clear what it does

catroid/src/main/java/org/catrobat/catroid/merge/MergeTask.java
```diff
@@ -60,21 +54,11 @@
 	private Project firstProject = null;
```
cheidenreich: I still don't like that the values of these Scenes and Projects depend on which Constructor is used in the first place.
For example if i create an instance with `MergeTask(Scene firstScene, Scene secondScene, Context context)` and then call `mergeProjects(String name)` firstProject and secondProject will be null but used in mergeProjects.

there is no clear border between those scenarios

cheidenreich:
any update on this?

PR #2101, opened 2017-02-11
CAT-2245 Fixes for large font

Manu91:


bernadettespieler:
also fix large font in drop down-options (variable list, broadcast, etc.)

PR #2105, opened 2017-02-13
CAT-FIX Use getCurrentScene in InternFormulaParser

robertpainsi:
Parsing only happens when editing a Formula through the
FormulaEditorFragment in the UI. So the InternFormulaParser uses
getCurrentScene only for now on.

Normally you can use getCurrentScene or getSceneToPlay in the UI. Both
are always refering to the same Scene Object when in the UI. However,
for correctness getCurrentScene is used because it represents the
currently opened Scene.


cheidenreich:
just out of curiosity, what's the actual difference between those two calls?

robertpainsi:
getCurrentScene = Current Scene in IDE
getSceneToPlay = Current Scene in Stage

https://github.com/robertpainsi/Catroid/commit/4b35742537e3d179f9afa44c9b073a4d588769fb#commitcomment-18707509 :smile:

robertpainsi:
test this please

PR #2107, opened 2017-02-14
Fix PointToAction special cases

robertpainsi:
This PR contains two fixes:
* Both Sprite on same position → direction = 90 (previously direction = 0)
* pointedSprite is invalid (e.g. doesn't exist) → don't change direction (previously direction = 0)

Base can also be changed to `release-0.9.28`

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/PointToAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/PointToAction.java
@@ -49,7 +49,7 @@ public class PointToAction extends TemporalAction {
 
 		double rotationDegrees;
 		if (spriteXPosition == pointedSpriteXPosition && spriteYPosition == pointedSpriteYPosition) {
-			rotationDegrees = 0;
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/PointToAction.java
```diff
@@ -27,23 +27,16 @@
...
-		final ArrayList<Sprite> spriteList = (ArrayList<Sprite>) ProjectManager.getInstance().getSceneToPlay()
-				.getSpriteList();
-		if (!spriteList.contains(pointedSprite)) {
-			pointedSprite = null;
-		}
-
-		if (pointedSprite == null) {
-			pointedSprite = this.sprite;
+		if (pointedSprite == null
+				|| !ProjectManager.getInstance().getSceneToPlay().getSpriteList().contains(pointedSprite)) {
```
robertpainsi: IMO it's not the job of the PointToAction to check if the Scene contains the pointedSprite. However, I kept it.

PR #2112, opened 2017-02-20
CAT-2245 Fixes for large font

Manu91:


cheidenreich:
just a general question:

if it's only the text size that needs to be larger on larger screens, cant we set the layout text size in different dimen files, like here:
https://android--examples.blogspot.co.at/2015/01/textview-font-size-in-android.html

that yould make programmatically changing it obsolete.

is there anything that prevents us from doing so?

Manu91:
Hi, I think it is not possible to set the layout text size in different dimen files for this problem because you have the possibility to set the text large or not (in the settings). It doesn't depend on the size of the screen.

cheidenreich:
ah i see, i missed that bit.
Is it possible to create a new class DynamicTextSizeArrayAdapter which is inherited from ArrayAdapter, override the getDropDownView() method there and use it whereever needed?

Manu91:
Thank you, this is a good suggestion. I will change it immediately.

PR #2114, opened 2017-02-21
CAT-2280 Fixes Edit Formula of bricks with multiple input fields

comeclarity:
In most of the switch statements, there was no implementation for the case that no input field is selected (which is the case if one wants to edit the formula by pressing "edit formula")


catroid/src/main/java/org/catrobat/catroid/content/bricks/GlideToBrick.java
```diff
@@ -169,6 +169,10 @@ public void showFormulaEditorToEditFormula(View view) {
 			case R.id.brick_glide_to_edit_text_duration:
 				FormulaEditorFragment.showFragment(view, this, BrickField.DURATION_IN_SECONDS);
 				break;
+
+			default:
+				FormulaEditorFragment.showFragment(view, this, BrickField.DURATION_IN_SECONDS);
+				break;
```
robertpainsi: When you move the `default` keyword to the corresponding `case` you don't have to implement it twice.
```java
case R.id.brick_glide_to_edit_text_duration:
default:
    // Code
```

comeclarity: thanks for the tip :)

robertpainsi:
@comeclarity, you missed a few bricks.
Can you please create a pull request for them? `release-0.9.28` as base. So it's going to be in the next release. :smile_cat:

* ArduinoSendDigitalValueBrick
* ArduinoSendPWMValueBrick
* DroneMoveBrick
* InsertItemIntoUserListBrick
* LegoEv3PlayToneBrick
* LegoNxtPlayToneBrick
* PhiroRGBLightBrick
* RaspiPwmBrick
* RaspiSendDigitalValueBrick
* SetTextBrick

PR #2117, opened 2017-02-22
CAT-2290 Updating and Outsourcing of Scratch server call tests

cfkh:


catroid/src/androidTest/java/org/catrobat/catroid/test/scratchconverter/ScratchServerCallsTest.java
```diff
@@ -0,0 +1,230 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
```
84n4n4: the copyright header in project hasnt been updated yet everywhere, so checkstyle complains that this one is not ok.
can you just change it to "2016" for this PR?

PR #2120, opened 2017-02-22
CAT-2322 Arduino output tests

jobermue:
Contains a few tests for ArduinoImpl (NOTE: only tests setDigitalArduinoPin() and setAnalogArduinoPin())

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
@@ -30,12 +30,22 @@ import org.catrobat.catroid.common.bluetooth.ConnectionDataLogger;
 import java.nio.ByteBuffer;
 
 import name.antonsmirnov.firmata.BytesHelper;
+import name.antonsmirnov.firmata.writer.AnalogMessageWriter;
+import name.antonsmirnov.firmata.writer.DigitalMessageWriter;
+import name.antonsmirnov.firmata.writer.ReportAnalogPinMessageWriter;
+import name.antonsmirnov.firmata.writer.ReportDigitalPortMessageWriter;
+import name.antonsmirnov.firmata.writer.SetPinModeMessageWriter;
 
 public final class FirmataUtils {
 
-	private static final int SET_PIN_MODE_COMMAND = 0xF4;
-	private static final int REPORT_ANALOG_PIN_COMMAND = 0xC0;
-	private static final int ANALOG_MESSAGE_COMMAND = 0xE0;
+	private static final int DIGITAL_MESSAGE_COMMAND = DigitalMessageWriter.COMMAND; // 0x90
+	private static final int ANALOG_MESSAGE_COMMAND = AnalogMessageWriter.COMMAND; // 0xE0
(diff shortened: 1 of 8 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
```diff
@@ -44,7 +51,24 @@ public FirmataUtils(ConnectionDataLogger logger) {
...
 
-	public FirmataMessage getAnalogMesageData() {
+	public FirmataMessage getDigitalMessageData() {
+
+		int portAndCommand = getNextMessage();
+
+		int port = getPinFromHeader(portAndCommand);
+		int command = getCommandFromHeader(portAndCommand);
+
+		Assert.assertEquals("This is no digital message command", DIGITAL_MESSAGE_COMMAND, command);
```
cheidenreich: I would not use asserts outside of tests. Those are designed to close the app in case of an unrecoverable state. Please use exceptions and handle them accordingly

catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
```diff
@@ -93,10 +131,10 @@ public int getNextMessage() {
 	}
 
 	private int getPinFromHeader(int header) {
-		return header & 15;
+		return header & 0x0F;
```
cheidenreich: please put those "magic numbers" in a static variable e.g. BITMASK_PIN

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -46,6 +46,9 @@
 
 	public static final int NUMBER_OF_DIGITAL_PINS = 14;
 	public static final int PINS_IN_A_PORT = 8;
+	public static final int NUMBER_OF_DIGITAL_PORTS = (NUMBER_OF_DIGITAL_PINS + PINS_IN_A_PORT - 1) / PINS_IN_A_PORT;
+
+	public static final int NUMBER_OF_ANALOG_PINS = 6;
```
cheidenreich: please reorder them and remove the empty line

catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
jobermue: This method is supposed to be used in tests only. The original methods in this file use a similar style. Also [BluetoothTestUtils.java](https://github.com/Catrobat/Catroid/blob/develop/catroid/src/androidTest/java/org/catrobat/catroid/common/bluetooth/BluetoothTestUtils.java) uses a similar style. If you still want me to use an exception here, could you suggest which exception to use?

catroid/src/androidTest/java/org/catrobat/catroid/common/firmata/FirmataUtils.java
jobermue: ok.

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
jobermue: ok.

PR #2135, opened 2017-02-24
Fixes lint issue.

mfuchs:
According to the linter the operator | should be on the next line here.


robertpainsi:
This is the 3rd time I see this code change :laughing:

1. ad482907b CAT-2216 fixes restard button within scenes
2. 81e5b6aa8 CAT-2280 Fixes Edit Formula of bricks with multiple input fields
3. d368b0fd9 According to the linter the operator | should be on the next line here.

PR #2141, opened 2017-02-26
CAT-2163 Enable reading Arduino pin 7

jobermue:
Firmata sends a value containg the aggregate status of all pins
of a given port that are configured as input pins. Thus that value
can be in the range of 0-255.
This fixes the problem of always reading low for pin 7 and
additionally fixes problems if multiple pins are set to inputs and
read HIGH (which might result in a value > 64, even without pin 7).


robertpainsi:
ok to test

PR #2143, opened 2017-03-01
CAT-2229 Apply profile always in main menu after restart

FlorianWeissensteiner:
CAT-2229
https://jira.catrob.at/browse/CAT-2229

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
@@ -24,6 +24,9 @@ package org.catrobat.catroid.ui;
 
 import android.app.ActionBar;
 import android.app.Activity;
+import android.app.AlarmManager;
+import android.app.PendingIntent;
+import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.net.Uri;
@@ -96,6 +99,9 @@ public class MainMenuActivity extends BaseActivity implements OnLoadProjectCompl
 	public static final String SHARED_PREFERENCES_SHOW_BROWSER_WARNING = "shared_preferences_browser_warning";
 	public static final int REQUEST_CODE_GOOGLE_PLUS_SIGNIN = 100;
 
+	public static final String RESTART_INTENT = "restart";
+	private final int restartDelayMsAfterAppClose = 500;
+
 	private static final String TYPE_FILE = "file";
 	private static final String TYPE_HTTP = "http";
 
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -96,6 +99,9 @@
 	public static final String SHARED_PREFERENCES_SHOW_BROWSER_WARNING = "shared_preferences_browser_warning";
 	public static final int REQUEST_CODE_GOOGLE_PLUS_SIGNIN = 100;
 
+	public static final String RESTART_INTENT = "restart";
+	private final int restartDelayAfterAppClose = 500;
```
cheidenreich: please specify the unit in the variables name.
i guess its milliseconds so it would be restartDelayMsAfterAppClose

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -113,6 +119,14 @@ protected void onCreate(Bundle savedInstanceState) {
 			return;
 		}
 
+		if (getIntent().getBooleanExtra(RESTART_INTENT, false)) {
+			Intent restartIntent = getApplicationContext().getPackageManager().getLaunchIntentForPackage(getApplicationContext().getPackageName());
+			PendingIntent intent = PendingIntent.getActivity(getApplicationContext(), 0, restartIntent, PendingIntent.FLAG_CANCEL_CURRENT);
+			AlarmManager manager = (AlarmManager) getApplicationContext().getSystemService(Context.ALARM_SERVICE);
+			manager.set(AlarmManager.RTC, System.currentTimeMillis() + restartDelayAfterAppClose, intent);
```
cheidenreich: what is the reason behind this delay?

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -113,6 +119,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
+			System.exit(2);
```
cheidenreich: why 2?
the android documentation states: Termination status. By convention, a nonzero status code indicates abnormal termination.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
FlorianWeissensteiner: thanks for taking a look. now the formatting/import errors showed up on my local test, so i must have been doing something wrong the last time.
i adjusted the variable name

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
FlorianWeissensteiner: The delay is there to give the app some time to terminate, before it is launched again. Launching the app while it is still terminating results in unexpected/unwanted behaviour

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
FlorianWeissensteiner: True that. changed to zero

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
cheidenreich: ok, can you guarantee that this already happened after 500 ms?

FlorianWeissensteiner: i cannot. The 500ms seemed to be plenty of time on the tested devices/emulators, but i can't guarantee that it will work on any device out there.

cheidenreich: maybe have a look at recreate(): https://developer.android.com/reference/android/app/Activity.html#recreate%28%29

Disclaimer: I haven't spent much time digging into this, but it may be more appropriate here

FlorianWeissensteiner: Unfortunately it just terminates the app without restarting it. Seems like that's what happens if you use recreate in the activity that is the lowest on the stack (Main Menu).

PR #2145, opened 2017-03-01
CAT-2231 Apply fontsize in formula editor with respect to device

FlorianWeissensteiner:
CAT-2231
https://jira.catrob.at/browse/CAT-2231

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
@@ -30,12 +30,19 @@ import android.text.Spannable;
 import android.text.SpannableString;
 import android.text.style.RelativeSizeSpan;
 import android.text.style.TextAppearanceSpan;
+import android.util.DisplayMetrics;
 import android.util.TypedValue;
+import android.view.Display;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.ViewGroup;
+import android.view.WindowManager;
+import android.widget.Button;
+import android.widget.TableLayout;
+import android.widget.TableRow;
 import android.widget.TextView;
 
+import org.catrobat.catroid.CatroidApplication;
 import org.catrobat.catroid.R;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
```diff
@@ -69,6 +80,38 @@ public static void enlargeViewGroup(ViewGroup viewGroup) {
...
+		}
+
+		Context context = CatroidApplication.getAppContext();
+		WindowManager wm = (WindowManager) context.getSystemService(Context.WINDOW_SERVICE);
+		Display display = wm.getDefaultDisplay();
+		DisplayMetrics metrics = new DisplayMetrics();
+		display.getMetrics(metrics);
+
+		if ((metrics.widthPixels / metrics.xdpi) > widthForIncreasedTextSizeInInches) {
+			modifier += modifierIncreaseDelta;
```
cheidenreich: as modifier is a member variable i see potential issues by modifying it here.
this affects all other methods in the class

catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
```diff
@@ -80,6 +123,17 @@ public static void enlargeTextView(TextView textView) {
 		}
 	}
 
+	public static void enlargeButtonText(Button button) {
+		if (modifier == 1.0f) {
+			return;
+		}
+
+		if (!enlargedButtons.contains(button)) {
```
cheidenreich: is this keeping of already modified buttons necessary?

catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
FlorianWeissensteiner: I increase the modifier by the delta, resize the text, and then decrease the modifier by the delta so it's back to normal again. This happens in the same function.
Unless i misunderstood you, I don't think this could pose a problem?

I changed the if statement to check a boolean for clarity, and moved the code to determine if it's a large device in a separate function.

catroid/src/main/java/org/catrobat/catroid/utils/TextSizeUtil.java
FlorianWeissensteiner: Good catch, this is indeed not necessary, I removed it.

PR #2152, opened 2017-03-03
CAT-2122 name check in ok button click event in NewDataDialog

KerschbaumerM:
Deleted the checkName function and postponed his functionality into the OK-Button press Event. Introduced another check too(if name consists only of white spaces then error dialog appears). 
Furthermore the OK-Button Event appears in the setOnShowListener. In this case the dialog will not be closed if the OK-Button is pressed and the Customer can digit(input) another name.

### DIFF
--- a/catroid/src/main/res/values/strings.xml
+++ b/catroid/src/main/res/values/strings.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8" standalone="no"?>
-<!--
+<?xml version="1.0" encoding="utf-8" standalone="no"?><!--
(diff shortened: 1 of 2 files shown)


catroid/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8" standalone="no"?>
-<!--
+<?xml version="1.0" encoding="utf-8" standalone="no"?><!--
```
cheidenreich: why? 😄

KerschbaumerM: hmmm...formatter did this O.o

PR #2154, opened 2017-03-04
CAT-FIX Updated license text

thmq:


robertpainsi:
Checkstyle will not report every file :disappointed: I'll take a look at it sometimes later. In the meantime:

```
modified:   catroid/build.properties
modified:   catroid/gradle/wrapper/gradle-wrapper.properties
modified:   gradle.properties
modified:   settings.gradle
```

I used the following script :wink: 
```
grep -rl --exclude-dir=.git --exclude-dir=build '2010-2016 The Catrobat Team' . | xargs sed -i 's/2010-2016 The Catrobat Team/2010-2017 The Catrobat Team/g'
```

thmq:
I already removed the files you listed 😉 there have been some problems with both the code style tasks as well as with our .gitignore configuration. also, we should think about which files really need a copyright header. we should probably have a chat about that somewhen.

PR #2155, opened 2017-03-05
CAT-2106: Username cannot be an Email ID

FleetAdmiral:
Fixes CAT-2106: Error message is shown when username is put as an email id.


FleetAdmiral:
Made a mistake with the labels on Jira, will come back to this once the issue has been moved to 'Ready for development'

PR #2156, opened 2017-03-07
CAT-2151 Fixed crash when whatsapp not installed

chandruscm:
Social sharing via the WhatsApp button on a details page in the webview crashes Pocket Code when the WhatsApp app is not installed on one's phone.

Added code to check WhatsApp installation before starting the share intent.


chandruscm:
Code format error fixed.

PR #2159, opened 2017-03-07
CAT-NOLB Additional tracking

sjaindl:
Adds tracking of:
- Project submission
- Broadcast message creation
- Enabling/Disabling of hints

+ increases versionCode for next release

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/devices/mindstorms/ev3/LegoEV3MotorTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/devices/mindstorms/ev3/LegoEV3MotorTest.java
@@ -59,8 +59,8 @@ public class LegoEV3MotorTest extends AndroidTestCase {
 		int inputSpeed = -70;
 		byte outputField = (byte) 0x01;
 
-		int expectedSpeed = (EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_MAX.getByte() & inputSpeed) |
-				EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_SIGN_NEGATIVE.getByte();
+		int expectedSpeed = (EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_MAX.getByte() & inputSpeed)
+				| EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_SIGN_NEGATIVE.getByte();
 
 		byte expectedOutputField = (byte) 0x01;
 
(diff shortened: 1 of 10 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/devices/mindstorms/ev3/LegoEV3MotorTest.java
```diff
@@ -59,8 +59,8 @@ public void testMotorMoveTest() {
 		int inputSpeed = -70;
 		byte outputField = (byte) 0x01;
 
-		int expectedSpeed = (EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_MAX.getByte() & inputSpeed) |
-				EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_SIGN_NEGATIVE.getByte();
+		int expectedSpeed = (EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_MAX.getByte() & inputSpeed)
+				| EV3CommandByte.EV3CommandParamByteCode.PARAM_SHORT_SIGN_NEGATIVE.getByte();
```
robertpainsi: Please don't add unrelated changes to commits, also the `done_button` changes. Afaik jenkins will still test your code even with Checkstyle warnings. If not, we definitely have to be more careful with merging unformatted code.
https://github.com/Catrobat/Catroid/pull/2135#issuecomment-283788639 :joy_cat:

An example how your commit message could also look like
```
Add tracking feature and increase versionCode to 7

Following actions will be tracked
* project submission
* creation of broadcast message
* enabling of hints
```
However, the `and` indicates that there are two things done in this commit. So I suggest to split it into two commits.

Split or not to split is always a hard decision.
e.g. your code changes also need to upgrade a library. So you keep everything in a single commit. Later on you find out that your code changes need to be reverted. But there are already other commits also depending on the library upgrade…

e.g. you implement a new feature and also add a library. So you split it into two commits. Later on this feature will be removed entirely. You just find the commit with the code and revert it leaving the commit with the library still in it. Since you always keep your libraries up-to-date, you always update an unnecessary library.

If it is other code instead of a library you're dealing with, reverting is much more painful. But it's not all about reverting. Actually it's nearly nothing about reverting and just easier to explain. It's all about jumping and searching through your history and also understand why the code is the way it currently is. However, I'll stop here because it's already after :sleeping_bed: :worried: 

Since the versionCode will be increased with or without this change, splitting is fine… I guess :laughing:

robertpainsi: Ah, it's not a new feature ^^ Then something like `Extend tracking feature…` or `Add new tracking actions` or `Track additional actions` or whatever you come up with.

PR #2162, opened 2017-03-08
Upgradle to 3.3

thmq:


robertpainsi:
I added an `Under discussion` label because a newer gradle job configuration is needed.
The current progress can be found in [JENKINS-75](https://jira.catrob.at/browse/JENKINS-75)

catroid/build.gradle
```diff
@@ -140,7 +140,6 @@ dependencies {
     natives "com.badlogicgames.gdx:gdx-box2d-platform:$gdxVersion:natives-x86"
 
     compile 'com.facebook.android:facebook-android-sdk:4.14.1'
-    compile 'com.android.support:support-v4:24.1.1'
```
robertpainsi: I'm not sure how gradle dependencies are handled and what are the best practices. There are four options here:

* Remove the explicitly defined support library `v4:24.1.1`. However, there will be code in our app using a support library. Which version used is defined by other dependencies, I guess.

* Downgrade the support library to `v4:23.4.0`. However, this is only a temporary solution until another library uses a different support library version. e.g. by updating the facebook version which uses `v4:25+`

* Keep or upgrade the android support library and suppress the lint error.

* Somehow exclude support libraries by using `exclude`.

So, what should we do?

robertpainsi:
@thmq, the PR title :rofl:

thmq:
green run [here](https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/1738/).

robertpainsi:
ping @mfuchs

PR #2172, opened 2017-03-09
CAT-2297: Changed background color of webview to that of application

FleetAdmiral:
Color of webView background has been changed to "#04222C" (or application_background_color), as mentioned in CAT-2297 (subtask under CAT-2296).


FleetAdmiral:
Updating upstream and making new pull request.

PR #2179, opened 2017-03-14
CAT-2169 Added nullcheck so StopAllScriptsAction won't run into nullpointer exception

rschluesselbauer:
When reset() method is called on a RepeatAction, the Action member is set to null, but is not checked for null the next time it is called - this null check stops the RepeatAction in case action member is null.


rschluesselbauer:
Original functionality never worked in the first place so I'm making some more changes.

PR #2183, opened 2017-03-17
CAT Refactor DataContainer

sjaindl:
This branch introduces a generic strategy pattern for the DataContainer with some factory methods.
 
I've also moved it to a new package and changed some methods, so that keys are passed as first argument and values as second.


catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+		List<V> data = getDataMap().get(key);
+
+		if (data == null) {
+			data = new ArrayList<>();
+			getDataMap().put(key, data);
+		}
+		return data;
+	}
+
+	V find(List<V> dataList, String name) {
```
cheidenreich: why not handing over the key here also? then you can use this in `exists(K key, String name)` as well

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+	Map<Sprite, List<V>> cloneForScene(Scene scene, Map<K, List<V>> dataMap) {
+		Map<Sprite, List<V>> clonedList = new HashMap<>();
+
+		for (Map.Entry<K, List<V>> entry : dataMap.entrySet()) {
+			List<V> newList = new ArrayList<>();
+			for (V data : entry.getValue()) {
+				newList.add(newInstance(getDataName(data)));
+			}
+
+			if (entry.getKey() == null) {
```
cheidenreich: can this happen?
why not checking before cloning the whole list?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+		V dataToDelete = find(dataList, name);
+		if (dataToDelete != null) {
+			dataList.remove(dataToDelete);
+			return true;
+		}
+		return false;
+	}
+
+	Map<Sprite, List<V>> cloneForScene(Scene scene, Map<K, List<V>> dataMap) {
+		Map<Sprite, List<V>> clonedList = new HashMap<>();
```
cheidenreich: this should be called `clonedMap`

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+
+			String keyName = getKeyName(entry.getKey());
+			Sprite clonedInstance = scene.getSpriteBySpriteName(keyName);
+			clonedList.put(clonedInstance, newList);
+		}
+
+		return clonedList;
+	}
+
+	V rename(K key, String newName, String oldName) {
```
cheidenreich: usually newName is the last parameter

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+			for (V data : entry.getValue()) {
+				newList.add(newInstance(getDataName(data)));
+			}
+
+			if (entry.getKey() == null) {
+				continue;
+			}
+
+			String keyName = getKeyName(entry.getKey());
+			Sprite clonedInstance = scene.getSpriteBySpriteName(keyName);
```
cheidenreich: `scene.getSpriteBySpriteName()` does NOT clone the instance, its simply a reference to the same object

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+		List<V> varList = getOrCreate(key);
+		V dataToRename = find(varList, oldName);
+		setDataName(dataToRename, newName);
+		return dataToRename;
+	}
+
+	void clean(K key) {
+		List<V> data = getDataMap().get(key);
+		if (data != null) {
+			data.clear();
```
cheidenreich: I think this is not necessary as the GC will remove it as soon as there is no more reference on the Object

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteVariableBehaviour.java
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	@Override
+	protected boolean isClone(Sprite key) {
+		return key.isClone();
+	}
+
+	private void removeSpriteVariableWithSameSpriteName(Sprite spriteToKeep) {
+		if (getDataMap() == null || spriteToKeep == null) {
```
cheidenreich: i would not check that, as it prevents Exceptions, which we want in this case to avoid follow up errors

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/UserBrickVariableBehaviour.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private DataContainer dataContainer;
+
+	UserBrickVariableBehaviour(DataContainer dataContainer) {
+		this.dataContainer = dataContainer;
+	}
+
+	@Override
+	protected void reset(List<UserVariable> dataList) {
+		//not supported
```
cheidenreich: throw `UnsupportedOperationException` ?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,202 @@
...
+
+	V add(K key, String name, Object value) {
+		List<V> dataList = getOrCreate(key);
+		V dataToAdd = newInstance(name, value);
+		dataList.add(dataToAdd);
+		return dataToAdd;
+	}
+
+	V addIfNotExists(K key, String name) {
+		List<V> list = getDataMap().get(key);
```
cheidenreich: you can use your find method here also

catroid/src/main/java/org/catrobat/catroid/content/bricks/UserBrick.java
```diff
@@ -297,6 +297,17 @@ public void onLayoutChanged(View currentView) {
 		}
 	}
 
+	public void deleteUserVariables(String userVariableName) {
+		List<UserScriptDefinitionBrickElement> currentElements = getUserScriptDefinitionBrickElements();
+		for (int id = 0; id < currentElements.size(); id++) {
+			if (currentElements.get(id).getText().equals(userVariableName) && currentElements.get(id).isVariable()) {
+				Context brickContext = getDefinitionBrick().view.getContext();
+				getDefinitionBrick().removeVariablesInFormulas(currentElements.get(id).getText(), brickContext);
+				currentElements.remove(id);
```
sjaindl: Well, I've deleted the method now anyway. It was already unused before - would possibly the best to refactor that method in the userbricks-branch in the future

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/ProjectDataBehaviour.java
```diff
@@ -0,0 +1,67 @@
...
+		for (V data : getDataList()) {
+			if (getDataName(data).equals(name)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	private V find(List<V> dataList, String name) {
+		if (dataList == null) {
```
sjaindl: just checked it, nope couldn't as a new ArrayList is created if the project lists or variables are null in getProjectLists() and getProjectVariables() in Project.
I've removed that check now.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/ProjectDataBehaviour.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	V rename(String newName, String oldName) {
+		V dataToRename = find(getDataList(), oldName);
+		setDataName(dataToRename, newName);
+		return dataToRename;
+	}
+
+	boolean exists(String name) {
+		for (V data : getDataList()) {
```
sjaindl: Good catch! Even shorter now ;)

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: Yes, makes a lot easier - I've removed some redundant code while refactoring this, too

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: There is some bug that causes null entries - I'm going to find out when I refactoring the whole process of copying/cloning/backpacking etc. as next step.
But true, for now I'm changed this by removing illegal entries beforehand.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: True, I'm not sure whether this is intentional?? 
This method is for cloning objects during stage execution (clone bricks) - I think that's the reason for the possibly misleading name here.
I've changed the name to "clonedInstanceOnStage" to be more clear

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: Yes, it should be enough to call 
getDataMap().remove(key);

sjaindl:
@cheidenreich Thx for reviewing - It's even more compact now.
After this is merged I'm going to refactor all the cloning/copying/backpacking etc. stuff

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,185 @@
...
+	boolean exists(K key, V value) {
+		List<V> list = getDataMap().get(key);
+		return list != null && list.contains(value);
+	}
+
+	boolean exists(K key, String name) {
+		if (key == null) {
+			return false;
+		}
+		List<V> list = getDataMap().get(key);
```
cheidenreich: you could use find here too

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
```diff
@@ -0,0 +1,185 @@
...
+		return null;
+	}
+
+	boolean exists(K key, V value) {
+		List<V> list = getDataMap().get(key);
+		return list != null && list.contains(value);
+	}
+
+	boolean exists(K key, String name) {
+		if (key == null) {
```
cheidenreich: i dont like this null check, should it be possible?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteVariableBehaviour.java
```diff
@@ -0,0 +1,117 @@
...
+	@Override
+	public List<UserVariable> getOrCreate(Sprite key) {
+		List<UserVariable> data = super.getOrCreate(key);
+		removeIllegalSpriteVariableEntries(key);
+		return data;
+	}
+
+	@Override
+	Map<Sprite, List<UserVariable>> cloneForScene(Scene scene, Map<Sprite, List<UserVariable>> originalData) {
+		removeIllegalSpriteVariableEntries(null);
```
cheidenreich: when do those illegal entries appear?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: Well, it could happen if current sprite is null due to any reason.

sjaindl: However, I think it shouldn't normally happen as this is only called within a sprite (rename and create dialogs). Otherwise it'll be another bug - so I'll remove that check ;)

cheidenreich: haha yeah, define any ;)

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteDataBehaviour.java
sjaindl: true, the good thing is it need to be changed only once now

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteVariableBehaviour.java
sjaindl: There is some bug that causes the creation of entries with null keys (null sprites) that leads to several other strange bugs. I haven't researched on that exactly yet, but IMO it should be refactored/fixed together with the clone/copy etc. methods - the refactoring branch that you have mentioned ;)

sjaindl:
> FYI the refactoring of cloning/copying/backpack is already in progress and in advanced state
Cool, maybe we can discuss next week at the meeting or so about further refactoring ;)

catroid/src/main/java/org/catrobat/catroid/formulaeditor/datacontainer/SpriteVariableBehaviour.java
cheidenreich: Sounds to me like the delete method is not called when it should be.
Was this happening before also?
Can you reproduce this easily?

cheidenreich: I really don't like this kind of bug workarounds, cause things like this leave a lot of unused code when the actual bug is fixed.

IMO since you're editing the module in which the bug appears it should be fixed right away.

sjaindl: I also don't like them - it's not a good idea to just "fight against the symptoms". 
I'm just not sure yet whether the problem is in the DataContainer itself (or an outside method that calls a DataContainer method). But I can have a look at it in this PR, too.
I'm not sure whether I can reproduce it easily (I can try different scenarios and see whether this bug still happens), but I don't think that deletion is the problem. I rather think that the problem is getOrCreate when null keys are passed in.

PR #2187, opened 2017-03-19
Changed progressCircle with progressDialog on loading webview

nishantt95:
It is a small and simple change where progressCircle is replaced with progressDialog. progressCircle wasn't clearly visible when webview loads some graphic data (loads half of the website but still loading rest of the data) and hence replaced it with progressDialog which is clearly visible with a message "Please Wait...".


robertpainsi:
Hi @nishantt95, thanks for contributing! :smiley_cat:

Please create a Jira Sub-Task [here](https://jira.catrob.at/browse/CAT-2296) first. Don't forget to add the reason for the change and an image of the new loading dialog. An UX member will give you feedback.

Closed for now. Will be reopened on demand.

PR #2189, opened 2017-03-20
CAT-FIX Adapt checkstyle license regex check for XML files

robertpainsi:
Checkstyle forced to use the start comment tokens in XML files on its
own line.

```
<?xml version="1.0" encoding="utf-8"?>
<!--
```

Creating new files with Android Studio also adds the comment token on
its own line. However, when reformatting the XML, the comment token will
be moved to the first line.

```
<?xml version="1.0" encoding="utf-8"?><!--
```

Since this is a bug of Android Studio and there is no fix for it,
Checkstyle allows both versions for now on.

https://github.com/Catrobat/Catroid/pull/2152#discussion_r106465282

___

This is only a temporary solution. We never configured the Android Studio XML formatter rules and also the CatrobatBot does only check java files.

\+ Developer creating new XML files won't have issues with the formatter.
\- Developer formatting whole XML files (instead of VCS changes only) will definitely create PR with license header changes.

I'm not sure if we should merge it… :joy_cat:


robertpainsi:
Meh, lets drop it. I prefer similarity anyway. :dancers:

PR #2193, opened 2017-03-22
CAT-XP6 [LS, MS] Change_ByNTests

M-Schiller:
refactored UI tests from Robotium to Espresso for bricks
* ChangeXByNBrick
* ChangeYByNBrick
* ChangeBrightessByNBrick


84n4n4:
ok to test

PR #2213, opened 2017-03-25
CAT-2169 Stop All Scripts Bugs

rschluesselbauer:
Added a nullcheck to RepeatAction (because it would run into null pointer when an action is reset).

Added a method to Look, that resets all broadcast- and when-scripts to what they were before running the "stop all scripts" block. Before that, stop all scripts would result in none of the scripts working anymore.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -43,10 +43,14 @@ import com.badlogic.gdx.utils.Array;
 import org.catrobat.catroid.common.Constants;
 import org.catrobat.catroid.common.DroneVideoLookData;
 import org.catrobat.catroid.common.LookData;
+import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.utils.TouchUtil;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 public class Look extends Image {
 	private static final float DEGREE_UI_OFFSET = 90.0f;
@@ -639,6 +643,13 @@ public class Look extends Image {
 			setUniformf(HUE_STRING_IN_SHADER, hue);
 			end();
 		}
+	}
+
+	public Map<String, List<String>> createScriptActions() {
+		this.setWhenParallelAction(null);
+		Map<String, List<String>> scriptActions = new HashMap<>();
+		sprite.createStartScriptActionSequenceAndPutToMap(scriptActions, false);
+		return scriptActions;
 	}
 
 	public Polygon[] getCurrentCollisionPolygon() {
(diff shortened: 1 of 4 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/RepeatAction.java
```diff
@@ -67,7 +67,7 @@ public boolean delegate(float delta) {
 		if (executedCount >= repeatCountValue && !isForeverRepeat) {
 			return true;
 		}
-		if (action.act(delta) && currentTime >= LOOP_DELAY) {
+		if (action != null && action.act(delta) && currentTime >= LOOP_DELAY) {
```
cheidenreich: do you knw why this can be null?

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -641,6 +645,13 @@ public void setHue(float hue) {
 		}
 	}
 
+	public void handleStopAllScripts() {
+		this.setWhenParallelAction(null);
+		Map<String, List<String>> scriptActions = new HashMap<>();
+		sprite.createStartScriptActionSequenceAndPutToMap(scriptActions, false);
+		StageActivity.stageListener.precomputeActionsForBroadcastEvents(scriptActions);
```
cheidenreich: I'd rather return the scriptActions from this method and call `precomputeActionsForBroadcastEvents()` from the `StopAllScriptsAction` since that already has the dependency to `StageActivity`. 
You would solve this without introducing the dependency here.

Just thinking: the Look shouldn't handle stage logic, the action does ;)

catroid/src/main/java/org/catrobat/catroid/content/actions/RepeatAction.java
robertpainsi: > do you knw why this can be null?

e.g. [StopOtherScriptsAction](https://github.com/Catrobat/Catroid/blob/2b637e7fefb7c91614ad79c3fb504e91a101b8e0/catroid/src/main/java/org/catrobat/catroid/content/actions/StopOtherScriptsAction.java#L47) calls reset which sets the action to null [here (libgdx.DelegateAction)](https://github.com/libgdx/libgdx/blob/b0163a4eaf95aca6cb04d300c82bd3d13f142dee/gdx/src/com/badlogic/gdx/scenes/scene2d/actions/DelegateAction.java#L55).

When and where is the action set again? Or is the whole RepeatAction recreated?

rschluesselbauer: The action is recreated by handleStopAllScripts() in the Look class.
However as @cheidenreich suggested I'll put that logic directly into the StopAllScriptsAction.

catroid/src/main/java/org/catrobat/catroid/content/Look.java
rschluesselbauer: I made the suggested change and squashed the commits.

PR #2214, opened 2017-03-25
CAT-2082 Crash related to multitouch sensors

rschluesselbauer:
Added a lot of nullchecks for many function types - this was not in the scope of this issue, but leaving it would create a lot more issues in the long run (better to fix it straight away).
Corrected some nullchecks.
Did some minor refactoring.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
+++ b/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
@@ -96,7 +96,7 @@ public class FormulaElement implements Serializable {
 	}
 
 	public List<InternToken> getInternTokenList() {
-		List<InternToken> internTokenList = new LinkedList<InternToken>();
+		List<InternToken> internTokenList = new LinkedList<>();
 
 		switch (type) {
 			case BRACKET:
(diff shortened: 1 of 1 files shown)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -500,19 +500,24 @@ private Object interpretFunction(Functions function, Sprite sprite) {
...
 				}
 				break;
 			case MULTI_FINGER_TOUCHED:
-				return TouchUtil.isFingerTouching(doubleValueOfLeftChild.intValue()) ? 1d : 0d;
+				return (doubleValueOfLeftChild != null && TouchUtil.isFingerTouching(doubleValueOfLeftChild.intValue
+						())) ? 1d : 0d;
 			case MULTI_FINGER_X:
-				return Double.valueOf(TouchUtil.getX(doubleValueOfLeftChild.intValue()));
+				return doubleValueOfLeftChild != null ? (double) TouchUtil.getX(doubleValueOfLeftChild
+						.intValue()) : 0d;
```
cheidenreich: can you put 0d to the top of this method as `final double defaultReturnValue = 0d` and use it everywhere. easier to read ;)

rschluesselbauer: I made the suggested change and squashed the commits.

PR #2219, opened 2017-03-28
CAT-2345 Multilingual Feature on Application Level

ndrnour:
The app must provide an interface that handles user input to switch languages independent of the smartphone’s global settings. This feature should list all languages supported by the app including languages not yet supported by the operating system.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/MultilingualTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/MultilingualTest.java
@@ -1,0 +1,186 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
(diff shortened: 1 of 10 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/ui/MultilingualTest.java
```diff
@@ -0,0 +1,186 @@
...
+		solo.scrollDown();
+		solo.scrollDown();
+		solo.scrollDown();
+		solo.scrollDown();
+		solo.clickInList(6);
+		solo.sleep(2000);
+		solo.waitForActivity(MainMenuActivity.class);
+		solo.assertCurrentActivity("MainMenu", MainMenuActivity.class);
+		String s = getActivity().getString(R.string.app_name);
+		assertEquals(s, "پاکت کد");
```
cheidenreich: can you please put all those button names and expected strings into static final string variables with english names (e.g. BUTTON_NAME_PROGRAMS) at the top of this test class, so developers who cannot read these languages can understand what this test does

PR #2231, opened 2017-03-30
CAT-1150 Deleted automatic screenshot file once manual screenshot has been taken

chandruscm:
**Change 1**: Deleted `checkIfAutomaticScreenshotShouldBeTaken` boolean
Redundant as the value of `checkIfAutomaticScreenshotShouldBeTaken` boolean is initialized to `true` and never changed.

**Change 2**: Using `&&` in place of `||`
`screenshotExists()` returns `false` if screenshot exists.

The result of:
````
makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME) || scene
				.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) || scene.screenshotExists
				(SCREENSHOT_MANUAL_FILE_NAME);
````
returns `true` if either the Automatic or Manual screenshot does not exist.

Required action: Take the automatic screenshot only if neither of them exist. (Use `&&` instead of `||`)
````
makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME) && scene
				.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) && scene.screenshotExists
				(SCREENSHOT_MANUAL_FILE_NAME);
````
**Change 3**: Deleted automatic screenshot when a manual screenshot has been made
````
if(saveScreenshot(this.screenshot, SCREENSHOT_MANUAL_FILE_NAME)) {
	deleteAutomaticScreenshot();
	return true;
} else {
	return false;
}

private void deleteAutomaticScreenshot() {
	FileHandle automaticScreenshot = Gdx.files.absolute(pathForSceneScreenshot + SCREENSHOT_AUTOMATIC_FILE_NAME);
	if(automaticScreenshot.exists()) {
		automaticScreenshot.delete();
	}
}
````

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -118,7 +118,6 @@ public class StageListener implements ApplicationListener {
 	private boolean reloadProject = false;
 	public boolean firstFrameDrawn = false;
 
-	private static boolean checkIfAutomaticScreenshotShouldBeTaken = true;
 	private boolean makeAutomaticScreenshot = false;
 	private boolean makeScreenshot = false;
 	private String pathForSceneScreenshot;
@@ -244,10 +243,9 @@ public class StageListener implements ApplicationListener {
 		}
 		axes = new Texture(Gdx.files.internal("stage/red_pixel.bmp"));
 		skipFirstFrameForAutomaticScreenshot = true;
-		if (checkIfAutomaticScreenshotShouldBeTaken) {
-			makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME) || scene
-					.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) || scene.screenshotExists(SCREENSHOT_MANUAL_FILE_NAME);
-		}
+		makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME) && scene
+				.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) && scene.screenshotExists
+				(SCREENSHOT_MANUAL_FILE_NAME);
 		if (drawDebugCollisionPolygons) {
 			collisionPolygonDebugRenderer.setProjectionMatrix(camera.combined);
 			collisionPolygonDebugRenderer.setAutoShapeType(true);
@@ -506,10 +504,10 @@ public class StageListener implements ApplicationListener {
 
 			/*
 			 * Necessary for UiTests, when EMMA - code coverage is enabled.
-			 * 
+			 *
 			 * Without setting DYNAMIC_SAMPLING_RATE_FOR_ACTIONS to false(via reflection), before
 			 * the UiTest enters the stage, random segmentation faults(triggered by EMMA) will occur.
-			 * 
+			 *
 			 * Can be removed, when EMMA is replaced by an other code coverage tool, or when a
 			 * future EMMA - update will fix the bugs.
 			 */
@@ -704,7 +702,20 @@ public class StageListener implements ApplicationListener {
 		while (makeScreenshot) {
 			Thread.yield();
 		}
-		return saveScreenshot(this.screenshot, SCREENSHOT_MANUAL_FILE_NAME);
+
+		if(saveScreenshot(this.screenshot, SCREENSHOT_MANUAL_FILE_NAME)) {
+			deleteAutomaticScreenshot();
+			return true;
+		} else {
+			return false;
+		}
+	}
+
+	private void deleteAutomaticScreenshot() {
+		FileHandle automaticScreenshot = Gdx.files.absolute(pathForSceneScreenshot + SCREENSHOT_AUTOMATIC_FILE_NAME);
+		if(automaticScreenshot.exists()) {
+			automaticScreenshot.delete();
+		}
 	}
 
 	private boolean saveScreenshot(byte[] screenshot, String fileName) {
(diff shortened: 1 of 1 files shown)


84n4n4:
the whole situation around the checkIfAutomaticScreenshotShouldBeTaken is kind of hacky. although its never directly set, it is modified by tests via reflection to turn it off.
a nicer way to go around all this would either be a setter annotated with @VisibleForTesting, or refactoring the whole thing to have the screenshotting done somewhere separate to be able to mock this for testing.

chandruscm:
@84n4n4 That explains! 
I will first try the adding the @VisibleForTesting annotation to a setter method. 
Do changes 2 and 3 look good?

84n4n4:
regarding  - Do changes 2 and 3 look good?

i honestly dont know, was there a bug/issue ticket of something breaking here?

-- screenshotExists() returns false if screenshot exists.
this is bad and should be fixed in the first place and refactored on all usages.

--Required action: Take the automatic screenshot only if neither of them exist.
although the name of the variable suggests that, im unsure that this is what should happen at that point. without digging deep into the code i cannot be sure that this isnt actually intended to work the way it does now, but only the names were chosen really bad. (see above function name...)
(i fear that quite a few components rely on this, especially the upload service, etc, and they maybe need an automatic screenshot in place even if there is a manual one...)

chandruscm:
> i honestly dont know, was there a bug/issue ticket of something breaking here?

There are no bugs and nor is anything breaking, [CAT-1150](https://jira.catrob.at/browse/CAT-1150) is a performance enhancement as mentioned in the ticket.

> screenshotExists() returns false if screenshot exists.

Should I add this as a sub task and refactor it in a new PR?

**My proposed solution**

Why are there 2 screenshot files in the first place? Let's just have one screenshot file. An automatic screenshot file is generated at the first instance and if the user were to make a manual screenshot, the screenshot file is overridden. This screenshot file can be used everywhere.

84n4n4:
>>There are no bugs and nor is anything breaking, CAT-1150 is a performance enhancement as mentioned in the ticket.

ah, ok.

>>Should I add this as a sub task and refactor it in a new PR?

i guess its fine to just do this in this PR and add as comment in the issue ticket that this has been cleaned up.

>>Why are there 2 screenshot files in the first place?

i honestly dont know, probably upload/ web / apk generator, etc. prefers one of them?

robertpainsi:
>> Why are there 2 screenshot files in the first place?

> i honestly dont know, probably upload/ web / apk generator, etc. prefers one of them?

[CAT-1150](https://jira.catrob.at/browse/CAT-1150) just states, that ["there can only be one"](http://www.imdb.com/title/tt0091203) screenshot _at the end_. However, it's important to have multiple screenshot file _names_.

* Each time you start the stage, a new automatic screenshot is taken automatically and _the old one is overwritten_, if there is no manual screenshot yet.
* If there is a manual screenshot, no automatic screenshot is taken.

If you would use a single screenshot file name, you wouldn't know if it's an automatic or manual screenshot and if it should be overwritten (automatic screenshot) or not (manual screenshot).
So a flag is needed to know if the current screenshot is a manual or automatic one. It's solved by using different screenshot file names (instead of e.g. saving it in the xml).

If this wasn't the question, just ignore my comment :smile_cat:

chandruscm:
Hey @robertpainsi thanks for getting back :) 

> If there is a manual screenshot, no automatic screenshot is taken.

This is not the case as of now, a new automatic screenshot is generated at the end of every game even if a manual screenshot exists.

This happens at the `finish()` method.
````
public void finish() {
		...
		if (thumbnail != null && !makeAutomaticScreenshot) {
			saveScreenshot(thumbnail, SCREENSHOT_AUTOMATIC_FILE_NAME);
		}
		...
	}
````
The condition `if (thumbnail != null && !makeAutomaticScreenshot)` is always `true`

Upon checking the following initialization for `makeAutomaticScreenshot`
````
makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME) || scene
					.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) || scene.screenshotExists(SCREENSHOT_MANUAL_FILE_NAME);
````

`project.manualScreenshotExists` returns `true`, i.e the manual screenshot does not exist even though it does exist. Please note the `screenshotExists()` function returns `false` if it exists, so I know this is getting pretty confusing at this point! Like @84n4n4 pointed, this needs to changed!

Because `project.manualScreenshotExists()` checks for the screenshot file in the location `/storage/emulated/0/Pocket Code/Galaxy War/manual_screenshot.png` i.e outside the scene folder, while all manual screenshots are saved inside the scene folder.

So this check needs to be removed right? or am I missing something here?

rschluesselbauer:
I think there might have been screenshots in the project folder at some point, but in the current version I couldn't find code that would store screenshots there. Seniors might know more about it.

This one looks odd though:
`makeAutomaticScreenshot = project.manualScreenshotExists(SCREENSHOT_MANUAL_FILE_NAME)
 && scene.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) 
&& scene.screenshotExists(SCREENSHOT_MANUAL_FILE_NAME);`

You only makeAutomaticScreenshot if:

- there is no project screenshot (let's assum that is always true) AND 
- no manual screenshot AND 
- no automatic screenshot

However I thought desired behaviour is to always update the automatic screenshot, unless there is a manual screenshot? If that is the case you should not check for this: `scene.screenshotExists(SCREENSHOT_AUTOMATIC_FILE_NAME) `

bernadettespieler:
Currently not mergeable. Please reopen if fixed

PR #2232, opened 2017-03-30
CAT-2356 Automatically logout non-NOLB users

sjaindl:


catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -390,6 +390,12 @@ public void handleUploadButton(View view) {
 		if (!viewSwitchLock.tryLock()) {
 			return;
 		}
+
+		if (BuildConfig.CREATE_AT_SCHOOL && !Utils.isCreateAtSchoolUser(this)) {
+			SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
+			sharedPreferences.edit().putString(Constants.TOKEN, Constants.NO_TOKEN).apply();
```
cheidenreich: (I know that this is already like this in other parts of the code)

if you edit multiple items you should call in this order:
```
SharedPreferences.Editor prefEditor = sharedPreferences.edit();
prefEditor.putString();
prefEditor.putString();
//... all other operations on this editor
prefEditor.apply();
```
for the following reasons: 

- apply() is an atomic write operation, meaning no part of the app could access the preferences in a weird state where for example Constants.TOKEN is written, but Constants.USERNAME still has the old value.
- Each time you call edit() a new SharedPreferences.Editor is created which creates a full copy of the original preferences and each apply() writes your modiefied copy back to the original -> performance

please check all your code bits which use this

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/IconsUtilTest.java
```diff
@@ -57,7 +57,7 @@ public void testAddIcons() {
 		assertNotNull("Control pos drawable not found.", correctDrawable);
 
 		Bitmap bitmap = ((BitmapDrawable) drawables[0]).getBitmap();
-		assertEquals("Icon size did not match the expected size.", IconsUtil.smallIconSizeBricks, drawables[0].getBounds());
+		assertEquals("Icon size did not match the expected size.", IconsUtil.getSmallIconSizeBricks(), drawables[0].getBounds());
```
cheidenreich: why is something icon related done in this PR? 😛

catroid/src/main/java/org/catrobat/catroid/web/ServerCalls.java
```diff
@@ -817,6 +817,16 @@ public boolean login(String username, String password, String token, Context con
 			String serverAnswer = jsonObject.optString(JSON_ANSWER);
 			SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
 
+			if (BuildConfig.CREATE_AT_SCHOOL) {
+				boolean createAtSchoolUser = jsonObject.optBoolean(Constants.CREATE_AT_SCHOOL_USER);
+				sharedPreferences.edit().putBoolean(Constants.CREATE_AT_SCHOOL_USER, createAtSchoolUser).apply();
+				if (!createAtSchoolUser) {
+					sharedPreferences.edit().putString(Constants.TOKEN, Constants.NO_TOKEN).apply();
```
cheidenreich: is there a reason not to call `Utils.logoutUser()`?

cheidenreich: @sjaindl please check that, after this we can merge, if your next PR is the bugfix for invalid entries ;)

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
sjaindl: That's a really good hint! I've changed it accordingly

catroid/src/main/java/org/catrobat/catroid/web/ServerCalls.java
sjaindl: Yes, now I remember the reason --> In the create@school-flavor there is an additional Tracking done in Utils.logoutUser() -- TrackingUtil.trackLogoutEndSessionEvent(context);
This should not be logged in the case of an error here.
Facebook and Google+ logout isn't relevant, as only native login is provided in this flavor. Furthermore, we don't want to show the Toast message "Logout successful". So it should be perfectly fine here to reset those 2 entries in the shared prefs.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/IconsUtilTest.java
sjaindl: Well, that was just a fix for not compiling tests. But yeah you're right, normally this should be a separate PR :P

catroid/src/main/java/org/catrobat/catroid/web/ServerCalls.java
sjaindl: Yes, next I'll have a look at the invalid entries bug ;)

PR #2234, opened 2017-04-01
MUS-225 Refactor of UI espresso tests

joseffilzmaier:
Apply changes suggested by Thomas.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
@@ -22,68 +22,76 @@
  */
 package org.catrobat.catroid.uiespresso.pocketmusic;
 
-import android.app.Activity;
-import android.content.pm.ActivityInfo;
-import android.support.test.espresso.Espresso;
-import android.support.test.espresso.IdlingResource;
+import android.content.Intent;
 import android.support.test.espresso.UiController;
 import android.support.test.espresso.ViewAction;
 import android.support.test.espresso.matcher.BoundedMatcher;
 import android.support.test.runner.AndroidJUnit4;
 import android.view.View;
 
+import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.R;
+import org.catrobat.catroid.common.SoundInfo;
+import org.catrobat.catroid.pocketmusic.PocketMusicActivity;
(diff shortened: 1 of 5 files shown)


joseffilzmaier:
@84n4n4 The UI tests are now much faster (<2sec on my emulator compared to 50+ sec before). Thanks for the tipps.
The jenkins checkstyle is failing due to some arduino files... are we allowed to change the year numbers of other teams?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
```diff
@@ -160,41 +136,38 @@ public void toggleRandomNoteViews() {
...
-		Activity currentActivity = getCurrentActivity();
-		currentActivity.setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
-		currentActivity.setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
-	}
+		List<SoundInfo> soundInfo = ProjectManager.getInstance().getCurrentSprite().getSoundList();
 
-	@After
-	public void tearDown() throws Exception {
+		Intent pocketMusicDataIntent = null;
+		if (soundInfo != null && !soundInfo.isEmpty()) {
```
84n4n4: i guess the test calling this will fail if this is null, but for ease of debugging if something breaks here wouldnt it be nicer to do something like:
assertNotNull(soundInfo)
assertFalse(soundInfo.isEmpty());
instead of the if(...)

84n4n4:
nice!
yes, would be nice if you would fix those arduino files...
thank you!

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/RecyclerViewMatcher.java
```diff
@@ -54,8 +54,7 @@ public void describeTo(Description description) {
 					try {
 						idDescription = this.resources.getResourceName(recyclerViewId);
 					} catch (Resources.NotFoundException var4) {
```
84n4n4: why var4?
Resources.NotFoundException e
would be a bit nicer..

joseffilzmaier: Yes it makes no sense. I'll correct it.

PR #2236, opened 2017-04-03
CAT-1152 Reduce size of screenshots

rschluesselbauer:
To get all screenshots we need to know which scenes the project has (screenshots are in their scenes' respective folders) - therefore we pass an array of scene names to the ProjectUploadService and scale all images accordingly. The scaling method had to be adjusted, because it didn't provide scaling to a specific width/height. Also the justScaleDown feature of getScaledBitmapFromPath(...) was implemented as the flag was previously passed without doing anything.

**ProjectAndSceneScreenshotLoader**: Made getScreenshotFile(...) method of inner class ScreenshotLoader accessible to public. This method is needed for retrieving the screenshot Files by their project and scene name.

**ProjectUploadService**: Retrieve an array of scene names from the intent, loop through all the scene names and get their screenshots with ProjectAndSceneScreenshotLoader's getScreenshotFile(...) method. Refactored if-chain to a switch statement. Added scene-name-array to the uploadBackupBundle (in case the upload fails).

**UploadProgressDialog**: Put the scene-name-array into the intent for the ProjectUploadService. Added small method for creating an array of scene names from a list of scenes.

**ImageEditing**: Overloaded scaleImageFile(...) method to scale with width and height in addition to a scaling factor. Changed getScaledBitmapFromPath(...) to use the justScaleDown boolean flag (instead of doing nothing with it).

**StatusBarNotificationManager**: Added scene-names-array to the reuploadIntent

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/io/ProjectAndSceneScreenshotLoader.java
+++ b/catroid/src/main/java/org/catrobat/catroid/io/ProjectAndSceneScreenshotLoader.java
@@ -105,6 +105,12 @@ public class ProjectAndSceneScreenshotLoader {
 		}
 	}
 
+	public File getScreenshotFile(String projectName, String sceneName, boolean isBackpackScene) {
+		ScreenshotData screenshotData = new ScreenshotData(projectName, sceneName, isBackpackScene, null);
+		ScreenshotLoader screenshotLoader = new ScreenshotLoader(screenshotData);
+		return screenshotLoader.getScreenshotFile();
+	}
+
 	class ScreenshotLoader implements Runnable {
 		ScreenshotData projectAndSceneScreenshotData;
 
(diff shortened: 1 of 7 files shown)


catroid/src/main/java/org/catrobat/catroid/io/ProjectAndSceneScreenshotLoader.java
```diff
@@ -192,6 +167,44 @@ public void run() {
 				}
 			});
 		}
+
+		File getScreenshotFile() {
```
cheidenreich: If you're motivated, please check the sanity of this code, and maybe come up with a better solution (if applicable)

catroid/src/main/java/org/catrobat/catroid/utils/ImageEditing.java
```diff
@@ -97,15 +97,17 @@ public static Bitmap getScaledBitmapFromPath(String imagePath, int outputRectang
...
-		if (resizeType == ResizeType.STRETCH_TO_RECTANGLE) {
-			newWidth = outputRectangleWidth;
-			newHeight = outputRectangleHeight;
-		} else if (resizeType == ResizeType.STAY_IN_RECTANGLE_WITH_SAME_ASPECT_RATIO) {
-			newWidth = (int) Math.floor(originalWidth / sampleSizeMaximum);
-			newHeight = (int) Math.floor(originalHeight / sampleSizeMaximum);
-		} else if (resizeType == ResizeType.FILL_RECTANGLE_WITH_SAME_ASPECT_RATIO) {
-			newWidth = (int) Math.floor(originalWidth / sampleSizeMinimum);
-			newHeight = (int) Math.floor(originalHeight / sampleSizeMinimum);
+		if (!(justScaleDown && outputRectangleHeight > originalHeight && outputRectangleWidth > originalWidth)) {
```
cheidenreich: maybe express this in a more readable way

catroid/src/main/java/org/catrobat/catroid/utils/ImageEditing.java
```diff
@@ -214,9 +216,14 @@ public static Bitmap rotateBitmap(Bitmap bitmap, int rotationDegree) {
 	public static void scaleImageFile(File file, double scaleFactor) throws FileNotFoundException {
 		String path = file.getAbsolutePath();
 		int[] originalBackgroundImageDimensions = getImageDimensions(path);
-		Bitmap scaledBitmap = ImageEditing.getScaledBitmapFromPath(path,
+		scaleImageFile(file,
```
cheidenreich: what if FileNotFoundException gets thrown?

cheidenreich:
one test is failing now:
junit.framework.AssertionFailedError: Loaded bitmap has incorrect height expected:<500> but was:<100>
at junit.framework.Assert.fail(Assert.java:50)
at junit.framework.Assert.failNotEquals(Assert.java:287)
at junit.framework.Assert.assertEquals(Assert.java:67)
at junit.framework.Assert.assertEquals(Assert.java:199)
at org.catrobat.catroid.test.utiltests.ImageEditingTest.testGetBitmap(ImageEditingTest.java:107)

rschluesselbauer:
Based on the comments I made the following additional changes:

ImageEditingTest: Edited test to test for onlyScaleDown flag (and fixed faulty part)

ImageEditing: Made if statement in getScaledBitmapFromPath more readable; 
fixed a bug in getScaledBitmapFromPath where height was assigned to width and the other way around (only noticable since onlyScaleDown is implemented); 
split up getScaledBitmapFromPath into two methods for better reusability; removed unused getScaledBitmapOfLoadedBitmap(...) and getImageDimensionsForLoadedImage(...)

UtilFile: Some cosmetic refactoring (I just happened to stumble accross the file and the formatting was off)

ProjectUploadService: Handled FileNotFoundException so that the upload is not stopped when scaling down a screenshot fails

ProjectAndSceneScreenshotLoader: Refactored getScreenshotFile to be a bit shorter

PR #2239, opened 2017-04-04
CAT-2333 Username stripped of whitespaces and test was added to check the issue

demonicode:


demonicode:
Thanks, will keep that in mind (new branch)
By the way, can I run this test server to check these warnings before making pull requests?

demonicode:
its done , new commit is added.

demonicode:
Can someone please verify it.

demonicode:
thanks for the info.

PR #2252, opened 2017-04-05
MUS-FIX Move jitpack definition

oliewa92:
Remove jitpack definition from parent gradle file and move it to lower one.


joseffilzmaier:
@thmq Is it okay like this?

PR #2256, opened 2017-04-06
CAT-2250 Don't display a think/say bubble when the text equals ""

rschluesselbauer:
Added a check in ThinkSayBubbleAction:
When the text equals "", don't create a new bubble actor and don't add it to the stage.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
@@ -47,14 +47,15 @@ public class ThinkSayBubbleAction extends TemporalAction {
 			Log.d(getClass().getSimpleName(), "Formula interpretation for this specific Brick failed.", interpretationException);
 			return;
 		}
-		ShowBubbleActor showBubbleActor = new ShowBubbleActor(textToDisplay, sprite, type);
-
 		if (StageActivity.stageListener.getBubbleActorForSprite(sprite) != null) {
 			StageActivity.stageListener.getStage().getActors().removeValue(StageActivity.stageListener.getBubbleActorForSprite(sprite), true);
 			StageActivity.stageListener.removeBubbleActorForSprite(sprite);
 		}
-		StageActivity.stageListener.addActor(showBubbleActor);
-		StageActivity.stageListener.putBubbleActor(sprite, showBubbleActor);
+		if (!textToDisplay.equals("")) {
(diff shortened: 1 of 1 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
```diff
@@ -47,14 +47,15 @@ protected void update(float delta) {
 			Log.d(getClass().getSimpleName(), "Formula interpretation for this specific Brick failed.", interpretationException);
 			return;
 		}
-		ShowBubbleActor showBubbleActor = new ShowBubbleActor(textToDisplay, sprite, type);
-
 		if (StageActivity.stageListener.getBubbleActorForSprite(sprite) != null) {
```
loibi93: If you have a program:

ShowBubble "Hello"
Wait 10 Seconds
ShowBubble ""

The Bubble with "Hello" would disappear, is this intended ?

robertpainsi: image
According to Scratch, yes

robertpainsi:
ok to test

rschluesselbauer:
There has been an addition to the requirements so this PR is on hold ATM.
A formula element that is completely empty should get the value of an empty string, if the brick expects a string.

To the people who know their way around Jenkins - I am very confused about the results of the build. Could you explain to me why I get checkstyle errors for files I didn't even touch?

robertpainsi:
@rschluesselbauer, about Checkstyle, don't worry about it :wink: Jenkins tests the merge of both branches, in this case `Catrobat:develop` and `rschluesselbauer:CAT-2250`, and there [have been Checkstyle warnings](https://github.com/Catrobat/Catroid/pull/2261) in the develop branch.

You already did the right thing :1st_place_medal:! Instead of fixing and adding unrelated code to this PR, which is a really bad thing to do, you mentioned that there is a problem.

About this PR, please also add a test for this behavior. If possible, a unit test.

robertpainsi:
> A formula element that is completely empty…

How can a formula element be completely empty or what do you mean by that?

rschluesselbauer:
By editing the speak bubble text and removing everything. Normally the resulting formula is not accepted, but according to the new requirements it should be accepted for specific blocks (like thinkbubble or speak or note) and replaced by a formula that represents an empty string.

robertpainsi:
@rschluesselbauer, ah, now I get it. Thanks for the update :smile_cat:

PR #2262, opened 2017-04-11
CAT-2355 Removing Hard-Coding String

ndrnour:
In the course of localizing software, all strings should be declared as resources and separated from the source code. In Pocket Code, program size is expressed in units of measurement (KB, MB, GB, etc.), these units are implemented as a hard-coded string.


ndrnour:
what's wrong!! 

was ist los!! @_@

robertpainsi:
* Whitespace changes are changes too.
* https://softwareengineering.stackexchange.com/questions/194788/why-doesnt-git-merge-adjacent-lines-without-conflict (imo the 2nd answer is the better one).

Just remove the changes on these lines in your commit entirely (https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits).

PR #2269, opened 2017-04-20
CAT-2250 Don't display a think/say bubble when the text equals ""

rschluesselbauer:
This is the second pull request for this issue, as I could not reopen the other one (https://github.com/Catrobat/Catroid/pull/2256) after force pushing.
I had to make quite a few changes to the code to make it testable, so please bear with me ;)

Added **ThinkSayBubbleActionTest** testing new behaviour (bubble disappears if ThinkSayBubble with text "" is created)
**ThinkSayBubbleAction**: moved BubbleActor creation to seperate method (for better testability); moved some code concerning adding/removing BubbleActors to StageListener
**WaitForBubbleBrickAction**:  moved some code concerning adding/removing BubbleActors to StageListener
**Brick**: added a new array that contains all BrickFields that expect a string value & a method that returns a boolean value if a field expects a string value
**ThinkBubbleBrick** & **ThinkForBubbleBrick**: adjusted the call to the ActionFactory method for creating a ThinkSayBubbleAction
**ActionFactory**: merged think & say methods into one
InternFormulaParser: added a constant; minor refactoring
**StageListener**: moved some code concerning adding/removing BubbleActors to StageListener from ThinkSayBubbleAction and WaitForBubbleBrickAction, minor refactoring
**FormulaEditorFragment**: used constants from InternFormulaParser instead of its own ones; now saves a formula with no input, if the BrickField expects a string
**build.gradle**: upped the version of com.google.dexmaker:dexmaker-mockito in order for Mockito.spy to work (needed for my test)

### DIFF
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -151,7 +151,7 @@ dependencies {
     compile fileTree(include: '*.jar', dir: 'src/main/libs-natives')
     androidTestCompile fileTree(include: '*.jar', dir: 'src/androidTest/libs')
     androidTestCompile 'com.jayway.android.robotium:robotium-solo:5.2.1'
-    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
+    androidTestCompile 'com.linkedin.dexmaker:dexmaker-mockito:1.5.1'
     androidTestCompile 'org.mockito:mockito-core:1.10.19'
 
     androidTestCompile ('com.android.support.test:runner:0.5') {
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
@@ -22,8 +22,6 @@
  */
 package org.catrobat.catroid.content.actions;
 
-import android.util.Log;
-
 import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
 
 import org.catrobat.catroid.content.Sprite;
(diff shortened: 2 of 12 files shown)


catroid/build.gradle
```diff
@@ -151,7 +151,7 @@ dependencies {
     compile fileTree(include: '*.jar', dir: 'src/main/libs-natives')
     androidTestCompile fileTree(include: '*.jar', dir: 'src/androidTest/libs')
     androidTestCompile 'com.jayway.android.robotium:robotium-solo:5.2.1'
-    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
+    androidTestCompile 'com.linkedin.dexmaker:dexmaker-mockito:1.5.1'
```
robotic-heidenreich: is this necessary?

catroid/src/main/java/org/catrobat/catroid/content/actions/ThinkSayBubbleAction.java
```diff
@@ -40,21 +38,28 @@
...
-		StageActivity.stageListener.addActor(showBubbleActor);
-		StageActivity.stageListener.putBubbleActor(sprite, showBubbleActor);
+		if (showBubbleActor != null) {
+			StageActivity.stageListener.setBubbleActorForSprite(sprite, showBubbleActor);
+		}
+	}
+
+	public ShowBubbleActor createBubbleActor() throws InterpretationException {
+		String textToDisplay = text == null ? "" : text.interpretString(sprite);
+		if (textToDisplay.equals("")) {
```
robotic-heidenreich: this can be replaced by `textToDisplay.isEmpty()`

robotic-heidenreich:
ok to test

robertpainsi:
> …as I could not reopen the other one (#2256) after force pushing.

[How to reopen a pull-request after a force-push?](https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c)

catroid/build.gradle
rschluesselbauer: It is necessary, if you want to use the 'spy'-feature of mockito which I did in my test. 1.2 was the highest available version by google, so this does seem like the only way to up the version.

However if we don't want this I will need to find another way to write a test.

cheidenreich: well, if it's only for tests, it's ok. in the worst scenario we have to rewrite the test if this stops working for any reason.

PR #2271, opened 2017-04-21
CAT-1611 X/Y axis flip with large (horizontal layout) tablets

rschluesselbauer:
Added new methods for getting default device rotation and for checking if the orientation needs to be flipped and used them for the appropriate functions (tilt, acceleration, ...) in the SensorHandler, removed unnecessary boxing.

rotateOrientation() will return 0 if the orientation should not be rotated, 1 & -1 if it should be rotated 90° in one or another direction.


thmq:
ok to test

PR #2291, opened 2017-05-03
CAT-2382 Numbering style for bidirectional languages

ndrnour:
Numbering styles vary depending on the region. bidirectional languages numbering style may be implemented by either Hindi digits or Arabic digits

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/controller/LookController.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/controller/LookController.java
@@ -68,6 +68,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.TreeSet;
 
 public final class LookController {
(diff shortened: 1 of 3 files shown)


84n4n4:
This branch is 1 commit ahead, 89 commits behind Catrobat:develop. 

this is not going to be mergable, please rebase and fix.

cheidenreich:
test this please

ndrnour:
need to close

catroid/src/main/java/org/catrobat/catroid/ui/controller/LookController.java
```diff
@@ -154,7 +155,10 @@ private void handleDetails(LookData lookData, LookViewHolder holder, LookBaseAda
 				holder.lookFileSizeTextView.setText(UtilFile.getSizeAsString(new File(lookData.getAbsolutePath())));
 			}
 			int[] measure = lookData.getMeasure();
-			String measureString = measure[0] + " x " + measure[1];
+			int Length = measure[0];
+			int Width = measure[1];
```
rschluesselbauer: This code is working correctly, but there is a mixup with the naming of local variables:
lookData.getMeasure() will return an array looking like this: {width, height}
So your 'Length' should be renamed to 'width' and 'Width' should be renamed to 'height'. (Those local variables should be written in camelCase, else our checkstyle will complain ;) )

rschluesselbauer:
I don't think that robotium tests are still added to our test suite - new tests should be written in Espresso.
@cheidenreich could you clarify if that is the case?

catroid/src/main/java/org/catrobat/catroid/ui/controller/LookController.java
ndrnour: First! thanks for your replay 
actually this PR is no longer available, instant of 
" 84n4n4 commented on May 24 
This branch is 1 commit ahead, 89 commits behind Catrobat:develop.
this is not going to be mergable, please rebase and fix."
I deleted my repository because I didn't know how to fix that and I made a new fork. 
could you close this PR and have a look at the new PR under the Link : 
https://github.com/Catrobat/Catroid/pull/2326

rschluesselbauer: Ok, please go ahead and close this PR, as I cannot do it!

PR #2294, opened 2017-05-09
CAT-2214 Layout Rendering for Right-to-Left languages

AyyalAwwad:
This pull request supports bidirectional languages in the following elements:
XML files for Pocket Code standards' layout.
Drawable files which have a specific directional orientation (back, next, undo, and redo).
RTL version for style file.

I have 10 errors and these errors occur since I use the paddingStart instead of paddingLeft and paddingEnd instead of using paddingRight specifically for RTL languages Just, which are included in layout-ldrtl package.

Android guidelines for RTL languages mentioned: to take advantage of RTL layout mirroring, we have to change all of the app's "left/right" layout properties to new "start/end" equivalents.

Note: The localization testing will be in the next PR


cheidenreich:
test this please

rschluesselbauer:
> I have 10 errors and these errors occur since I use the paddingStart instead of paddingLeft and paddingEnd instead of using paddingRight specifically for RTL languages Just, which are included in layout-ldrtl package.

Are you planning on fixing those or is the min api version going to be turned up to 17?
In case the min api version is turned up, you should look into changing those "drawableLeft/Right" to "drawableStart/End". Wouldn't we avoid maintaining two versions of the layout files that way?

AyyalAwwad:
NO need to change drawableLeft/Right" to "drawableStart/End" since drawableLeft/Right" are used in a different style file which is created for Arabic language

rschluesselbauer:
Right, but would it not be good if we could do all this in ONE file? It is probably way easier to maintain one set of style files instead of two. I would not work on this though until anyone can confirm that a switch to api 17 is done. @cheidenreich  please have a look.

AyyalAwwad:
But for some UI elements attributes such as gravity: left or right, or LeftOF or RightOf attributes we have to create a different style file for RTL languages.

rschluesselbauer:
It looks like api 17 solves those too though, but I don't know if upgrading to api 17 is even a talking point.

bernadettespieler:
R2L feature already covered in merged PR

PR #2298, opened 2017-05-11
CAT-2213 Intro for formula editor

FlorianWeissensteiner:
#CAT-2213
https://jira.catrob.at/browse/CAT-2213


FlorianWeissensteiner:
Applied the same changes as suggested in https://github.com/Catrobat/Catroid/pull/2299

PR #2299, opened 2017-05-11
CAT-2213 Intro for formula editor

FlorianWeissensteiner:
#CAT-2213
https://jira.catrob.at/browse/CAT-2213

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
@@ -1,0 +1,261 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.utils;
+
+import android.app.ActionBar;
+import android.app.Activity;
+import android.graphics.Bitmap;
+import android.graphics.Canvas;
+import android.graphics.Color;
+import android.graphics.ColorFilter;
+import android.graphics.LightingColorFilter;
+import android.graphics.Matrix;
+import android.graphics.Paint;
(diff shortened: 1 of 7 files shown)


FlorianWeissensteiner:
fixing the checkstyle errors...

FlorianWeissensteiner:
fixed.

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
```diff
@@ -0,0 +1,261 @@
...
+		summaryStringIds[idx++] = R.string.formula_editor_intro_summary_logic;
+		summaryStringIds[idx++] = R.string.formula_editor_intro_summary_device;
+		summaryStringIds[idx++] = R.string.formula_editor_intro_summary_data;
+	}
+
+	private static void setParamsOfView(View view, int highlightRectangleIndex, int widthModifier, int heightModifier) {
+		int width = 1;
+		int height = 1;
+		int[] coords = new int[2];
+		coords[0] = 0;
```
cheidenreich: this initialization is not necessary: http://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5

> Each class variable, instance variable, or array component is initialized with a default value when it is created

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
```diff
@@ -0,0 +1,261 @@
...
+		coords[1] = 0;
+
+		if (view != null) {
```
cheidenreich: is this possible?

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
```diff
@@ -0,0 +1,261 @@
...
+		screenshot = Bitmap.createBitmap(screenShotView.getMeasuredWidth(), screenShotView
+				.getMeasuredHeight(), Bitmap.Config.ARGB_8888);
+		Canvas canvas = new Canvas(screenshot);
+		screenShotView.draw(canvas);
+		screenShotView.setDrawingCacheEnabled(false);
+
+		screenshotDark = screenshot.copy(Bitmap.Config.ARGB_8888, true);
+		canvas = new Canvas(screenshotDark);
+		Paint paint = new Paint(Color.WHITE);
+		ColorFilter filter = new LightingColorFilter(0xFF5F5F5F, 0x00000000);
```
cheidenreich: those colors could also be either final static values at the top of the class, or values in the colors.xml

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
FlorianWeissensteiner: Removed the redundant initialization (its necessary for width and height).

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
FlorianWeissensteiner: yes. See line '157' I pass it as an argument because the first page of the intro does not have a hightlight area.

catroid/src/main/java/org/catrobat/catroid/utils/FormulaEditorIntroUtil.java
FlorianWeissensteiner: Made final static values for the colours.

FlorianWeissensteiner:
@cheidenreich thanks for reviewing!

PR #2300, opened 2017-05-11
CAT-FIX Fix checkstyle and pmd warnings

robertpainsi:


catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
```diff
@@ -630,6 +630,7 @@ public boolean saveFormulaIfPossible() {
 				if (Brick.BrickField.isExpectingStringValue(currentBrickField)) {
 					return saveValidFormula(new FormulaElement(FormulaElement.ElementType.STRING, "", null));
 				}
+				// fallthrough
```
robertpainsi: @rschluesselbauer, @DoenerMaschine, @robotic-heidenreich
Please review if this fallthrough is on purpose (https://github.com/Catrobat/Catroid/pull/2269)

rschluesselbauer: Yes, this fallthrough is on purpose. 
If the BrickField we are editin the formula for is not expecting a string-value we want to return without saving when we get passed an empty string.

PR #2305, opened 2017-05-16
CAT-2408 Bidirectional Localization Testing

AyyalAwwad:
The Bidirectional localization testing checks the following screens:
Main Menu 
Program Menu 
Bricks Categories
For bi-directional testing, we need to use the following:
LAYOUT_DIRECTION_RTL`
getLayoutDirection
The warnings appear because of using the LAYOUT_DIRECTION_RTL and getLayoutDirection.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/BricksCategoriesTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/BricksCategoriesTest.java
@@ -1,0 +1,176 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.uiespresso.localization;
+
+import android.support.test.espresso.Espresso;
+import android.support.test.espresso.ViewInteraction;
+import android.support.test.espresso.action.ViewActions;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.ProjectManager;
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Project;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.ui.ProgramMenuActivity;
+import org.catrobat.catroid.uiespresso.util.BaseActivityInstrumentationRule;
+import org.catrobat.catroid.uiespresso.util.UiTestUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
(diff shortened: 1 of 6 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/BricksCategoriesTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+	@Rule
+	public BaseActivityInstrumentationRule<ProgramMenuActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(ProgramMenuActivity.class, true, false);
+
+	@Before
+	public void setUp() throws Exception {
+		UiTestUtils.createProject("BricksCategoriesTest");
+		baseActivityTestRule.launchActivity(null);
+		Espresso.onView(withId(R.id.program_menu_button_scripts)).perform(ViewActions.click());
```
84n4n4: if you need the scriptActivity, then start it with the Rule instead of navigating there

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/BricksCategoriesTest.java
```diff
@@ -0,0 +1,176 @@
...
+		Espresso.onView(withId(R.id.button_add)).perform(ViewActions.click());
+		brickCategoryEvent = Espresso.onView(withText(R.string.category_event));
```
84n4n4: - i think its bad practice to have UI operations going on during setUp. if a view isnt found its actually setUp failing and not the test, so all in all, i think its bad practice...

- if categories are added at a later time in the project, to include them, the setUp and every single test have to be touched.

- use onData for list views like the category brick view, since this will most likely break on smaller display devices.

so i would propose:
- define a new function that returns List<ViewInteractions> getBrickCategoryViewInteractions()
- that uses onData to collect all the viewInteractions your interested in
- call this at the beginning of each test, instead of in setUp

84n4n4: and of course implement a function that takes a viewAssertion and the List\<ViewInteractions\> as parameters and performs it on all of the items in the list.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/MainMenuActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+
+import static org.catrobat.catroid.uiespresso.localization.assertions.LayoutDirectionAssertions.isLayoutDirectionRTL;
+import static org.catrobat.catroid.uiespresso.localization.assertions.TextDirectionAssertions.isTextDirectionRTL;
+import static org.catrobat.catroid.uiespresso.localization.assertions.VisibilityAssertions.isVisible;
+import static org.hamcrest.core.IsNull.notNullValue;
+
+@RunWith(AndroidJUnit4.class)
+
+public class MainMenuActivityTest extends ActivityInstrumentationTestCase2<MainMenuActivity> {
```
84n4n4: ActivityInstrumentationTestCase2 is deprecated and not to be used anymore. all espresso tests have to use ActivityTestRule.
just like you did in the test above.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/MainMenuActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+	public MainMenuActivityTest() {
+		super(MainMenuActivity.class);
+	}
+
+	@Before
+	public void setUp() throws Exception {
+		super.setUp();
+		injectInstrumentation(InstrumentationRegistry.getInstrumentation());
+		mainMenuActivity = getActivity();
+		mainMenuContinue = Espresso.onView(withId(R.id.main_menu_button_continue));
```
84n4n4: here also same applies as i commented on the test above. do not do extensive ui interactions in set up

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/localization/BricksCategoriesTest.java
AyyalAwwad: Thank you, If it is possible I need a real example.

bernadettespieler:
R2L feature already covered in merged PR

PR #2311, opened 2017-05-19
CAT-2270 Correcting the negative zero(-0) and positive zero(0) evaluation

nishantt95:
CAT-2270 Evaluating -0 = 0 to True 

Earlier -0 and 0 evaluates to False (-0 = 0 is False) and now after the correction -0 and 0 evaluates to True (-0 = 0 is True).

Note: This pull request also includes test file.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/formulaeditor/ParserTestOperators.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/formulaeditor/ParserTestOperators.java
@@ -27,6 +27,7 @@ import android.test.AndroidTestCase;
 import org.catrobat.catroid.content.SingleSprite;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.formulaeditor.FormulaElement;
+import org.catrobat.catroid.formulaeditor.FormulaElement.ElementType;
 import org.catrobat.catroid.formulaeditor.Functions;
 import org.catrobat.catroid.formulaeditor.InternFormulaParser;
 import org.catrobat.catroid.formulaeditor.InternToken;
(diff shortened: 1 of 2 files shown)


robertpainsi:
LGTM! :+1:
You just need to improve your commit message header. :wink:
(https://help.github.com/articles/changing-a-commit-message/)

`CAT - 2270 Issue : earlier (-0 = 0) computes to False (CAT - 2270).`
* No spaces in the ticket identifier (before and after the minus sign `-`).
* Do not end the subject line with a period.
* You don't need to reference the ticket twice. The ticket identifier at the beginning is sufficient.
* Normally commit headers start with a verb (add, remove, fix, update, refactor, etc.). However, to me just using `Negative zero equals positive zero (-0 = 0)` is better.

→ `CAT-2270 Negative zero equals positive zero (-0 = 0)`

You can also use `CAT-2270 -0 = 0 evaluates to True`, `…is True`, or whatever you come up with.

Often it's a good idea not to use the Jira ticket title as commit message header. Like in this case. So it's great that you didn't use it. Your commit message body is also fine.

rschluesselbauer:
+1
Looking good :-)

robertpainsi:
commit message header !== pull request header :wink:

The commit message header is the first line of the commit message and separated by an empty newline from the body. `git log --oneline` will print the log output but only show the commit message headers.
There is a list on how to format your commit message [here](https://github.com/Catrobat/Catroid/wiki/Commit-Message-Guidelines). Please adapt your commit message accordingly.

Both, your old and new pull request header (title) were fine. So I'm not that picky :laughing: However, your current commit message needs improvements.

robertpainsi:
ok to test

PR #2313, opened 2017-05-20
Add template tag & code.xml entry + template remix information

sjaindl:


catroid/src/main/java/org/catrobat/catroid/transfers/GetTagsTask.java
```diff
@@ -87,6 +91,18 @@ protected void onPostExecute(String response) {
...
 		}
 
+		if (BuildConfig.CREATE_AT_SCHOOL) {
+			addTemplateTag(tags);
+		}
+
 		return Collections.unmodifiableList(tags);
 	}
+
+	private void addTemplateTag(List<String> tags) {
```
cheidenreich: since this is used only once: instead of making a separate method for this I'd put the code directly into line 95.

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -33,7 +33,7 @@
 	// Reflection in testcases needed
 	// http://stackoverflow.com/questions/1615163/modifying-final-fields-in-java?answertab=votes#tab-top
 
-	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = Float.valueOf(0.993f);
+	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = Float.valueOf(0.994f);
```
cheidenreich: doesn't this need a conversion action from older language version programs in `ProjectManager.loadProject()`?

sjaindl: It doesn't need a conversion, there's simply no template remix information for older programs.
There was already a check for languageVersion 0.993 to 0.994 --> I've added a brief comment in ProjectManager.loadProject()

sjaindl:
EDIT: It shouldn't be merged yet as discussed today with @bernadettespieler --> no languageVersion changes should be made for the upcoming create@school release

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
cheidenreich: Side note: even if no changes are required, it should still raise the old project's version number, otherwise there would be problems for future language version conversions as every conversion would be stuck on 0.993

sjaindl: Well, I think this is a point of discussion since it can also be raised together with future language version changes, when it really is necessary.
Whatever, I've added that line now ;)

84n4n4:
just fyi, language version has been increased for release 0.9.32 due to arduino pmw values changes....

PR #2314, opened 2017-05-20
CAT-2347 Rewrite BroadcastSequence(Wait)Map as a member of Sprite

rschluesselbauer:
Make BroadcastSequenceMap and BroadcastWaitSequenceMap a member of each sprite, adjust Broadcast unit tests, rewrite Broadcast regression tests in espresso

- BroadcastSequenceMap is changed to be a member of Sprite instead
  of one static map (this way we know which Sprite should react on
  which Broadcast)
- BroadcastWaitSequenceMap is changed to be a member of Sprite instead
  of one static map (for consistency with BroadcastSequenceMap), it now maps from Sprite to BroadcastSequenceMap (this is done so we know which sprites we want to do which actions)
- Updated calls to new Broadcast(Wait)SequenceMap in BroadcastHandler and StageDialog
- Removed for-loop from BroadcastHandler method handleActionFromBroadcastWait(..) that did not loop
- Regression Tests for Broadcasts are rewritten in espresso
- Added ProjectSetupUtils for putting together test projects more easily
- Added methods to UiTestUtils to efficiently wait for a UserVariable to have a specific value


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
```diff
@@ -203,4 +206,28 @@ public static boolean comparePixelRgbaArrays(byte[] firstArray, byte[] secondArr
 		}
 		return true;
 	}
+
+	public static void waitForVariableValueEquals(UserVariable userVariable, double expectedValue, int timeout) {
```
84n4n4: istead of the hard to read and comprehend version with the return; and assert in bottom, wouldnt it be nicer to have this function return a boolean, and rename it to something like "userVariableEqualsWithinTimeout" to then do something like:
assertTrue(userVariableEqualsValueWithinTimeout(uv, ev, t))

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
```diff
@@ -203,4 +206,28 @@ public static boolean comparePixelRgbaArrays(byte[] firstArray, byte[] secondArr
...
+			if (Math.abs((double) userVariable.getValue() - expectedValue) < epsilon) {
+				return;
+			}
+			onView(isRoot()).perform(CustomActions.wait(step));
+		}
+		Assert.assertEquals(expectedValue, (double) userVariable.getValue(), epsilon);
+	}
+
+	public static void waitForVariableValueGreaterThan(UserVariable userVariable, double expectedValue, int timeout) {
+		double epsilon = 0.001;
```
84n4n4: make epsilon a constant member of the utils class

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
```diff
@@ -203,4 +206,28 @@ public static boolean comparePixelRgbaArrays(byte[] firstArray, byte[] secondArr
...
+		int step = 10;
+		for (; timeout > 0; timeout -= step) {
+			if ((double) userVariable.getValue() > (expectedValue + epsilon)) {
```
84n4n4: shouldnt that be (expectedValue - epsilon)?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
```diff
@@ -0,0 +1,59 @@
...
+		return broadcastScript;
+	}
+
+	public static Script addSpriteWithScript(Project project, String spriteName, Script script) {
+		Sprite sprite = new Sprite(spriteName);
+		project.getDefaultScene().addSprite(sprite);
+		return addScriptToSprite(script, sprite);
+	}
+
+	public static Script addScriptToSprite(Script script, Sprite sprite) {
```
84n4n4: why tho?
i dont see the point of this function, caller has to know script and sprite and can just call sprite.addScript himself.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
```diff
@@ -0,0 +1,59 @@
...
+		SetVariableBrick setVariableBrickAfterBroadcast = new SetVariableBrick(new Formula(value), userVariable);
+		broadcastScript.addBrick(setVariableBrickAfterBroadcast);
+		sprite.addScript(broadcastScript);
+		return broadcastScript;
+	}
+
+	public static Script addSpriteWithScript(Project project, String spriteName, Script script) {
+		Sprite sprite = new Sprite(spriteName);
+		project.getDefaultScene().addSprite(sprite);
+		return addScriptToSprite(script, sprite);
```
84n4n4: see comment at function definition below.
this does not make it easier to see whats going on, especially with statements like this.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
```diff
@@ -0,0 +1,59 @@
+/*
```
84n4n4: since all of the functions in this class are very specifically designed for the two new tests, move the class over to /stage/ and rename to StageTestUtils. (eg. theres also a BrickTestUtils that also contains project setup stuff but again has different requirements than those for the stage)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
```diff
@@ -0,0 +1,59 @@
...
+	public static Script addBroadcastScriptSettingVariable(Sprite sprite, String message, UserVariable userVariable,
+			double value) {
+		Script broadcastScript = new BroadcastScript(message);
+		SetVariableBrick setVariableBrickAfterBroadcast = new SetVariableBrick(new Formula(value), userVariable);
+		broadcastScript.addBrick(setVariableBrickAfterBroadcast);
+		sprite.addScript(broadcastScript);
+		return broadcastScript;
+	}
+
+	public static Script addSpriteWithScript(Project project, String spriteName, Script script) {
```
84n4n4: rework your function names, that addSpriteWithScript returns a Script is not at all clear from its name.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+		baseActivityTestRule.launchActivity(null);
+		Thread.sleep(1500);
+
+		for (int i = 0; i < variables.size(); i++) {
+			UiTestUtils.waitForVariableValueEquals(variables.get(i), values[i], 2000);
+		}
+	}
+
+	private List<UserVariable> createProject(int[] values) {
```
84n4n4: please rename this function. for the reader its not at all clear that "createProject" returns a List  of UserVariables.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
```diff
@@ -0,0 +1,94 @@
...
+		Project project = UiTestUtils.createEmptyProject("test");
+		DataContainer dataContainer = project.getDefaultScene().getDataContainer();
+
+		ArrayList<UserVariable> userVariables = new ArrayList<>();
+		for (int i = 0; i < values.length; i++) {
+			Sprite sprite = new SingleSprite("sprite" + i);
+			project.getDefaultScene().addSprite(sprite);
+			UserVariable userVariable = dataContainer.addProjectUserVariable("var" + i);
+
+			Script startScript1 = ProjectSetupUtils.addScriptToSprite(new StartScript(), sprite);
```
84n4n4: please rename startScript**1**

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
```diff
@@ -0,0 +1,94 @@
...
+		baseActivityTestRule.launchActivity(null);
+		Thread.sleep(1500);
+
+		for (int i = 0; i < variables.size(); i++) {
+			UiTestUtils.waitForVariableValueEquals(variables.get(i), values[i], 2000);
+		}
+	}
+
+	private List<UserVariable> createProject(int[] values) {
+		Project project = UiTestUtils.createEmptyProject("test");
```
84n4n4: it would probably be nicer to have your own createEmptyProject function here, where you already get the data container so you dont have to dig for it.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Test
+	public void testSendMultipleBroadcasts() throws InterruptedException {
+
+		int[] values = new int[] {30, 60, -30, -60};
+		List<UserVariable> variables = createProject(values);
+
+		baseActivityTestRule.launchActivity(null);
+		Thread.sleep(1500);
```
84n4n4: is this necessary?
if this is still required despite the waitForVariableValueEquals function below, maybe the timeout for it is too short, or you need to still need to add a idlingResource to the stage if this is due to boot up time of it. (i doubt that, it works on the simpleStageTest without anything.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import org.catrobat.catroid.formulaeditor.Formula;
+import org.catrobat.catroid.formulaeditor.UserVariable;
+
+public final class ProjectSetupUtils {
+	// Suppress default constructor for noninstantiability
+	private ProjectSetupUtils() {
+		throw new AssertionError();
+	}
+
+	public static Script addBroadcastScriptSettingVariable(Sprite sprite, String message, UserVariable userVariable,
```
84n4n4: rename this function

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
```diff
@@ -203,4 +206,28 @@ public static boolean comparePixelRgbaArrays(byte[] firstArray, byte[] secondArr
 		}
 		return true;
 	}
+
```
84n4n4: please move both functions over to a .../stage/StageTestUtils.java class.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/BroadcastReceiverRegressionTest.java
```diff
@@ -0,0 +1,232 @@
...
+	private Project project;
+
+	@Rule
+	public BaseActivityInstrumentationRule<StageActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(StageActivity.class, true, false);
+
+	@Before
+	public void setUp() throws Exception {
+		BroadcastHandler.clearActionMaps();
+		project = UiTestUtils.createEmptyProject("test");
```
84n4n4: make your own function, so you dont have to dig for the userVariable, and since you add a sprite and a start script in all of the below tests include it there aswell.

catroid/src/main/java/org/catrobat/catroid/common/BroadcastSequenceMap.java
```diff
@@ -28,43 +28,43 @@
...
+		return broadcastSequenceMap.containsKey(sceneName) && broadcastSequenceMap.get(sceneName).containsKey(key);
 	}
 
-	public static boolean containsKey(String key, String sceneName) {
-		if (!BroadcastSequenceMap.broadcastSequenceMap.containsKey(sceneName)) {
-			return false;
+	public ArrayList<SequenceAction> get(String key, String sceneName) {
+		Map<String, ArrayList<SequenceAction>> map = broadcastSequenceMap.get(sceneName);
+		if (map == null) {
+			map = new HashMap<>();
```
84n4n4: is this change required?
before BroadcastSequenceMap behave similar to a normal java Map, returning null if key is not found... now creation of new items in map is handeled in "get".

catroid/src/main/java/org/catrobat/catroid/common/BroadcastSequenceMap.java
```diff
@@ -28,43 +28,43 @@
...
-		return BroadcastSequenceMap.broadcastSequenceMap.get(sceneName).containsKey(key);
+		return map.get(key);
 	}
 
-	public static ArrayList<SequenceAction> get(String key, String sceneName) {
-		if (!BroadcastSequenceMap.broadcastSequenceMap.containsKey(sceneName)) {
-			return null;
-		}
-		return BroadcastSequenceMap.broadcastSequenceMap.get(sceneName).get(key);
+	public boolean add(String sceneName, String key, SequenceAction value) {
```
84n4n4: why rename to add? up until now BroadcastSequenceMap although not extending Map kind of behaved like what one would expect from something called *Map.

catroid/src/main/java/org/catrobat/catroid/common/BroadcastSequenceMap.java
```diff
@@ -28,43 +28,43 @@
...
-	public static void clear() {
-		BroadcastSequenceMap.broadcastSequenceMap.clear();
+	public void clear() {
+		broadcastSequenceMap.clear();
 	}
 
-	public static void clear(String sceneName) {
-		if (!BroadcastSequenceMap.broadcastSequenceMap.containsKey(sceneName)) {
+	public void clear(String sceneName) {
+		if (!broadcastSequenceMap.containsKey(sceneName)) {
```
84n4n4: please change to:
if (broadcastSequenceMap.containsKey(sceneName)) {
    broadcastSequenceMap.get(sceneName).clear();
}
its a lot easier to read...

catroid/src/main/java/org/catrobat/catroid/common/BroadcastWaitSequenceMap.java
```diff
@@ -25,68 +25,66 @@
...
-public final class BroadcastWaitSequenceMap {
-	private static Map<String, HashMap<String, ArrayList<SequenceAction>>> broadcastWaitSequenceMap = new HashMap<>();
-	private static BroadcastEvent currentBroadcastEvent = null;
+public class BroadcastWaitSequenceMap {
+	private BroadcastEvent currentBroadcastEvent = null;
+	private Map<Sprite, BroadcastSequenceMap> waitSequenceMap;
 
-	private BroadcastWaitSequenceMap() {
-		throw new AssertionError();
+	public BroadcastWaitSequenceMap() {
```
84n4n4: why initialization different than in BroadcastSequenceMap?

catroid/src/main/java/org/catrobat/catroid/common/BroadcastWaitSequenceMap.java
```diff
@@ -25,68 +25,66 @@
...
+		waitSequenceMap = new HashMap<>();
 	}
 
-	public static boolean containsKey(String key, String sceneName) {
-		if (!broadcastWaitSequenceMap.containsKey(sceneName)) {
-			return false;
-		}
-		return BroadcastWaitSequenceMap.broadcastWaitSequenceMap.get(sceneName).containsKey(key);
+	public void add(String sceneName, String key, Sprite sprite, ArrayList<SequenceAction> values) {
```
84n4n4: see comments in BroadcastSequenceMap

catroid/src/main/java/org/catrobat/catroid/content/BroadcastHandler.java
```diff
@@ -51,48 +51,52 @@ private BroadcastHandler() {
...
 			Script scriptOfAction = actionScriptMap.get(action);
 
 			if (!handleAction(action, scriptOfAction)) {
 				addOrRestartAction(look, action);
 			}
 		}
 
-		if (BroadcastWaitSequenceMap.containsKey(broadcastMessage, sceneName)) {
-			for (SequenceAction action : BroadcastWaitSequenceMap.get(broadcastMessage, sceneName)) {
+		if (event.getSenderSprite().getBroadcastWaitSequenceMap().containsKey(broadcastMessage, sceneName, look.sprite)) {
```
84n4n4: this long dig: event.getSenderSprite().getBroadcastWaitSequenceMap(). show up a lot.
maybe make a local variable for it and rename the broadcastSequenceMap and this new one so that its clear whos who.

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -106,7 +106,7 @@ public boolean touchDown(InputEvent event, float x, float y, int pointer, int bu
 		this.addListener(new BroadcastListener() {
 			@Override
 			public void handleBroadcastEvent(BroadcastEvent event, String broadcastMessage) {
-				doHandleBroadcastEvent(broadcastMessage);
+				doHandleBroadcastEvent(event, broadcastMessage);
```
84n4n4: event is handed through a few functions just to be used for getting the map (event.getSenderSprite().getBroadcastWaitSequenceMap())
maybe pass the map through?

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -379,8 +376,7 @@ public Sprite cloneForCloneBrick() {
 	}
 
 	public Sprite cloneForScene() {
-		Sprite clone = clone();
-		return clone;
+		return clone();
```
84n4n4: find usages of this functions and change them to call clone() directly and then remove this function

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
rschluesselbauer: I don't think so.
With greater than we want the two values to be at least epsilon (plus a value >0) apart.
Saying value > expectedValue-epsilon would have the effect that value could be smaller than expectedValue.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/ProjectSetupUtils.java
rschluesselbauer: It saves a line when creating scripts for a sprite like this:
`Script startScript = ProjectSetupUtils.addScriptToSprite(new StartScript(), sprite);`

I'll still remove it, as you are right and the benefit is minimal.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
rschluesselbauer: Sorry, that's something I overlooked while refactoring. My bad.

catroid/src/main/java/org/catrobat/catroid/common/BroadcastSequenceMap.java
rschluesselbauer: Concerning the map changes - I think I got a bit carried away there.
I'll change it back to what you would expect of a map.
Thanks for the elaborate feedback!

rschluesselbauer:
Thanks for the elaborate feedback!
I have incoorporated almost all of it, except for the one with epsilon (shouldnt that be (expectedValue - epsilon)?) in the method userVariableGreaterThanWithinTimeout(..) in StageTestUtils.
I also fixed the PhysicsCollisionBaseTest, as it didn't add sprite2 to the project before (resulting in a failing build)
I also added a nullcheck for senderSprite, as it is null in case of a CollisionBroadcast.

84n4n4:
hi!

regarding the epsilon thing. that depends, if you approach it mathematically:
when comparing two floats, two floats can be considered equal if they are closer than epsilon together. so when a is growing, and b is your requested limit, above condition would be reached when a == b - epsilon
on the other hand, if you really want to make sure that a > b then stating a > b is sufficient. (because in this code snippet, a could be already bigger than b while still within epsilon...)

rschluesselbauer:
Ok, so when a is growing and b is the limit
a > (b - epsilon) is the same thing as a **greaterOrEqual** to b
a > (b + epsilon) is the same thing as a **greaterAndNotEqual** to b

I am going with the second one, because I don't want the condition to be true when a and b are equal (within the epsilon).

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/MultipleBroadcastsTest.java
```diff
@@ -0,0 +1,100 @@
...
+	public void setUp() throws Exception {
+		BroadcastHandler.clearActionMaps();
+		createProject();
+	}
+
+	@Test
+	public void testSendMultipleBroadcasts() throws InterruptedException {
+
+		int[] values = new int[] {30, 60, -30, -60};
+		List<UserVariable> variables = setUpProjectAndReturnUserVariables(values);
```
84n4n4: now theres a createProject and a setUpProject.. function. since theres only one test here, please move everything that is actually setting up for this test into the setUp function. (and merge createProject and setUpProject... into one)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/StageTestUtils.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	public static Script addScriptToAnonymousSprite(Project project, String spriteName, Script script) {
+		Sprite sprite = new Sprite(spriteName);
+		project.getDefaultScene().addSprite(sprite);
+		sprite.addScript(script);
+		return script;
+	}
+
+	public static boolean userVariableEqualsWithinTimeout(UserVariable userVariable, double expectedValue, int timeout) {
```
84n4n4: can you please rename timeout to timeoutMs or something like this to make it clear that its in ms.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/StageTestUtils.java
```diff
@@ -0,0 +1,84 @@
...
+		int step = 10;
```
84n4n4: please rename step to checkIntervalMs (or something suitable that communicates that this is a time step and its in ms)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/StageTestUtils.java
```diff
@@ -0,0 +1,84 @@
...
+public final class StageTestUtils {
+
+	private static final double EPSILON = 0.001;
+
+	// Suppress default constructor for noninstantiability
+	private StageTestUtils() {
+		throw new AssertionError();
+	}
+
+	public static Script addBroadcastScriptSettingUserVariableToValue(Sprite sprite, String message, UserVariable userVariable,
```
84n4n4: can you please refactor this.
if a function is called addAtoB i expect it to do just that without returning another object. so either refactor into a different function structure, or rename to make it clear what this does (if its getting hard to put a name on your functions because they do so much at once, splitting them up is maybe a good idea)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/StageTestUtils.java
```diff
@@ -0,0 +1,84 @@
...
+			double value) {
+		Script broadcastScript = new BroadcastScript(message);
+		SetVariableBrick setVariableBrickAfterBroadcast = new SetVariableBrick(new Formula(value), userVariable);
+		broadcastScript.addBrick(setVariableBrickAfterBroadcast);
+		sprite.addScript(broadcastScript);
+		return broadcastScript;
+	}
+
+	public static Script addScriptToAnonymousSprite(Project project, String spriteName, Script script) {
```
84n4n4: same as last comment

rschluesselbauer:
All suggested changes have been made. Anything else I should improve on?

84n4n4:
test this please

84n4n4:
@cheidenreich  can you please have a look.
for me this looks fine.

cheidenreich:
thanks @84n4n4 for the review.

I had another look on the code and tested the functionality manually on my device, even with heavy copy/paste/delete nothing breaks, looks good

cheidenreich:
test this please

cheidenreich:
please run espresso tests

cheidenreich:
test this please

PR #2323, opened 2017-05-29
CAT-2348 ComeToFront and MoveNSteps simplification

nishantt95:
This pull request simplify ComeToFront and MoveNSteps

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/ComeToFrontAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/ComeToFrontAction.java
@@ -38,14 +38,6 @@ public class ComeToFrontAction extends TemporalAction {
 	protected void update(float delta) {
 
 		List<Sprite> spriteList = ProjectManager.getInstance().getCurrentProject().getSpriteListWithClones();
-		int actualSpriteZIndex = sprite.look.getZIndex();
-
-		for (int i = 0; i < spriteList.size(); i++) {
-			if (spriteList.get(i).look.getZIndex() > actualSpriteZIndex) {
-				spriteList.get(i).look.setZIndex(spriteList.get(i).look.getZIndex()
-						+ Constants.Z_INDEX_NUMBER_VIRTUAL_LAYERS - 1);
-			}
-		}
 		sprite.look.setZIndex(spriteList.size() + Constants.Z_INDEX_NUMBER_VIRTUAL_LAYERS - 1);
 	}
 
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
@@ -24,15 +24,13 @@ package org.catrobat.catroid.content.actions;
 
 import android.util.Log;
 
+import com.badlogic.gdx.scenes.scene2d.Action;
 import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
 
-import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.common.Constants;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.formulaeditor.Formula;
 import org.catrobat.catroid.formulaeditor.InterpretationException;
-
-import java.util.List;
 
 public class GoNStepsBackAction extends TemporalAction {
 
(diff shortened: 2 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/ComeToFrontAction.java
```diff
@@ -38,15 +37,7 @@
...
-		int actualSpriteZIndex = sprite.look.getZIndex();
-
-		for (int i = 0; i < spriteList.size(); i++) {
-			if (spriteList.get(i).look.getZIndex() > actualSpriteZIndex) {
-				spriteList.get(i).look.setZIndex(spriteList.get(i).look.getZIndex()
-						+ Constants.Z_INDEX_NUMBER_VIRTUAL_LAYERS - 1);
-			}
-		}
-		sprite.look.setZIndex(spriteList.size() + Constants.Z_INDEX_NUMBER_VIRTUAL_LAYERS - 1);
+		sprite.look.setZIndex(spriteList.size());
```
robertpainsi: I guess we can't remove `…+ Constants.Z_INDEX_NUMBER_VIRTUAL_LAYERS - 1` here. Even though `Z_INDEX_NUMBER_VIRTUAL_LAYERS` is currently `1`, it may change in the future.

This case is different from `spriteList.get(i).look.setZIndex(spriteList.get(i).look.getZIndex() +…` where the new z-index is calculated relatively. When you use `getZIndex`, you already have an index which considers all virtual layers.

catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
```diff
@@ -53,23 +53,16 @@ protected void update(float delta) {
...
-		for (int i = 0; i < spriteList.size(); i++) {
-			if (spriteList.get(i).look.getZIndex() > actualSpriteZIndex) {
-				spriteList.get(i).look.setZIndex(spriteList.get(i).look.getZIndex() - 1);
-			}
-		}
-		sprite.look.setZIndex(spriteList.size() - 1);
+	private void toFront(float delta) {
+		ComeToFrontAction mComeToFrontAction = new ComeToFrontAction();
+		mComeToFrontAction.setSprite(sprite);
+		mComeToFrontAction.update(delta);
```
robertpainsi: Use `sprite.getActionFactory().createComeToFrontAction(sprite)` instead.

catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
```diff
@@ -79,14 +72,13 @@ private void goNStepsBack(int steps) {
...
-				}
+
+			int currentSpriteZIndex = spriteList.get(i).look.getZIndex();
+
+			if (steps > 0 && currentSpriteZIndex >= newSpriteZIndex && currentSpriteZIndex < zPosition) {
+				spriteList.get(i).look.setZIndex(currentSpriteZIndex + 1);
+			} else if (currentSpriteZIndex <= newSpriteZIndex && currentSpriteZIndex > zPosition) {
+				spriteList.get(i).look.setZIndex(currentSpriteZIndex - 1);
 			}
 		}
```
robertpainsi: Does this loop really do something useful? When the z-index of a look is changed, all others are automatically updated correctly as well. So it should be sufficient to set the z-index of the current sprite only. Please recheck.

catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
```diff
@@ -53,42 +53,20 @@ protected void update(float delta) {
...
-
-		for (int i = 0; i < spriteList.size(); i++) {
-			if (spriteList.get(i).look.getZIndex() > actualSpriteZIndex) {
-				spriteList.get(i).look.setZIndex(spriteList.get(i).look.getZIndex() - 1);
-			}
-		}
-		sprite.look.setZIndex(spriteList.size() - 1);
+	private void toFront(float delta) {
+		ComeToFrontAction mComeToFrontAction = (ComeToFrontAction) sprite.getActionFactory().createComeToFrontAction(sprite);
+		mComeToFrontAction.update(delta);
```
robertpainsi: Please do not prefix variables with a 'm'. Neither member variables nor local variables.
I would call the method `act` instead of `update` because the Stage does also call `act` and you can remove the casting. The execution sequence of `act` is `begin`, `update`, `end`.

robertpainsi:
test this please

robertpainsi:
I just have to look something up… tomorrow :smile:

robertpainsi:
> I just have to look something up… tomorrow

Jupp, found a problem [CAT-2430](https://jira.catrob.at/browse/CAT-2430). However, it's not caused by this Pull Request.

PR #2327, opened 2017-05-30
CAT-XP6 [KL,LS] PhiroIfBrickTest

kirshanluhana:


84n4n4:
ok to test

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
```diff
@@ -105,4 +106,16 @@ public static void createUserListFromDataFragment(String userListName, boolean f
...
+
+	public static void deleteBrickAtPosition(int position, int intBrickResouce) {
+		onScriptList().atPosition(position).onChildView(withText(intBrickResouce)).perform(click());
+		onView(withText(R.string.brick_context_dialog_delete_brick))
+				.perform(click());
+		onView(withText(R.string.yes))
+				.perform(click());
+	}
+
+	public static void checkBrickNotExists(int intResourceId) {
```
84n4n4: remove this.

no functions with only 1 line of code.
this is obfuscating the code, it makes it only harder to read.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
```diff
@@ -105,4 +106,16 @@ public static void createUserListFromDataFragment(String userListName, boolean f
 		onView(withText(userListName))
 				.check(matches(isDisplayed()));
 	}
+
+	public static void deleteBrickAtPosition(int position, int intBrickResouce) {
```
84n4n4: thers a few things wrong here:
1. no functions in brickTestUtils that are only used in one test. put it in the testclass
2. name of function is not clear enough, if it would take (int position) as only parameter it would be fine, but this way nobody can just go there and use it without reading the whole function first.
3. brickPosition must be sufficient to delete a brick
4. intBrickResource implies to me that this is going to be the layout id of the brick, but when reading the function i see that this is a stringResource id.
-- all of the above will make this unusable for others.

PR #2335, opened 2017-06-02
CAT-1678 & CAT-1676 Implementation of two sensor for EV3

MightyMendor:
CAT-1678 Nxt Temperature Sensor Support for the EV3
CAT-1676 Hi Technic Color Sensor Support for the EV3 - Basic Mode


catroid/src/main/java/org/catrobat/catroid/devices/mindstorms/ev3/sensors/TemperatureSensor.java
```diff
@@ -0,0 +1,49 @@
...
+	private static final int DEFAULT_VALUE = 0;
+	private static final int SENSOR_VALUE_READ_LENGTH = 4;
+
+	public TemperatureSensor(int port, MindstormsConnection connection, EV3SensorMode mode) {
+		super(port, EV3SensorType.NXT_TEMPERATURE, mode, connection);
+		lastValidValue = DEFAULT_VALUE;
+	}
+
+	@Override
+	public float getValue() {
```
cheidenreich: I assume this is the only reason why the return value of `getValue()` changed to float.

I'm thinking if there is a more elegant solution.

We could make the base class generic `EV3Sensor<T>` and define the type either `int` or `float` in each derrived class.
@robertpainsi @84n4n4 what is your feeling about that?

MightyMendor: Just as info why I didn't hesitate too much about that change: in formulaeditor/SensorHandler.java where the values are used, all of them are cast to double.

cheidenreich: i see, valid point

84n4n4:
can you please rebase this?

84n4n4:
can you please rebase this?

PR #2336, opened 2017-06-06
CAT-2145 change range for PWM values from [0-100] to [0-255]

jobermue:
This removes an inconsistency in the code and creates compatibility
with Arduino's value range.


robertpainsi:
Lets see if there is a test which fails :wink:

test this please

robertpainsi:
> Lets see if there is a test which fails

YES!
But this isn't the test I'm looking for :worried: I'm missing a test for the range limits. So if you pass in a number <0 or >255.

[ArduinoImplTest.testSetAnalogArduinoPinRandomly](https://github.com/Catrobat/Catroid/blob/1f8609002b582e4140d0c98f767ecc1bd249254f/catroid/src/androidTest/java/org/catrobat/catroid/test/devices/arduino/ArduinoImplTest.java#L194) also fails. I don't see any value in this test anyway. It's already handled by [ArduinoImplTest.testSetAnalogArduinoPin](https://github.com/Catrobat/Catroid/blob/1f8609002b582e4140d0c98f767ecc1bd249254f/catroid/src/androidTest/java/org/catrobat/catroid/test/devices/arduino/ArduinoImplTest.java#L165).

robertpainsi:
@jobermue, please fix and adapt (test values between [0-255]) `ArduinoImplTest.testSetAnalogArduinoPin` and remove `ArduinoImplTest.testSetAnalogArduinoPinRandomly`.

jobermue:
Hi,
I updated the code to be in line with the results of our discussion on JIRA (removed all range checks). I furthermore cleaned up the relevant test cases (`ArduinoImplTest.testSetAnalogArduinoPin` passes and `ArduinoImplTest.testSetAnalaogArduinoPinRandomly` was removed). The test checks values in the range of [0-16383] and the behaviour when using out-of-range values.
Please take a look and let me know if it's ok.

robertpainsi:
test this please

robertpainsi:
please run espresso tests

robertpainsi:
@cheidenreich, @bernadettespieler, to me, this fix looks fine. However, I'm not really familiar with Firmata. If there is no one else looking at this, I'm fine with merging it.

cheidenreich:
please run espresso tests

cheidenreich:
test this please

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
```diff
@@ -256,7 +245,7 @@ public static int getIndexOfPinOnPort(int pin) {
 
 	private void sendAnalogFirmataMessage(int pin, int value) {
 		sendFirmataMessage(new SetPinModeMessage(pin, SetPinModeMessage.PIN_MODE.PWM.getMode()));
-		sendFirmataMessage(new AnalogMessage(pin, castValue(value)));
+		sendFirmataMessage(new AnalogMessage(pin, value));
```
cheidenreich: does this change the behaviour of existing programs?

robertpainsi: Yes! (`Behavior Change` label)

Previously values were clamped to [0-100] and then mapped to [0-255].
Now values are just passed to Firmata and Arduino.

So the language version needs to be increased for the next release, I guess.

robertpainsi:
test this please

catroid/src/main/java/org/catrobat/catroid/devices/arduino/ArduinoImpl.java
robertpainsi: Updates?
This is a really important change for the Arduino feature. The sooner it's merged, the better :wink:.

84n4n4:
test this please

84n4n4:
please run espresso tests

84n4n4:
@robertpainsi @thmq @cheidenreich 
since 2573 and 2558 have been just mergedn, wouldnt this be a good time to merge this aswell and up the langauge version?

PR #2337, opened 2017-06-06
CAT-2420 Replace Firebase with Crashlytics

nishantt95:
This pull request includes the implementation of crashlytics and
replaced it with firebase crash report.

Note: This request includes the test files


catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static junit.framework.Assert.assertTrue;
+
+public class CrashReporterTest {
+
+	@Test
+	public void testCrashlyticsUninitializedOnAnonymousReportDisabled() {
+		Context context = InstrumentationRegistry.getTargetContext();
+		SharedPreferences.Editor editor = PreferenceManager.getDefaultSharedPreferences(context).edit();
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
+		editor.clear();
```
robertpainsi: In some tests you're using `editor.clear()`, in some you don't. I guess it's better to use it in all tests. Assume that the tests are executed in a random order and should also start in a clean state. So using a [@Before method (setUp)](http://junit.sourceforge.net/javadoc/org/junit/Before.html) to clear the `SharedPreferences` seems like a good idea. Maybe create a static util method so others can benefit from it as well.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+	@Test
+	public void testCrashlyticsInitializedOnAnonymousReportEnabled() {
+		Context context = InstrumentationRegistry.getTargetContext();
+		SharedPreferences.Editor editor = PreferenceManager.getDefaultSharedPreferences(context).edit();
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true);
+		editor.clear();
+		editor.commit();
+		CrashReporter crashReporter = new CrashReporter(context);
+		assertTrue(crashReporter.initialize());
```
robertpainsi: Structuring test methods in
```
void testMethod() {
  // setup

  // execute

  // assert
}
```
will often improve the readability of tests. Mind the newlines between each block. Rarely no newline between the setup/execute block and assert block improves readability.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,139 @@
...
+			Fabric.with(context, new Crashlytics());
+			Log.d(TAG, "INITIALIZED!");
+			return true;
+		}
+		Log.d(TAG, "INITIALIZATION FAILED!");
+		return false;
+	}
+
+	@VisibleForTesting
+	public void setIsAnonymousCrashReportEnable(boolean value) {
```
robertpainsi: Please remove. It's not used for testing anymore. The whole method can be inlined.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,139 @@
...
+		isAnonymousReportEnabled = value;
+	}
+
+	@VisibleForTesting
+	public void setIsCrashReportEnabled(boolean value) {
+		isCrashReportEnabled = value;
+	}
+
+	@VisibleForTesting
+	public void setIsRecoveredFromCrash(boolean value) {
```
robertpainsi: Can be inlined as well

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		SharedPreferences.Editor editor = sharedPreferences.edit();
+
+		Exception exception = new RuntimeException("Error");
+		CrashReporter crashReporter = new CrashReporter(context);
+		crashReporter.setIsCrashReportEnabled(true);
+		crashReporter.storeUnhandledException(exception);
+		//editor.clear();
+		editor.commit();
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").length() > 0);
```
robertpainsi: Android Studio suggests to use `!…isEmpty()` instead. It's up to you :wink:

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	@Test
+	public void testSharedPreferencesClearedAfterLoggingException() {
+
+		Context context = InstrumentationRegistry.getTargetContext();
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		SharedPreferences.Editor editor = sharedPreferences.edit();
+
+		Exception exception = new RuntimeException("Error");
```
robertpainsi: You can inline the variable (in such cases). IMO there is no need for a local variable.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,139 @@
...
+
+import io.fabric.sdk.android.Fabric;
+
+public class CrashReporter {
+
+	Context context;
+	private SharedPreferences preferences;
+	public boolean isAnonymousReportEnabled;
+	public boolean isCrashReportEnabled;
+	public boolean isRecoveredFromCrash;
```
robertpainsi: Can be private.

In most cases, the Android Studio code inspection tool will highlight such cases for you.

However, there are cases where public (or if it's meant to be overwritten, protected) is fine. e.g.
```
public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
```

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+public class CrashReporterTest {
+
+	private Context context;
+	private SharedPreferences sharedPreferences;
+	private SharedPreferences.Editor editor;
+	private Exception exception;
+
+	@Before
+	public void setup() {
```
robertpainsi: Please renamte to `setUp`, uppercase u, to be consistent.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,142 @@
...
+import io.fabric.sdk.android.Fabric;
+
+public class CrashReporter {
+
+	Context context;
+	private SharedPreferences preferences;
+	private static boolean isAnonymousReportEnabled;
+	private static boolean isCrashReportEnabled;
+	private static boolean isRecoveredFromCrash;
+	public static boolean isCrashlyticsInitialized;
```
robertpainsi: You can change this member to private too :wink:

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,142 @@
...
+
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	private static final String TAG = CrashReporter.class.getSimpleName();
```
robertpainsi: To be consistent, the TAG should always be one of or the first variables in a class like the `serialVersionUID`.
I would also move `RECOVERED_FROM_CRASH` and  `EXCEPTION_FOR_REPORT` up, right before Context.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,142 @@
...
+
+	public CrashReporter(Context context) {
+		this.context = context;
+		preferences = PreferenceManager.getDefaultSharedPreferences(context);
+		isAnonymousReportEnabled = preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
+		isRecoveredFromCrash = preferences.getBoolean(RECOVERED_FROM_CRASH, false);
+		setIsCrashReportEnabled(BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED);
+		isCrashlyticsInitialized = false;
```
robertpainsi: Changing `isCrashlyticsInitialized` in the constructor may lead to errors.
e.g. if you create a new instance of the CrashReporter and call initializie which sets the `isCrashlyticsInitialized` to true. The next time you create a new instance, `isCrashlyticsInitialized` will be set to false even though Crashlytics has been initialized.

All flags (the state  of the CrashReporter) are stored in static members. This hints that there should be only a single CrashReporter object. Let's try to make the CrashReporter static. Start by removing the constructor and pass in a Context to `initialize`. Depending on the result, you might also store the SharedPreferences in a static variable. Then you don't need to store `isAnonymousReportEnabled` and `isRecoveredFromCrash` in a variable since you can always get them from the SharedPreferences. Maybe create helper methods to get these flags. e.g. `isAnonymousReportEnabled() { return preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false)}`;

You may also need to use [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) in some places. e.g. `if (!initalized) return;`.

AFAIS, the only place where you can't pass the Context as parameter, is, when you manually log an exception. In all other places you can because you always passed in the Context in the constructor already.

Give it a try and let me know how it worked out.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,148 @@
...
+
+	private CrashReporter() {
+	}
+
+	public static boolean initialize(Context context) {
+
+		CrashReporter.context = context;
+		preferences = PreferenceManager.getDefaultSharedPreferences(context);
+		setIsCrashReportEnabled(BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED);
+		isCrashlyticsInitialized = false;
```
robertpainsi: Please remove, `isCrashlyticsInitialized` is set later in this method to either `true` or `false`

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -368,6 +385,14 @@ public static void setAutoCrashReportingEnabled(Context context, boolean value)
 		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
 		editor.putBoolean(SETTINGS_CRASH_REPORTS, value);
 		editor.commit();
+
+		if (!value) {
```
robertpainsi: Please rename `value` to `isEnabled` or `enabled`

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -368,6 +385,14 @@ public static void setAutoCrashReportingEnabled(Context context, boolean value)
...
+			CrashReporter.disableCrashlytics();
+			Log.e(TAG, "Crashlytics Disabled");
+		} else {
+			CrashReporter.initialize(context);
+			Log.e(TAG, "Crashlytics Enabled");
```
robertpainsi: Please don't log it as `e`.
'Crashlytics Enabled' may be misleading. Crashlytics may be enabled depending on the flags in the CrashReporter. So it would be better to move the logging up one line in this block and rename it to `Enabling Crashlytics`.

However, the CrashReporter already does some logging. I'm not sure if it's necessary to log twice…

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,148 @@
...
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	static Context context;
+	private static SharedPreferences preferences;
+	private static boolean isCrashReportEnabled;
+	private static boolean isCrashlyticsInitialized;
+
+	private CrashReporter() {
+	}
```
robertpainsi: Nice trick :wink:

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,148 @@
...
+		Gson gson = new Gson();
+		String json = preferences.getString(EXCEPTION_FOR_REPORT, "");
+		Throwable exception = gson.fromJson(json, Throwable.class);
+		logException(exception);
+		preferences.edit().remove(EXCEPTION_FOR_REPORT).commit();
+	}
+
+	public static void storeUnhandledException(Throwable exception) {
+		SharedPreferences.Editor prefsEditor = preferences.edit();
+		if (isCrashReportEnabled) {
```
robertpainsi: I think this should be `isCrashReportEnabled && isAnonymousReportEnabled()`, right?

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,135 @@
...
+	private Exception exception;
+
+	@Before
+	public void setUp() {
+		context = InstrumentationRegistry.getTargetContext();
+		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		editor = sharedPreferences.edit();
+		editor.clear();
+		editor.commit();
+		exception = new RuntimeException("Error");
```
robertpainsi: I think you can also add `CrashReporter.setIsCrashReportEnabled(true);` to the setUp and remove it from every test. This way every test is set up to behave like in production. Keep all `CrashReporter.setIsCrashReportEnabled(false);`

Maybe we should also add a `@AfterClass` aka `tearDownClass` to set the `isCrashReportEnabled` to `false` at the end.

What do you think?

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@VisibleForTesting
+	public static void setIsCrashReportEnabled(boolean value) {
+		isCrashReportEnabled = value;
+	}
+
+	public static boolean logException(Throwable t) {
+
+		if (isAnonymousReportEnabled() && isCrashReportEnabled) {
+			Log.e(TAG, "LOG_EXCEPTION : " + t);
```
robertpainsi: I'm not 100% sure if we should log here as well. In many places the exception is already logged. So every exception is logged twice. Maybe just log that an exception is logged without logging the exception itself? Or remove it entirely…

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+import org.catrobat.catroid.ui.SettingsActivity;
+import org.catrobat.catroid.utils.CrashReporter;
+import org.junit.Before;
+import org.junit.Test;
+
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+
+public class CrashReporterTest {
```
robertpainsi:
```java
… extends AndroidTestCase
```

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
nishantt95: You are right. We already did the logging at initialize() method so I'll be removing it from here.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
nishantt95: Logically it should be but its just the previous implementation and I didn't change its logic. But now I'll implement it.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
nishantt95: Yes, that's logical.

nishantt95: Do I have to add after method() or after class?

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,148 @@
...
+
+	public static void disableCrashlytics() {
+		if (isCrashlyticsInitialized) {
+			Crashlytics crashlytics = new Crashlytics.Builder()
+					.core(new CrashlyticsCore.Builder().disabled(BuildConfig.DEBUG).build())
+					.build();
+			Fabric.with(context, crashlytics);
+			isCrashlyticsInitialized = false;
+		}
+	}
```
robertpainsi: Please remove, Crashlytics can't be disabled during runtime.
https://stackoverflow.com/questions/32168754/how-to-disable-crashlytics-fabric-at-runtime-when-user-changes-preferences/32867527#32867527 I can confirm that.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
robertpainsi: This check (`isCrashReportEnabled && isAnonymousReportEnabled()`) is used often. Please create a method for it.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
robertpainsi: Since it isn't an expensive operation after each test would be better

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,132 @@
...
+		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		editor = sharedPreferences.edit();
+		editor.clear();
+		editor.commit();
+		exception = new RuntimeException("Error");
+		CrashReporter.setIsCrashReportEnabled(true);
+	}
+
+	@After
+	public void onFinish() {
```
robertpainsi: Please rename to `tearDown()`. It's just convention for junit tests.
`@Before` is normally named `setUp()`
`@After` is normally named `tearDown()`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,132 @@
...
+		CrashReporter.setIsCrashReportEnabled(false);
+	}
+
+	@Test
+	public void testCrashlyticsUninitializedOnAnonymousReportDisabled() {
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
+		editor.commit();
+
+		Assert.assertFalse(CrashReporter.initialize(context));
```
robertpainsi: See https://github.com/Catrobat/Catroid/pull/2337#discussion_r122846606, then you don't need to use `Assert.`.

catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
```diff
@@ -24,43 +24,26 @@
...
 		java.lang.Thread.UncaughtExceptionHandler {
 
 	private static final String TAG = BaseExceptionHandler.class.getSimpleName();
 
 	public static final int EXIT_CODE = 10;
-	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
-	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
-
-	private final SharedPreferences preferences;
+	private Context context;
```
robertpainsi: Unused

catroid/build.gradle
```diff
@@ -101,7 +117,7 @@ android {
             buildConfigField "String", "PROJECT_NAME", "\"${(String) getProjectName()}\"";
             buildConfigField "String", "START_PROJECT", "\"${(String) getProgramId(true) }\""
             buildConfigField "boolean", "FEATURE_APK_GENERATOR_ENABLED", "true"
-            buildConfigField "boolean", "FIREBASE_CRASH_REPORT_ENABLED", "false"
+            buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"
```
robertpainsi: The standalone version didn't use Firebase. Don't know why. Let's keep that for now.

catroid/build.gradle
```diff
@@ -66,8 +75,10 @@ android {
             buildConfigField "boolean", "FEATURE_ARDUINO_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_RASPI_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_POCKETMUSIC_ENABLED", "true"
-            buildConfigField "boolean", "FIREBASE_CRASH_REPORT_ENABLED", "false"
+            buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"
```
robertpainsi: We somehow have to know if we should initialize the real Crashlytics stuff or just use a debug version. So I guess we've to keep `false` here

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,147 @@
...
+
+		if (isReportingEnabled()) {
+
+			Crashlytics crashlytics = new Crashlytics.Builder()
+					.core(new CrashlyticsCore.Builder().disabled(BuildConfig.DEBUG).build())
+					.build();
+
+			Fabric.with(context, crashlytics);
+
+			Fabric.with(context, new Crashlytics());
```
robertpainsi: You can't initialize Fabric twice.

robertpainsi:
> Error:(63, 0) /{PATH}/Catroid/keystore.properties (No such file or directory)

https://medium.com/@cassioso/a-strategy-to-secure-your-api-keys-using-gradle-b9c107272860 won't quite work.

I'm not familiar with gradle at all. Mostly using c&p code from the internet. However, I found a file called `catroid/google-services.json`. It uses fake data. So I guess that's what we should do as well. Just add a file keystore.properties to git and add a fake api key.

The person who creates the release will have to replace the fake api key with the real one.

If there is a better solution, just go for it :wink:

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
robertpainsi: …however, you have to initialize either
* the debug Fabric **or**
* the real Fabric

robertpainsi: Oh, my fault. You're using a flag now. I see…

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,144 @@
...
+
+	private static final String TAG = CrashReporter.class.getSimpleName();
+
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	static Context context;
+	private static SharedPreferences preferences;
+	private static boolean isCrashReportEnabled;
+	private static boolean isCrashlyticsInitialized;
```
robertpainsi: Not used anymore :wink:

nishantt95:
`debug{
...
buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"
}`

will result in the failure of multiple test cases because we are using `setIsCrashReportEnabled(BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED);` 
inside `initialize(Context context)`

Now we cannot alter it before the initialization of Crashlytics as it will always take the original value from the build.gradle and will not initialize Crashlytics because `isCrashReportEnabled -> setIsCrashReportEnabled(BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED);` will always be false (original value from the build.gradle) where it needs to be true for initializing Crashlytics.

robertpainsi:
`isCrashReportEnabled` was never intended to be (re)set. What do you think about?

`private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;`

nishantt95:
It worked! Thanks

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -364,10 +380,14 @@ public static boolean isRaspiSharedPreferenceEnabled(Context context) {
 		return getBooleanSharedPreference(false, SETTINGS_SHOW_RASPI_BRICKS, context);
 	}
 
-	public static void setAutoCrashReportingEnabled(Context context, boolean value) {
+	public static void setAutoCrashReportingEnabled(Context context, boolean isEnable) {
```
cheidenreich: I'd name the boolean parameter `isEnabled` or simply `enabled`

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,135 @@
...
+	private static boolean isRecoveredFromCrash() {
+		return preferences != null && preferences.getBoolean(RECOVERED_FROM_CRASH, false);
+	}
+
+	private static boolean isReportingEnabled() {
+		return isAnonymousReportEnabled() && isCrashReportEnabled;
+	}
+
+	@VisibleForTesting
+	public static void setIsCrashReportEnabled(boolean value) {
```
cheidenreich: Please also reconsider the naming of this parameter. `value` is not as expressive as `enabled` or `isEnabled`

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,135 @@
...
+			Fabric.with(context, new Crashlytics());
+			Log.d(TAG, "INITIALIZED!");
+			return true;
+		}
+		Log.d(TAG, "INITIALIZATION FAILED! [ Anonymous Report : " + isAnonymousReportEnabled() + " Crash Report : " + isCrashReportEnabled + " ]");
+		return false;
+	}
+
+	private static boolean isAnonymousReportEnabled() {
+		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
```
cheidenreich: At first glance i was concerned about accessing the shared preference every time this is called, but it seems to be fine.
Just for info: Turns out SharedPreferences are loaded from the disk into memory as a HashMap the first time they are accessed, so any further read access does not imply disk IO.

nishantt95: Thanks for the info. So, I think we will keep it as it is?

cheidenreich: yes :)

catroid/build.gradle
robertpainsi: My fault. I meant don't change it. Maybe one day we will enable it for standalone as well, but
not today

nishantt95: `buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"` is true for now. Should I change it to false?

robertpainsi: Yes, `CRASHLYTICS_CRASH_REPORT_ENABLED` should be `false` here.

catroid/src/main/res/values/strings.xml
```diff
@@ -1577,5 +1577,6 @@
     <string name="crash_dialog_always_send">Always send</string>
     <string name="crash_dialog_send">Send</string>
     <string name="plus" translatable="false">+</string>
+    <string name="com.crashlytics.android.build_id">RANDOM_UUID</string>
```
robertpainsi: Is this a leftover or am I missing something?

nishantt95: An exception was getting raised during the test execution and no tests were being performed.

`Test running failed: Instrumentation run failed due to 'io.fabric.sdk.android.services.concurrency.UnmetDependencyException'`

So I looked out for the last solution here and it worked:- [solution](https://stackoverflow.com/questions/28762624/com-crashlytics-android-crashlyticsmissingdependencyexception-during-gradle-tes )

robertpainsi: Ah, k. Please add a comment to this line so people know why this is needed.

Oh, and add `translatable="false"` as well.
```xml
<string name="com.crashlytics.android.build_id" translatable="false">RANDOM_UUID</string>
```

robertpainsi:
test this please

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,135 @@
...
+import io.fabric.sdk.android.Fabric;
+
+public final class CrashReporter {
+
+	private static final String TAG = CrashReporter.class.getSimpleName();
+
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	static Context context;
```
robertpainsi: Unused AFAICS

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -85,13 +86,16 @@
 
 	public static final String SETTINGS_CRASH_REPORTS = "setting_enable_crash_reports";
 
+	private static final String TAG = SettingsActivity.class.getSimpleName();
```
robertpainsi: unused

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
nishantt95: What's unused?
```
public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
static Context context;
```
They all are being used.

robertpainsi: Rephrasing :smile_cat:: Field 'context' is assigned but never accessed.

nishantt95: Ohh.. My mistake. I didn't saw this `initialize(Context context)`. On it.

robertpainsi: Thanks!
np, should've used the phrase in my second comment anyway.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,133 @@
...
+	@VisibleForTesting
+	public static void setIsCrashReportEnabled(boolean isEnabled) {
+		isCrashReportEnabled = isEnabled;
+	}
+
+	public static boolean logException(Throwable t) {
+
+		if (isReportingEnabled()) {
+			Crashlytics.logException(t);
+			preferences.edit().remove(EXCEPTION_FOR_REPORT).commit();
```
robertpainsi: Please remove this line from here. `logException` has nothing to to with the `EXCEPTION_FOR_REPORT`.

A solution I can think of is
```java
…
preferences.edit().remove(EXCEPTION_FOR_REPORT).commit();
return logException(exception);
```
What do you think?

robertpainsi:
test this please

(this is just a Jenkins test. Will be removed. Just ignore for now)

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
robertpainsi: Rephrasing: A solution in the method `sendUnhandledCaughtException` I can think of is

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+public class CrashReporterTest extends AndroidTestCase {
+
+	private Context context;
+	private SharedPreferences sharedPreferences;
+	private SharedPreferences.Editor editor;
+	private Exception exception;
+
+	@Before
+	public void setUp() {
```
robertpainsi: Oh, since we're extending `AndroidTestCase` now, change to
```java
@Override
```
same for `tearDown()`.

`@Before` and `@After` is currently only used in Espresso tests :smile:

fabric.properties
```diff
@@ -0,0 +1 @@
+apiKey=0123456789012345678901234567890123456789
```
robertpainsi:
```
$ ./gradlew clean check
…
ERROR - Crashlytics Developer Tools error.
java.lang.IllegalArgumentException: Crashlytics found an invalid API key: null.
…
```

This file has to be in the catroid folder.
Since you've to update it anyway, please add a newline to the end of the file. This will remove the 'No newline at end of file' info.

catroid/src/main/res/values/strings.xml
```diff
@@ -1577,5 +1577,8 @@
     <string name="crash_dialog_always_send">Always send</string>
     <string name="crash_dialog_send">Send</string>
     <string name="plus" translatable="false">+</string>
+
+    <!--Solution to handle fabric's UnmetDependencyException during the test -->
+    <string name="com.crashlytics.android.build_id" translatable="false">RANDOM_UUID</string>
```
robertpainsi: Once the fabric.properties file is moved to the catroid folder, remove this string.

```
./gradlew clean check
…
* What went wrong:
Execution failed for task ':catroid:mergeCatroidReleaseResources'.
[string/com.crashlytics.android.build_id] …/catroid/src/main/res/values/strings.xml
[string/com.crashlytics.android.build_id] …/catroid/build/generated/fabric/res/catroid/release/values/com_crashlytics_build_id.xml:
Error: Duplicate resources
…
```

catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
```diff
@@ -23,45 +23,25 @@
...
-		SharedPreferences.Editor prefsEditor = preferences.edit();
-		if (BuildConfig.FIREBASE_CRASH_REPORT_ENABLED) {
-			Gson gson = new Gson();
-			String check = preferences.getString(BaseExceptionHandler.EXCEPTION_FOR_REPORT, "");
-			if (check.isEmpty()) {
-				String json = gson.toJson(exception);
-				prefsEditor.putString(EXCEPTION_FOR_REPORT, json);
-				prefsEditor.commit();
-			}
+		if (!BuildConfig.DEBUG) {
```
robertpainsi: Please remove the if condition (keep the content (exit, kill)). We are handling this in https://github.com/Catrobat/Catroid/pull/2365#pullrequestreview-46144841
There seems to be a problem if we don't exit, kill the process. More details on https://github.com/Catrobat/Catroid/pull/2365#pullrequestreview-46144841 will follow

catroid/src/main/res/values/strings.xml
```diff
@@ -1577,5 +1577,6 @@
     <string name="crash_dialog_always_send">Always send</string>
     <string name="crash_dialog_send">Send</string>
     <string name="plus" translatable="false">+</string>
+
```
robertpainsi: please remove.
No changes needed in strings.xml

robertpainsi:
Thanks for updating!
Please rebase your branch onto the latest develop branch. There are some Checkstyle/PMD violations which aren't from your code. Rebasing will fix those violations and Jenkins will hopefully have a green testrun. :crossed_fingers:

robertpainsi:
test this please

PR #2349, opened 2017-06-09
JENKINS-89 Added androidTest sourceSet

vinzynth:
JENKINS-89 Added androidTest sourceSet to provide excluded testcases to the gradle build.

This sourceSet is used to provide a list with excluded testcases on the Jenkins-System.

### DIFF
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -228,6 +228,14 @@ android {
         xmlOutput file("build/reports/lint.xml")
         htmlOutput file("build/reports/lint.html")
     }
+
+    sourceSets {
+        androidTest {
+            if (file('../testexclusions.txt').exists()) {
+                java.exclude file('../testexclusions.txt').readLines()
+            }
+        }
+    }
 }
 
 task copyAndroidNatives() {


mfuchs:
Could anyone please review and merge this commit?

mfuchs:
@84n4n4 could you please check this PR?

84n4n4:
as already discussed with matthias, im not a huge fan of blacklisting/excluding failing tests, especially if its not broken tests per se, but tests failing due to known bugs.

more people should be involved in this discussion / decision, i will bring it on the table on todays core meeting.

does this fix exclude testcases, or classes?

analog to this we could also introduce a new annotation, eg. "@KnownBug" or "@KnownToFail", then the codebase is searchable using this tag. (which could also include a string parameter with the jira ticketnumber) so to make it easier to locate all tests failing due to a certain bug, to ensure all of the are whitelisted again when someone fixes the bug.
i guess thats easier to handle in the long run, if multiple tests failing due to different reasons are accumulating, if one can search for a tag with ticket id in the codebase instead of one file listing all failing (due to different bugs) tests.

but all in all, id still be happy to just keep the tests just failing in the testruns, so everybodys aware that something has to be fixed....

84n4n4:
ok, so after discussion at team meeting - we do not see the benefit of showing a green testrun while something is broken but just hidden. everyone heres on the same page that **we do want broken stuff showing up as failing tests**.

mfuchs:
This PR is not per-se for ignoring broken tests.
It is also a necessity to allow running tests in parallel using the [Parallel Test Executor Plugin](https://wiki.jenkins.io/display/JENKINS/Parallel+Test+Executor+Plugin).

I am not a fan of ignoring tests on Jenkins, as that would really be the last step.
That we even though about this step was caused by having broken tests for years.
Having red tests for ages counteracts benefits of a CI system and is worst practice.
This is worsened by tests that might crash the emulator, leading to aborting the Jenkins job.

I understand that especially high-level tests (like all of the tests in androidTest) take a while to get working.
They are like an acceptance test that can be specified before the feature exists.
One way to handle such cases is to mark them with e.g. @Ignore("CAT-123").
Then they are skipped and it is possible to mark a build as unstable if there are skipped tests.
Of course it would also be possible to create new annotations as you mentioned. 

In any case it is beneficial to differentiate between tests of not implemented features and broken tests.
Currently that is not the case.

mfuchs:
What could also be possible is to skip certain tests on Jenkins, via custom annotations, while not skipping them locally.

84n4n4:
can you please come to the next core team meeting on wednesday, because i guess its easier to discuss in person.

>> It is also a necessity to allow running tests in parallel using the Parallel Test Executor Plugin.

i dont understand this. if a failing test (-> one failing assert in one test) is hindering parallel execution then something else must be majorly wrong.

>> That we even though about this step was caused by having broken tests for years.

currently there are NO broken tests, just failing tests.

>> Having red tests for ages counteracts benefits of a CI system and is worst practice.

i would argue that removing red stuff without fixing is worst practice. having red tests is not that bad.

>>I understand that especially high-level tests (like all of the tests in androidTest) take a while to get working.

the tests ARE working. the ones that currently are red are due to bugs in prod code.

>>They are like an acceptance test that can be specified before the feature exists.
One way to handle such cases is to mark them with e.g. @ignore("CAT-123").

but thats not test driven development then. by far not.

>>In any case it is beneficial to differentiate between tests of not implemented features and broken tests.
Currently that is not the case.

again, there are currently NO broken tests. theres also, and never should be tests for not implemented features in develop branch (again, not tdd), and if we end up having test that BROKEN tests, this should be dealt with by fixing the test instead of blacklisting it.

on a historical note - i was told by older members that the blacklisting practice was one of the things that led to the completely unusable test setup in the past years. because it ended up beeing easier to exclude tests than to fix them for good and in the end everybody lost track of whats broken....

mfuchs:
Just responding now. Zero time the last few days.
We already discussed some parts of the stuff below, but I also had some other ideas in the 

> the tests ARE working

I was referring to the multitude of failing and crashy Robotium tests.
They are end-of-life though as you mentioned so the solution is to not run them anymore for nightlies.
I completely agree here on not doing any test-exclusions on the Jenkins part.

I think we can also combine both our views in regard of TDD and get the best of them: Keeping track of not-yet-implemented features or bugs, while having at least unstable builds.
I think it would be possible to do that via a custom anotation like @knownfailure.
Locally (not on Jenkins) this should not interfere in regular test-execution, i.e. the test will fail.
On Jenkins though this instruction should run the test and

- mark the build as unstable if the test fails
- mark the build as failed if the test worked -> it was supposed to fail but did not, maybe the feature is done now and the attribute can be removed.

I hope we can discuss this in more detail tomorrow.

84n4n4:
hi!

after we had a clarifying talk about the status of the robotium tests and the nightly runs, do we still need this? or can this PR be closed?

mfuchs:
This PR was also intended to speed-up jobs by parallelising test-execution.
Yet that way is not really possible with test-suites.

So I think this PR can be closed.

84n4n4:
closed for reason stated by matthias above in last comment

PR #2350, opened 2017-06-12
[GSoC 2017] Base Version for Recycler View and Storage Handling Changes

thmq:


thmq:
@cheidenreich I tried to request a review from you, but apparently you are not in the group of people I can request a review from 😮 . Anyway, I'd be nice if you could have a look at it.

thmq:
ping @cheidenreich

re-cheid:
oh, i didn't see this, sorry
will look at it today after work

catroid/src/main/java/org/catrobat/catroid/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+import org.catrobat.catroid.copypaste.ListItem;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class RecyclerViewAdapter<T> extends RecyclerView.Adapter<RecyclerViewAdapter.ViewHolder> {
+
+	public static final String TAG = RecyclerViewAdapter.class.getSimpleName();
+
+	private static final int SELECTED_BG_COLOR = 0xFFDDDDDD;
```
cheidenreich: I'm not sure if here is the best place for this.

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/RenameDialog.java
```diff
@@ -0,0 +1,76 @@
...
+	@Override
+	protected boolean handlePositiveButtonClick() {
+		String newName = input.getText().toString().trim();
+
+		if (newName.equals(previousText)) {
+			renameInterface.clearSelection();
+			return true;
+		}
+
+		boolean newNameConsistsOfSpacesOnly = newName.isEmpty();
```
cheidenreich: `newName.isEmpty()` can be used directly in the if clause

catroid/src/main/java/org/catrobat/catroid/ui/fragment/RecyclerViewListFragment.java
```diff
@@ -0,0 +1,239 @@
...
+			prefix = prefix.trim();
+			String suffix = name.substring(name.lastIndexOf('(') + 1, name.lastIndexOf(')'));
+			i = Integer.parseInt(suffix);
+			i++;
+		} catch (Exception e) {
+			prefix = name;
+			i = 1;
+		}
+
+		return getUniqueItemName(prefix + " (" + i + ")");
```
cheidenreich: i think this recursion can be resolved with a simple while loop

catroid/src/main/java/org/catrobat/catroid/copypaste/ClipboardHandler.java
```diff
@@ -0,0 +1,67 @@
...
+
+	public static final String TAG = ClipboardHandler.class.getSimpleName();
+
+	private static ClipboardHandler instance;
+
+	private Clipboard clipboard;
+
+	private static final DirectoryPathInfo CLIPBOARD_DIRECTORY = new DirectoryPathInfo(new DirectoryPathInfo(Constants.DEFAULT_ROOT), "clipboard");
+
+	public static ClipboardHandler getInstance() {
```
thmq: well, we need the member clipboard to be persistent (through different activities). I'm going to look into it.

catroid/src/main/java/org/catrobat/catroid/ui/adapter/RecyclerViewAdapter.java
thmq: this has to be changed once the UX team's mind is made up about the colours (so this is basically just a temporary way to set it). additionally there is probably more the UX team wants to change for the selected items, so I thought for the time being, this should suffice.

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/RenameDialog.java
thmq: yes, absolutely. I kinda missed that one after all the changes.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/RecyclerViewListFragment.java
thmq: yes, I did not see the advantage in a loop, but now that I thought about it - A loop can get a stopping condition, which would prevent us from recursing forever if something goes wrong - good point. 😼

thmq:
In general, this is not a finished (i.e. mergeable) version, i want to introduce some code that is actually executed first. furthermore, there are some minor aspects that have to be discussed with the UX team in more detail, but thank you @cheidenreich for the first review.

catroid/src/main/java/org/catrobat/catroid/copypaste/ClipboardHandler.java
cheidenreich: if i'm not missing anything completely you can just make it private static

catroid/src/main/java/org/catrobat/catroid/ui/fragment/RecyclerViewListFragment.java
cheidenreich: yes, plus you don't propagate the result back through the complete call stack

catroid/src/main/java/org/catrobat/catroid/copypaste/ListItem.java
```diff
@@ -0,0 +1,50 @@
...
+
+import android.graphics.drawable.Drawable;
+
+import org.catrobat.catroid.io.storage.DirectoryPathInfo;
+
+import java.io.IOException;
+
+public interface ListItem extends Cloneable {
+
+	int THUMBNAIL_WIDTH = 80;
```
thmq: this is already static final as it is an interface.

catroid/src/main/java/org/catrobat/catroid/data/LookInfo.java
```diff
@@ -0,0 +1,125 @@
...
+	public void setFilePathInfo(FilePathInfo filePathInfo) {
+		this.filePathInfo = filePathInfo;
+	}
+
+	@Override
+	public void createThumbnail() {
+		String imagePath = filePathInfo.getAbsolutePath();
+
+		BitmapFactory.Options options = new BitmapFactory.Options();
+		options.inPreferredConfig = Bitmap.Config.ARGB_8888;
```
cheidenreich: the Bitmap config may also be moved to the interface

PR #2352, opened 2017-06-14
[CAT-2099] Fix crash when Sound/Look is renamed

rschluesselbauer:
Removes the member 'name/title' from the **hashCode** function of **LookData** and **SoundInfo**.
Reason: LookData & SoundInfo are used as keys in a HashMap. 
If the hash changes when renaming a look or a sound, we will not be able to find the object in the HashMap.
This resulted in a NPE for renaming a sound. There was a workaround in place for renaming a Look - it has now been replaced with a proper solution (not changing the hash value of the object).

This is not an issue for Sprites, Scenes, Project or NfcTags, as their hashcode function is not overridden or does not include their names.

Very minor cosmetic changes.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/adapter/LookAdapter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/adapter/LookAdapter.java
@@ -58,15 +58,7 @@ public class LookAdapter extends LookBaseAdapter implements ActionModeActivityAd
 		if (position < 0 || position >= idMap.size()) {
 			return INVALID_ID;
 		}
-
 		LookData item = getItem(position);
-		if (!idMap.containsKey(item)) {
-			idMap.clear();
-			for (int i = 0; i < getCount(); i++) {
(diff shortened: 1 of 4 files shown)


rschluesselbauer:
test this please

cheidenreich:
test this please

cheidenreich:
test this please

catroid/src/main/java/org/catrobat/catroid/ui/adapter/LookAdapter.java
```diff
@@ -58,15 +58,7 @@ public long getItemId(int position) {
 		if (position < 0 || position >= idMap.size()) {
 			return INVALID_ID;
 		}
-
 		LookData item = getItem(position);
-		if (!idMap.containsKey(item)) {
```
cheidenreich: apparently removing this causes a crash when editing a look in PocketPaint, renamin still works with it.
i know this is ugly code, but is there a specific reason this was removed?

PR #2359, opened 2017-06-20
CAT-2314 Rewrite WhenNfcBrickTest for Espresso

JoachimLesser:
Refactor and rewrite whenNfcBrickTest for Espresso

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/WhenNfcBrickStageTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/WhenNfcBrickStageTest.java
@@ -1,0 +1,180 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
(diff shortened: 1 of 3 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiNFCTestUtils.java
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	public static void gotoNfcFragment(int nfcBrickPosition) {
+		onBrickAtPosition(nfcBrickPosition).onSpinner(R.id.brick_when_nfc_spinner)
+				.perform(click());
+		onView(withText(R.string.new_nfc_tag))
+				.perform(click());
+	}
+
+	public static void clickSpinnerValueOnBrick(int spinnerResourceId, int position, int stringResourceId) {
```
84n4n4: lets sit together this week and refactor all those spinner utility functions out of here. (90% of those are unnecessary by now (like this one here should be already available in BrickDataInteractionWrapper, and the remaining ones we just add to wrappers)

PR #2364, opened 2017-06-23
CAT-2324 Removed Phiro label from SetPenColor Brick

cheidenreich:


catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
```diff
@@ -154,29 +154,54 @@ public void showFormulaEditorToEditFormula(View view) {
...
+		Formula formulaRed;
+		Formula formulaGreen;
+		Formula formulaBlue;
+
+		try {
+			formulaRed = getFormulaWithBrickField(BrickField.PHIRO_LIGHT_RED);
+			formulaGreen = getFormulaWithBrickField(BrickField.PHIRO_LIGHT_GREEN);
+			formulaBlue = getFormulaWithBrickField(BrickField.PHIRO_LIGHT_BLUE);
+		} catch (IllegalArgumentException ex) {
+			// brick is already correct
```
rschluesselbauer: I'm not sure about using exceptions for flow control, especially since the case that the brick is already correct is not really an exception (should happen quite a bit). What do you think?

cheidenreich: yeah not the cleanest way, any suggestions?

cheidenreich:
changed handling now to a replace the brick field if its present in the formula map.

cheidenreich:
please run espresso tests

cheidenreich:
can someone please approve this? @84n4n4 @thmq @robertpainsi ?

cheidenreich:
@84n4n4 @thmq @robertpainsi anyone please? this is release blocking

PR #2365, opened 2017-06-24
CAT-2466 Test for BaseExceptionHandler.uncaughtException

nishantt95:
This pull request includes the implementation of test to check if an exception is stored in CrashReporter after uncaughtException.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
@@ -1,0 +1,78 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.ui;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+import android.support.test.InstrumentationRegistry;
+import android.test.AndroidTestCase;
+
+import org.catrobat.catroid.ui.BaseExceptionHandler;
+import org.catrobat.catroid.ui.SettingsActivity;
+import org.catrobat.catroid.utils.CrashReporter;
+import org.junit.Before;
+
+import static org.catrobat.catroid.utils.CrashReporter.EXCEPTION_FOR_REPORT;
+
+public class BaseExceptionHandlerTest extends AndroidTestCase {
+
+	private SharedPreferences sharedPreferences;
+	private SharedPreferences.Editor editor;
+
+	@Before
+	public void setUp() throws Exception {
+		super.setUp();
+		Context context = InstrumentationRegistry.getTargetContext();
+		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		editor = sharedPreferences.edit();
+		editor.clear();
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true);
+		editor.commit();
+		CrashReporter.setIsCrashReportEnabled(true);
+		CrashReporter.initialize(context);
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		CrashReporter.setIsCrashReportEnabled(false);
+		editor.clear();
+		editor.commit();
+		super.tearDown();
+	}
+
+	public void testExceptionStoredOnCallingUncaughtException() {
+		assertTrue(sharedPreferences.getString(EXCEPTION_FOR_REPORT, "").isEmpty());
+
+		BaseExceptionHandler baseExceptionHandler = new BaseExceptionHandler() {
+			@Override
+			protected void exit() {
+			}
+		};
+
+		baseExceptionHandler.uncaughtException(null, new RuntimeException("Test Error"));
+
+		assertFalse(sharedPreferences.getString(EXCEPTION_FOR_REPORT, "").isEmpty());
+	}
+}
--- a/catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
@@ -23,20 +23,19 @@
 
 package org.catrobat.catroid.ui;
 
-import android.util.Log;
-
 import org.catrobat.catroid.utils.CrashReporter;
 
 public class BaseExceptionHandler implements
 		java.lang.Thread.UncaughtExceptionHandler {
 
-	private static final String TAG = BaseExceptionHandler.class.getSimpleName();
-
 	private static final int EXIT_CODE = 10;
 
 	public void uncaughtException(Thread thread, Throwable exception) {
-		Log.e(TAG, "unhandled exception", exception);
 		CrashReporter.storeUnhandledException(exception);
+		exit();
+	}
+
+	protected void exit() {
 		System.exit(EXIT_CODE);
 		android.os.Process.killProcess(android.os.Process.myPid());
 	}


catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true);
+		editor.commit();
+		CrashReporter.setIsCrashReportEnabled(true);
+		CrashReporter.initialize(context);
+	}
+
+	public void testExceptionStoredOnCallingUncaughtException() {
+		assertTrue(sharedPreferences.getString(EXCEPTION_FOR_REPORT, "").isEmpty());
+
+		new BaseExceptionHandler().uncaughtExceptionTest(new RuntimeException("Test Error"));
```
robertpainsi: To test `uncaughtException` you need to call it :wink:
Let's see what you come up with. Let me know if you need help or a hint.

catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
```diff
@@ -23,45 +23,38 @@
...
-				String json = gson.toJson(exception);
-				prefsEditor.putString(EXCEPTION_FOR_REPORT, json);
-				prefsEditor.commit();
-			}
+		if (!sharedPreferences.getBoolean(IS_TESTING_BASE_EXCEPTION_HANDLER, false)) {
+			System.exit(EXIT_CODE);
+			android.os.Process.killProcess(android.os.Process.myPid());
+		} else {
+			editor.putBoolean(IS_TESTING_BASE_EXCEPTION_HANDLER, false);
 		}
```
robertpainsi: Interesting solution. :+1:

However, it changes the production code in a way I don't like. Here are two solutions I would prefer. If you come up with another solution, you can definitely give it a try :wink:

We only exit and kill the process when not in debug mode.
```java
public void uncaughtException(Thread thread, Throwable exception) {
  CrashReporter.storeUnhandledException(exception);
  if (!BuildConfig.DEBUG) {
    System.exit(EXIT_CODE);
    android.os.Process.killProcess(android.os.Process.myPid());
  }
}
```
or we override the method which exits and kills the process. This is as well known idiom in testing.
```java
// BaseExceptionHandler.java
public void uncaughtException(Thread thread, Throwable exception) {
  CrashReporter.storeUnhandledException(exception);
  exit();
}

protected void exit() {
  System.exit(EXIT_CODE);
  android.os.Process.killProcess(android.os.Process.myPid());
}

// BaseExceptionHandlerTest.java
BaseExceptionHandler baseExceptionHandler = new BaseExceptionHandler() {
  @Override
  protected void exit() {
  }
}
…
baseExceptionHandler.uncaughtException(null, new RuntimeException("Test Error"));
…
```

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+	public void testCrashlyticsUninitializedOnCrashReportDisabled() {
+		CrashReporter.setIsCrashReportEnabled(false);
+
+		assertFalse(CrashReporter.initialize(context));
+	}
+
+	public void testUnhandledExceptionStoredOnCrashReportEnabled() {
+		CrashReporter.storeUnhandledException(exception);
+
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: This could be `assertFalse(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());` for better readability.

This is valid for all further occurrences.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
+	}
+
+	public void testUnhandledExceptionStoredOnNoPreviousExceptionStored() {
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
+
+		CrashReporter.storeUnhandledException(exception);
+
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: `assertFalse(...);`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	public void testUnhandledExceptionNotStoredOnPreviousExceptionStored() {
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
+
+		exception = new RuntimeException("Error 1");
+		CrashReporter.storeUnhandledException(exception);
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: `assertFalse(...);`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+		String error1Data = sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "");
+
+		exception = new RuntimeException("Error 2");
+		CrashReporter.storeUnhandledException(exception);
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: `assertFalse(...);`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").equals(error1Data));
+	}
+
+	public void testSharedPreferencesClearedAfterLoggingException() {
+		CrashReporter.storeUnhandledException(exception);
+
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: `assertFalse(...);`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+		CrashReporter.sendUnhandledCaughtException();
+
+		assertTrue(sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
+	}
+
+	public void testSharedPreferencesClearedOnLoggingFailed() {
+		CrashReporter.storeUnhandledException(exception);
+
+		assertTrue(!sharedPreferences.getString(CrashReporter.EXCEPTION_FOR_REPORT, "").isEmpty());
```
cheidenreich: `assertFalse(...);`

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -0,0 +1,134 @@
...
+			}
+		}
+		return false;
+	}
+
+	public static void sendUnhandledCaughtException() {
+		Log.d(TAG, "AFTER_EXCEPTION : sendCaughtException()");
+		Gson gson = new Gson();
+		String json = preferences.getString(EXCEPTION_FOR_REPORT, "");
+		Throwable exception = gson.fromJson(json, Throwable.class);
```
cheidenreich: what happens if the String is empty?

keystore.properties
```diff
@@ -0,0 +1 @@
+FabricApiKey = "4901b94d39bb066fcdc1349bc4e9d47d50d24bec"
```
cheidenreich: is this the real one?

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
nishantt95: I think this issue is supposed to be handle [here #2366](https://github.com/Catrobat/Catroid/pull/2366). I will be updating it soon.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java
nishantt95: This issue is handled here[ CAT-2420 Replace Firebase with Crashlytics #2337](https://github.com/Catrobat/Catroid/pull/2337) and will be updating it here soon.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
robertpainsi: @cheidenreich, all pull requests build on #2337. So you only need to review the last (or latest) commits, the first one can always be skipped. e.g. here you only need to review [pull/2365@b57e2d2](https://github.com/Catrobat/Catroid/pull/2365/commits/b57e2d285eea8bc768897a9e03a05c3c870ea97e) (commit may change and link may be broken in the future). Once #2337 is merged, we will do a rebase and remove the first commit from the PRs. We just need it now to build the project and fix all CrashReporter related tickets.

Yes, it will be handled in #2366.
This code is from #2337 where we kept as much code as possible from the original Firebase implementation. To create a better git history and make it easier to spot bugs later on.
In #2337 it isn't a problem because this code will only be executed if there is an exception stored and the SendReportDialog isn't currently used at all (we should remove it sometimes later on :wink:). In #2366 it definitely will be a problem :smile_cat:

But still, good catch!

keystore.properties
robertpainsi: not from pocket code, no

catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
robertpainsi: If we choose option one, manually test (or use) the app and the app crashes, the app hangs completely (black screen, no other feedback, not even an App has crashed or restart). I don't like that at all. If the app crashes, the app should restart (or show an App has crashed notification). So option one is off the table.

nishantt95: Started working with option 2

catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Before;
+
+import static org.catrobat.catroid.utils.CrashReporter.EXCEPTION_FOR_REPORT;
+
+public class BaseExceptionHandlerTest extends AndroidTestCase {
+
+	private SharedPreferences sharedPreferences;
+
+	@Before
+	public void setUp() {
```
robertpainsi:
```java
@Override
	super.setUp();
```

robertpainsi:
test this please

catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
robertpainsi: @nishantt95, please update the method signature and add a `super` call.

catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
```diff
@@ -0,0 +1,68 @@
...
+		Context context = InstrumentationRegistry.getTargetContext();
+		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		SharedPreferences.Editor editor = sharedPreferences.edit();
+		editor.clear();
+		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true);
+		editor.commit();
+		CrashReporter.setIsCrashReportEnabled(true);
+		CrashReporter.initialize(context);
```
robertpainsi: **Not** in this PR, but in it's own follow-up.
Create methods for clearing the shared preferences or enabling crash reporting to remove a lot of duplicate code.

robertpainsi:
Don't forget to disable the CrashReporter after the testrun (tearDown)

robertpainsi:
Thanks for updating!

test this please

catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		super.tearDown();
+		CrashReporter.setIsCrashReportEnabled(false);
+		editor.clear();
+		editor.commit();
+	}
```
robertpainsi: `super.tearDown();` should be at the end of the method. It's the same principle as with the [pmd rule Android.CallSuperLast](http://pmd.sourceforge.net/pmd-4.3.0/rules/android.html#CallSuperLast).

Best explained with an example.
```
Parent {
  setUp() {
    database = new Database();
    database.open();
  }

  tearDown() {
    database.close();
    database = null;
  }
}

Child extends Parent {
  setUp() {
    super.setUp();
    database.add(item1);
  }

  tearDown() {
    #1 super.tearDown();
    database.remove(item1);
    #2 super.tearDown();
  }
}
```
If you add `super.tearDown` at position \#1 in `Child.tearDown`, you'll see that the database will be closed and set to null. So ` database.remove(item1);` in the Child will not work. Even if you don't set the database object to null in `Parent.tearDown`, the database has been closed already.

Position \#2 will work perfectly. You remove every resource created/initialized in the Child first and then the Parent will release it's resources.

However, you should never work with a real database in unit tests and therefore shouldn't care about removing the item anyway. It's just the first example which came into my mind and I think it perfectly describes the scenario.

robertpainsi:
test this please

robertpainsi:
test this please

PR #2366, opened 2017-06-24
CAT-2467 CrashReporter.CheckIfCrashRecoveryAndFinishActivity

nishantt95:
This pull request includes the modified version of CrashReporter.checkIfCrashRecoveryAndFinishActivity where the unrelated code is removed.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
@@ -27,6 +27,7 @@ import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
@@ -36,6 +37,7 @@ import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.R;
 import org.catrobat.catroid.ui.dialogs.AboutDialogFragment;
 import org.catrobat.catroid.ui.dialogs.TermsOfUseDialogFragment;
+import org.catrobat.catroid.utils.CrashReporter;
 import org.catrobat.catroid.utils.ToastUtil;
 import org.catrobat.catroid.utils.Utils;
 
@@ -44,6 +46,7 @@ public abstract class BaseActivity extends Activity {
 	private boolean returnToProjectsList;
 	private String titleActionBar;
 	private boolean returnByPressingBackButton;
+	private static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
 
 	@Override
 	protected void onCreate(Bundle savedInstanceState) {
@@ -52,7 +55,7 @@ public abstract class BaseActivity extends Activity {
 		returnToProjectsList = false;
 		returnByPressingBackButton = false;
 		Thread.setDefaultUncaughtExceptionHandler(new BaseExceptionHandler());
-		Utils.checkIfCrashRecoveryAndFinishActivity(this);
+		checkIfCrashRecoveryAndFinishActivity(this);
 		getActionBar().setDisplayHomeAsUpEnabled(true);
 	}
 
@@ -121,6 +124,21 @@ public abstract class BaseActivity extends Activity {
 				break;
 		}
 		return super.onOptionsItemSelected(item);
+	}
+
+	private void checkIfCrashRecoveryAndFinishActivity(final Activity context) {
+		if (isRecoveredFromCrash()) {
+			CrashReporter.sendUnhandledCaughtException();
+			if (!(context instanceof MainMenuActivity)) {
+				context.finish();
+			} else {
+				PreferenceManager.getDefaultSharedPreferences(this).edit().putBoolean(RECOVERED_FROM_CRASH, false).commit();
+			}
+		}
+	}
+
+	private boolean isRecoveredFromCrash() {
+		return PreferenceManager.getDefaultSharedPreferences(this).getBoolean(RECOVERED_FROM_CRASH, false);
 	}
 
 	// Taken from http://stackoverflow.com/a/11270668
--- a/catroid/src/main/java/org/catrobat/catroid/ui/dialogs/SendReportDialog.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/dialogs/SendReportDialog.java
@@ -37,7 +37,6 @@ import android.widget.CompoundButton;
 import org.catrobat.catroid.R;
 import org.catrobat.catroid.ui.SettingsActivity;
 import org.catrobat.catroid.utils.CrashReporter;
-import org.catrobat.catroid.utils.Utils;
 
 public class SendReportDialog extends DialogFragment {
 	public static final String TAG = "dialog_send_report";
@@ -58,7 +57,7 @@ public class SendReportDialog extends DialogFragment {
 				.setPositiveButton(R.string.crash_dialog_send, new DialogInterface.OnClickListener() {
 					@Override
 					public void onClick(DialogInterface dialog, int id) {
-						Utils.sendCaughtException(getActivity());
+						CrashReporter.sendUnhandledCaughtException();
 					}
 				})
 				.setNegativeButton(getString(R.string.no), new DialogInterface.OnClickListener() {
--- a/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
@@ -23,7 +23,6 @@
 
 package org.catrobat.catroid.utils;
 
-import android.app.Activity;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
@@ -35,7 +34,6 @@ import com.crashlytics.android.core.CrashlyticsCore;
 import com.google.gson.Gson;
 
 import org.catrobat.catroid.BuildConfig;
-import org.catrobat.catroid.ui.MainMenuActivity;
 import org.catrobat.catroid.ui.SettingsActivity;
 
 import io.fabric.sdk.android.Fabric;
@@ -44,10 +42,8 @@ public final class CrashReporter {
 
 	private static final String TAG = CrashReporter.class.getSimpleName();
 
-	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
-	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
-
 	private static SharedPreferences preferences;
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
 	private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;
 
 	private CrashReporter() {
@@ -65,20 +61,12 @@ public final class CrashReporter {
 			Log.d(TAG, "INITIALIZED!");
 			return true;
 		}
-		Log.d(TAG, "INITIALIZATION FAILED! [ Anonymous Report : " + isAnonymousReportEnabled() + " Crash Report : " + isCrashReportEnabled + " ]");
+		Log.d(TAG, "INITIALIZATION FAILED! [ Report : " + isReportingEnabled() + "]");
 		return false;
 	}
 
-	private static boolean isAnonymousReportEnabled() {
-		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
-	}
-
-	private static boolean isRecoveredFromCrash() {
-		return preferences != null && preferences.getBoolean(RECOVERED_FROM_CRASH, false);
-	}
-
 	private static boolean isReportingEnabled() {
-		return isAnonymousReportEnabled() && isCrashReportEnabled;
+		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false) && isCrashReportEnabled;
 	}
 
 	@VisibleForTesting
@@ -96,28 +84,13 @@ public final class CrashReporter {
 		return false;
 	}
 
-	public static boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
-		Log.d(TAG, "AFTER_EXCEPTION : checkIfCrashRecoveryAndFinishActivity()");
-		if (isRecoveredFromCrash()) {
-			if (isReportingEnabled()) {
-				sendUnhandledCaughtException();
-			}
-			if (!(context instanceof MainMenuActivity)) {
-				context.finish();
-			} else {
-				preferences.edit().putBoolean(RECOVERED_FROM_CRASH, false).commit();
-				return true;
-			}
-		}
-		return false;
-	}
-
 	public static void sendUnhandledCaughtException() {
-		Log.d(TAG, "AFTER_EXCEPTION : sendCaughtException()");
-		Gson gson = new Gson();
 		String json = preferences.getString(EXCEPTION_FOR_REPORT, "");
-		Throwable exception = gson.fromJson(json, Throwable.class);
-		if (logException(exception)) {
+		if (isReportingEnabled() && !json.isEmpty()) {
+			Log.d(TAG, "AFTER_EXCEPTION : sendCaughtException()");
+			Gson gson = new Gson();
+			Throwable exception = gson.fromJson(json, Throwable.class);
+			logException(exception);
 			preferences.edit().remove(EXCEPTION_FOR_REPORT).commit();
 		}
 	}
@@ -133,6 +106,5 @@ public final class CrashReporter {
 				prefsEditor.commit();
 			}
 		}
-		preferences.edit().putBoolean(RECOVERED_FROM_CRASH, true).commit();
 	}
 }
--- a/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
@@ -130,14 +130,6 @@ public final class Utils {
 		return true;
 	}
 
-	public static boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
-		return CrashReporter.checkIfCrashRecoveryAndFinishActivity(context);
-	}
-
-	public static void sendCaughtException(Context context) {
-		CrashReporter.sendUnhandledCaughtException();
-	}
-
 	public static boolean isNetworkAvailable(Context context, boolean createDialog) {
 		ConnectivityManager connectivityManager = (ConnectivityManager) context
 				.getSystemService(Context.CONNECTIVITY_SERVICE);


catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -94,17 +88,9 @@ public static boolean logException(Throwable t) {
...
-			} else {
-				preferences.edit().putBoolean(RECOVERED_FROM_CRASH, false).commit();
-				return true;
-			}
+		if (isReportingEnabled()) {
+			sendUnhandledCaughtException();
+			return true;
 		}
 		return false;
 	}
```
robertpainsi: It looks like we don't need this method at all. Updating sendUnhandledCaughtException accordingly will make this method obsolete. You will need to write new or improve tests for sendUnhandledCaughtException as well.

nishantt95: What I have understood is that checkIfCrashRecoveryAndFinishActivity() method should be removed and its features (which is required) should be implemented in sendUnhandledCaughtException(). So that's what I have to do?

robertpainsi: Not quite.

Everything related to RECOVERED_FROM_CRASH in checkIfCrashRecoveryAndFinishActivity ([ebce06f CrashReporter.java#L97](https://github.com/nishantt95/Catroid/blob/ebce06f4d6c4302c7a99c9edb71635190123dee5/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java#L97) and [ebce06f CrashReporter.java#L101-L106](https://github.com/nishantt95/Catroid/blob/ebce06f4d6c4302c7a99c9edb71635190123dee5/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java#L101-L106)) should be removed from the CrashReporter.java file and put somewhere else.

`CrashReporter.checkIfCrashRecoveryAndFinishActivity` is only called from `Utils.checkIfCrashRecoveryAndFinishActivity` which is only called from `BaseActivity.onCreate`. So to me it looks like moving the code related to RECOVERED_FROM_CRASH to the BaseActivity is the best place for two reason

* A method in Utils for that doesn't make sense since it's only used once and won't be used more often.
* The name checkIfCrashRecoveryAndFinishActivity indicates it does two things (note the *_And_* in the method name).

The remaining code in `CrashReporter.checkIfCrashRecoveryAndFinishActivity` is just the code from above. It mostly calls `sendUnhandledCaughtException`. However, there is still an if isReportingEnabled check which needs to be moved to sendUnhandledCaughtException as well.

And last, sendUnhandledCaughtException also needs to handle the case when EXCEPTION_FOR_REPORT is empty and `gson.fromJson(…)` may throw an exception.

So you need to update sendUnhandledCaughtException accordingly and write new tests.

Let me know if this clarifies it ^^

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +130,25 @@ public boolean onOptionsItemSelected(MenuItem item) {
 		return super.onOptionsItemSelected(item);
 	}
 
+	public static boolean isRecoveredFromCrash() {
+		return sharedPreferences != null && sharedPreferences.getBoolean(RECOVERED_FROM_CRASH, false);
```
robertpainsi: Remove static

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +130,25 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+	}
+
+	public static boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
```
robertpainsi: remove static

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -45,15 +48,19 @@
 	private String titleActionBar;
 	private boolean returnByPressingBackButton;
 
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+	private static SharedPreferences sharedPreferences;
```
robertpainsi: Remove class variable `sharedPreferences`

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -45,15 +48,19 @@
...
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		titleActionBar = null;
 		returnToProjectsList = false;
 		returnByPressingBackButton = false;
 		Thread.setDefaultUncaughtExceptionHandler(new BaseExceptionHandler());
-		Utils.checkIfCrashRecoveryAndFinishActivity(this);
+		checkIfCrashRecoveryAndFinishActivity(this);
 		getActionBar().setDisplayHomeAsUpEnabled(true);
+		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
```
robertpainsi: Inline occurrences

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -35,16 +34,16 @@
 import com.google.gson.Gson;
 
 import org.catrobat.catroid.BuildConfig;
-import org.catrobat.catroid.ui.MainMenuActivity;
 import org.catrobat.catroid.ui.SettingsActivity;
 
 import io.fabric.sdk.android.Fabric;
 
+import static org.catrobat.catroid.ui.BaseActivity.RECOVERED_FROM_CRASH;
```
robertpainsi: Code related to RECOVERED_FROM_CRASH should be in the BaseExceptionHandler

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
robertpainsi: or remove entirely?

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
nishantt95: What exactly I have to do in this line?

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +130,25 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 	}
 
+	public static boolean isRecoveredFromCrash() {
+		return sharedPreferences != null && sharedPreferences.getBoolean(RECOVERED_FROM_CRASH, false);
+	}
+
+	public static boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
+		if (isRecoveredFromCrash()) {
+
+			CrashReporter.sendUnhandledCaughtException();
```
robertpainsi: I would like to move this line to `onCreate`.
This removes the dependency between crash recovering and crash reporting entirely.

So the CrashReporter is like: "I don't care if the activity is recovering from a crash or not. If there is an exception stored, I'll report it."
And the CrashRecovering is like: "I don't care about the CrashReporter at all. I'll just finish every activity till I'm back in the MainMenuActivity."

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
robertpainsi: https://www.jetbrains.com/help/idea/inline.html

However, the inline refactoring of Android Studio doesn't work here.
Here by inline I mean replace `sharedPreferences` with `PreferenceManager.getDefaultSharedPreferences(this)` everywhere.

I think we don't need a member nor class member sharedPreferences. The only method sharedPreferences is used is in checkIfCrashRecoveryAndFinishActivity.

Since sharedPreferences != null can be removed as well, you can either remove isRecoveredFromCrash or pass in the sharedPreferences as parameter.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -35,16 +34,16 @@
 import com.google.gson.Gson;
 
 import org.catrobat.catroid.BuildConfig;
-import org.catrobat.catroid.ui.MainMenuActivity;
 import org.catrobat.catroid.ui.SettingsActivity;
 
 import io.fabric.sdk.android.Fabric;
 
+import static org.catrobat.catroid.ui.BaseExceptionHandler.RECOVERED_FROM_CRASH;
+
```
robertpainsi: Code related to crash recovering should me moved to other classes/files.

I can't comment on line `preferences.edit().putBoolean(RECOVERED_FROM_CRASH, true).commit();`. This is the only line depending on RECOVERED_FROM_CRASH. When you move this line to the BaseExceptionHandler, everything related to the crash recovering is removed from the CrashReporter :man_dancing:

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -73,11 +72,7 @@ private static boolean isAnonymousReportEnabled() {
 		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
 	}
 
-	private static boolean isRecoveredFromCrash() {
-		return preferences != null && preferences.getBoolean(RECOVERED_FROM_CRASH, false);
-	}
-
-	private static boolean isReportingEnabled() {
+	public static boolean isReportingEnabled() {
```
robertpainsi: Can be private

catroid/src/main/java/org/catrobat/catroid/ui/BaseExceptionHandler.java
```diff
@@ -30,6 +30,8 @@
 public class BaseExceptionHandler implements
 		java.lang.Thread.UncaughtExceptionHandler {
 
+	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
+
```
robertpainsi: To be consistent with other classes, the order of occurrences

1.  ```private static final long serialVersionUID = 1L;```
2. ```private static final String TAG = BaseExceptionHandler.class.getSimpleName();```
3. Anything else, e.g. ```public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";```

(1., 2. and 3. are each optional, no newline between 1. and 2., newline between 2. and 3.).

If files are very similar, it's easier for developer to work with any file and also they don't have to waste time on context switching.

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +128,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
 		return super.onOptionsItemSelected(item);
 	}
 
+	public boolean isRecoveredFromCrash() {
+		return PreferenceManager.getDefaultSharedPreferences(this).getBoolean(RECOVERED_FROM_CRASH, false);
+	}
```
robertpainsi: Please make private and put it after the `checkIfCrashRecoveryAndFinishActivity` method.

> We like to put the private utilities called by a public function right after the public function itself. This follows the stepdown rule and helps the program read like a newspaper article. -Clean Code|Chapter 10: Classes

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +128,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+
+	public boolean checkIfCrashRecoveryAndFinishActivity(final Activity context) {
```
robertpainsi: Please make private, change return type to void and remove the parameter.

Previously I was thinking about passing in the context (or SharedPreferences) into `isRecoveredFromCrash`. However, having this duplicate code (`PreferenceManager.getDefaultSharedPreferences(this)`) doesn't look bad after all. So, no need to pass in the context here nor pass any parameter to `isRecoveredFromCrash`.

nishantt95: I think parameter (context) is still required.

```
if (!(context instanceof MainMenuActivity)) {
	context.finish();
}
```

robertpainsi: `context == this` because it's called like `checkIfCrashRecoveryAndFinishActivity(this);`
However, it still looks nicer with the parameter context. So just keep it.

Maybe you can come up with a better name for `checkIfCrashRecoveryAndFinishActivity`?

Btw, it's definitely a good idea to declare the context parameter as final, good job! :+1:

nishantt95: It was not my idea to put context as final. I think is was same as before or I think IDE did the trick there. 😅

robertpainsi:
Jenkins, test this please :balloon::cactus::fearful:

robertpainsi:
test this please :balloon::cactus::fearful:

robertpainsi:
test this please

robertpainsi:
test this please

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -51,9 +55,10 @@ protected void onCreate(Bundle savedInstanceState) {
 		titleActionBar = null;
 		returnToProjectsList = false;
 		returnByPressingBackButton = false;
-		Thread.setDefaultUncaughtExceptionHandler(new BaseExceptionHandler());
-		Utils.checkIfCrashRecoveryAndFinishActivity(this);
+		Thread.setDefaultUncaughtExceptionHandler(new BaseExceptionHandler(this));
+		checkIfCrashRecoveryAndFinishActivity(this);
 		getActionBar().setDisplayHomeAsUpEnabled(true);
+		CrashReporter.sendUnhandledCaughtException();
```
84n4n4: this is really obvuscating whats going on here, it says sendUnhandledCaughtException and is called EVERY time on create of the activity.
why not move the whole thing to checkIfCrashRecovery or into exceptionHandler or somewhere else...

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/SendReportDialog.java
```diff
@@ -58,7 +58,7 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
 				.setPositiveButton(R.string.crash_dialog_send, new DialogInterface.OnClickListener() {
 					@Override
 					public void onClick(DialogInterface dialog, int id) {
-						Utils.sendCaughtException(getActivity());
+						Utils.sendCaughtException();
```
84n4n4: please directly call CrashReporter.sendUnhandledCaughtException(); and remove the utility function.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -73,10 +69,6 @@ private static boolean isAnonymousReportEnabled() {
 		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
```
84n4n4: are both isAnonymousReportEnabled and isReportingEnabled necessary? can we please combine this into one function when already touching this class.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -44,10 +42,8 @@
 
 	private static final String TAG = CrashReporter.class.getSimpleName();
 
-	public static final String RECOVERED_FROM_CRASH = "RECOVERED_FROM_CRASH";
-	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
-
 	private static SharedPreferences preferences;
+	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
 	private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;
```
84n4n4: please remove this variable or make it all upper case, because this is hiding a constant behind a variable name.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -96,29 +88,15 @@ public static boolean logException(Throwable t) {
...
-		Gson gson = new Gson();
 		String json = preferences.getString(EXCEPTION_FOR_REPORT, "");
-		Throwable exception = gson.fromJson(json, Throwable.class);
-		if (logException(exception)) {
-			preferences.edit().remove(EXCEPTION_FOR_REPORT).commit();
+		if (isReportingEnabled() && !json.isEmpty()) {
+			Log.d(TAG, "AFTER_EXCEPTION : sendCaughtException()");
+			Gson gson = new Gson();
+			Throwable exception = gson.fromJson(json, Throwable.class);
+			if (logException(exception)) {
```
84n4n4: please make logException return void.
- the boolean return value is only utilized here.
- its always going to be true, since in line 93  you made sure that its only reached when isReportingEnabled() is true

nishantt95: We can not remove the return value because it is [utilized in the unit test](https://github.com/nishantt95/Catroid/blob/7a78857c1d88f12c5260f071abb5663c781b130f/catroid/src/androidTest/java/org/catrobat/catroid/test/utils/CrashReporterTest.java#L134-L147). But yes, we are already using `isReportingEnabled()` so we don't need to apply if condition here.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
nishantt95: I think we can't combine them. There can be some situation when isAnonymousReportEnabled() is True and isReportingEnabled() is False. 

For example: While the application is in standalone mode or debug mode then (isAnonymousReportEnabled is enabled from the debug.apk) but reporting is set to false. This is to prevent the reporting during the development mode. (if we will combine them then we will get emails from fabric for all the crashes from all the contributor of this application 😨).

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
nishantt95: Can't say anything about it. I do need some additional supporting comments here on its usage. Meanwhile understanding BaseActivity.java because I haven't done any work here.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
nishantt95: For making a variable upper case it needs to be a final (and for obvious reasons checkstyle showed the warnings). Now we cannot set it to upper case as this variable can't be set to final and we are changing its value for [the testing purpose](https://github.com/nishantt95/Catroid/blob/7a78857c1d88f12c5260f071abb5663c781b130f/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java#L76-L79).

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
robertpainsi: I like the idea of decoupling crash reporting and crash recovering a lot. It also makes it easier to test. https://github.com/Catrobat/Catroid/pull/2366#discussion_r125107704

@nishantt95, for now please move this method back to the checkIfCrashRecovery. Moving `CrashReporter.sendUnhandledCaughtException();` somewhere else is out of the scope of this ticket anyway.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
robertpainsi: I guess @84n4n4 means
```java
private static boolean isReportingEnabled() {
    return isCrashReportEnabled &&  preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false);
}
```

> This is to prevent the reporting during the development mode

Actually, crashes aren't reported to fabric during development mode anyway because
* Crashlytics is [initialized in development mode](https://github.com/Catrobat/Catroid/blob/5b3c137387d7190c99adab52313e46ea1d0f9931/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java#L61-L63) (and also disabled through [gradle](https://github.com/Catrobat/Catroid/blob/5b3c137387d7190c99adab52313e46ea1d0f9931/catroid/build.gradle#L90))
* The [real api keys](https://github.com/Catrobat/Catroid/blob/develop/catroid/fabric.properties) aren't commited.

However, as @nishantt95 already mentioned, the standalone version has disabled crash reporting by default (don't ask me why).

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -123,6 +127,21 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 
+	private void checkIfCrashRecoveryAndFinishActivity(final Activity context) {
+		if (isRecoveredFromCrash()) {
+			if (!(context instanceof MainMenuActivity)) {
+				context.finish();
+			} else {
+				PreferenceManager.getDefaultSharedPreferences(this).edit().putBoolean(RECOVERED_FROM_CRASH, false).commit();
+			}
+		}
+		CrashReporter.sendUnhandledCaughtException();
```
robertpainsi: Please move the line in the if isRecoveredFromCrash block.
```java
if (isRecoveredFromCrash()) {
  CrashReporter.sendUnhandledCaughtException();
  …
```

This way we don't change the behavior in any kind. We will discuss where to put it later on.

robertpainsi:
test this please

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
84n4n4: ah, ok, i get it now...

robertpainsi:
@nishantt95, please do me a last favor and reword your commit message. Maybe something like
```
CAT-2467 Move crash recovering to BaseActivity

Move crash recovering related code from the CrashReporter to the BaseActivity.
```

@84n4n4, do you agree on merging it?
In this PR we just want to move the crash recovering out of the CrashReporter. We can do further improvements in a different PR.

84n4n4:
ok for me.

robertpainsi:
est this please

robertpainsi:
test this please

robertpainsi:
@nishantt95, please compile project and fix errors

nishantt95:
Which errors? I am not getting any errors/warnings here. 😕 
- No compile error.
- No build error.
- Application is working on emulator.
- Application is working on physical android devices.

robertpainsi:
BaseExceptionHandlerTest has compile errors

catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
Error:(68, 47) error: constructor BaseExceptionHandler in class BaseExceptionHandler cannot be applied to given types;
required: Context
found: no arguments
reason: actual and formal argument lists differ in length
Error:(69, 4) error: method does not override or implement a method from a supertype
Error:Execution failed for task ':catroid:compileCatroidDebugAndroidTestJavaWithJavac'.
\> Compilation failed; see the compiler error output for details.

robertpainsi:
test this please

robertpainsi:
Please rebase this branch onto the latest develop and recompile again.

nishantt95:
test this please

robertpainsi:
add to whitelist

robertpainsi:
test this please

PR #2370, opened 2017-06-29
CAT-2315 GSOC[TH] Test categories

84n4n4:


84n4n4:
@cheidenreich 
@robertpainsi 
@thmq 

can someone please have a quick look?

84n4n4:
this isnt as big of a pull request as it seems, all tests have been touched to add the category annotations implemented. so you can skip on the test files, unless you want to have a look of what tests have been put in which categories.

theres also some minor changes to the NfcBrickTest, just to fix some stuff ive overlooked when reviewing it.

and then finally, the most eyesore in this pull request is the bash script that generates the AllEspressoTestsSuite...
theres no JUnit4 way of getting all tests in a single test suite besides explicitly listing their classes. there are some third party libraries to do exactly this, i tried multiple versions of both takari-cpsuite (ClasspathSuite) and junittoolbox.WildcardPatternSuite, and after two days of not getting any results i ditched the idea and reached for the hammer (bash)...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/testsuites/generateAllEspressoTestsSuite.sh
```diff
@@ -0,0 +1,42 @@
...
+#put packagename
+echo "package org.catrobat.catroid.uiespresso.testsuites;\n" >> AllEspressoTestsSuite.java
+
+#put imports
+find .. -name '*Test.java' | sed 's/\.\.\//import org.catrobat.catroid.uiespresso./' | sed 's/\//\./g' | sed 's/.java/;/' | sort >> AllEspressoTestsSuite.java
+
+#put annotations
+echo "import org.junit.runner.RunWith;\nimport org.junit.runners.Suite;\n\n@RunWith(Suite.class)\n@Suite.SuiteClasses({" >> AllEspressoTestsSuite.java
+
+#put classes to be importet
```
cheidenreich: typo: imported

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/testsuites/Level.java
```diff
@@ -0,0 +1,35 @@
...
+package org.catrobat.catroid.uiespresso.testsuites;
+
+public interface Level {
+	interface Detailed {
+	}
+
+	interface Functional {
+	}
+
+	interface Smoke {
```
cheidenreich: I'm still not sure what you mean by smoke :D

84n4n4: https://de.wikipedia.org/wiki/Smoke_testing

84n4n4: ...  is preliminary testing to reveal simple failures severe enough to (for example) reject a prospective software release. A smoke tester will select and run a subset of test cases that cover the most important functionality of a component or system ...

cheidenreich: ah, right, thanks 😃

cheidenreich:
so the bash script has to be executed manually or will this happen automatically on Jenkins?

84n4n4:
currently the bash script is run manually, but one could add this to a jenkins job. anyways, this is only to be updated when a new test class is added.

cheidenreich:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/testsuites/generateAllEspressoTestsSuite.sh
```diff
@@ -0,0 +1,42 @@
...
+#put annotations
+echo "import org.junit.runner.RunWith;\nimport org.junit.runners.Suite;\n\n@RunWith(Suite.class)\n@Suite.SuiteClasses({" >> AllEspressoTestsSuite.java
+
+#put classes to be imported
+find .. -name '*Test.java' | sed 's/^.*\///' | sed 's/java/class/' | sed 's/$/,/' | sed '$s/,//' | sed 's/^/		/' >> AllEspressoTestsSuite.java
+
+#put class
+echo "})\npublic class AllEspressoTestsSuite {\n}" >> AllEspressoTestsSuite.java
+
+exit
```
robertpainsi: Maybe this file should be executable (`chmod +x generateAllEspressoTestsSuite.sh`)? So everyone can easily execute it.
However, echo in bash doesn't convert `\n` into new line characters by default. You either have to use `#!/bin/sh`, `echo -e` or `printf`. Or if you like inlines or variables instead of comments, take a look at [these alternatives](https://gist.github.com/robertpainsi/4b4267bc8ff96dc5c689d058f8e32149).

Interesting to see that your SuiteClasses (classes to be imported) is completely different from mine :laughing:

84n4n4: ill take it! thanks..

interestingly my version with the \n's works for me.

regarding the sort order, yeah, well, thats still kind of a problem because the generated file will lead to lint error due to wrong import order, because an import of a class higher in the package hierarchy has to be higher up. this is why the androidStudio generated sort order looks like this:

import org.catrobat.catroid.uiespresso.SmokeTest;
import org.catrobat.catroid.uiespresso.annotations.FlakyTestTest;
...

and the one generated from the script looks like this:
...
import org.catrobat.catroid.uiespresso.pocketmusic.PocketMusicActivityTest;
import org.catrobat.catroid.uiespresso.SmokeTest;
import org.catrobat.catroid.uiespresso.stage.StageSimpleTest;
...

i honestly dont really know if that bash script is such a good idea, but on the other hand, i have no alternatives right now... (except writing it in python :-) )

its really a shame that jUnit4 does not have anything build in to just do that. (jUnit3 had as far as i could google up), but if that helps, googling this shows that more people are banging their heads against the wall with this...

robertpainsi: We could exclude this file entirely from Checkstyle [catroid/gradle/code_quality_tasks.gradle#L28](https://github.com/Catrobat/Catroid/blob/be9b08440d257bdb9adbd6cf142aae5b41a8e8d0/catroid/gradle/code_quality_tasks.gradle#L28).
However, if you reformat the whole project through AS, the import order will always be changed.

We should also add a comment to the java file that it's auto generated and that you should use the script to create it (and then run the formatter, if we don't go with excluding the file).

I think we should add a `| sort` [here](https://github.com/Catrobat/Catroid/blob/be9b08440d257bdb9adbd6cf142aae5b41a8e8d0/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/testsuites/generateAllEspressoTestsSuite.sh#L4). Otherwise the order is completely different depending on the OS. Not sure why find is different on Ubuntu :anguished:

```diff
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
-        PocketMusicActivityTest.class,
+        FormulaEditorKeyboardTest.class,
+        FormulaEditorTest.class,
         SmokeTest.class,
         FlakyTestTest.class,
-        WhenNfcBrickStageTest.class,
-        LegoEv3PlayToneBrickTest.class,
-        ComeToFrontBrickTest.class,
-        HideBrickTest.class,
-        ChangeTransparencyByNBrickTest.class,
+        PocketMusicActivityTest.class,
…
```

Since this task is executed quickly, maybe we add it to gradle to always run before any build?

PR #2371, opened 2017-06-29
CAT-2468 Testing that the CrashReporter is initialized when enabled through settings

nishantt95:
This pull request includes the test file for settings activity which tests the enable and disable functionality of Crash Report system.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/utils/SettingsActivityCrashReporterTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/utils/SettingsActivityCrashReporterTest.java
@@ -1,0 +1,70 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.utils;
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+
+import org.catrobat.catroid.ui.SettingsActivity;
+import org.catrobat.catroid.utils.CrashReporter;
+import org.catrobat.catroid.utils.CrashReporterInterface;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SettingsActivityCrashReporterTest {
+
+	@Mock
+	CrashReporterInterface reporter;
+
+	@Before
+	public void setUp() throws Exception {
+		CrashReporter.setIsCrashReportEnabled(true);
+		CrashReporter.setCrashReporterInterface(reporter);
+	}
+
+	@Test
+	public void testAutoCrashReportEnabling() {
+		Context context = InstrumentationRegistry.getTargetContext();
+		SettingsActivity.setAutoCrashReportingEnabled(context, false);
+
+		verify(reporter, times(0)).initialize(context);
+
+		SettingsActivity.setAutoCrashReportingEnabled(context, true);
+
+		verify(reporter, times(1)).initialize(context);
+	}
+
+	@After
+	public void tearDown() throws Exception {
+		CrashReporter.setIsCrashReportEnabled(false);
+	}
+}
--- a/catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
@@ -401,7 +401,6 @@ public class SettingsActivity extends PreferenceActivity {
 		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
 		editor.putBoolean(SETTINGS_CRASH_REPORTS, isEnabled);
 		editor.commit();
-
 		if (isEnabled) {
 			CrashReporter.initialize(context);
 		}
--- a/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
@@ -29,40 +29,22 @@ import android.preference.PreferenceManager;
 import android.support.annotation.VisibleForTesting;
 import android.util.Log;
 
-import com.crashlytics.android.Crashlytics;
-import com.crashlytics.android.core.CrashlyticsCore;
 import com.google.gson.Gson;
 
 import org.catrobat.catroid.BuildConfig;
 import org.catrobat.catroid.ui.SettingsActivity;
 
-import io.fabric.sdk.android.Fabric;
-
 public final class CrashReporter {
 
 	private static final String TAG = CrashReporter.class.getSimpleName();
 
-	private static SharedPreferences preferences;
 	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	protected static SharedPreferences preferences;
 	private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;
+	private static CrashReporterInterface reporter = new CrashlyticsCrashReporter();
 
 	private CrashReporter() {
-	}
-
-	public static boolean initialize(Context context) {
-
-		preferences = PreferenceManager.getDefaultSharedPreferences(context);
-
-		if (isReportingEnabled()) {
-			Crashlytics crashlyticsKit = new Crashlytics.Builder()
-					.core(new CrashlyticsCore.Builder().disabled(BuildConfig.DEBUG).build())
-					.build();
-			Fabric.with(context, crashlyticsKit);
-			Log.d(TAG, "INITIALIZED!");
-			return true;
-		}
-		Log.d(TAG, "INITIALIZATION FAILED! [ Report : " + isReportingEnabled() + "]");
-		return false;
 	}
 
 	private static boolean isReportingEnabled() {
@@ -70,18 +52,36 @@ public final class CrashReporter {
 	}
 
 	@VisibleForTesting
-	public static void setIsCrashReportEnabled(boolean isEnabled) {
-		isCrashReportEnabled = isEnabled;
+	public static void setCrashReporterInterface(CrashReporterInterface crashReporterInterface) {
+		reporter = crashReporterInterface;
 	}
 
-	public static boolean logException(Throwable t) {
+	public static boolean initialize(Context context) {
+
+		preferences = PreferenceManager.getDefaultSharedPreferences(context);
 
 		if (isReportingEnabled()) {
-			Crashlytics.logException(t);
+			reporter.initialize(context);
+			Log.d(TAG, "INITIALIZED!");
 			return true;
 		}
 
+		Log.d(TAG, "INITIALIZATION FAILED! [ Report : " + isReportingEnabled() + "]");
 		return false;
+	}
+
+	public static boolean logException(Throwable exception) {
+
+		if (isReportingEnabled()) {
+			reporter.logException(exception);
+			return true;
+		}
+		return false;
+	}
+
+	@VisibleForTesting
+	public static void setIsCrashReportEnabled(boolean isEnabled) {
+		isCrashReportEnabled = isEnabled;
 	}
 
 	public static void sendUnhandledCaughtException() {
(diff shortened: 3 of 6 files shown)


robertpainsi:
Will the [PreferenceMatchers](https://developer.android.com/reference/android/support/test/espresso/matcher/PreferenceMatchers.html) help here?

nishantt95:
I did try to use PreferenceMatchers and it doesn't seem to work because 

```
org.hamcrest.Matcher<android.view.View>
//is different from
org.hamcrest.Matcher<android.preference.Preference>
```
PreferenceMatchers would have helped in getting the checkbox with the help of key (instead of with matching it with thier text value) but it's not working as we want it to.

84n4n4:
hi!

i implemented a test for all settings and that they are reflected in the shared preferences correctly here:
https://github.com/Catrobat/Catroid/pull/2378
the flag for crash reporting is also covered there.

just as an info, the PreferenceMatchers methods work as follows for this task:

onData(PreferenceMatchers.withTitle(displayedTitleResourceString)).perform(click());

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -384,14 +384,16 @@ public static boolean isRaspiSharedPreferenceEnabled(Context context) {
 		return getBooleanSharedPreference(false, SETTINGS_SHOW_RASPI_BRICKS, context);
 	}
 
-	public static void setAutoCrashReportingEnabled(Context context, boolean isEnabled) {
+	public static boolean setAutoCrashReportingEnabled(Context context, boolean isEnabled) {
```
robertpainsi: Adding a return value just for testing should always be prevented. Additionally setters with a return value are always problematic. You never know what they are returning.

Stateful static classes like the CrashReporter are always hard to test. If code uses such classes, it even gets more complicated.

I've to do more research on this topic (maybe it's better to do DI and use interfaces? Don't know yet :laughing:) but my first thought to test it is. If you come up with something better, that would be great!
* Disable crash reporting in the setup process
* Logging an exception fails
* Enable crash reporting
* Successfully log an exception

nishantt95:
Conflicting file is completely changed. Should I build test over it?
catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityTest.java

robertpainsi:
Please create a new test file.

catroid/build.gradle
```diff
@@ -86,7 +86,7 @@ android {
             buildConfigField "boolean", "FEATURE_RASPI_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_POCKETMUSIC_ENABLED", "true"
             buildConfigField "boolean", "FEATURE_CAST_ENABLED", "true"
-            buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "false"
+            buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"
```
robertpainsi: Please create a separate pull request for this.
Also explain why this change is necessary, what the problem was and the effects of the patch. Also mention that the Jenkins Espresso testrun is green because the auto-generated AllEspressoTestsSuite.java file isn't up-to-date and doesn't contain the SettingsActivityTest. See [information in commit messages](https://github.com/Catrobat/Catroid/wiki/Commit-Message-Guidelines#information-in-commit-messages).

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -45,6 +45,7 @@
 	private static SharedPreferences preferences;
 	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
 	private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;
+	public static boolean IS_CRASH_REPORTER_INITIALIZED = false;
```
robertpainsi: just name it `initialized` or `isInitialized`.
* lowercase because it isn't a constant
* ~`CRASH_REPORTER`~ because it's known from the context

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -401,10 +401,7 @@ public static void setAutoCrashReportingEnabled(Context context, boolean isEnabl
 		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
 		editor.putBoolean(SETTINGS_CRASH_REPORTS, isEnabled);
 		editor.commit();
-
-		if (isEnabled) {
-			CrashReporter.initialize(context);
-		}
+		CrashReporter.IS_CRASH_REPORTER_INITIALIZED = CrashReporter.initialize(context);
```
robertpainsi: Instead of assigning the variable here, update the variable in the `CrashReporter.initialize` method itself.

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -401,10 +400,6 @@ public static void setAutoCrashReportingEnabled(Context context, boolean isEnabl
 		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
 		editor.putBoolean(SETTINGS_CRASH_REPORTS, isEnabled);
 		editor.commit();
-
-		if (isEnabled) {
-			CrashReporter.initialize(context);
-		}
```
robertpainsi: Thanks for letting me know that the testrun will be red without these lines of code :smile:

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityCrashReporterTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.ui.activity;
```
robertpainsi: Since you removed interaction with the UI, you can move it to a more appropriate package.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityCrashReporterTest.java
```diff
@@ -0,0 +1,62 @@
...
+		assertFalse(CrashReporter.isInitialized);
+
+		SettingsActivity.setAutoCrashReportingEnabled(context, true);
+
+		assertTrue(CrashReporter.isInitialized);
+	}
+
+	@After
+	public void tearDown() throws Exception {
+		CrashReporter.isInitialized = false;
```
robertpainsi: In this case, the setUp and tearDown should do the same thing (call the same methods) I guess.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityCrashReporterTest.java
nishantt95: To ->test->utils?

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -45,6 +45,7 @@
 	private static SharedPreferences preferences;
 	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
 	private static boolean isCrashReportEnabled = BuildConfig.CRASHLYTICS_CRASH_REPORT_ENABLED;
+	public static boolean isInitialized = false;
```
robertpainsi: Afaik `@VisibleForTesting` only works for methods.
Please create getter and setter.

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
nishantt95: Oh! 😶 😜

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityCrashReporterTest.java
robertpainsi: Since the `CrashReporterTest` is located there as well, :+1:

Maybe we should create a package `crashreporter` later on.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -59,12 +60,24 @@ public static boolean initialize(Context context) {
...
 
+	@VisibleForTesting
+	public static void setIsInitialized(boolean initialized) {
+		isInitialized = initialized;
+	}
+
+	@VisibleForTesting
+	public static boolean getIsInitialized() {
+		return isInitialized;
+	}
```
robertpainsi: A boolean getter is normally called `isX` or `hasX`.
Method names like `isInitialized` and `setInitialized` are better imo.

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/SettingsActivityCrashReporterTest.java
```diff
@@ -0,0 +1,63 @@
...
+		assertFalse(CrashReporter.getIsInitialized());
+
+		SettingsActivity.setAutoCrashReportingEnabled(context, true);
+
+		assertTrue(CrashReporter.getIsInitialized());
+	}
+
+	@After
+	public void tearDown() throws Exception {
+		CrashReporter.setIsInitialized(false);
```
robertpainsi: Don't forget to disable the CrashReporter as well :wink:

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -401,10 +401,7 @@ public static void setAutoCrashReportingEnabled(Context context, boolean isEnabl
...
+		CrashReporter.initialize(context);
```
robertpainsi: Please undo as was. Calling initialize even if you don't want to initialize the CrashReporter is really really odd.

robertpainsi:
As mentioned in the mails, I would still prefer something like https://gist.github.com/robertpainsi/240b60d47eda87ad912c39575f0b6211 (I recreated it from all other gist samples)

This way the used CrashReporter is completely independent from the CrashReporter logic itself. It also can be easily replaced, tested and also the crash reporter logic connected to the crash recovering logic will be testable as well (we won't test the crash recovering logic in this pr).

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
nishantt95: `editor.putBoolean(SETTINGS_CRASH_REPORTS, isEnabled);` will determine if CrashReporter will initialize or not. I thought checking the same thing twice will be redundant. But yes, it makes sense that calling the function even when it is not required is odd. Ok, will change it 👍

nishantt95:
Why do we need
```
@VisibleForTesting
  static void setReporter(ICrashReporter crashReporter) {
    reporter = crashReporter;
  }
```

if we are already using `private static ICrashReporter reporter = new CrashlyticsCrashReporter();`

It does work well with all the test cases.

robertpainsi:
Oh, I missed to tell you that. Now comes the sweet sugar chocolate candy bar part.
You only have to check if CrashReporter.reporter.initialize has been called (or not). You can do this easily with mocks. No `isInitialized` boolean or related code needed.

Just image that if the Firebase crash reporter would've been implemented like you implemented the Crashlytics crash reporter. Replacing Firebase with Crashlytics would've been easy as pie.

> Do I have to work on CrashRecovering? If yes then in this PR?

No and no :smile_cat: 
I just wanted to mention that this change isn't only related to the initialize stuff but also to the crash recovering.

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -29,61 +29,56 @@
...
-import io.fabric.sdk.android.Fabric;
-
 public final class CrashReporter {
 
 	private static final String TAG = CrashReporter.class.getSimpleName();
 
-	private static SharedPreferences preferences;
 	public static final String EXCEPTION_FOR_REPORT = "EXCEPTION_FOR_REPORT";
+
+	public static CrashReporterInterface reporter = new CrashlyticsCrashReporter();
```
robertpainsi: Please make private and add a `@VisibleForTesting` setter.
I still don't know why `@VisibleForTesting` only works for methods and classes :worried:

catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
```diff
@@ -29,61 +29,56 @@
...
 	public static boolean initialize(Context context) {
 
 		preferences = PreferenceManager.getDefaultSharedPreferences(context);
 
-		if (isReportingEnabled()) {
-			Crashlytics crashlyticsKit = new Crashlytics.Builder()
-					.core(new CrashlyticsCore.Builder().disabled(BuildConfig.DEBUG).build())
-					.build();
-			Fabric.with(context, crashlyticsKit);
+		if (CrashReporter.isReportingEnabled()) {
```
robertpainsi: ~`CrashReporter.`~`isReportingEnabled()`

catroid/src/androidTest/java/org/catrobat/catroid/test/utils/SettingsActivityCrashReporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+	@Test
+	public void testAutoCrashReportEnabling() {
+		Context context = InstrumentationRegistry.getTargetContext();
+		SettingsActivity.setAutoCrashReportingEnabled(context, false);
+
+		verify(CrashReporter.reporter, Mockito.times(0)).initialize(context);
+
+		SettingsActivity.setAutoCrashReportingEnabled(context, true);
+
+		verify(CrashReporter.reporter, Mockito.times(1)).initialize(context);
```
robertpainsi: `verify(`~`CrashReporter.`~`reporter, `~`Mockito.`~`times(1)).initialize(context);`

robertpainsi:
Overall looking good! I like those improvements. Especially using JUnit.4 and Mockito annotations!

robertpainsi:
test this please

PR #2373, opened 2017-07-02
Generalize string id's and translations for flavored versions

sjaindl:
Refactoring/Renaming according to @wslany 

+ 1 new string added: formula_editor_device_sensors (this is the title of the FE sensor category, text on keyboard stays the same)


cheidenreich:
please squash into one commit, since the last commit is only a single word changed

sjaindl:
@cheidenreich ok, I've squashed it to a single commit

PR #2379, opened 2017-07-05
CAT-2471 Adding parameters to sent Cloud Message

nishantt95:
This pull request includes the implementation of cloud messaging with custom parameters.

### DIFF
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -148,6 +148,7 @@ dependencies {
     compile 'com.google.code.gson:gson:2.7'
     compile 'com.github.johnpersano:supertoasts:2.0@aar'
     compile 'com.github.mrengineer13:snackbar:1.2.0'
+    compile 'com.google.firebase:firebase-messaging:9.4.0'
 
     compile 'com.koushikdutta.async:androidasync:2.+'
     compile 'com.sothree.slidinguppanel:library:3.3.0'
@@ -177,8 +178,8 @@ dependencies {
     compile fileTree(include: '*.jar', dir: 'src/main/libs-natives')
     androidTestCompile fileTree(include: '*.jar', dir: 'src/androidTest/libs')
     androidTestCompile 'com.jayway.android.robotium:robotium-solo:5.2.1'
-    androidTestCompile 'com.linkedin.dexmaker:dexmaker-mockito:1.5.1'
-    androidTestCompile 'org.mockito:mockito-core:1.10.19'
+    androidTestCompile 'com.linkedin.dexmaker:dexmaker-mockito:2.2.0'
+    androidTestCompile 'org.mockito:mockito-core:2.8.47'
 
     androidTestCompile ('com.android.support.test:runner:0.5') {
         exclude group: 'com.android.support'
@@ -187,6 +188,9 @@ dependencies {
         exclude group: 'com.android.support'
     }
     androidTestCompile ('com.android.support.test.espresso:espresso-contrib:2.2.2') {
+        exclude group: 'com.android.support'
+    }
+    androidTestCompile ('com.android.support.test.espresso:espresso-intents:2.2') {
         exclude group: 'com.android.support'
     }
 
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
@@ -1,0 +1,130 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.cloudmessaging;
+
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.media.RingtoneManager;
+import android.support.test.InstrumentationRegistry;
+import android.support.v4.app.NotificationCompat;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.cloudmessaging.CloudMessage;
+import org.catrobat.catroid.cloudmessaging.CloudMessaging;
+import org.catrobat.catroid.cloudmessaging.NotificationBuilderProvider;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.util.HashMap;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class CustomFirebaseMessagingServiceTest {
+
+	private NotificationBuilderProvider mockNotificationBuilderProvider;
+	private NotificationCompat.Builder mockNotificationBuilder;
+	private Notification mockNotification;
+	private NotificationManager mockNotificationManager;
+
+	@Before
+	public void setUp() {
+		mockNotificationBuilder = Mockito.mock(NotificationCompat.Builder.class, Mockito.RETURNS_SELF);
+		mockNotification = Mockito.mock(Notification.class);
+		mockNotificationBuilderProvider = Mockito.mock(NotificationBuilderProvider.class);
+		when(mockNotificationBuilderProvider.get()).thenReturn(mockNotificationBuilder);
+		when(mockNotificationBuilder.build()).thenReturn(mockNotification);
+		mockNotificationManager = Mockito.mock(NotificationManager.class);
+	}
+
+	@Test
+	public void testNotificationIsBuildAndShown() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.TITLE, "Title");
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "https://www.catrobat.org");
+
+		Context context = InstrumentationRegistry.getTargetContext();
+		CloudMessage cloudMessage = new CloudMessage(data);
+
+		CloudMessaging cloudMessaging = new CloudMessaging(context, mockNotificationBuilderProvider, mockNotificationManager);
+		cloudMessaging.showNotification(cloudMessage);
+
+		verify(mockNotificationBuilder).setContentTitle(eq("Title"));
+		verify(mockNotificationBuilder).setContentText(eq("Message"));
+		verify(mockNotificationBuilder).setSmallIcon(R.drawable.ic_launcher);
+		verify(mockNotificationBuilder).setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION));
+		verify(mockNotificationBuilder).setContentIntent(any(PendingIntent.class));
+		verify(mockNotificationManager).notify(anyInt(), eq(mockNotification));
+	}
+
+	@Test
+	public void testNotificationIsNotShownOnInvalidUrl() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.TITLE, "Title");
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "Invalid Url");
+
+		checkNotificationNotShownForInvalidData(data);
+	}
+
+	@Test
+	public void testNotificationIsNotShownOnRequiredFieldEmpty() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.TITLE, "");
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "https://www.catrobat.org");
+
+		checkNotificationNotShownForInvalidData(data);
+	}
+
+	@Test
+	public void testNotificationIsNotShownOnRequiredFieldMissing() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "https://www.catrobat.org");
+
+		checkNotificationNotShownForInvalidData(data);
+	}
+
+	private void checkNotificationNotShownForInvalidData(HashMap<String, String> data) {
+
+		Context context = InstrumentationRegistry.getTargetContext();
+		CloudMessage cloudMessage = new CloudMessage(data);
+
+		CloudMessaging cloudMessaging = new CloudMessaging(context, mockNotificationBuilderProvider, mockNotificationManager);
+		cloudMessaging.showNotification(cloudMessage);
+
+		verify(mockNotificationBuilder, times(0)).setContentTitle(anyString());
+		verify(mockNotificationManager, times(0)).notify(anyInt(), any(Notification.class));
+	}
+}
--- a/catroid/src/main/AndroidManifest.xml
+++ b/catroid/src/main/AndroidManifest.xml
@@ -244,6 +244,13 @@
         <service android:name="com.parrot.freeflight.service.DroneControlService" />
         <service android:name=".utils.StatusBarNotificationManager$NotificationActionService" />
         <service android:name=".cast.CastService" android:exported="false"/>
+        <service
+            android:exported="false"
+            android:name=".cloudmessaging.CustomFirebaseMessagingService">
+             <intent-filter>
+                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
+             </intent-filter>
+        </service>
 
     </application>
 </manifest>
--- a/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessage.java
+++ b/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessage.java
@@ -1,0 +1,93 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.cloudmessaging;
+
+import android.support.annotation.VisibleForTesting;
+import android.webkit.URLUtil;
+
+import com.google.firebase.messaging.RemoteMessage;
+
+import java.util.HashMap;
+
+public class CloudMessage {
+
+	public static final String TITLE = "title";
+	public static final String MESSAGE = "message";
+	public static final String WEB_PAGE_URL = "link";
+
+	private String title;
+	private String message;
+	private String url;
+
+	@VisibleForTesting
+	public CloudMessage() {
+	}
+
+	@VisibleForTesting
+	public CloudMessage(HashMap<String, String> data) {
+		setNotificationTitle(data.get(TITLE));
+		setNotificationMessage(data.get(MESSAGE));
+		setNotificationUrl(data.get(WEB_PAGE_URL));
+	}
+
+	public CloudMessage(RemoteMessage remoteMessage) {
+		setNotificationTitle(remoteMessage.getNotification().getTitle());
+		setNotificationMessage(remoteMessage.getNotification().getBody());
+		setNotificationUrl(remoteMessage.getData().get(WEB_PAGE_URL));
+	}
+
+	public String getTitle() {
+		return title;
+	}
+
+	public String getMessage() {
+		return message;
+	}
+
+	public String getWebPageUrl() {
+		return url;
+	}
+
+	public void setNotificationTitle(String title) {
+		this.title = title;
+	}
+
+	public void setNotificationMessage(String message) {
+		this.message = message;
+	}
+
+	public void setNotificationUrl(String url) {
+		if (URLUtil.isHttpUrl(url) || URLUtil.isHttpsUrl(url)) {
+			this.url = url;
+		}
+	}
+
+	public boolean isValidData() {
+		return !(isStringNullOrEmpty(getTitle()) || isStringNullOrEmpty(getMessage()) || isStringNullOrEmpty(getWebPageUrl()));
+	}
+
+	private boolean isStringNullOrEmpty(String string) {
+		return string == null || string.isEmpty();
+	}
+}
--- a/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
+++ b/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
@@ -1,0 +1,71 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.cloudmessaging;
+
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.media.RingtoneManager;
+import android.net.Uri;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.utils.StatusBarNotificationManager;
+
+public class CloudMessaging {
+
+	private final Context context;
+	private final NotificationBuilderProvider notificationBuilderProvider;
+	private final NotificationManager notificationManager;
+
+	public CloudMessaging(Context context, NotificationBuilderProvider notificationBuilderProvider,
+			NotificationManager notificationManager) {
+		this.context = context;
+		this.notificationBuilderProvider = notificationBuilderProvider;
+		this.notificationManager = notificationManager;
+	}
+
+	public void showNotification(CloudMessage cloudMessage) {
+		if (cloudMessage.isValidData()) {
+			Notification notification = buildNotification(cloudMessage);
+			notificationManager.notify(StatusBarNotificationManager.getInstance().getUniqueNotificationId(), notification);
+		}
+	}
+
+	private Notification buildNotification(CloudMessage cloudMessage) {
+		Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(cloudMessage.getWebPageUrl()));
+		intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+		PendingIntent pendingIntent = PendingIntent.getActivity(context, 0, intent, PendingIntent.FLAG_ONE_SHOT);
+		Uri defaultSoundUri = RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION);
+		return notificationBuilderProvider.get()
+				.setSmallIcon(R.drawable.ic_launcher)
+				.setContentTitle(cloudMessage.getTitle())
+				.setContentText(cloudMessage.getMessage())
+				.setAutoCancel(true)
+				.setSound(defaultSoundUri)
+				.setContentIntent(pendingIntent)
+				.build();
+	}
+}
--- a/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CustomFirebaseMessagingService.java
+++ b/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CustomFirebaseMessagingService.java
@@ -1,0 +1,41 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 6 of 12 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseInstanceIDService.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.catrobat.catroid.ui.cloudmessaging;
+
+import android.util.Log;
+
+import com.google.firebase.iid.FirebaseInstanceId;
+import com.google.firebase.iid.FirebaseInstanceIdService;
+
+public class CustomFirebaseInstanceIDService extends FirebaseInstanceIdService {
```
robertpainsi: Will this file be used in any further updates?

nishantt95: If we want to send the notification to specific users in future then yes. This token will be stored in the database under the userId. Though this was the part of the installation.

robertpainsi: > If we want to send the notification to specific users in future then yes

k. We won't send notifications to specific users :wink:

catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseMessagingService.java
```diff
@@ -0,0 +1,118 @@
...
+			URL url = new URL(strURL);
+			HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+			connection.setDoInput(true);
+			connection.connect();
+			InputStream input = connection.getInputStream();
+			return BitmapFactory.decodeStream(input);
+		} catch (IOException e) {
+			Log.e(TAG, "Error : " + e);
+			return null;
+		}
```
robertpainsi: Instead of returning null, just add IOException to the signature.
null checks are always a little nasty.

catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseMessagingService.java
```diff
@@ -0,0 +1,115 @@
...
+				.setSmallIcon(R.drawable.ic_launcher)
+				.setLargeIcon(image)
+				.setContentTitle(title)
+				.setContentText(message)
+				.setAutoCancel(true)
+				.setSound(defaultSoundUri)
+				.setContentIntent(pendingIntent);
+
+		NotificationManager notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+		notificationManager.notify(1, notification.build());
```
robertpainsi: id 1 will always overwrite the current notification.

There is a [StatusBarNotificationManager.java](https://github.com/Catrobat/Catroid/blob/be9b08440d257bdb9adbd6cf142aae5b41a8e8d0/catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java) but I'm not sure if it's well suited. It handles any kind of Notification and I'm definitely not happy with that. The id is also not thread safe and (maybe) should be saved persistently.

We will have to improve this and the StatusBarNotificationManager to work well with notifications. However, for now just keep this code and start the notification ids at 1000. No thread safety or persistence needed.

catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseMessagingService.java
```diff
@@ -0,0 +1,115 @@
...
+			URL url = new URL(link);
+			HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+			connection.setDoInput(true);
+			connection.connect();
+			InputStream input = connection.getInputStream();
+			return BitmapFactory.decodeStream(input);
+		} catch (IOException e) {
+			Log.e(TAG, "Failed to download image. Using default icon: " + e);
+			return BitmapFactory.decodeResource(this.getResources(), R.drawable.ic_launcher);
+		}
```
robertpainsi: Even better! :+1:

btw Log.e has also a version with three parameters: `Log.e(TAG, "Failed to download image. Using default icon", e);`

catroid/src/main/AndroidManifest.xml
```diff
@@ -243,6 +243,11 @@
...
         <service android:name=".utils.StatusBarNotificationManager$NotificationActionService" />
-
-    </application>
+        <service
+            android:name=".ui.cloudmessaging.CustomFirebaseMessagingService">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT"/>
+            </intent-filter>
+        </service>
+        </application>
```
robertpainsi: Please indent correctly

catroid/src/main/AndroidManifest.xml
```diff
@@ -243,6 +243,11 @@
...
         <service android:name="com.parrot.freeflight.service.DroneControlService" />
         <service android:name=".utils.StatusBarNotificationManager$NotificationActionService" />
-
-    </application>
+        <service
+            android:name=".ui.cloudmessaging.CustomFirebaseMessagingService">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT"/>
+            </intent-filter>
+        </service>
```
robertpainsi: Is it possible to somehow deactivate firebase-messaging entirely?
So that the service isn't register in the standalone version.

catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseMessagingService.java
robertpainsi: updated previous comment: "ids at 10.000 (or so)"

catroid/src/main/AndroidManifest.xml
nishantt95: Well, there are ways to subscribe and unsubscribe to the topic but not to the service itself (not available in its API). Though I can search for disabling the service (if there is one available).

robertpainsi: Thanks. I'll create a ticket for that. For now, it has low priority :wink:

Standalone versions are just apks with a single Catrobat program and they start directly in the stage (or so). So no IDE or Editor.
The apk generator will create apps (standalones) of each program. So if you've installed lots of standalone apks, they will all start-up and drain battery. They'll also always listen to FCM messages. I'm not sure if it's a big problem, but if we can fix it smoothly or easily, I would go with that. However, as said, low priority.

catroid/src/main/java/org/catrobat/catroid/ui/cloudmessaging/CustomFirebaseMessagingService.java
robertpainsi: The overall id handling in the StatusBarNotificationManager is definitely a mess. If you come up with a better solution, go with that. I'm also fine with a workaround (for now) like described above. :wink:

catroid/src/main/AndroidManifest.xml
robertpainsi: :arrow_up: this :wink: not sure why the AS formatter doesn't fix this automatically…
`</application>` should be one tab to the left

catroid/build.gradle
```diff
@@ -112,6 +114,7 @@ android {
             buildConfigField "String", "PROJECT_NAME", "\"No Standalone Project\""
             buildConfigField "boolean", "FEATURE_APK_GENERATOR_ENABLED", "false"
             buildConfigField "boolean", "CRASHLYTICS_CRASH_REPORT_ENABLED", "true"
+            buildConfigField "boolean", "FIREBASE_CLOUD_MESSAGING_ENABLED", "true"
```
robertpainsi: Please rename to `"FEATURE_CLOUD_MESSAGING_ENABLED"`
`FEATURE…` to be consistent with all other flags.
~`FIREBASE…`~ because it's not important what service we're using ^^. Previously FCM was called GCM and we would have need to update that everywhere if it would've been implemented already. It also feels more abstract :wink:

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -0,0 +1,101 @@
...
+	public Bitmap getBitmap(String link) {
+		try {
+			URL url = new URL(link);
+			HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+			connection.setDoInput(true);
+			connection.connect();
+			InputStream input = connection.getInputStream();
+			return BitmapFactory.decodeStream(input);
+		} catch (IOException e) {
+			Log.e(TAG, "Failed to download image. Using default icon: " + e);
```
robertpainsi: See https://source.android.com/source/code-style#log-sparingly. I like these logging guidelines a lot.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -115,6 +116,15 @@
 	@Override
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
+		if (getIntent().getExtras() != null) {
+			try {
+				String link = getIntent().getExtras().get("link").toString();
```
robertpainsi: `get("link").toString()` → `getString("link")`

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -115,6 +116,15 @@
...
+				startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(link)));
```
robertpainsi: https://developer.android.com/guide/components/intents-common.html#ViewUrl (see example code)
> If there are no apps on the device that can receive the implicit intent, your app will crash when it calls startActivity(). To first verify that an app exists to receive the intent, call resolveActivity() on your Intent object. If the result is non-null, there is at least one app that can handle the intent and it's safe to call startActivity(). If the result is null, you should not use the intent and, if possible, you should disable the feature that invokes the intent.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -115,6 +116,15 @@
...
+			} catch (NullPointerException e) {
```
robertpainsi: Replace the exception by an if statement :wink:

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -115,6 +116,15 @@
...
+				Log.e(TAG, "Url not found with notification", e);
+			}
+		}
```
robertpainsi: Please create a method.
___
There is a problem if the link doesn't start with http:// or https://. The app stays in an infinite loop of creating the MainMenuActivity, crashing and restarting the App.
___
Open any Activity besides the MainMenuActivity, e.g. the ScriptActivity. Move the app into the background. Send yourself a message and tap on the message, the browser opens. Press back. The MainMenuAtivity is shown. Press back again, the ScriptActivity is shown.
The MainMenuActivity shouldn't be shown when pressing back the first time.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
robertpainsi: When I find a hard-coded string during code review
Unacceptable

:stuck_out_tongue_winking_eye:

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CustomFirebaseMessagingService.java
```diff
@@ -0,0 +1,89 @@
...
+		if (!cloudMessaging.isValidData(notificationData)) {
+			return;
+		}
+
+		Bitmap imageBitmap = cloudMessaging.getBitmap(imageUrl);
+
+		sendNotification(title, message, imageBitmap, webPageUrl);
+	}
+
+	private void sendNotification(String title, String message, Bitmap image, String link) {
```
robertpainsi: Not sure why it's called `sendNotification` everywhere online, wouldn't `showNotification` be a better name?

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -116,15 +117,7 @@
...
-		if (getIntent().getExtras() != null) {
-			try {
-				String link = getIntent().getExtras().get("link").toString();
-				startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(link)));
-			} catch (NullPointerException e) {
-				Log.e(TAG, "Url not found with notification", e);
-			}
-		}
-
+		getUrlFromCloudMessage();
```
robertpainsi: `getX` without returning something isn't normally expected. I first thought about `openUrlFromCloudMessage` but that's not a good name. What do you think about `checkCloudMessage` or `handleOpenedCloudMessage`? Maybe you can come up with something better.

nishantt95: `checkCloudMessage` looks good

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CustomFirebaseMessagingService.java
```diff
@@ -0,0 +1,83 @@
...
+		NotificationCompat.Builder notification = new NotificationCompat.Builder(this)
+				.setSmallIcon(R.drawable.ic_launcher)
+				.setContentTitle(title)
+				.setContentText(message)
+				.setAutoCancel(true)
+				.setSound(defaultSoundUri)
+				.setContentIntent(pendingIntent);
+
+		NotificationManager notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+		notificationManager.notify(0, notification.build());
```
robertpainsi: The [StatusBarNotificationManager](https://github.com/Catrobat/Catroid/blob/da79ee54c412e73e98f005add7691831b77ff31b/catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java#L53) currently handles all notification. We won't add it there.
However, the StatusBarNotificationManager also handles the notification ids. Something like https://stackoverflow.com/a/25713229/2774399 should be sufficient for us (no class needed, just a function)

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -123,10 +126,12 @@ protected void onCreate(Bundle savedInstanceState) {
 		PreferenceManager.setDefaultValues(this, R.xml.preferences, true);
 		UtilUi.updateScreenWidthAndHeight(this);
 
+		Log.e(TAG, "Token : " + FirebaseInstanceId.getInstance().getToken());
+
```
robertpainsi: I think it definitely comes handy to log the token. Just as verbose, add `Firebase` to the text and move it to the CatroidApplication.create method.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CustomFirebaseMessagingService.java
robertpainsi: Ah, sorry. This may have been a little bit confusing.

Add the `getId` (camelcase) method to the StatusBarNotificationManager (replace notificationId).

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -79,7 +81,7 @@ public int createUploadNotification(Context context, String programName) {
 		Intent uploadIntent = new Intent(context, MainMenuActivity.class);
 		uploadIntent.setAction(Intent.ACTION_MAIN);
 		uploadIntent = uploadIntent.setFlags(Intent.FLAG_ACTIVITY_RESET_TASK_IF_NEEDED);
-		PendingIntent pendingIntent = PendingIntent.getActivity(context, notificationId, uploadIntent,
+		PendingIntent pendingIntent = PendingIntent.getActivity(context, getId(), uploadIntent,
```
robertpainsi: `getId() ≠ notificationId`, `notificationId = ATOMIC_INTEGER.get()`
Always be careful not to change behavior.

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -50,7 +52,7 @@
 
 	private static final StatusBarNotificationManager INSTANCE = new StatusBarNotificationManager();
 
-	private int notificationId;
+	private static final AtomicInteger ATOMIC_INTEGER = new AtomicInteger(0);
```
robertpainsi: Please rename.

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -331,4 +334,8 @@ private void closeNotificationBar() {
 			sendBroadcast(it);
 		}
 	}
+
+	public static int getId() {
+		return ATOMIC_INTEGER.incrementAndGet();
+	}
```
robertpainsi: I can't comment on line #150 (`return notificationId++;`) on GitHub. You've to use 
```java
notificationId = getId();
return notificationId;
```
on line #150.

Please rename the member `ATOMIC_INTEGER`. Maybe `UNIQUE_NOTIFICATION_ID_COUNTER` or something like that. And the method `getId` to `getUniqueNotificationId` or so.

nishantt95:
```
data.setNotificationBuilder(notificationBuilder);
notificationDataMap.put(notificationId, data);

return notificationId;
```
This way?

robertpainsi:
```java
notificationId++
```
translates to
```java
notificationId = getUniqueNotificationId();
return notificationId;
```

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -126,7 +128,7 @@ protected void onCreate(Bundle savedInstanceState) {
 		if (STANDALONE_MODE) {
 			/*requestWindowFeature(Window.FEATURE_NO_TITLE);
 			getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN,
-				WindowManager.LayoutParams.FLAG_FULLSCREEN);*/
+			WindowManager.LayoutParams.FLAG_FULLSCREEN);*/
```
robertpainsi: undo

catroid/build.gradle
```diff
@@ -189,6 +190,9 @@ dependencies {
     androidTestCompile ('com.android.support.test.espresso:espresso-contrib:2.2.2') {
         exclude group: 'com.android.support'
     }
+    androidTestCompile ('com.android.support.test.espresso:espresso-intents:2.2'){
```
robertpainsi: missing space between `){`

catroid/build.gradle
```diff
@@ -189,6 +190,9 @@ dependencies {
     androidTestCompile ('com.android.support.test.espresso:espresso-contrib:2.2.2') {
         exclude group: 'com.android.support'
     }
+    androidTestCompile('com.android.support.test.espresso:espresso-intents:2.2') {
```
robertpainsi: Now the space is missing between `androidTestCompile(`…

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -49,8 +51,9 @@
 	public static final String ACTION_CANCEL_UPLOAD = "cancel_upload";
 
 	private static final StatusBarNotificationManager INSTANCE = new StatusBarNotificationManager();
+	private static AtomicInteger uniqueNotificationIdCounter = new AtomicInteger(0);
```
robertpainsi: The `StatusBarNotificationManager` is a Singleton. Let's us it that way. Remove static here (adapt code), readd final.

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -331,4 +334,11 @@ private void closeNotificationBar() {
 			sendBroadcast(it);
 		}
 	}
+
+	public static int getUniqueNotificationId() {
+		if (uniqueNotificationIdCounter == null) {
+			uniqueNotificationIdCounter = new AtomicInteger(0);
+		}
```
robertpainsi: Why is this if needed?

catroid/build.gradle
nishantt95: Oh. Though I did select it completely and then hit (Ctrl + Alt + L) and still missed.

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
nishantt95: As I was calling `getUniqueNotificationId()` in `CustomFirebaseMessagingService `class then:-

When application was in background - Everything worked perfectly

When application was in foreground - The application was getting stopped unexpectedly (on receiving notification) without any error message but with a log [System.exit() is called with status 10]. After a while I realized that it was because of calling `getUniqueNotificationId()`. I did some changes and everything worked well.

 Seems like `uniqueNotificationIdCounter` wasn't getting initialized on directly calling static method  `getUniqueNotificationId()`.

cheidenreich: since it's not a static method anymore with the latest changes, it might be removed again, am i right?

nishantt95: removed again? why and how? 😕

robertpainsi: https://github.com/Catrobat/Catroid/pull/2379#discussion_r128917046 includes to remove static from getUniqueNotificationId as well
Instead of calling `StatusBarNotificationManager.getUniqueNotificationId()` you use
`StatusBarNotificationManager.getInstance().getUniqueNotificationId()`
Let me know if you have further questions.

nishantt95: Yes, I know that the method has been changed. Still why and how it will be removed?

robertpainsi: Hmm, we're just talking about the if block, not the whole method. Does this answer your questions?

```java
public int getUniqueNotificationId() {
	return uniqueNotificationIdCounter.incrementAndGet();
}
```

nishantt95: > Hmm, we're just talking about the if block, not the whole method. Does this answer your questions?

@robertpainsi  Oh, Ok 👍 

> since it's not a static method anymore with the latest changes, it might be removed again, am i right?

@cheidenreich  Yes

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.cloudmessaging;
+
+import android.test.AndroidTestCase;
+
+import org.catrobat.catroid.cloudmessaging.CloudMessage;
+
+public class CloudMessagingTest extends AndroidTestCase {
```
robertpainsi: `CloudMessageTest`

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,94 @@
...
+		assertFalse(cloudMessage.isValidData());
+
+		cloudMessage.setNotificationTitle("Test Notification Title");
+
+		assertFalse(cloudMessage.isValidData());
+
+		cloudMessage.setNotificationMessage("Test Notification Message");
+		cloudMessage.setNotificationUrl("https://www.catrobat.org/");
+
+		assertTrue(cloudMessage.isValidData());
```
robertpainsi: Actually, you don't need to "fix" the cloud message. The test just needs to test that the CloudMessage is invalid if required fields are missing. Please adapt everywhere.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,94 @@
...
+		cloudMessage.setNotificationTitle("");
+
+		assertFalse(cloudMessage.isValidData());
+
+		cloudMessage.setNotificationTitle("Test Notification Title");
+		cloudMessage.setNotificationMessage("Test Notification Message");
+		cloudMessage.setNotificationUrl("https://www.catrobat.org/");
+
+		assertTrue(cloudMessage.isValidData());
+	}
```
robertpainsi: This test may also only test if a required field is missing. Please fix

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+	public void testRequiredFieldIsInValid() {
```
robertpainsi: `testNotificationUrlIsInvalid`

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.mockito.Mockito;
+
+import java.util.HashMap;
+
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class CustomFirebaseMessagingServiceTest {
```
robertpainsi: `CloudMessagingTest`

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+		data.put(CloudMessage.WEB_PAGE_URL, "http://TEST.COM");
+
+		Context context = InstrumentationRegistry.getTargetContext();
+		CloudMessage cloudMessage = new CloudMessage(data);
+
+		CloudMessaging cloudMessaging = new CloudMessaging(context, mockNotificationBuilderProvider, mockNotificationManager);
+		cloudMessaging.showNotification(cloudMessage);
+
+		verify(mockNotificationBuilder).setContentTitle(eq("Title"));
+		verify(mockNotificationBuilder).setContentText(eq("Message"));
```
robertpainsi: icon, sound, etc are missing.
If the webPageUrl isn't testable, test that setContentIntent is called with a non-null value.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+		verify(mockNotificationManager).notify(anyInt(), eq(mockNotification));
+	}
+}
```
robertpainsi: Also test that _no_ notification is shown if the CloudMessage is invalid.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessage.java
```diff
@@ -0,0 +1,93 @@
...
+	private String message;
+	private String url;
+
+	@VisibleForTesting
+	public CloudMessage() {
+
+	}
+
+	@VisibleForTesting
+	public CloudMessage(HashMap<String, String> data) {
```
robertpainsi: Please remove this constructor

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -331,4 +334,11 @@ private void closeNotificationBar() {
 			sendBroadcast(it);
 		}
 	}
+
+	public int getUniqueNotificationId() {
+		if (uniqueNotificationIdCounter == null) {
+			uniqueNotificationIdCounter = new AtomicInteger(0);
+		}
```
robertpainsi: Is the if block still needed?

nishantt95: Yes. As earlier, it was returning some exception and directly using `uniqueNotificationIdCounter = new AtomicInteger(0);` will always return 1.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessage.java
nishantt95: Required [here](https://github.com/nishantt95/Catroid/blob/20a3eaab1f048d71e6afb6d5222978d9a8b0e6a1/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java#L71-L74) or else I need to set the data one by one. This will help when the number of parameters will grow in future.

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
robertpainsi: The method isn't static anymore. Maybe it has been resolved by using `StatusBarNotificationManager.getInstance().getUniqueNotificationId()`? Please give it another test.

robertpainsi: Thanks for retesting!

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -546,6 +548,21 @@ private void startStageProject() {
...
+			Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(link));
+			if (isValidUrl(link) && intent.resolveActivity(getPackageManager()) != null) {
+				startActivity(intent);
+				finish();
+			}
+		}
+	}
+
+	private boolean isValidUrl(String url) {
+		return url.startsWith("http://") || url.startsWith("https://");
```
robertpainsi: Please use the `URLUtil` methods here as well.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
robertpainsi: Just `setSmallIcon` and `setSound` is missing imo

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+	@Test
+	public void testNotificationIsNotShownOnRequiredFieldMissing() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "http://TEST.COM");
+
+		commonTestForInValidData(data);
+	}
+
+	private void commonTestForInValidData(HashMap<String, String> data) {
```
robertpainsi: What do you think about `checkNotificationNotShownForInvalidData` or so?
btw, In**v**alid

robertpainsi:
test this please

robertpainsi:
* 3 Checkstyle warnings
* 2 Lint issues
* 6 failing tests
  * WebSocketClientTest.testAsyncConnectAndAuthenticateWhetherWebSocketObjectSettersAreCalledCorrectly
  * WebSocketClientTest.testReceivedOnBaseMessageEventWithErrorMessageWhenClientIsConnectedAndAuthenticated
  * StatusBarNotificationManagerTest.testCreateCopyNotification
  * StatusBarNotificationManagerTest.testCreateDownloadNotification
  * StatusBarNotificationManagerTest.testCreateUploadNotification
  * StatusBarNotificationManagerTest.testUploadRejectedNotification

nishantt95:
WebSocketClientTest.testAsyncConnectAndAuthenticateWhetherWebSocketObjectSettersAreCalledCorrectly
WebSocketClientTest.testReceivedOnBaseMessageEventWithErrorMessageWhenClientIsConnectedAndAuthenticated

Are still there and I am working on it. But the strange thing is, I am not getting any checkstyle warning.

robertpainsi:
Here you go
```
[ant:checkstyle] [WARN] /home/robert/Projects/Catrobat/Catroid/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessageTest.java:65: Next line is an unnecessary newline [RegexpMultiline]
[ant:checkstyle] [WARN] /home/robert/Projects/Catrobat/Catroid/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessageTest.java:75: Next line is an unnecessary newline [RegexpMultiline]
[ant:checkstyle] [WARN] /home/robert/Projects/Catrobat/Catroid/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessage.java:44: Next line is an unnecessary newline [RegexpMultiline]
```

robertpainsi:
Test this please

robertpainsi:
Test this please

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+		cloudMessaging.showNotification(cloudMessage);
+
+		verify(mockNotificationBuilder).setContentTitle(eq("Title"));
+		verify(mockNotificationBuilder).setContentText(eq("Message"));
+		verify(mockNotificationBuilder).setContentIntent(any(PendingIntent.class));
+		verify(mockNotificationManager).notify(anyInt(), eq(mockNotification));
+	}
+
+	@Test
+	public void testNotificationIsNotShownOnInValidUrl() {
```
robertpainsi: `…Invalid…`

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CustomFirebaseMessagingServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+		checkNotificationNotShownForInvalidData(data);
+	}
+
+	@Test
+	public void testNotificationIsNotShownOnRequiredFieldEmpty() {
+		HashMap<String, String> data = new HashMap<>();
+		data.put(CloudMessage.TITLE, "");
+		data.put(CloudMessage.MESSAGE, "Message");
+		data.put(CloudMessage.WEB_PAGE_URL, "http://TEST.COM");
```
robertpainsi: Please use a more normal looking address. e.g. http://test.com https://www.catrobat.org

nishantt95:
> setSmallIcon and setSound is missing imo

Working on it. Please don't test

robertpainsi:
Test this please

PR #2380, opened 2017-07-05
CAT-2323 Fixed Backpack-Unpacking bug

alt-13:
See
https://jira.catrob.at/browse/CAT-2323
for detailed description

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
@@ -45,9 +45,9 @@ public abstract class FormulaBrick extends BrickBaseType implements View.OnClick
 	ConcurrentFormulaHashMap formulaMap;
 
 	@XStreamOmitField
-	private List<BackPackedListData> backPackedListData;
+	private List<BackPackedListData> backPackedListData = new ArrayList<>();
 	@XStreamOmitField
-	private List<BackPackedVariableData> backPackedVariableData;
+	private List<BackPackedVariableData> backPackedVariableData = new ArrayList<>();
 
 	public Formula getFormulaWithBrickField(BrickField brickField) throws IllegalArgumentException {
 		if (formulaMap != null && formulaMap.containsKey(brickField)) {
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoEv3MotorMoveBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoEv3MotorMoveBrick.java
@@ -106,11 +106,6 @@ public class LegoEv3MotorMoveBrick extends FormulaBrick {
 	}
 
 	@Override
-	public Brick clone() {
-		return new LegoEv3MotorMoveBrick(motorEnum, getFormulaWithBrickField(BrickField.LEGO_EV3_SPEED).clone());
-	}
-
-	@Override
 	public void showFormulaEditorToEditFormula(View view) {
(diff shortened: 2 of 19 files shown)


mfuchs:
Please ignore me rerunning the tests, this is just to check some Jenkins behavior.
test this please

catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoEv3MotorMoveBrick.java
```diff
@@ -151,8 +147,11 @@ public void onItemSelected(AdapterView<?> arg0, View arg1, int position, long ar
 			public void onNothingSelected(AdapterView<?> arg0) {
 			}
 		});
-
+		if (motorEnum == null) {
+			readResolve();
+		}
 		motorSpinner.setSelection(motorEnum.ordinal());
+		motorSpinner.setGravity(Gravity.CENTER);
```
alt-13: It does not have to be done in code but I wanted to add it so all LegoBricks look alike. This was the only brick that did not set the gravity of the spinner to center.
Option one: change all xmls of the bricks with spinners or
Option two: add one line in one file to make them all look alike.
I guess you would have liked option one. If you insist on it to be done that way let me know.

alt-13: That was only an attempt to increase the uniformity but as I recognise now: there is no uniform way spinners are shown anyways across all bricks.
If you like I could remove this line or even all these setGravity lines from the other bricks.

cheidenreich: can you do this in the layout XML files?
I don't know if the XML files will change due to refactoring, but definitely not in the first iteration.
I see potential problems if part of the appearance is defined in the XML and part of it in code.

cheidenreich: or did i miss your point?

alt-13: the point is: should I change these lines in all affected brick's xmls?

cheidenreich: yes please,
we should keep layout definitions in the XMLs.

thanks

alt-13: done

catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
```diff
@@ -162,10 +162,10 @@ public void storeDataForBackPack(Sprite sprite) {
 	}
 
 	public List<BackPackedListData> getBackPackedListData() {
-		return backPackedListData;
+		return backPackedListData != null ? backPackedListData : new ArrayList<BackPackedListData>();
```
thmq: It would be better to initialise the member variable at the point of declaration (i.e. 

``` List<BackpackedListData> backPackedListData = new ArrayList<>(); ```

 instead of checking here. you can still overwrite the list at other points and you save yourself this (unnecessary) check.

alt-13: Hopefully it is ok now!

cheidenreich:
please run espresso tests

PR #2383, opened 2017-07-05
CAT-2476 Fix CloneBrick grouping functionality

sjaindl:
As this fix is not only relevant to feature/NOLB this fix should be merged into develop, too.
see #2374

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/content/SceneTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/content/SceneTest.java
@@ -27,8 +27,12 @@ import android.test.AndroidTestCase;
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.content.Project;
 import org.catrobat.catroid.content.Scene;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.SingleSprite;
+import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.content.StartScript;
 import org.catrobat.catroid.content.bricks.AddItemToUserListBrick;
+import org.catrobat.catroid.content.bricks.CloneBrick;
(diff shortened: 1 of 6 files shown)


cheidenreich:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/BrickCloneTest.java
```diff
@@ -82,11 +85,40 @@
...
+	private Sprite secondSprite;
 
 	@Override
 	protected void setUp() throws Exception {
 		super.setUp();
 		sprite = new SingleSprite("testSprite");
+		secondSprite = new SingleSprite("secondSprite");
+	}
+
+	public void testSameSpriteReferenceAfterCloning() {
```
cheidenreich: in my opinion this doesn't really test the clone brick

sjaindl: > please run espresso tests
Is there already a Jenkins job for running the espresso tests that I can start?

sjaindl: Well, I think it doesn't specifically test the clone brick .. but this problem isn't only relevant to the clone brick (also to the PointToBrick - that's also why I've made an interface for these bricks). It tests whether the method refreshSpriteReferences() really does what it should do. It should set the correct reference to a sprite in these bricks (e.g. after shallow cloning when there is still the old sprite reference). 
That was my intention for this test ;)

sjaindl: But maybe I should move it to another test class..

sjaindl:
ok to test

cheidenreich:
please run espresso tests

PR #2384, opened 2017-07-05
CAT-FIX Add dedicated PullRequestTrigger test suite

84n4n4:
that excludes upcoming sensor hardware test box tests, and other
tests that require additional hardware or permission action by user.


84n4n4:
hi!

so, if this is merged, can someone please change the config for CatroidTriggerPullRequestEspressoTest to run this test suite (PullRequestTriggerSuite) instead of AllEspressoTestsSuite.

Thanks!

PR #2386, opened 2017-07-10
CAT-2327: adds new SetLookByIndexBrick

comeclarity:
adds the functionality to set a look by index (via formula)
same for setBackground and setBackgroundAndWait


catroid/src/androidTest/java/org/catrobat/catroid/test/content/actions/SetLookByIndexActionTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+		ActionFactory factory = sprite.getActionFactory();
+		Action action = factory.createSetLookByIndexAction(sprite, formula);
+		action.act(1.0f);
+		assertEquals("Action didn't set the first LookData", firstLookData, sprite.look.getLookData());
+
+		formula = new Formula(2);
+		action = factory.createSetLookByIndexAction(sprite, formula);
+		action.act(1.0f);
+		assertEquals("Action didn't set the second LookData", secondLookData, sprite.look.getLookData());
```
cheidenreich: we might also want to test formula values outside the looks range here

catroid/src/main/java/org/catrobat/catroid/content/actions/SetLookByIndexAction.java
```diff
@@ -0,0 +1,66 @@
...
+public class SetLookByIndexAction extends SetLookAction {
+
+	private Formula formula;
+
+	@Override
+	protected void doLookUpdate() {
+
+		updateLookFromFormula();
+		if (wait) {
+			BackgroundWaitHandler.addObserver(look, this);
```
cheidenreich: should this really be done before the check below?
what happens if the observer gets added, but the `setLookData()` never gets called?

cheidenreich:
this introduces new bricks, isn't a language version increment is required?

cheidenreich:
please run espresso tests

comeclarity:
Thanks for the feedback. I updated the pull request.

cheidenreich:
please run espresso tests

PR #2400, opened 2017-07-17
CAT-FIX Set Crashlytics flag to True in debug mode

nishantt95:
Earlier the flag CRASHLYTICS_CRASH_REPORT_ENABLED was set to false in
debug mode. In SettingsActivity the checkbox button (which
enables/disables the Crashlytics crash reporting) is either visible or
hidden depending upon the flag value (visible if true and hidden if
false).

As the checkbox was hidden (hidden for false), it resulted in failure of
different espresso test cases where checkbox was required to perform the
UI test. Moreover, espresso testrun is green because the auto-generated
AllEspressoTestsSuite.java file isn't up-to-date and doesn't contain the
SettingsActivityTest.


robertpainsi:
please run espresso tests

robertpainsi:
please test this

robertpainsi:
test this please

PR #2401, opened 2017-07-17
CAT-FIX Fix infinite loop when app crashes

robertpainsi:
When the app crashes, the RECOVERED_FROM_CRASH flag has to be set to
true and then every Activity is finished until the app is back in the
MainMenuActivity. If the RECOVERED_FROM_CRASH flag is not set, the app
will go into an infinite loop, where it tries to start an activity over
and over again because the activity crashes but never finises. The
problem was only reproducible in specific activities, e.g. when throwing
an exception in ScriptActivity.create.

Setting the RECOVERED_FROM_CRASH flag to true when the app crashes fixes
the infinite loop. It has been unintentionally removed during improving
the Crash Reporter[1].

[1] https://github.com/Catrobat/Catroid/pull/2366/files#diff-d61e5ef6cdb376657aae60af754655d0L136


cheidenreich:
👍 thanks for the quick fix

PR #2414, opened 2017-07-24
NOLB Move strings to "main-flavor"

sjaindl:
Move strings to "main-flavor" and adapt lint checks


sjaindl:
+ add new template strings

PR #2416, opened 2017-07-25
Add Phiro-Flavor

sjaindl:


catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+	private void checkBrickInProject() {
+		Project project = ProjectManager.getInstance().getCurrentProject();
+		Script script = project.getDefaultScene().getSpriteList().get(0).getScript(0);
+		Brick brick = script.getBrick(script.getBrickList().size() - 1);
+		assertTrue("Brick has been deleted!", brick instanceof NoteBrick);
+	}
+
+	private void removePhiroPrograms() {
+		for (String project : PhiroMainMenuActivity.phiroProjects) {
```
84n4n4: there should never be any programs existing in this state (before a test is started, when setup is called) why is this necessary?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+		removePhiroPrograms();
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@Test
+	public void phiroProgramsInitiallyLoadedTest() {
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+
+		ViewInteraction viewInteraction = onView(withId(android.R.id.list));
```
84n4n4: please dont call this "viewInteraction" because pretty much everything in espresso is either a view / data Interaction and this does not help the reader in any way.

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
```diff
@@ -0,0 +1,84 @@
...
+			@Override
+			public void describeTo(Description description) {
+				description.appendText("with index: ");
+				description.appendValue(index);
+				matcher.describeTo(description);
+			}
+
+			@Override
+			public boolean matchesSafely(View view) {
+				return matcher.matches(view) && currentIndex++ == index;
```
84n4n4: what is the "++" good for? (also, currentIndex is first compared with index and incemented AFTERWARDS, so its not doing anything here (according to java guru cheidenreich))

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
```diff
@@ -0,0 +1,84 @@
...
+import android.widget.Adapter;
+import android.widget.AdapterView;
+
+import org.catrobat.catroid.common.ProjectData;
+import org.catrobat.catroid.ui.adapter.ProjectListAdapter;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+public final class AdapterViewMatchers {
```
84n4n4: this name is so broad that it does not add any information or hint on what this is exactly good for.

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
```diff
@@ -0,0 +1,84 @@
...
+	private AdapterViewMatchers() {
+		throw new AssertionError();
+	}
+
+	public static Matcher<View> withAdaptedDataAt(final String expectedContent, final int listPosition) {
```
84n4n4: calling a view matcher withAdapted**Data**At is not a good idea, since in espresso we got view- and data- Matchers/Actions/etc, and those are different things

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
```diff
@@ -0,0 +1,84 @@
...
+
+			@Override
+			public boolean matchesSafely(View view) {
+				if (!(view instanceof AdapterView)) {
+					return false;
+				}
+
+				Adapter adapter = ((AdapterView) view).getAdapter();
+				String listItem;
+				if (adapter instanceof ProjectListAdapter) {
```
84n4n4: i have the feeling that all of this shouldnt be done with a custom view matcher but with onData functionality of espresso in the first place.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+		onView(withId(android.R.id.home))
+				.perform(click());
+
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+
+		checkBrickInProject();
+	}
+
+	private void addBrickToProject(String projectName) {
```
84n4n4: what is this good for in the first place?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+
+		onView(withId(android.R.id.list))
+				.check(matches(withAdaptedDataAt(PhiroMainMenuActivity.phiroProjects[0], 0)));
+	}
+
+	@Test
+	public void phiroProgramsUnchangedWhenModifiedTest() {
```
84n4n4: write a separate test class that sets up a project, and then checks if it is properly displayed, and add/remove bricks there.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+
+		checkBrickInProject();
+	}
+
+	private void addBrickToProject(String projectName) {
+		try {
+			ProjectManager.getInstance().loadProject(projectName, InstrumentationRegistry.getTargetContext());
+		} catch (LoadingProjectException e) {
```
84n4n4: why catch exceptions in a test without having it fail?!?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+			Log.e(TAG, "project version outdated: " + e.getMessage());
+		} catch (CompatibilityProjectException e) {
+			Log.e(TAG, "project not compatible: " + e.getMessage());
+		}
+
+		Project project = ProjectManager.getInstance().getCurrentProject();
+		project.getDefaultScene().getSpriteList().get(0).getScript(0).addBrick(new NoteBrick("I was added!"));
+	}
+
+	private void checkBrickInProject() {
```
84n4n4: please rename this to provide some info on what this is actually doing

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+		Project project = ProjectManager.getInstance().getCurrentProject();
+		Script script = project.getDefaultScene().getSpriteList().get(0).getScript(0);
+		Brick brick = script.getBrick(script.getBrickList().size() - 1);
+		assertTrue("Brick has been deleted!", brick instanceof NoteBrick);
```
84n4n4: what is that assertion message supposed to tell me?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
```diff
@@ -0,0 +1,119 @@
...
+		ProjectManager.getInstance().setProject(project);
+		ProjectManager.getInstance().setCurrentSprite(sprite);
+
+		return project;
+	}
+
+	private void preparePhiroSettings() {
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+
+		SharedPreferences.Editor editor = sharedPreferences.edit();
```
84n4n4: when changing shared preferences in a test you should reset them to what they were before the testrun in a teardown method.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
```diff
@@ -0,0 +1,119 @@
...
+				.perform(click());
+
+		onView(withId(R.id.formula_editor_keyboard_sensors))
+				.perform(click());
+
+		onView(withIndex(withId(android.R.id.list), 1))
+				.check(matches(withAdaptedDataAt("phiro_front_left_sensor", 0)));
+	}
+
+	public static Project createProject(String projectName) {
```
84n4n4: is there a reason for "public static Project" signature? if not so please make private void

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+	public void phiroProgramsInitiallyLoadedTest() {
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+
+		ViewInteraction viewInteraction = onView(withId(android.R.id.list));
+
+		Integer position = 0;
+		for (String program : PhiroMainMenuActivity.phiroProjects) {
+			viewInteraction
+					.check(matches(withAdaptedDataAt(program, ++position)));
```
84n4n4: please add a separate line for incrementing the position.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	@Test
+	public void phiroProgramsRestoredWhenDeletedTest() {
```
84n4n4: is this test necessary?
shouldnt it be enough to check that projects are generated on program start?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+	@Test
+	public void phiroIsFirstInFormulaEditorDeviceCategoryTest() {
+		onView(withId(R.id.brick_set_x_edit_text))
+				.perform(click());
+
+		onView(withId(R.id.formula_editor_keyboard_sensors))
+				.perform(click());
+
+		onView(withIndex(withId(android.R.id.list), 1))
```
84n4n4: please use onData

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+	@Before
+	public void setUp() throws Exception {
+		SharedPreferences sharedPreferences = PreferenceManager
+				.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+
+		for (String setting : allSettings) {
+			initialSettings.put(setting, sharedPreferences.getBoolean(setting, false));
+		}
+		initialSettings.put(SETTINGS_SHOW_PHIRO_BRICKS, sharedPreferences.getBoolean(SETTINGS_SHOW_PHIRO_BRICKS, false));
```
84n4n4: please include SETTINGS_SHOW_PHIRO_BRICKS in allSettings to make this work as intended (and also to have the teardown work as expected, also deactivating SETTINGS_SHOW_PHIRO_BRICKS again!!!)

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,172 @@
...
+		for (String setting : initialSettings.keySet()) {
+			sharedPreferencesEditor.putBoolean(setting, initialSettings.get(setting));
+		}
+		sharedPreferencesEditor.apply();
+		initialSettings.clear();
+	}
+
+	@Test
+	public void basicSettingsTest() {
+		checkPreference(R.string.preference_title_enable_arduino_bricks, SETTINGS_SHOW_ARDUINO_BRICKS);
```
84n4n4: this is already tested in SettingsActivityTest, is there any point in duplicating this test?

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: In the case of Phiro there may be programs:
There are programs (placed in .zip-Files in the assets-folder) that are extracted when MyProjectsActivity is opened. So this is necessary to ensure that no programs exist in the first place.. It could be done in in teardown, alternatively

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -375,16 +387,10 @@ public void handleWebButton(View view) {
...
-			SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
-			String username = sharedPreferences.getString(Constants.USERNAME, Constants.NO_USERNAME);
-			String token = sharedPreferences.getString(Constants.TOKEN, Constants.NO_TOKEN);
 
-			String url = Constants.CATROBAT_TOKEN_LOGIN_URL + username + Constants.CATROBAT_TOKEN_LOGIN_AMP_TOKEN + token;
-			startWebViewActivity(url);
-		} else {
-			startWebViewActivity(Constants.BASE_URL_HTTPS);
-		}
+		String url = Constants.FLAVORED_BASE_URL_HTTPS;
```
84n4n4: isnt this going to break the webview login if above code is removed?

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
sjaindl: I'm not quite sure about when exactly matchesSafely is called (you are the Espresso Guru :)), but it's a member variable. So I think it might be to iterate through a list, for example. In the first pass it's 0, then 1 etc. until it matches..
 (I've taken and adapted that matcher from https://stackoverflow.com/questions/29378552/in-espresso-how-to-choose-one-the-view-who-have-same-id-to-avoid-ambiguousviewm)
So, I'm not quite sure, but I'll try without incrementing whether the test passes

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
sjaindl: I've moved that code to a Utils function, as this is used in other places, too.
--> This is now done in: Utils.addUsernameAndTokenInfoToUrl(..)

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
sjaindl: That's intentionally. I want to test whether the setting is initially activated - So I don't want it to be set to true in setAllSettingsTo(true) .. but true, it should be restored in teardown, nevertheless. I'll add that!

84n4n4: if the test is about checking out the default / initial state, why touch shared preferences in the first place? 
also this should be verified via UI, since you have to give a fallback when asking sharedPreferences, and if you do sharedPreferences.getBoolean(SETTINGS_SHOW_PHIRO_BRICKS, false) in your test, doesnt mean that someone might have used sharedPreferences.getBoolean(SETTINGS_SHOW_PHIRO_BRICKS, true) in production code, (would make a difference if run for the first time and key isnt there yet)

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
84n4n4: but you do have an adapter view, and should use onData in the first place, instead of writing a matcher that does something that espresso can do for you out of the box.
the stackoverflow answer has some stuff not in a adapterView and uses this approach because of this.

write a matcher that ensures you got the right list on your hands similar to this one here:
https://github.com/Catrobat/Catroid/blob/develop/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/matchers/BrickCategoryListMatchers.java

and then use it like this:

onData(allOf(is(instanceOf(String.class)), is(UiTestUtils.getResourcesString(categoryStringResource))))
					.inAdapterView(BrickCategoryListMatchers.isBrickCategoryView())

found in https://github.com/Catrobat/Catroid/blob/develop/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/BrickValueParameterTest.java

or even add a wrapper that calls above chunk of code and gives you a data interaction like this:
public static BrickDataInteractionWrapper onBrickAtPosition(int position) {
		return new BrickDataInteractionWrapper(onData(instanceOf(Brick.class))
				.inAdapterView(ScriptListMatchers.isScriptListView())
				.atPosition(position));
	}

https://github.com/Catrobat/Catroid/blob/develop/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickDataInteractionWrapper.java

and when using espresso functionality instead of your own matcher that does "atPosition", "inAdapterView", "onChildView", etc...


and then

84n4n4:
btw, whats up with the hundreds of filepath changes? 

Catroid/catroid/src/androidTest/java/org/catrobat/catroid/... 
  to
Catroid/catroid/src/androidTest/java/org.catrobat.catroid/...

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
sjaindl: Thanks for your suggestions, they will be useful - I'm currently rewriting it to use onData(..)

84n4n4:
had a further look into the files that had their paths changed, was this intentional:
catroid/src/**androidTest**/assets/Air_fight_0.5.catrobat 
→ 
catroid/src/**androidTestCatroid**/assets/Air_fight_0.5.catrobat

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,172 @@
...
+				.atPosition(0)
+				.check(matches(isDisplayed()));
+	}
+
+	@Test
+	public void phiroSettingsTest() {
+		onData(PreferenceMatchers.withTitle(R.string.preference_title_enable_phiro_bricks)).perform(click());
+		checkPreference(R.string.preference_title_enable_phiro_bricks, SETTINGS_SHOW_PHIRO_BRICKS);
+
+		String phiroLink = InstrumentationRegistry.getTargetContext().getString(R.string.phiro_preference_link);
```
84n4n4: theres a 
public static String getResourcesString(int stringId) 

in espresso UiTestUtils that does this.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FragmentFactory.java
```diff
@@ -0,0 +1,64 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.ui.fragment;
+
+import android.app.Fragment;
+import android.content.Context;
+
+final class FragmentFactory {
```
84n4n4: very general name if this is only meant to provide BrickCategoryFragment and FormulaEditorCategoryListFragment.
although its named factory, but its not really analog to a GOF factory pattern.

wouldnt it be nicer to use static factory methods in those fragments, that provide the right implementation when called (http://www.androiddesignpatterns.com/2012/05/using-newinstance-to-instantiate.html)

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FragmentFactory.java
```diff
@@ -0,0 +1,64 @@
...
+		return (FormulaEditorCategoryListFragment) FragmentFactory.createFragment(fragmentName, context);
+	}
+
+	static BrickCategoryFragment createBrickCategoryFragment(Context context) {
+		String fragmentName = BrickCategoryFragment.flavoredTag;
+		return (BrickCategoryFragment) FragmentFactory.createFragment(fragmentName, context);
+	}
+
+	private static Fragment createFragment(String className, Context context) {
+		String classNameWithPackage = context.getPackageName() + "." + className;
```
84n4n4: i dont know a lot about the whole flavoring thing, but is this really supposed to be done like this? (like building the full class and package as string and the loading it that way)

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FragmentFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+		if (!classNameExists(classNameWithPackage, context)) {
```
84n4n4: i see this rather problematic, this could make bugs really hard to find, also this is something that shouldnt happen if everything is correct in the first place, right?
if a classname / packagename would be malformed or not existing, this would still make sure that everything compiles and runs, but is this really what we want? (id rather see this break on runtime in any test, then accidentially rolling this out some time and having to find out that the wrong fragment is displayed)

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FragmentFactory.java
84n4n4: ok, i read up on android flavoring, and all of this code, including the factory and deciding on runtime what implementation to initiate is not how flavoring is meant to be used.
please change structure.
https://stackoverflow.com/questions/23698863/build-flavors-for-different-version-of-same-class

84n4n4:
and same goes for all tests, the common test base stays in androidTest and the flavor specific ones go into their flavor androidTestPhiro and androidTestCatroid (while androidTestCatroid would be empty at this point)
https://stackoverflow.com/questions/22373655/test-package-for-different-flavors-in-android-studio
http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Testing1

catroid/src/main/java/org/catrobat/catroid/content/bricks/ConcurrentFormulaHashMap.java
```diff
@@ -30,6 +32,7 @@
 
 	private static final long serialVersionUID = 9030965461744658052L;
 
+	@TargetApi(24)
```
cheidenreich: why is this necessary?

catroid/src/main/java/org/catrobat/catroid/scratchconverter/ScratchConversionManager.java
```diff
@@ -150,6 +150,13 @@ private void closeAllActivities() {
 				} catch (ClassNotFoundException e) {
 					Log.e(TAG, e.getMessage());
 				}
+			} else if (BuildConfig.PHIRO_CODE) {
+				try {
```
cheidenreich: duplicate code, you can just set the string as a variable

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -553,7 +579,7 @@ public static boolean isNfcSharedPreferenceEnabled(Context context) {
 	public static void setNfcSharedPreferenceEnabled(Context context, boolean value) {
 		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
 		editor.putBoolean(SETTINGS_SHOW_NFC_BRICKS, value);
-		editor.commit();
+		editor.apply();
```
cheidenreich: `putBoolean()` returns the editor again, so you can just append the `.apply()` above

catroid/src/main/java/org/catrobat/catroid/ui/fragment/BrickCategoryFragment.java
```diff
@@ -71,6 +75,61 @@ public void setBrickAdapter(BrickAdapter brickAdapter) {
 	public void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setHasOptionsMenu(true);
+		initBrickCategories();
+	}
+
+	private void initBrickCategories() {
```
cheidenreich: is this used somewhere else? otherwise it's unnecessary to put in a separate method

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: This tests the following scenario:
1. An existing Phiro program is changed
2. User navigates back to MainMenu
3. User navigates to ProjectList
4. The changed Phiro program should be in the same state (not overridden)

catroid/src/androidTest/java/org.catrobat.catroid/uiespresso/util/matchers/AdapterViewMatchers.java
sjaindl: Yes, it's now done with onData and custom view matchers/wrappers

sjaindl:
I've refactored the flavoring of fragments, should be fine now. 
Adaptations from the main fragment versions only need to be done in the flavor files (if there are no adaptations, the files are empty). The main versions are abstract so that they can't be instantiated.

catroid/src/androidTestCreateatschool/AndroidManifest.xml
```diff
@@ -0,0 +1,53 @@
...
+  ~
+  ~ This program is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  ~ GNU Affero General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<manifest package="org.catrobat.catroid.phiro.test"
```
cheidenreich: seems like this file is not meant to be in the folder `catroid/src/androidTestCreateatschool` if its for the test package `org.catrobat.catroid.phiro.test` or am i missing something?

catroid/src/main/java/org/catrobat/catroid/ui/fragment/MainFormulaEditorCategoryListFragment.java
```diff
@@ -606,28 +562,44 @@ private void closeCategoryListFragment() {
...
 	}
 
-	public class FormulaEditorCategory {
+	protected class FormulaEditorCategory {
 		private int[] categoryItemIds = new int[] {};
 		private int[] categoryParameterIds = new int[] {};
 		private Map<Integer, String> headerOrderMap = new TreeMap<>();
 
-		public void addHeader(String text) {
+		boolean addCategoryIfDoesNotExist(String header, int[] items, int[] parameters) {
```
cheidenreich: why do you have a `boolean` return value if you never check it?

catroid/src/phiro/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
```diff
@@ -0,0 +1,54 @@
...
+public class FormulaEditorCategoryListFragment extends MainFormulaEditorCategoryListFragment {
+
+	@Override
+	public void onCreate(Bundle savedInstanceState) {
+		initCategories();
+		super.onCreate(savedInstanceState);
+	}
+
+	private void initCategories() {
+		initSensors();
```
cheidenreich: this method is not necessary if the only thing it does is one method call

cheidenreich:
@thmq @84n4n4 any feedback for the latest commit? https://github.com/Catrobat/Catroid/pull/2416/commits/46da464df7220496e2fa428674acdb5402179939

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/FormulaEditorCategoryListDataInteractionWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+
+public class FormulaEditorCategoryListDataInteractionWrapper extends DataInteractionWrapper {
+
+	public FormulaEditorCategoryListDataInteractionWrapper(DataInteraction dataInteraction) {
+		super(dataInteraction);
+	}
+
+	public static FormulaEditorCategoryListDataInteractionWrapper onFormulaEditorCategoryItemWithNameAtPosition(String formulaCategoryItem, int position) {
```
84n4n4: i want to have the dataInteraction and viewInteraction wrappers to handle and feel the same as the erspresso internal ones, please have a look at BrickDataInteractionWrapper for example.

so to have every method having only one parameter, i dont want to combine the component and position in one method.
eg.

`public static FormulaEditorCategoryListDataInteractionWrapper onFormulaEditorCategory(String category) {
return new FormulaEditorCategoryListDataInteractionWrapper(onData(allOf(instanceOf(String.class), withFormulaEditorCategoryItemName(formulaCategoryItem)))
inAdapterView(isFormulaEditorCategoryListView());
}`

which then can be used as follows:
instead of 
`onFormulaEditorCategoryItemWithNameAtPosition("someCategory", 5)`
that would be 
`onFormulaEditorCategoryItemWithName("someCategory").atPosition(5)`

but then again, depending on if this is used in multiple test classes or not, it could be even put there as a function instead of separate class...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/ScriptListDataInteractionWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+
+public class ScriptListDataInteractionWrapper extends DataInteractionWrapper {
+
+	public ScriptListDataInteractionWrapper(DataInteraction dataInteraction) {
+		super(dataInteraction);
+	}
+
+	public static ScriptListDataInteractionWrapper onBrickCategoryWithNameAtPosition(String brickCategory, int position) {
```
84n4n4: same here as for FormulaEditorCategoryListDataInteractionWrapper

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+		onView(withId(android.R.id.home))
+				.perform(click());
+
+		openProjectList();
+
+		checkAddedNoteBrickIsStillInFirstScript();
+	}
+
+	private void openProjectList() {
```
84n4n4: since all tests start with a call to "openProjectList", refactor the test to open the fragment or activity you actually want to test instead of the main menu activity and delete this function.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,153 @@
...
+		onView(withText(R.string.main_menu_programs))
+				.perform(click());
+	}
+
+	private void addNoteBrickToFirstScript(String projectName) {
```
84n4n4: is there a reason to do this via the UI instead of just modifying the project directly?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/FormulaEditorCategoryListDataInteractionWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+		return new FormulaEditorCategoryListDataInteractionWrapper(onData(allOf(instanceOf(String.class), withFormulaEditorCategoryItemName(formulaCategoryItem)))
```
84n4n4: also to be noted, if this whole class will contain only this one static function, then theres no point in having it in a wrapper class in the first place.
the wrappers are intended to provide a way to perform more complicated actions on a certain object while keeping the espresso "look and feel". please have a look at:
https://confluence.catrob.at/display/CAT/Espresso+testing+-+Using+wrappers+instead+of+utility+functions

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
```diff
@@ -0,0 +1,126 @@
...
+	public void phiroIsFirstInFormulaEditorDeviceCategoryTest() {
+		onView(withId(R.id.brick_set_x_edit_text))
+				.perform(click());
+
+		onView(withId(R.id.formula_editor_keyboard_sensors))
+				.perform(click());
+
+		String firstPhiroSensorString = UiTestUtils.getResourcesString(R.string.formula_editor_phiro_sensor_front_left);
+
+		onFormulaEditorCategoryItemWithNameAtPosition(firstPhiroSensorString, 0)
```
84n4n4: ok, regarding what ive wrote further up in the wrapper function. now seeing how this is actually applied, it would be even nicer and way more readable if one would just do:
`onFormulaEditorCategory().atPosition(0).check(matches(withText("someCategoryName")))`

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+		Intents.init();
+		Matcher<Intent> expectedIntent = allOf(hasAction(Intent.ACTION_VIEW), hasData(phiroLink));
+		intending(expectedIntent).respondWith(new Instrumentation.ActivityResult(0, null));
+		onData(PreferenceMatchers.withTitle(R.string.phiro_preference_title)).perform(click());
+		intended(expectedIntent);
+		Intents.release();
+	}
+
+	private void openSettings() {
```
84n4n4: use BaseActivityInstrumentationRule<SettingsActivity> to open directly the activity under test.
DO NOT use the ui to navigate to the activity that you want to test. this practice was one of the main reasons the old ui test setup was slow as fuck and flaky.

please have a look at:
https://confluence.catrob.at/display/CAT/Espresso+testing+-+How+to
theres a whole paragraph about this "Open the activity you want to test directly"

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -174,6 +178,15 @@ private boolean isRecoveredFromCrash() {
 		return PreferenceManager.getDefaultSharedPreferences(this).getBoolean(RECOVERED_FROM_CRASH, false);
 	}
 
+	private Intent createSettingsIntent() {
```
84n4n4: since this change in funcitonality is depending solely on the flavor, why not cover this also by moving the SettingsActivity into the flavor folders?

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -194,6 +194,13 @@ public void onPageStarted(WebView view, String urlClient, Bitmap favicon) {
 					} catch (ClassNotFoundException e) {
 						Log.e(TAG, e.getMessage());
 					}
+				} else if (BuildConfig.PHIRO_CODE) {
```
84n4n4: same here, why not use flavored classes to do this instead of Class.forName based on a build flag?

catroid/src/main/java/org/catrobat/catroid/utils/TrackingUtil.java
```diff
@@ -61,7 +61,7 @@ private TrackingUtil() {
 	}
 
 	public static void trackCreateProgram(String projectName, Boolean landscapeMode, boolean exampleProgram) {
```
84n4n4: just looking at the div below, i see the very same if clauses about 50 times.
why not implement a trackerClass that is then instantiated ONCE based on the build flag at startup, either as activeTracker, or as "blackHole" or whatever thats just swallowing tracking calls?

catroid/src/androidTestCreateatschool/AndroidManifest.xml
sjaindl: good catch. This Manifest actually isn't necessary anymore .. I've removed it

catroid/src/main/java/org/catrobat/catroid/ui/fragment/MainFormulaEditorCategoryListFragment.java
sjaindl: The return value is checked multiple times in MainFormulaEditorCategoryListFragment and FormulaEditorCategoryListFragment of the Phiro flavor

catroid/src/phiro/java/org/catrobat/catroid/ui/fragment/FormulaEditorCategoryListFragment.java
sjaindl: I've refactored that now so that the code is even more maintainable using the template method pattern. 

This means MainFormulaEditorCategoryListFragment knows the "recipe" of which categories to setup. The "recipe" is initCategories:

	private void initCategories() {
		initObjects();
		initFunctions();
		initLogic();
		initSensors();
	}
These init-methods are protected, so that the subclass can override it to setup the categories with help of polymorphism. 

So, all that needs to be done now in the "Phiro-"FormulaEditorCategoryListFragment is override the method initSensors() and call super.initSensors()

Even much shorter and better maintainable.

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
sjaindl: Yes, that's one thing I've planned to refactor, too.
OK, then I'll refactor it directly in this branch.

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
sjaindl: yes, same as above, this won't be necessary anymore, after all.

catroid/src/main/java/org/catrobat/catroid/utils/TrackingUtil.java
sjaindl: An even better approach might be to use a protection proxy.
I could implement a TrackingInvocationHandler that proxies the TrackingUtil class.
By doing that, all that needs to be done is to check:
if (BuildConfig.DATA_TRACKING) {
	return method.invoke(trackingUtil, method, args);
}
once in the invocation handlers invoke-method. So I don't even need a "blackHole"-Tracker.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
sjaindl: OK, this is possible now for both, FE and Brick categories:
onFormulaEditorCategory()
				.atPosition(0)
				.check(matches(withText(firstPhiroSensorString)));
..and:
onBrickCategory()
				.atPosition(0)
				.check(matches(withText(expectedCategoryString)));

I've left them in wrapper functions since I had to check for the childView(..) of the respective LinearLayout, too. I think therefore it's worth to put it inside a wrapper.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: There was a reason for doing so.
In MainMenuActivity (Phiro-flavor file) was some initialization logic going on. The Phiro programs have been loaded there when the app is started.

I've moved that code to the flavored ProjectListFragment, where the phiro programs are loaded async now. The downside is a bit of delay until programs are available.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: No there's no reason to this via the UI. This is just to test whether programs are overwritten.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
sjaindl: There was the same reason as above. I've moved the initialization code to the flavored version of SettingsActivity now.

sjaindl:
OK, next round.
I've refactored the other activities, too. As well as tests (I had to move some initialization code - see comments to outdated diffs) and introduced some design patterns where useful (proxy pattern for tracking, template method for FE-Category initialization).

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/FormulaEditorCategoryListDataInteractionWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.catrobat.catroid.uiespresso.util.matchers.FormulaEditorCategoryListMatchers.isFormulaEditorListItem;
+import static org.hamcrest.Matchers.instanceOf;
+
+public class FormulaEditorCategoryListDataInteractionWrapper extends DataInteractionWrapper {
+
+	public FormulaEditorCategoryListDataInteractionWrapper(DataInteraction dataInteraction) {
+		super(dataInteraction);
+	}
+
+	public static DataInteraction onFormulaEditorCategory() {
```
84n4n4: i cant see my comments anymore from before, but this seems to be unchanged. theres no point in having a wrapper if it doesnt provide additional functionality. this is merely a utility function class, that provides one method to instantiate it:
onFormulaEditorCategory()

please at least rename it to FormulaEditorCategoryListDataInteractionUtils, since this is NOT a wrapper

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/ProjectListDataInteractionWrapper.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+
+public class ProjectListDataInteractionWrapper extends DataInteractionWrapper {
+
+	public ProjectListDataInteractionWrapper(DataInteraction dataInteraction) {
+		super(dataInteraction);
+	}
+
+	public static ProjectListDataInteractionWrapper onProjectWithName(String projectName) {
```
84n4n4: same here as with FormulaEditorCategoryListDataInteractionWrapper

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/ScriptListDataInteractionWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.catrobat.catroid.uiespresso.util.matchers.BrickCategoryListMatchers.isBrickCategoryView;
+import static org.hamcrest.Matchers.instanceOf;
+
+public class ScriptListDataInteractionWrapper extends DataInteractionWrapper {
+
+	public ScriptListDataInteractionWrapper(DataInteraction dataInteraction) {
+		super(dataInteraction);
+	}
+
+	public static DataInteraction onBrickCategory() {
```
84n4n4: and here aswell...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/matchers/FormulaEditorCategoryListMatchers.java
```diff
@@ -0,0 +1,68 @@
...
+				description.appendText("FormulaEditorCategoryListView");
+			}
+		};
+	}
+
+	public static Matcher<View> isFormulaEditorListItem() {
+		return new TypeSafeMatcher<View>() {
+			@Override
+			protected boolean matchesSafely(View item) {
+				return item.getId() == R.id.fragment_formula_editor_list_item;
```
84n4n4: why not user withId(R.id.fragment_formula_editor_list_item) since this is only used once?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/matchers/FormulaEditorCategoryListMatchers.java
```diff
@@ -0,0 +1,68 @@
...
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+public final class FormulaEditorCategoryListMatchers {
+	private FormulaEditorCategoryListMatchers() {
+		throw new AssertionError();
+	}
+
+	public static Matcher<View> isFormulaEditorCategoryListView() {
```
84n4n4: this is doing exactly the same as util.matchers.BrickCategoryListMatchers

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/matchers/ProjectListMatchers.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public static Matcher<Object> withProjectName(String expectedProjectName) {
+		return withProjectName(equalTo(expectedProjectName));
+	}
+	private static Matcher<Object> withProjectName(final Matcher<String> expectedName) {
+		return new BoundedMatcher<Object, ProjectData>(ProjectData.class) {
+
+			@Override
+			public boolean matchesSafely(final ProjectData actualObject) {
+				return expectedName.matches(actualObject.projectName);
```
84n4n4: please rename actualObject to actualProjectData

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,126 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.ui.activity;
```
84n4n4: can you put the phiro tests in a separate package please

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+		checkAddedNoteBrickIsStillInFirstScript();
+	}
+
+	private void relaunchActivity() {
+		baseActivityTestRule.getActivity().finish();
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	private void addNoteBrickToFirstScript(String projectName) {
```
84n4n4: please remove everything from this function besides:
 +		project.getDefaultScene().getSpriteList().get(0).getScript(0).addBrick(new NoteBrick("I was added!")); 

since unnecessary switching around activities leads to problems with espresso tests when running in bulk. also, i think that testing if the project is really opened when clicked, should be tested in a seperate test dealing with basic functionality of the projectManager

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroScriptActivityTest.java
```diff
@@ -0,0 +1,130 @@
...
+	public void teardown() throws Exception {
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+		SharedPreferences.Editor editor = sharedPreferences.edit();
+
+		editor.putBoolean(SETTINGS_SHOW_PHIRO_BRICKS, initialPreferences.get(SETTINGS_SHOW_PHIRO_BRICKS));
+		editor.putBoolean(SETTINGS_SHOW_HINTS, initialPreferences.get(SETTINGS_SHOW_HINTS));
+		editor.apply();
+	}
+
+	private static void createProject(String projectName) {
```
84n4n4: theres a createProjectAndGetStartScript in content.brick.utils.BrickTestUtils that could be used instead:
eg:
BrickTestUtils.createProjectAndGetStartScript(projectname).addBrick(new SetXBrick());

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,151 @@
...
+		for (String setting : allSettings) {
+			initialSettings.put(setting, sharedPreferences.getBoolean(setting, false));
+		}
+		setAllSettingsTo(false);
+
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@Test
+	public void phiroIsInitiallyActivatedTest() {
```
84n4n4: in lite 86 in setup all settings are set to false: setAllSettingsTo(false)

so this test does merely verify the setup method of itself

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,151 @@
...
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+		assertTrue("Phiro preference was not initially enabled!", sharedPreferences.getBoolean(SETTINGS_SHOW_PHIRO_BRICKS, false));
+	}
+
+	@Test
+	public void phiroIsFirstVisibleSettingTest() {
+		onData(PreferenceMatchers.withTitle(R.string.preference_title_enable_phiro_bricks))
+				.atPosition(0)
```
84n4n4: why is the atPosition necessary here? there shouldnt be more than one preferences with the same title, if so that should fail rightfully

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,151 @@
...
+		SharedPreferences.Editor sharedPreferencesEditor = PreferenceManager
+				.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext()).edit();
+
+		for (String setting : allSettings) {
+			sharedPreferencesEditor.putBoolean(setting, value);
+		}
+		sharedPreferencesEditor.commit();
+	}
+
+	private void checkPreference(int displayedTitleResourceString, String sharedPreferenceTag) {
```
84n4n4: sinc this function is idential with the one in SettingsActivityTest, please move it to a utility class

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,151 @@
...
+		SharedPreferences.Editor sharedPreferencesEditor = PreferenceManager
+				.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext()).edit();
+		for (String setting : initialSettings.keySet()) {
+			sharedPreferencesEditor.putBoolean(setting, initialSettings.get(setting));
+		}
+		sharedPreferencesEditor.commit();
+		initialSettings.clear();
+	}
+
+	private void setAllSettingsTo(boolean value) {
```
84n4n4: sinc this function is idential with the one in SettingsActivityTest, please move it to a utility class

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
```diff
@@ -0,0 +1,151 @@
...
+		for (String setting : allSettings) {
+			sharedPreferencesEditor.putBoolean(setting, value);
+		}
+		sharedPreferencesEditor.commit();
+	}
+
+	private void checkPreference(int displayedTitleResourceString, String sharedPreferenceTag) {
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+
+		onData(PreferenceMatchers.withTitle(displayedTitleResourceString)).perform(click());
```
84n4n4: please reformat longer espresso calls like this to format:
onWhatever(.......)
     .doWhatever(....)

eg. 
onData(PreferenceMatchers.withTitle(displayedTitleResourceString))
      .perform(click());

84n4n4:
test this please

84n4n4:
please add the additional espresso tests to testsuites.AllEspressoTestsSuit.java to have them run with the pullrequest trigger...

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
ok, so to make the jenkins espresso PR trigger work, feature/NOLB would have to be rebased on develop. (since the PullRequestTriggerSuite is missing here...)

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
sjaindl: This is necessary here because I want to explicitly check whether this preference is the first in the list. Without .atPosition(0) the test would pass, but shouldn't!

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
sjaindl: It does, in setup all settings are set to false. This checks whether it has been set to true again when starting MyProjectsActivity!

sjaindl:
OK, I have refactored the Espresso tests according to your last comments. 
Concerning the 2 tests in PhiroSettingsActivityTest.java, please see my comments (diffs outdated because I've changed the package). 
I can't rebase on develop now because we want don't want to have this branch ahead of the release branch for the upcoming Phiro release (I have to do that later to make PR-Trigger-Tests work).

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: I can't directly say that, because there's no project loaded at that stage. So, I had to set the project first..

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/FormulaEditorCategoryListDataInteractionWrapper.java
84n4n4: bump.
i did not meant just renaming this, but leaving it extending the DataInteractionWrapper.

please take
FormulaEditorCategoryListDataInteractionUtils.java
ProjectListDataInteractionUtils.java
ScriptListDataInteractionUtils.java

and move the one contained functions somewhere meaningful, in this case, since each of them is only used in single test classes, put them in the testclasses.

so in general, if a utility function is used in - 
- one test class, put it there
- multiple test classes in the same package, make a utility function class in this package, make it package private
- multiple test classes spanning multiple distinct packages, make a global utility file with a meaningful name in the utils packag.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
84n4n4: ok

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroSettingsActivityTest.java
84n4n4: ok

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
84n4n4: ok

84n4n4: but in general ive got a bad feeling with this test, since my experience has shown that manually killing / relaunching / and especially switching inbetween activities leads to really weird test failures when the test is run in conjunction with other tests. (did that a few times already, was a painful learning experience)

ill try to have this run on jenkins to see if this causes other tests to fail.

84n4n4: ok, so heres the run on your repository:
https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidSinglePackageEmulatorTest/19319/

and here the one on the feature/Nolb
https://jenkins.catrob.at/view/LeeroyCatroid/job/CatroidSinglePackageEmulatorTest/19320/

since most of the tests are in another package, they arent run here in the first place...
weve got to think up a proper way on how to deal with the flavored tests i guess...

84n4n4: both look similar, but also the phiroActivity tests arent run there...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/utils/FormulaEditorCategoryListDataInteractionWrapper.java
sjaindl: OK, I've moved them to the respective test classes.

catroid/src/androidTestPhiro/java/org/catrobat/catroid/uiespresso/ui/activity/PhiroMyProjectsActivityTest.java
sjaindl: Doesn't sound so well. What was your solution to those cases where you had to restart/switch activities?
OK, so the same tests are failing.. are they failing in develop, too? Since they shouldn't have something to do with feature/NOLB in the first place. 
Ooh yeah we have to think about that.. maybe tests could be build and executed on a per flavor basis on Jenkins?

sjaindl: Hmm, at least in feature/nolb there are really no flavored Espresso tests yet

84n4n4: yeah, the ones above are failing in develop aswell, there is actually a bug causing them to fail.

anyways, i dont really have a solution for the activity switching issue. with this tests here weve got a change that its going to be fine since you start / finish / restart the activity under test using the test rule.

the problems i encountered up to this point was not that problematic tests were failing, but they made other tests fail. and to make things more dificult, not the next ones, but there was some constellation that the next ones running main menu activity were failing.
although ive read about other people having similar problems, its worse in our project due to the huge number of activities, and us messing with the back stack with certain activities, and also the sometimes kind of crappy code in some of the activity lifecylce related methods (onCreate, onDestroy, etc...)
so in the end i just started removing tests from the testsuites to see when the other test was working again and could narrow it down.

so all in all, i dont really have a solution for this problem.....
so its just something i try to avoid in the first place...

catroid/src/main/java/org/catrobat/catroid/ui/fragment/BaseProjectListFragment.java
```diff
@@ -128,7 +127,7 @@ public int compare(ProjectData project1, ProjectData project2) {
 			}
 		});
 
-		projectAdapter = new ProjectListAdapter(getActivity(), R.layout.list_item, projectList);
+		projectAdapter = new ProjectListAdapter(context, R.layout.list_item, projectList);
```
cheidenreich: `context` is null at this time, this causes the app to crash at older Android versions, changing it back to `getActivity()` works.

sjaindl: Ok, I see. However, I can't rely on getActivity() here, because CopyPhiroProjectsTask in the ProjectListFragment of the Phiro-flavor causes the Fragment to be detached from it's activity. So, the context/activity must be remembered. 
So, I'm sure the issue is that, because we are not using the SupportFragmentManager, onAttach(Context) is simply not called. 
I think the best approach would be to override onAttach(Context) & onAttach(Activity) for API < 23, as described here: https://stackoverflow.com/questions/32077086/android-onattachcontext-not-called-for-api-23

PR #2419, opened 2017-07-26
CAT-2490 Create methods for setup things for CrashReporter

nishantt95:
In this PR common methods are created for both CrashReporterTest and
BaseExceptionHandlerTest.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
@@ -25,12 +25,10 @@ package org.catrobat.catroid.test.ui;
 
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.preference.PreferenceManager;
-import android.support.test.InstrumentationRegistry;
 import android.test.AndroidTestCase;
 
+import org.catrobat.catroid.test.utils.CrashReporterTest;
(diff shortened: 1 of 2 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/ui/BaseExceptionHandlerTest.java
```diff
@@ -46,14 +44,10 @@
...
-		editor = sharedPreferences.edit();
-		editor.clear();
-		editor.putBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true);
-		editor.commit();
-		CrashReporter.setIsCrashReportEnabled(true);
-		CrashReporter.initialize(context);
+		CrashReporterTest.setUpMethod();
+		context = CrashReporterTest.findContext();
+		sharedPreferences = CrashReporterTest.getSharedPreferences();
+		editor = CrashReporterTest.getEditor();
```
robertpainsi: That's a little over the top imo. I guess you thought that as well when creating the getters.
I think, the best code that can be shared is a util method which enables and initialize crash reporting. However, I'm not sure if it's worth it. What do you think?

nishantt95: > However, I'm not sure if it's worth it. What do you think?

Yes, you are right. Creating a method just for enabling and initializing crash reporting isn't worth it.

robertpainsi:
Ticket closed

PR #2423, opened 2017-07-27
fixed receiving the same broadcast multiple times

comeclarity:
The bug appeared when there were multiple receive broadcast bricks with the same broadcast message in the same sprite.
The list of actions was overriden, instead of of being extended, so only the script of the last added receive brick was executed.


cheidenreich:
what do I see there? Is that a Unit Test? <3

PR #2425, opened 2017-07-27
CAT-2491 Fixed bug with unsupported bricks in cast programs

eisfux:


catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -415,9 +415,12 @@
 		looksBrickList.add(new SetBackgroundByIndexBrick(BrickValues.SET_LOOK_BY_INDEX));
 		looksBrickList.add(new SetBackgroundAndWaitBrick());
 		looksBrickList.add(new SetBackgroundByIndexAndWaitBrick(BrickValues.SET_LOOK_BY_INDEX));
-		looksBrickList.add(new CameraBrick());
-		looksBrickList.add(new ChooseCameraBrick());
-		looksBrickList.add(new FlashBrick());
+		//only add these bricks to the list if the current program is NOT a cast program
+		if (!ProjectManager.getInstance().getCurrentProject().isCastProject()) {
```
cheidenreich: can you utilize your list here as well?

eisfux: I don't think so because it makes sense to use the list when bricks are added in some kind of loop, but here the bricks are added "manually".

PR #2428, opened 2017-07-29
CAT-2494 Remove dependency of Test-classes

redeamer:
Resolve references between classes with actual tests. This allows to run only a
subset of tests by exluding some test-classes from the build without generating
build errors.
This is needed for parallel testing on Jenkins.

- SetLookByIndexActionTest (containing just 2 tests, without other
    modifications) was derived from SetLookActionTest which defined tests
    itself. Moved the tests into the base class.
- PhysicsCollisionScriptInteractionTest used the class-name of
     PhysicsSpriteCloneTest as log tag (copy and paste issue)
- PhysicsWorldTest used the class-name of PhysicsObjectTest as log tag
     (copy and paste issue)


84n4n4:
test this please

PR #2332, opened 2017-05-31
CAT-2345 Multilingual Feature on Application Level

ndrnour:
The app must provide an interface that handles user input to switch languages independent of the smartphone’s global settings. This feature should list all languages supported by the app including languages not yet supported by the operating system.

PR #2430, opened 2017-07-29
CAT-2473 Setting for enable disable cloud notifications

nishantt95:
Added the feature for enabling/disabling the notifications through settings. It uses the firebase cloud messaging's topics subscribe/unsubscribe feature to enable/disable notifications.

Note: This PR also includes the test file.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
@@ -1,0 +1,91 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.cloudmessaging;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.cloudmessaging.CloudMessaging;
+import org.catrobat.catroid.cloudmessaging.CloudMessagingInterface;
+import org.catrobat.catroid.cloudmessaging.FirebaseCloudMessaging;
+import org.catrobat.catroid.ui.SettingsActivity;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(AndroidJUnit4.class)
+public class CloudMessagingTest {
+
+	private Context context;
+	private SharedPreferences sharedPreferences;
+	private CloudMessagingInterface messagingService;
+
+	@Before
+	public void setUp() {
+		sharedPreferences = mock(SharedPreferences.class);
+		context = mock(Context.class);
+		messagingService = mock(FirebaseCloudMessaging.class);
+		when(PreferenceManager.getDefaultSharedPreferences(context)).thenReturn(sharedPreferences);
+		when(sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false)).thenReturn(true);
+		CloudMessaging.setIsCloudNotificationsEnabled(true);
+		CloudMessaging.setCloudMessagingInterface(messagingService);
+	}
+
+	@Test
+	public void testCloudMessagingUninitializedOnNotificationsDisabled() {
+		CloudMessaging.setIsCloudNotificationsEnabled(false);
+
+		assertFalse(CloudMessaging.initialize(context));
+	}
+
+	@Test
+	public void testUserSubscribedToTopicsOnCloudMessagingInitialized() {
+		assertTrue(CloudMessaging.initialize(context));
+		verify(messagingService, times(1)).subscribe();
+	}
+
+	@Test
+	public void testUserUnsubscribedToTopicsOnCloudMessagingInitializationFailed() {
+		when(sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false)).thenReturn(false);
+
+		assertFalse(CloudMessaging.initialize(context));
+		verify(messagingService, times(1)).unsubscribe();
+	}
+
+	@After
+	public void tearDown() {
+		CloudMessaging.setIsCloudNotificationsEnabled(false);
+	}
+}
--- a/catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
+++ b/catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
@@ -27,8 +27,10 @@ import android.support.multidex.MultiDex;
 import android.support.multidex.MultiDexApplication;
 import android.util.Log;
 
+import com.google.firebase.iid.FirebaseInstanceId;
 import com.parrot.freeflight.settings.ApplicationSettings;
 
+import org.catrobat.catroid.cloudmessaging.CloudMessaging;
 import org.catrobat.catroid.utils.CrashReporter;
 
 public class CatroidApplication extends MultiDexApplication {
@@ -46,7 +48,9 @@ public class CatroidApplication extends MultiDexApplication {
 	public void onCreate() {
 		super.onCreate();
 		CrashReporter.initialize(this);
+		CloudMessaging.initialize(this);
 		Log.d(TAG, "CatroidApplication onCreate");
+		Log.v(TAG, "Firebase Cloud Messaging Token : " + FirebaseInstanceId.getInstance().getToken());
 		settings = new ApplicationSettings(this);
 		CatroidApplication.context = getApplicationContext();
 	}
--- a/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
+++ b/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
@@ -30,8 +30,12 @@ import android.content.Context;
 import android.content.Intent;
 import android.media.RingtoneManager;
 import android.net.Uri;
+import android.preference.PreferenceManager;
+import android.support.annotation.VisibleForTesting;
 
+import org.catrobat.catroid.BuildConfig;
 import org.catrobat.catroid.R;
+import org.catrobat.catroid.ui.SettingsActivity;
 import org.catrobat.catroid.utils.StatusBarNotificationManager;
 
 public class CloudMessaging {
@@ -39,12 +43,39 @@ public class CloudMessaging {
 	private final Context context;
 	private final NotificationBuilderProvider notificationBuilderProvider;
 	private final NotificationManager notificationManager;
+	private static boolean isCloudNotificationsEnabled;
+	private static CloudMessagingInterface messagingService = new FirebaseCloudMessaging();
 
 	public CloudMessaging(Context context, NotificationBuilderProvider notificationBuilderProvider,
 			NotificationManager notificationManager) {
 		this.context = context;
 		this.notificationBuilderProvider = notificationBuilderProvider;
 		this.notificationManager = notificationManager;
+		isCloudNotificationsEnabled = BuildConfig.FEATURE_CLOUD_MESSAGING_ENABLED;
+	}
+
+	public static boolean initialize(Context context) {
+		boolean isEnabled = isCloudNotificationsEnabled(context);
+		if (isEnabled) {
+			messagingService.subscribe();
+		} else {
+			messagingService.unsubscribe();
+		}
+		return isEnabled;
+	}
+
+	@VisibleForTesting
+	public static void setCloudMessagingInterface(CloudMessagingInterface cloudMessagingInterface) {
+		messagingService = cloudMessagingInterface;
+	}
+
+	@VisibleForTesting
+	public static void setIsCloudNotificationsEnabled(boolean isEnabled) {
+		isCloudNotificationsEnabled = isEnabled;
+	}
+
+	private static boolean isCloudNotificationsEnabled(Context context) {
+		return PreferenceManager.getDefaultSharedPreferences(context).getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false) && isCloudNotificationsEnabled;
 	}
 
 	public void showNotification(CloudMessage cloudMessage) {
--- a/catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
@@ -39,6 +39,7 @@ import android.view.MenuItem;
 
 import org.catrobat.catroid.BuildConfig;
 import org.catrobat.catroid.R;
+import org.catrobat.catroid.cloudmessaging.CloudMessaging;
 import org.catrobat.catroid.common.DroneConfigPreference;
 import org.catrobat.catroid.devices.mindstorms.ev3.sensors.EV3Sensor;
 import org.catrobat.catroid.devices.mindstorms.nxt.sensors.NXTSensor;
@@ -87,6 +88,7 @@ public class SettingsActivity extends PreferenceActivity {
 	public static final String RASPI_VERSION_SPINNER = "setting_raspi_version_preference";
 
 	public static final String SETTINGS_CRASH_REPORTS = "setting_enable_crash_reports";
+	public static final String SETTINGS_CLOUD_NOTIFICATIONS = "setting_enable_cloud_notifications";
 
 	@SuppressWarnings("deprecation")
 	@Override
@@ -96,6 +98,7 @@ public class SettingsActivity extends PreferenceActivity {
 		addPreferencesFromResource(R.xml.preferences);
 
 		setAnonymousCrashReportPreference();
+		setCloudNotificationPreference();
 		setNXTSensors();
 		setEV3Sensors();
 		setDronePreferences();
@@ -160,6 +163,12 @@ public class SettingsActivity extends PreferenceActivity {
 			crashlyticsPreference.setEnabled(false);
 			screen.removePreference(crashlyticsPreference);
 		}
+
+		if (!BuildConfig.FEATURE_CLOUD_MESSAGING_ENABLED) {
+			CheckBoxPreference cloudMessagingPreference = (CheckBoxPreference) findPreference(SETTINGS_CLOUD_NOTIFICATIONS);
+			cloudMessagingPreference.setEnabled(false);
+			screen.removePreference(cloudMessagingPreference);
+		}
 	}
 
 	@SuppressWarnings("deprecation")
@@ -169,6 +178,18 @@ public class SettingsActivity extends PreferenceActivity {
 			@Override
 			public boolean onPreferenceChange(Preference preference, Object isChecked) {
 				setAutoCrashReportingEnabled(getApplicationContext(), (Boolean) isChecked);
+				return true;
+			}
+		});
+	}
+
+	@SuppressWarnings("deprecation")
+	private void setCloudNotificationPreference() {
+		CheckBoxPreference reportCheckBoxPreference = (CheckBoxPreference) findPreference(SETTINGS_CLOUD_NOTIFICATIONS);
+		reportCheckBoxPreference.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
+			@Override
+			public boolean onPreferenceChange(Preference preference, Object isChecked) {
+				setCloudNotificationsEnabled(getApplicationContext(), (Boolean) isChecked);
 				return true;
 			}
 		});
@@ -404,6 +425,12 @@ public class SettingsActivity extends PreferenceActivity {
 		if (isEnabled) {
 			CrashReporter.initialize(context);
 		}
+	}
+
+	public static void setCloudNotificationsEnabled(Context context, boolean isEnabled) {
+		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
+		editor.putBoolean(SETTINGS_CLOUD_NOTIFICATIONS, isEnabled).commit();
+		CloudMessaging.initialize(context);
 	}
 
 	private static void setBooleanSharedPreference(boolean value, String settingsString, Context context) {
--- a/catroid/src/main/res/values/strings.xml
+++ b/catroid/src/main/res/values/strings.xml
@@ -131,6 +131,8 @@
     <string name="preference_description_nfc_bricks">Allow to use NFC tags in programs</string>
     <string name="preference_description_crash_reports">Allow to send crash reports automatically</string>
     <string name="preference_title_enable_crash_reports">Send anonymous crash reports</string>
+    <string name="preference_title_enable_cloud_notifications">Enable Cloud Notifications</string>
+    <string name="preference_description_cloud_notifications">Allow to receive cloud notifications</string>
     <!-- -->
     <string name="preference_description_arduino_bricks">Allow to control Arduino boards</string>
     <string name="preference_title_enable_raspi_bricks">Raspberry Pi bricks</string>
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/SettingsActivityCloudNotificationTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/SettingsActivityCloudNotificationTest.java
@@ -1,0 +1,63 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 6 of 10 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -406,6 +427,15 @@ public static void setAutoCrashReportingEnabled(Context context, boolean isEnabl
 		}
 	}
 
+	public static void setCloudNotificationsEnabled(Context context, boolean isEnabled) {
+		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
+		editor.putBoolean(SETTINGS_CLOUD_NOTIFICATIONS, isEnabled);
+		editor.commit();
+		if (isEnabled) {
+			CloudMessaging.initialize(context);
+		}
```
robertpainsi: Disabling notifications will not unsubscribe from FCM (immediately).

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,76 @@
...
+	private SharedPreferences.Editor editor;
+
+	@Before
+	public void setUp() {
+		context = InstrumentationRegistry.getTargetContext();
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		editor = sharedPreferences.edit();
+		editor.clear();
+		editor.putBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, true);
+		editor.commit();
```
robertpainsi: Hmm, can we mock the Context and SharedPreferences everywhere?
Otherwise the appstore and debug/testing version share the same SharedPreferences (afaik).

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -28,17 +28,27 @@
...
 import org.catrobat.catroid.utils.StatusBarNotificationManager;
 
 public class CloudMessaging {
 
+	public static final String DEFAULT_TOPIC = "DEFAULT_TOPIC";
+
 	private final Context context;
 	private final NotificationBuilderProvider notificationBuilderProvider;
 	private final NotificationManager notificationManager;
+	private static boolean isCloudNotificationsEnabled = BuildConfig.FEATURE_CLOUD_MESSAGING_ENABLED;
```
robertpainsi: We implemented the CrashReporter the same way. I haven't found a better solution yet. Can you please also do some research here?

robertpainsi:
Currently only a global topic is supported. Since we will support multiple languages, can you please adapt the code to support this as well. Reliably un-/subscribing is definitely a challenge. Unfortunately there is no way to get a list of subscribed topics afaik.

nishantt95:
> Currently only a global topic is supported. Since we will support multiple languages, can you please adapt the code to support this as well. Reliably un-/subscribing is definitely a challenge. Unfortunately there is no way to get a list of subscribed topics afaik.

Languages will completely on the country or there is a different setting for that (Allowing user to select the language)?

nishantt95:
> Currently only a global topic is supported. Since we will support multiple languages, can you please adapt the code to support this as well. Reliably un-/subscribing is definitely a challenge. Unfortunately there is no way to get a list of subscribed topics afaik.

Languages will completely on the country or there is a different setting for that (Allowing user to select the language)?

Now the idea is to subscribe user to 2 topics, 
1. DEFAULT_TOPIC (global) that will determine if notifications are enabled or not.
2. COUNTRY_CODE (could be any country's code) will identify the country (and the language).

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
nishantt95: Right now, the only solution I can think of is to place `isCloudNotificationsEnabled = BuildConfig.FEATURE_CLOUD_MESSAGING_ENABLED;` inside the constructor.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
nishantt95: Can we use a different sharedpreference. Like `context.getSharedPreferences(<ONLY_FOR_TESTING>, Private mode)`

robertpainsi:
Drop the country implementation for now. Supporting language is best for now.

catroid/src/main/res/values/strings.xml
```diff
@@ -131,6 +131,8 @@
     <string name="preference_description_nfc_bricks">Allow to use NFC tags in programs</string>
     <string name="preference_description_crash_reports">Allow to send crash reports automatically</string>
     <string name="preference_title_enable_crash_reports">Send anonymous crash reports</string>
+    <string name="preference_description_cloud_notifications">Allow to receive cloud notifications</string>
+    <string name="preference_title_enable_cloud_notifications">Enable Cloud Notifications</string>
```
robertpainsi: Please swap these lines.
1. title
2. description

Please create a PR to correct the order of the crash_reporter strings as well.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
robertpainsi: I'm not 100% sure… You want to pass in the SharedPreferences, right? I guess I need to see it implemented to evaluate it.

___
JFI:
```java
PreferenceManager.getDefaultSharedPreferences(context)
```
==
```java
public static SharedPreferences getDefaultSharedPreferences(Context context) {
    return context.getSharedPreferences(getDefaultSharedPreferencesName(context),
            getDefaultSharedPreferencesMode());
}
```

Context can be mocked with Mocktito or [MockContext](https://developer.android.com/reference/android/test/mock/MockContext.html)

So it's easily possible to return a custom SharedPreferences object. However, creating a fully functional SharedPreferences object isn't that easy because you've to do it yourself. I'd hate that.

Mocking everything is also an option.

Many people suggest to use Dagger but I just don't want to introduce it now.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -54,6 +69,50 @@ public void showNotification(CloudMessage cloudMessage) {
...
+		FirebaseMessaging.getInstance().unsubscribeFromTopic(getCountryCode(context));
+	}
+
+	private static boolean isCloudNotificationsEnabled(Context context) {
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		return sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false) && isCloudNotificationsEnabled;
+	}
+
+	private static String getLanguageCode(Context context) {
+		String languageCode = context.getResources().getConfiguration().locale.getCountry();
```
robertpainsi: `getCountry` = `US`
`getLanguage` = `en`

Please do some research on how the language code is standardized. See [getLanguage](https://developer.android.com/reference/java/util/Locale.html#getLanguage())

catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
```diff
@@ -47,6 +48,7 @@
 	public void onCreate() {
 		super.onCreate();
 		CrashReporter.initialize(this);
+		CloudMessaging.initialize(this);
```
robertpainsi: Is this really needed?

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -28,23 +28,38 @@
...
 public class CloudMessaging {
 
+	private static final String TAG = CloudMessaging.class.getSimpleName();
+
+	public static final String DEFAULT_TOPIC = "DEFAULT_TOPIC";
+
 	private final Context context;
 	private final NotificationBuilderProvider notificationBuilderProvider;
 	private final NotificationManager notificationManager;
+	private static boolean isCloudNotificationsEnabled;
```
robertpainsi: Currently, `isCloudNotificationsEnabled` is not needed because the default value of `SETTINGS_CLOUD_NOTIFICATIONS` is false anyway, I guess.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -54,6 +69,50 @@ public void showNotification(CloudMessage cloudMessage) {
...
+		return languageCode;
+	}
+
+	private static String getCountryCode(Context context) {
+		TelephonyManager telephonyManager = (TelephonyManager) context.getSystemService(Context.TELEPHONY_SERVICE);
+		String countryCode = telephonyManager.getNetworkCountryIso();
+		Log.v(TAG, "Country Code : " + countryCode);
+		return countryCode;
+	}
+
```
robertpainsi: CloudMessaging handles
* Building and showing notifications
* Un-/subscribing to topics

IMO this can easily be split into two classes.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
robertpainsi: Ah, it may needed, my fault

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
nishantt95: Yes, I was thinking about `context.getSharedPreferences('cloud_testing', 0);` but it doesn't seem to work as expected. I did some work on Mocktito and will try to implement the mock sharedpreferences.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
nishantt95: Applied `getDisplayLanguage()`. That will be more appropriate and standardized.

catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
nishantt95: Hmm. With current configuration (checkbox is false by default) I guess not. But IMO the checkbox value should be set to true by default (because generally, users don't go to the settings tab unless required.) 

So, if you want me to remove it then let me know.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
nishantt95: > ...appstore and debug/testing version share the same SharedPreferences (afaik).

Even though we are using same SharedPreferences, we are still clearing to remove the previous data.
https://stackoverflow.com/a/17791944/4590536

Do we still need to mock?

robertpainsi: Clearing the SharedPreferences will clear all the settings on the smartphone. The next time the developer uses the app, all settings are set to the default values. That's definitely a pita because 'Show hints' will be enabled as well :unamused:

> …creating a fully functional SharedPreferences object isn't that easy because you've to do it yourself. I'd hate that.

Hmm, maybe it isn't that bad after all if we just create a partially functional SharedPreferences implementation. Just with HashMaps and to get/set values synchronously. Every more complex method will just throw an `UnsupportedOperationException`.
We could also provide a method to store and restore the SharedPreferences. This can be used [here](https://github.com/Catrobat/Catroid/pull/2377/files#diff-c41b068c0724c4fe613f723e96219ba1R81). Btw, such implementation is officially called a Fake :wink:

What do you think?

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
robertpainsi: That's more or less the same (if you don't mean by DI)

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
robertpainsi: I haven't tested it but according to the docs, [` getDisplayLanguage()`](https://developer.android.com/reference/java/util/Locale.html#getDisplayLanguage()) is localized.

default DISPLAY locale is en_US => French
default DISPLAY locale is fr_FR => anglais

I'm not sure if there is a reliable way to get the standardized language code. However, we need a reliable way ^^

https://stackoverflow.com/a/8691091/2774399 looks interesting.

There is also a PR https://github.com/Catrobat/Catroid/pull/2332/files#diff-7a2a77ec325d68cbe50a2c6e57b0468aR53 already defining a list of all languages. However, it's far from final and maybe won't be merged that way.

The Firebase console also automatically provides languages. User properties can be used to check if a user should receive notifications. See
image
The problems, it can't be combined with topics and I don't know if it uses the app language or OS language :disappointed: There may be different type of notifications the user can subscribe to (events, news, hot projects, etc).

nishantt95: > The problems, it can't be combined with topics...

I think we can't use it then. But on the other hand

> https://stackoverflow.com/a/8691091/2774399 looks interesting.

We can use it. So here we will create a key-value pair between the code and the language. As far as I have understood the concept we will be using it like this (en = "English", fr = "French") and we will be subscribing the users to values (English, French). I think key-value pairs will be created manually in strings.xml.

Another problem is subscribing/unsubscribing during the language change. We can solve in 2 ways:-
1. We can store the current locale on sharedPreferences and whenever the user opens the device we can check if stored locale is equal to the current locale.
 OR
2. https://stackoverflow.com/a/43584678/4590536 [we can use locale change listener]

Now, 
- Do I have to create key-value pairs for strings.xml?
- Which of the 2 methods (or any other method) should I use? (for un/subscribing)

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
nishantt95: Yes it is.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
nishantt95: I think 
`PreferenceManager.getDefaultSharedPreferences(context);` is different from `context.getSharedPreferences('cloud_testing', 0);` as their names are different so different files will be created (values will not be shared) (do correct me if I am wrong)

and with this approach, I was getting an issue that [this preference](https://github.com/nishantt95/Catroid/blob/f0bc32f9378506e671a38c161c1a42986a6df309/catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java#L45-45) (actual implementation -> context.getSharedPreferences('cloud_testing', 0);) is different from this [sharedPreference](https://github.com/nishantt95/Catroid/blob/f0bc32f9378506e671a38c161c1a42986a6df309/catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java#L79-79) and I was not able to set the sharedPrefernce of CloudMessaging.java from CloudMessagingTest.java.

> SharedPreferences implementation. Just with HashMaps and to get/set values synchronously. Every more complex method will just throw an UnsupportedOperationException.
> We could also provide a method to store and restore the SharedPreferences. This can be used here. Btw, such implementation is officially called a Fake 😉

Implementation part is still not clear as (used [here](https://github.com/Catrobat/Catroid/pull/2377/files#diff-c41b068c0724c4fe613f723e96219ba1R81)) is the normal one and can't figure out that how this one is different from the one we have used normally.

robertpainsi:
```java
…
sharedPreferences = new FakeSharedPreferences();
context = mock(Context.class);
when(context.getSharedPreferences(anyString(), anyInt())).thenReturn(sharedPreferences);
…
```

If the FakeSharedPreferences and FakeSharedPreferences.Editor are partially implemented by simply using Maps, everything else can stay the same.

> as (used here) is the normal one

The difference is that you have to restore the SharedPreferences when using the "normal" method.
If a test crashes or you debug through a test and press the stop button, `tearDown` is never called and therefore the SharedPreferences aren't restored.

Maybe mentioning the [_normal_ one](https://github.com/Catrobat/Catroid/pull/2377/files#diff-c41b068c0724c4fe613f723e96219ba1R81) was a little confusing. I just thought that saving and restoring the (default) SharedPreferences can also be a feature of the FakeSharedPreferences :wink:

robertpainsi:
Don't forget to update https://github.com/Catrobat/Catroid/blob/fa7f5e73e3024fc4f2c10a0df290a1994b73cfe8/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/SettingsActivityTest.java#L73 as well

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/NotificationSubscriber.java
```diff
@@ -0,0 +1,43 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.cloudmessaging;
+
+import com.google.firebase.messaging.FirebaseMessaging;
+
+public final class NotificationSubscriber {
+
+	private static final String DEFAULT_TOPIC = "DEFAULT_TOPIC";
```
robertpainsi: Hmm, `DEFAULT_TOPIC` isn't really meaningful. What to you think about `notification_enabled`?
Then you can send it to e.g. `'notification_enabled' in topics && 'language_en' in topics`

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
robertpainsi: Or maybe we just provide the functionality to save and restore SharedPreferences?

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
robertpainsi: > Do I have to create key-value pairs for strings.xml?

No, Crowdin would remove these strings anyway.
We can make sure that there is a language tag in every strings.xml file by using a unit test.

> Which of the 2 methods (or any other method) should I use? (for un/subscribing)

Both, I guess. With an additional change
- "whenever the user opens the ~app~ _MainMenuActivity_"

Changing the app language through the app (https://github.com/Catrobat/Catroid/pull/2332) will finish all activities and open the MainMenuActivity.

@cheidenreich, what do you think as well?

robertpainsi:
There is currently no test if CloudMessaging.initialize subscribes to the topics.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
nishantt95: In the current update, I have used mockito. Don't know if that's the right way. Do let me know 😕

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,83 @@
...
+public class CloudMessagingTest extends AndroidTestCase {
+
+	private Context context;
+	private SharedPreferences sharedPreferences;
+
+	@Before
+	public void setUp() {
+		sharedPreferences = mock(SharedPreferences.class);
+		context = mock(Context.class);
+		when(PreferenceManager.getDefaultSharedPreferences(context)).thenReturn(sharedPreferences);
```
robertpainsi: image

It… really… works!

k, I first thought maybe mocks don't work but stubs will. So I quickly tested it
```java
assertTrue(Foo.get());
when(Foo.get()).thenReturn(false);
assertFalse(Foo.get());
```
→ Mockito Exception :confused:

The reason why this works is because `PreferenceManager.getDefaultSharedPreferences` internally calls `context.getSharedPreferences` and context is mocked. So Mockito is smart enough to handle this case :smile:

That's great! This code is much cleaner! I like that a lot! Good work!

Maybe we can do some mocking in the CrashReporterTest as well instead of using the real SharedPreferences? However, it may need a fake (partially functional) SharedPreferences object.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/NotificationSubscriber.java
```diff
@@ -0,0 +1,43 @@
...
+package org.catrobat.catroid.cloudmessaging;
+
+import com.google.firebase.messaging.FirebaseMessaging;
+
+public final class NotificationSubscriber {
+
+	private static final String DEFAULT_TOPIC = "notification_enabled";
+
+	private NotificationSubscriber() {
+
```
robertpainsi: Please remove empty line (Checkstyle warning)

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -54,6 +60,24 @@ public void showNotification(CloudMessage cloudMessage) {
 		}
 	}
 
+	public static boolean initialize(Context context) {
+		if (isCloudNotificationsEnabled(context)) {
+			NotificationSubscriber.subscribeToFirebaseServices();
+			return true;
```
cheidenreich: i personally like it when there are as least exit points as possible.
so you could save the `isCloudNotificationsEnabled(context)` return value in a variable, then either sub or unsub and then return the variable value in the end.

@robertpainsi do you agree or doesn't it matter?

catroid/src/main/java/org/catrobat/catroid/ui/SettingsActivity.java
```diff
@@ -406,6 +427,13 @@ public static void setAutoCrashReportingEnabled(Context context, boolean isEnabl
 		}
 	}
 
+	public static void setCloudNotificationsEnabled(Context context, boolean isEnabled) {
+		SharedPreferences.Editor editor = getSharedPreferences(context).edit();
+		editor.putBoolean(SETTINGS_CLOUD_NOTIFICATIONS, isEnabled);
```
cheidenreich: you can directly put `.commit()` at the end of this. Editor calls always return the Editor again ;)

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
robertpainsi: How would you name such variable?

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -54,6 +60,28 @@ public void showNotification(CloudMessage cloudMessage) {
...
+
+		if (isCloudNotificationsEnabled(context)) {
+			isInitialized = true;
+			NotificationSubscriber.subscribeToFirebaseServices();
+		} else {
+			isInitialized = false;
+			NotificationSubscriber.unsubscribeFromFirebaseServices();
+		}
+		return isInitialized;
+	}
```
robertpainsi:
```java
public static boolean initialize(Context context) {
  return isCloudNotificationsEnabled(context);
}
```
will create a green testrun as well.

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
cheidenreich: exactly

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,103 @@
...
+	public void testCloudMessagingUninitializedOnNotificationsDisabledByUser() {
+		when(sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false)).thenReturn(false);
+
+		assertFalse(CloudMessaging.initialize(context));
+	}
+
+	@Test
+	public void testUserSubscribedToTopicsOnCloudMessagingInitialized() {
+		CloudMessaging.initialize(context);
+		verify(messagingService, times(1)).subscribe();
```
robertpainsi: Please merge `testCloudMessagingInitializedOnNotificationsEnabled` and `testCloudMessagingInitializedOnNotificationsEnabled` to a single test

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	@Test
+	public void testUserUnsubscribedToTopicsOnCloudMessagingInitializationFailed() {
+		when(sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false)).thenReturn(false);
+		CloudMessaging.initialize(context);
+
+		verify(messagingService, times(1)).unsubscribe();
```
robertpainsi: Please merge `testCloudMessagingUninitializedOnNotificationsDisabledByUser` and `testUserUnsubscribedToTopicsOnCloudMessagingInitializationFailed` to a single test

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
robertpainsi: c&p error
`testCloudMessagingInitializedOnNotificationsEnabled` and `testUserSubscribedToTopicsOnCloudMessagingInitialized` :wink:

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/NotificationSubscriber.java
```diff
@@ -0,0 +1,52 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.cloudmessaging;
+
+import android.support.annotation.VisibleForTesting;
+
+import com.google.firebase.messaging.FirebaseMessaging;
+
+public final class NotificationSubscriber {
```
robertpainsi: This class isn't needed anymore.

catroid/src/androidTest/java/org/catrobat/catroid/test/cloudmessaging/CloudMessagingTest.java
```diff
@@ -0,0 +1,91 @@
...
+	public void testUserSubscribedToTopicsOnCloudMessagingInitialized() {
+		assertTrue(CloudMessaging.initialize(context));
+		verify(messagingService, times(1)).subscribe();
+	}
+
+	@Test
+	public void testUserUnsubscribedToTopicsOnCloudMessagingInitializationFailed() {
+		when(sharedPreferences.getBoolean(SettingsActivity.SETTINGS_CLOUD_NOTIFICATIONS, false)).thenReturn(false);
+		assertFalse(CloudMessaging.initialize(context));
+
```
robertpainsi: Move this empty newline one line up. :wink:

catroid/src/main/java/org/catrobat/catroid/cloudmessaging/CloudMessaging.java
```diff
@@ -54,6 +61,30 @@ public void showNotification(CloudMessage cloudMessage) {
 		}
 	}
```
robertpainsi: Please move the `showNotification` to the `buildNotification` method. Things that belongs together shouldn't be separated.
I consider to split the show/build notification and the initialization into two classes. However, for now it's fine to me.

robertpainsi: Please don't inline the `showNotification` method. The `showNotification` method should be above the `buildNotification`


```java
…some unrelated methods

…more unrelated methods

showNotification()

buildNotification()
```

PR #2434, opened 2017-07-31
Refactored Pocket Paint Intent

thmq:


catroid/src/main/java/org/catrobat/catroid/ui/controller/LookController.java
```diff
@@ -468,35 +471,64 @@ public void loadPictureFromLibraryIntoCatroid(String filePath, Activity activity
...
-					});
-			AlertDialog alert = builder.create();
-			alert.show();
-			return false;
-		}
-		return true;
+		return !packages.isEmpty();
+	}
+
+	public void installAndOpenPocketPaint(final Intent pocketPaintIntent, final Activity activity, final int
```
thmq: I really hate having to pass through all these parameters, but the current structure (with the controller/fragment/adapter and dialogs) doesn't really allow for a more elegant way. The whole setup should (and hopefully will) be changed anyway, so I did not really see any reason to spend too much time on it.

catroid/src/androidTest/java/org/catrobat/catroid/uitest/ui/dialog/NewSpriteDialogTest.java
```diff
@@ -67,20 +63,6 @@ protected void tearDown() throws Exception {
 	}
 
 	public void testNewSpriteDialogStep1() throws Exception {
-		UiTestUtils.getIntoSpritesFromMainMenu(solo);
```
thmq: I have removed the ui tests for now. If I've understood the priority of this PR correctly, there was no time to add new espresso tests for the functionality. @cheidenreich, is that so? or should I add before we merge it?

catroid/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,7 @@
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
+    <uses-permission android:name="android.permission.REORDER_TASKS" />
```
thmq: This permission allows the app to reorder applications on the android activity stack. For devices using an API version below ~ 5.0 (not sure about the exact version) this is necessary in order to put pocket paint to the correct position on the stack. The permission is not considered a (great risk)[https://developer.android.com/guide/topics/permissions/normal-permissions.html] by Google, so I'm sure we'll be fine using it.

cheidenreich:
Hmm I just tried this on my phone, unfortunately it does not work entirely :(
If i have NOT installed Pocket Paint on my device and i want to draw something, it correctly asks if i want to install it -> yes -> play store opens and i install it -> pocket paint opens and I draw something -> go back -> save? -> yes
Pocket Code reopens, but it still shows the "New Object" dialog and my drawing is lost.
Any further attempt when pocket paint is installed works as expected.

thmq:
I have to have another look at it, I was expecting some problems with the dialogs. I just wanted to push this first version and ask about the urgency of this pr. 😉

cheidenreich:
would have been nice to have in this release, if it's easy to do, but if it's too much work i don't think it's worth it to wait (also it might be double work with the refactoring 😺)

how do you think about this?

thmq:
As far as I have checked, it now works, so we could have a look at it today and if everything is fine, we can still put it into the release.

wslany:
Please, yes, in the release...

re-cheid:
OK, I just tested it again, the bug from before is fixed :)

However a small thing is still not working entirely:
Add object -> Draw image -> play store opens -> press back (without installing) and click draw image again -> now install -> it opens, draw something, go back and save -> another pocket paint instance seems to get opened -> back -> pocket code gets the first drawing.

cheidenreich:
OK, I just tested it again, the bug from before is fixed :)

However a small thing is still not working entirely:
Add object -> Draw image -> play store opens -> press back (without installing) and click draw image again -> now install -> it opens, draw something, go back and save -> another pocket paint instance seems to get opened -> back -> pocket code gets the first drawing.

cheidenreich:
We also have to handle the case where no Google Play is installed on the device

cheidenreich:
implemented said changes on top of this and opened a new PR #2465

PR #2442, opened 2017-08-05
CAT-FIX NullPointerException At ProjectListFragment

nishantt95:
Fixed NullPointerException of onLoadProjectSuccess

Error:- Exception java.lang.Throwable: Attempt to invoke virtual method
'java.lang.String android.content.Context.getPackageName()' on a null
object reference

Error url:-
https://console.firebase.google.com/u/0/project/firebase-catrobat/monitoring/app/android:org.catrobat.catroid/cluster/077bd744?duration=2592000000

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/MyProjectsActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/MyProjectsActivity.java
@@ -153,4 +153,10 @@ public class MyProjectsActivity extends BaseCastActivity {
 
 		item.setTitle(showDetails ? R.string.hide_details : R.string.show_details);
 	}
+
+	@Override
+	public void onBackPressed() {
+		projectListFragment.cancelLoadProjectTask();
+		super.onBackPressed();
+	}
 }
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
@@ -80,6 +80,7 @@ public class ProjectListFragment extends ListActivityFragment implements LoadPro
 	private List<ProjectData> projectList;
 	private ProjectData projectToEdit;
 	private int selectedProjectPosition;
+	private LoadProjectTask loadProjectTask;
 
 	@Override
 	public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
@@ -231,7 +232,7 @@ public class ProjectListFragment extends ListActivityFragment implements LoadPro
 
 	@Override
 	public void handleOnItemClick(int position, View view, ProjectData listItem) {
-		LoadProjectTask loadProjectTask = new LoadProjectTask(getActivity(), listItem.projectName, true, false);
+		loadProjectTask = new LoadProjectTask(getActivity(), listItem.projectName, true, false);
 		loadProjectTask.setOnLoadProjectCompleteListener(this);
 		getActivity().findViewById(R.id.fragment_container).setVisibility(View.GONE);
 		loadProjectTask.execute();
@@ -398,5 +399,11 @@ public class ProjectListFragment extends ListActivityFragment implements LoadPro
 	@Override
 	public void onLoadProjectFailure() {
 		getActivity().findViewById(R.id.fragment_container).setVisibility(View.VISIBLE);
+	}
+
+	public void cancelLoadProjectTask() {
+		if (loadProjectTask != null) {
+			loadProjectTask.cancel(true);
+		}
 	}
 }


robertpainsi:
Please let me know how to reproduce the error in the first place

robertpainsi:
Refer to the crash report in the commit message

nishantt95:
Error: 
Exception java.lang.Throwable: Attempt to invoke virtual method 'java.lang.String android.content.Context.getPackageName()' on a null object reference

error at file ProjectListFragment.java on `Intent intent = new Intent(getActivity(), ProjectActivity.class);` where `getActivity()` was getting null and hence the above mentioned error.

robertpainsi:
Do you know a step by step guide on how to reproduce the error on a device?

> Also refer to the crash report in the _**commit message**_

nishantt95:
No. I just found the reason of such error on the web and then tried to solve it.

robertpainsi:
I think the fix is not correct.

`onLoadProjectSuccess` is called after a project has been loaded. The `LoadProjectTask` (which is started in `handleOnItemClick`) triggers this callback. So the list was created ([`getActivity != null`](https://github.com/Catrobat/Catroid/blob/26ee9c28c2c656065c9abd7fac6044b14d32da17/catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java#L125)) and the user clicked on an item.
The problem is, that `onLoadProjectSuccess` is called after the Fragment has been detached from the Activity. In other words, the callback is called when the user is already in another activity. e.g. if the loading takes too long and the user goes back to the MainMenuActivity.

nishantt95:
> The problem is, that onLoadProjectSuccess is called after the Fragment has been detached from the Activity. In other words, the callback is called when the user is already in another activity. e.g. if the loading takes too long and the user goes back to the MainMenuActivity.

In that case, just handling the exception (and not performing anything) will work as the user is no longer interested in that project?

robertpainsi:
I don't know which event is best to use to cancel the `LoadProjectTask`. Maybe `onDetach`.
Or you early exit in `onLoadProjectFailure` and `onLoadProjectSuccess`

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -392,15 +391,13 @@ public void run() {
 
 	@Override
 	public void onLoadProjectSuccess(boolean startProjectActivity) {
-		if (context == null && getActivity() == null) {
-			Log.e(TAG, "Null Value at onLoadProjectSuccess");
+		if (getActivity() == null) {
+			Log.e(TAG, "Null Value of getActivity() at onLoadProjectSuccess");
```
robertpainsi: You don't need to check for `null` because the Task which triggers this callback should have been canceled anyway.
If the problem still exists, Crashlytics will tell us.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -399,4 +400,10 @@ public void onLoadProjectSuccess(boolean startProjectActivity) {
 	public void onLoadProjectFailure() {
 		getActivity().findViewById(R.id.fragment_container).setVisibility(View.VISIBLE);
 	}
+
+	@Override
+	public void onDetach() {
+		super.onDetach();
+		loadProjectTask.cancel(true);
```
robertpainsi: Not tested, but does this throw a NullPointerException if the user doesn't load a project and goes back to the MainMenuActivity?

nishantt95: Yes, it throws NullPointerException. Applied a null checker for loadProjectTask.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -399,4 +400,12 @@ public void onLoadProjectSuccess(boolean startProjectActivity) {
...
 		getActivity().findViewById(R.id.fragment_container).setVisibility(View.VISIBLE);
 	}
+
+	@Override
+	public void onDetach() {
+		super.onDetach();
+		if (loadProjectTask != null) {
+			loadProjectTask.cancel(true);
+		}
+	}
```
robertpainsi: `onDestroy` or `onDetach`? ping @cheidenreich

robertpainsi: Please fix the problem by using `MyProjectsActivity.onBackPressed`

catroid/src/main/java/org/catrobat/catroid/ui/MyProjectsActivity.java
```diff
@@ -153,4 +153,13 @@ private void handleShowDetails(boolean showDetails, MenuItem item) {
 
 		item.setTitle(showDetails ? R.string.hide_details : R.string.show_details);
 	}
+
+	@Override
+	public void onBackPressed() {
+		if (projectListFragment != null) {
+			projectListFragment.cancelLoadProjectTask();
+			Log.d(TAG, "Back press detected!");
+		}
```
robertpainsi: projectListFragment is always set, no null check needed
Please remove log output

robertpainsi:
test this please

robertpainsi:
please run espresso tests

robertpainsi:
test this please

robertpainsi:
please run espresso tests

robertpainsi:
test this please

robertpainsi:
please run espresso tests

PR #2443, opened 2017-08-05
CAT-FIX Resolve ShowDialog Error On ProjectManager

nishantt95:
Fixed error which was occured on calling Utils.showErrorDialog at
ProjectManager.java

Error url:- https://console.firebase.google.com/u/0/project/firebase-catrobat/monitoring/app/android:org.catrobat.catroid/cluster/09ea9d7c?duration=2592000000&appVersions=35

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
@@ -399,8 +399,8 @@ public final class Utils {
 			public void onClick(DialogInterface dialog, int which) {
 			}
 		});
-		Dialog errorDialog = builder.create();
-		errorDialog.show();
+		Dialog dialog = builder.create();
+		showDialog(context, dialog);
 	}
 
 	public static void showErrorDialog(Context context, String msg, int errorTitleId) {
@@ -412,8 +412,8 @@ public final class Utils {
 			public void onClick(DialogInterface dialog, int which) {
 			}
 		});
-		Dialog errorDialog = builder.create();
-		errorDialog.show();
+		Dialog dialog = builder.create();
+		showDialog(context, dialog);
 	}
 
 	public static void showErrorDialog(Context context, int errorTitleId, int errorMessageId) {
@@ -425,8 +425,8 @@ public final class Utils {
 			public void onClick(DialogInterface dialog, int which) {
 			}
 		});
-		Dialog errorDialog = builder.create();
-		errorDialog.show();
+		Dialog dialog = builder.create();
+		showDialog(context, dialog);
 	}
 
 	public static String md5Checksum(File file) {
@@ -477,6 +477,19 @@ public final class Utils {
 	public static double round(double value, int precision) {
 		final int scale = (int) Math.pow(10, precision);
 		return (double) Math.round(value * scale) / scale;
+	}
+
+	private static void showDialog(Context context, final Dialog dialog) {
+		if (context instanceof Activity) {
+			((Activity) context).runOnUiThread(new Runnable() {
+				@Override
+				public void run() {
+					dialog.show();
+				}
+			});
+		} else {
+			dialog.show();
+		}
 	}
 
 	private static String toHex(byte[] messageDigest) {


robertpainsi:
Is there a test which tests if an error is shown if the default project couldn't be initialized?

nishantt95:
No

robertpainsi:
Ah, k. Please add one :wink:

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -314,7 +314,7 @@ public void saveProject(Context context) {
 		}
 	}
 
-	public boolean initializeDefaultProject(Context context) {
+	public boolean initializeDefaultProject(final Context context) {
```
robertpainsi: > Things to avoid when creating commits
>
> * Mixing two unrelated functional changes.
> * …

If you fix a bug, always only add all necessary changes to the commit but nothing more.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -479,6 +476,15 @@ public static double round(double value, int precision) {
 		return (double) Math.round(value * scale) / scale;
 	}
 
+	private static void displayDialog(Context context, final Builder builder) {
+		((Activity) context).runOnUiThread(new Runnable() {
```
robertpainsi: The cast is just a quick-fix and will throw an exception if the Context can't be casted to `Activity`. Please fix appropriately.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -479,6 +476,15 @@ public static double round(double value, int precision) {
...
+			public void run() {
+				Dialog errorDialog = builder.create();
```
robertpainsi: Instead of creating the Dialog on the UiThread, pass a Dialog object instead of a Builder object.
If you want to pass in a Builder, just overload the `displayDialog` function :wink:

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -479,6 +481,15 @@ public static double round(double value, int precision) {
 		return (double) Math.round(value * scale) / scale;
 	}
 
+	private static void displayDialog(final Dialog errorDialog) {
```
robertpainsi: What do you think about `showDialog`? It's consistent with the dialog interface, `dialog.show()`
Btw, just name the parameter dialog, the method is not just limited to an error dialog.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
robertpainsi: I took another look at the changes. The quick-fix is currently the only solution. Please use `((Activity) context).runOnUiThread`.

The problem is that methods like `getCurrentProjectName`, `copyImageToCatroid` and so on interact with the ui. They definitely should *not* interact with the ui. Fixing it properly would need a looot of rewrite. So I'm fine with the quick-fix for now. Please revert to your initial version. Sorry!
No tests will be needed! :wink:

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -479,6 +479,15 @@ public static double round(double value, int precision) {
...
 
+	private static void showDialog(Context context, final Dialog dialog) {
+		((Activity) context).runOnUiThread(new Runnable() {
+			@Override
+			public void run() {
+				dialog.show();
+			}
+		});
+	}
+
```
robertpainsi: Maybe we should do
```java
if (context instanceof Activity) {
  ((Activity) context).runOnUiThread(…);
} else {
  dialog.show();
}
```

* If the context is an activity, we can call `runOnUiThread`. `runOnUiThread` will either call the code immediately, if the current thread is the UiThread, or postpone it. (this fixes the crash)
* If the context isn't an activity but we're currently executing the code on the UiThread, we can call `dialog.show()`
* If the context isn't an activity and we are not on the UiThread, an exception will be thrown.

This is currently just a dirty, non-oo quick-fix.

nishantt95:
Sorry! closed it by mistake 😞

robertpainsi:
test this please

robertpainsi:
please run espresso tests

robertpainsi:
Same tests fail on develop branch.
[default @develop](https://jenkins.catrob.at/job/CatroidTriggerPullRequestTest/2506)
[espresso tests @develop](https://jenkins.catrob.at/job/CatroidTriggerPullRequestEspressoTest/51/)

PR #2455, opened 2017-08-20
CAT-FIX ProgressDialog Error On Activity Finished

nishantt95:
progressDialog.show() was throwing an exception on activity finished.

Error URL:
https://console.firebase.google.com/u/0/project/firebase-catrobat/monitoring/app/android:org.catrobat.catroid/cluster/e0c21001

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
@@ -337,4 +337,10 @@ public class WebViewActivity extends BaseActivity {
 			return false;
 		}
 	}
+
+	@Override
+	protected void onDestroy() {
+		webView.setDownloadListener(null);
+		super.onDestroy();
+	}
 }


catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -337,4 +345,24 @@ private boolean isWhatsappInstalled() {
...
+
+	@Override
+	protected void onDestroy() {
+		super.onDestroy();
+		if (progressDialog == null) {
+			return;
+		}
+		dismissProgressDialog();
+		progressDialog = null;
+		webView.setDownloadListener(null);
```
robertpainsi: IMO only setting the DownloadListener to null is sufficient at first. If the error still occurs after the fix, we need to take a look at it again. What do you think?

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -337,4 +337,10 @@ private boolean isWhatsappInstalled() {
 			return false;
 		}
 	}
+
+	@Override
+	protected void onDestroy() {
+		super.onDestroy();
```
robertpainsi: Please move `super.onDestroy` to the end of the method.

robertpainsi:
test this please

robertpainsi:
please run espresso tests

PR #2461, opened 2017-08-22
CAT-FIX Null Value Of getActivity() at ProjectListFragment

nishantt95:
getActivity() is null while calling runOnUiThread() method.

Error URL:
https://console.firebase.google.com/u/0/project/firebase-catrobat/monitoring/app/android:org.catrobat.catroid/cluster/e70cc96e

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
@@ -22,8 +22,10 @@
  */
 package org.catrobat.catroid.ui.fragment;
 
+import android.app.Activity;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.os.AsyncTask;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.util.Log;
@@ -81,6 +83,7 @@ public class ProjectListFragment extends ListActivityFragment implements LoadPro
 	private ProjectData projectToEdit;
 	private int selectedProjectPosition;
 	private LoadProjectTask loadProjectTask;
+	private AsyncTask projectInitializer;
 
 	@Override
 	public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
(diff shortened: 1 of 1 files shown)


robertpainsi:
You can reproduce the error by adding a wait
```java
…
projectManager.initializeDefaultProject(getActivity());
long stop = System.currentTimeMillis() + 5000;
while (System.currentTimeMillis() < stop) ;
getActivity().runOnUiThread(new Runnable() {
…
```
* Delete all you programs
* A loading spinner appears. Immediately press back.
* Back in the MainMenuActivity just wait a few seconds. :boom: `getActivity() == null`.

So I guess that `projectManager.initializeDefaultProject(getActivity());` just takes too long on some devices.
Maybe use a AsyncTask instead and cancel it onDestroy. Please do some research here.

robertpainsi:
[`onPreExecute()`](https://developer.android.com/reference/android/os/AsyncTask.html#onPreExecute()) and [`onPostExecute(Result result)`](https://developer.android.com/reference/android/os/AsyncTask.html#onPostExecute(Result)) will come handy for updating the UI because these methods run on the UI thread. You can use them to show/hide the loading spinner logic.

Please inline the variable `projectManager`. It is only used once.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -369,24 +372,23 @@ public boolean onOptionsItemSelected(MenuItem item) {
 	}
 
 	private void initializeDefaultProjectAfterDelete() {
-		final ProjectManager projectManager = ProjectManager.getInstance();
 		getActivity().findViewById(R.id.fragment_container).setVisibility(View.GONE);
 		getActivity().findViewById(R.id.progress_circle).setVisibility(View.VISIBLE);
```
robertpainsi: IMO the code is easier to understand when using onPreExecute() for this code.
* pre (Show loading spinner)
* run (init default project and update list)
* post (Hide loading spinner)

What do you think?

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -369,24 +372,23 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-		Runnable r = new Runnable() {
+
+		asyncTask = new AsyncTask<Void, Void, Activity>() {
```
robertpainsi: Please rename `asyncTask` to something more meaningful.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -369,24 +372,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+				return getActivity();
+			}
+
+			@Override
+			protected void onPostExecute(Activity activity) {
+				activity.findViewById(R.id.fragment_container).setVisibility(View.VISIBLE);
+				activity.findViewById(R.id.progress_circle).setVisibility(View.GONE);
+				initializeList();
+			}
+		}.execute();
```
robertpainsi: I took a look at this. IMO the best way to handle it is using a variable for the Activity.
```java
final Activity activity = getActivity();
projectInitializer = new AsyncTask<Void, Void, Void>() {
…
```
In `onPreExecute` and `onPostExecute` there is no null check needed because these methods run on the ui thread.
`doInBackground` is only used as context.

However and JFI, using `getActivity()` in `doInBackground` may result in a NPE.

nishantt95: So, should I use a variable "activity" and apply the null checker in "doInBackground'?

robertpainsi: variable - yes, null check - no.

The benefit of using a variable here is that `initializeDefaultProject` will only use the activity as context. So passing in a finished activity should be fine.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ProjectListFragment.java
```diff
@@ -369,24 +372,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+			@Override
+			protected void onPreExecute() {
+				activity.findViewById(R.id.fragment_container).setVisibility(View.GONE);
+				activity.findViewById(R.id.progress_circle).setVisibility(View.VISIBLE);
+			}
+
+			@Override
+			protected Activity doInBackground(Void... params) {
+				ProjectManager.getInstance().initializeDefaultProject(activity);
+				return activity;
```
robertpainsi: IMO there is no value in returning activity.
The return value of `doInBackground` is the result of the Task. Since the task has no result (just side-effects), it's fine to return `null`. Don't forget to update `onPostExecute`.

robertpainsi:
test this please

robertpainsi:
please run espresso tests

PR #2466, opened 2017-08-24
CAT-2430 Negative Value Sets Object In Front Of PasspartOut

nishantt95:
Negative value sets at GoBackN makes the object in front of PasspartOut
and in actual no object needs to be in fron of the passpartout.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
@@ -54,6 +54,7 @@ public class GoNStepsBackAction extends TemporalAction {
 			toFront(delta);
 		} else {
 			goNStepsBack(stepsValue.intValue());
+			toFront(delta);
 		}
 	}
 


catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java
```diff
@@ -54,6 +54,7 @@ protected void update(float delta) {
 			toFront(delta);
 		} else {
 			goNStepsBack(stepsValue.intValue());
+			toFront(delta);
```
robertpainsi: This won't fix the problem. This will just move the object to the front. Every other object will be behind.
However, the user may just want to move the object one single layer (back or forth).

robertpainsi: This will also screw up Go back with a positiv number.

nishantt95: > This will also screw up Go back with a positive number.

You mean that positive number will also cause the same problem? or I should try with making it a positive number? (allowing the positive numbers only)

robertpainsi: The Go back brick should work like this
```
front
|P|o|o|o|o|X|o|o|o|o|
Go back 1
|P|o|o|o|o|o|X|o|o|o|
Go back -2
|P|o|o|o|X|o|o|o|o|o|
Go back -100
|P|X|o|o|o|o|o|o|o|o|

X = The object executing the Go back brick.
P = Passpartout
o = Any object
```
There is also special handling for e.g. `Go back 100`. The background object and also a Stamp (or pen?) object are always the last objects. [GoNStepsBackAction.java#L51-L52](https://github.com/Catrobat/Catroid/blob/d9efbd44e3b91c33e452153188cdffffcf521656/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java#L51-L52)

The following is just an overflow check. [GoNStepsBackAction.java#L53-L54](https://github.com/Catrobat/Catroid/blob/d9efbd44e3b91c33e452153188cdffffcf521656/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java#L53-L54)

From the [CAT-2430](https://jira.catrob.at/browse/CAT-2430)
> Instead of fixing the GoNStepsBack brick, maybe it's possible to use a Group instead. This will fix the problem with the Passpartout once and for all.

IMO a quick-fix is fine but a real fix is much more appreciated. There have been a few problems with the Passpartout and Background z-index already.

nishantt95: I checked `goNStepsBack()` function and [here](https://github.com/Catrobat/Catroid/blob/d9efbd44e3b91c33e452153188cdffffcf521656/catroid/src/main/java/org/catrobat/catroid/content/actions/GoNStepsBackAction.java#L67) if we set a negative value (-ve value set to goNStepsBack() function) then it actually adds up the value in positive number and sets that as its index.

For Example:-

If I set goBackNSteps to -200 then it sets the value to 202 index.
`int newSpriteZIndex = Math.max(zPosition - steps, Constants.Z_INDEX_FIRST_SPRITE);`

-> `Math.max(2 - (-200), const...)` -> 202 as newSpriteZIndex.

As I applied Math.abs() then  `Math.max(2 - Math.abs(-200), const...)` -> 2 [const...] as newSpriteZIndex.

So, Is `Math.abs()`  is the right solution? [Though it worked and image is behind the passpartout]

robertpainsi: With `Math.abs` the object can only go back (GoBack(-2) == GoBack(2)).
When using GoBack with a negative value, the object should go forward.

`Z_INDEX_FIRST_SPRITE` is the sprite index after the background (and stamp) not the passpartout.
The background has always z-index 0, the stamp z-index 1.
Sprites start at z-index 2. The passpartout has z-index #actors.

robertpainsi:
Does not solve the problem → Closed

PR #2467, opened 2017-08-24
CAT-2521 Dynamic textsize and icons for bricks

FlorianWeissensteiner:
#CAT-2521
https://jira.catrob.at/browse/CAT-2521

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/ThinkForBubbleBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/ThinkForBubbleBrick.java
@@ -25,7 +25,6 @@ package org.catrobat.catroid.content.bricks;
 import android.content.Context;
 import android.util.Log;
 import android.view.View;
-import android.view.ViewGroup;
 import android.widget.BaseAdapter;
 import android.widget.TextView;
 
@@ -39,8 +38,6 @@ import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.formulaeditor.Formula;
 import org.catrobat.catroid.formulaeditor.InterpretationException;
 import org.catrobat.catroid.ui.fragment.FormulaEditorFragment;
-import org.catrobat.catroid.utils.IconsUtil;
-import org.catrobat.catroid.utils.TextSizeUtil;
 import org.catrobat.catroid.utils.Utils;
 
 import java.util.List;
@@ -101,14 +98,6 @@ public class ThinkForBubbleBrick extends FormulaBrick {
 		view = View.inflate(context, layoutId, null);
 		view = BrickViewProvider.setAlphaOnView(view, alphaValue);
 
-		if (type == Constants.SAY_BRICK) {
-			IconsUtil.addIcon(context, (TextView) view.findViewById(R.id.brick_say_for_bubble_label),
-					context.getString(R.string.category_looks));
-		} else {
-			IconsUtil.addIcon(context, (TextView) view.findViewById(R.id.brick_think_for_bubble_label),
-					context.getString(R.string.category_looks));
-		}
-
 		setCheckboxView(checkboxId);
 
 		TextView editText = (TextView) view.findViewById(editTextId);
@@ -140,7 +129,6 @@ public class ThinkForBubbleBrick extends FormulaBrick {
 		}
 
 		editDuration.setOnClickListener(this);
-		TextSizeUtil.enlargeViewGroup((ViewGroup) view);
 		return view;
 	}
 
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
@@ -90,6 +90,7 @@ import org.catrobat.catroid.content.bricks.LegoNxtMotorStopBrick;
 import org.catrobat.catroid.content.bricks.LegoNxtMotorTurnAngleBrick;
 import org.catrobat.catroid.content.bricks.LegoNxtPlayToneBrick;
 import org.catrobat.catroid.content.bricks.MoveNStepsBrick;
+import org.catrobat.catroid.content.bricks.NestingBrick;
 import org.catrobat.catroid.content.bricks.NextLookBrick;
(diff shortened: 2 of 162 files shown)


catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
```diff
@@ -102,6 +106,25 @@ public static void mapIconSizesToDeviceSize() {
 		largeIconBottomBar = (int) (metrics.ydpi * (LARGE_ICON_SIZE_BOTTOM_BAR_IN_CM * INCH_TO_CM));
 	}
 
+	public static void addIconToLabel(ViewGroup parent, Brick brick, Sprite sprite, Context context) {
+		for (int i = 0; i < parent.getChildCount(); i++) {
+			View child = parent.getChildAt(i);
+			if (child instanceof LinearLayout || child instanceof RelativeLayout || child instanceof BrickLayout) {
```
sjaindl: You could simply check for ViewGroup here (see my comment on #2485 )

catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
```diff
@@ -102,6 +106,25 @@ public static void mapIconSizesToDeviceSize() {
...
+				addIconToLabel((ViewGroup) child, brick, sprite, context);
+			}
+			if (child instanceof TextView) {
+				TextView textView = (TextView) child;
+				CategoryBricksFactory categoryBricksFactory = new CategoryBricksFactory();
+
+				String id = context.getResources().getResourceEntryName(textView.getId());
+				String subString = id.subSequence(id.length() - LABEL_KEYWORD.length(), id.length()).toString();
```
sjaindl: This is potentially subject to IndexOutOfBoundsException if someone introduces a TextView with a really short id. Please add a check whether id size > LABEL_KEYWORD.length() and ignore these entries.

sjaindl:
please rebase

catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
sjaindl: or can this be applied to each BrickTextView?
This would make it easier since you could just check for instance of BrickTextView and get rid of dealing with substrings.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -737,6 +742,16 @@ public String getBrickCategory(Brick brick, Sprite sprite, Context context) {
...
+		} else if (brick instanceof IfLogicElseBrick) {
+			category = res.getString(R.string.category_control);
+		} else if (brick instanceof IfLogicEndBrick) {
+			category = res.getString(R.string.category_control);
+		} else if (brick instanceof IfThenLogicEndBrick) {
+			category = res.getString(R.string.category_control);
+		} else if (brick instanceof LoopEndlessBrick) {
+			category = res.getString(R.string.category_control);
+		} else if (brick instanceof LoopEndBrick) {
+			category = res.getString(R.string.category_control);
```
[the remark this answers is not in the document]
FlorianWeissensteiner: True that. Fixed.

catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
FlorianWeissensteiner: Fixed².

catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
FlorianWeissensteiner: Not to each because some bricks have multiple BrickTextViews. But it just dawned on me that I can simply only apply it to the first BrickTextView in every Brick, instead of applying it to the one with a "_label" keyword. I did that, should be the safer option.

FlorianWeissensteiner:
@sjaindl Thanks again for the review.
Same question, should the espresso test be an extra ticket so this can be merged?

sjaindl:
Same answer ;)
IMO the test(s) should bei implemented within this PR

catroid/src/main/java/org/catrobat/catroid/content/bricks/ThinkForBubbleBrick.java
```diff
@@ -95,20 +92,12 @@ public View getView(Context context, int brickId, BaseAdapter baseAdapter) {
 				.brick_think_for_bubble_edit_text_text;
 		int editDurationId = type == Constants.SAY_BRICK ? R.id.brick_say_for_bubble_edit_text_duration : R.id
 				.brick_think_for_bubble_edit_text_duration;
-		int thinkSaySecondsLabelId = type == Constants.SAY_BRICK ? R.id.brick_say_for_bubble_seconds_label : R.id
-				.brick_think_for_bubble_seconds_label;
+		int thinkSaySecondsLabelId = type == Constants.SAY_BRICK ? R.id.brick_say_for_bubble_seconds_text_view : R.id
+				.brick_think_for_bubble_seconds_text_view;
```
cheidenreich: why did the other ids changed from _text_view to _ label and this one the other way around?

cheidenreich: please resolve the conflicts

FlorianWeissensteiner: My first approach with the icons was to add them to the TextView with the '_label' keyword. The ThinkForBubbleBrick however, had two TextViews with the '_label' keyword, so it had two icons. To resolve this i removed the '_label' from the second TextView.
Since then I changed it, so that it just adds the icon to the first TextView instead of looking for the keyword. 
So now it doesn't matter anymore if it has a '_label' in it or not, but for consistency's sake it doesn't hurt for the first TextView to have that keyword, as most TextViews in the Bricks have it. I can of course change it back again if you wish.

Conflicts resolved.

84n4n4:
im not really up to speed on what the status of this is...
as christophs comment seems to be gone after last push, is this ok to merge now?

PR #2481, opened 2017-09-03
CAT-2523 Restricted login dialog (create@school)

sjaindl:


catroid/build.gradle
```diff
@@ -106,6 +106,7 @@ android {
     }
 
     defaultConfig {
+        testApplicationId 'org.catrobat.catroid.test'
```
84n4n4: why do we that?

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+		for (String setting : allSettings) {
+			initialSettings.put(setting, sharedPreferences.getBoolean(setting, false));
+		}
+
+		SharedPreferences.Editor editor = sharedPreferences.edit();
+		editor.putBoolean(Constants.RESTRICTED_LOGIN_ACCEPTED, false);
+		editor.putBoolean(Constants.CREATE_AT_SCHOOL_USER, true);
+		editor.apply();
+
+		Reflection.setPrivateField(Constants.class, "DEFAULT_ROOT", Environment.getExternalStorageDirectory()
```
84n4n4: why does this have to run in the default root again? if thats really necessary, theres probably nicer ways to achieve that.

the default root is reflected out (the only reflection that is currently in place in the whole espresso setup) in the BaseActivityInstrumentationRule to be "pocket code ui test". 
i did this because in the old setup, the setup went to the project folder and zipped and deleted all projects, and then in teardown unzipped everything again and deleted the zip.
i thought this isnt actually that nice because it leads to a lot of storage operations and zipping on every single test, also if the instrumenter crashes, your projects probably going to stay zipped in your folder, etc...

so i switched to reflecting the DEFAULT_ROOT to "pocket code ui test" folder, so that every test just deletes this before the test runs. which has the nice effect also that if you run a single test, you can inspect the project xmls, as they are still in place after the test has finished, so only the next test would clean that up in startup.

if for some reason you do really really need to use the production "pocket code" folder, you should write your own ActivityInstrumentationRule that does not do the reflection and deals with existing projects itself.

so in short:
as you do here is not nice, first, because you could just get rid of two reflections by making your own ActivityInstrumentationRule, and second, which also applies if you do your own rule, you have to deal with existing projects and to cleanup. (keep in mind cornercases where you already have a project with exactly this name, etc...)
so if theres no real good reason to do the reflection here, just keep the "pocket code uitests" one and work around it another way.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+import android.support.test.espresso.intent.Intents;
+import android.support.test.espresso.matcher.ViewMatchers;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.common.Constants;
+import org.catrobat.catroid.createatschool.ui.CreateAtSchoolMainMenuActivity;
+import org.catrobat.catroid.test.utils.Reflection;
+import org.catrobat.catroid.uiespresso.util.BaseActivityInstrumentationRule;
+import org.catrobat.catroid.uitest.util.UiTestUtils;
```
84n4n4: please NO imports and usages of solo utility functions. its a mess over there...
please refactor / reimplement the utility function you need and put it where you need it. (so either in your test package or even in the test itself, depending on how many tests need it)

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+		Reflection.setPrivateField(Constants.class, "DEFAULT_ROOT", Environment.getExternalStorageDirectory()
+				.getAbsolutePath() + "/Pocket Code");
+
+		baseActivityTestRule.launchActivity(null);
+
+		idlingResource = baseActivityTestRule.getActivity().getIdlingResource();
+		Espresso.registerIdlingResources(idlingResource);
+
+		UiTestUtils.setFakeToken(baseActivityTestRule.getActivity(), false);
```
84n4n4: see my comment on the import at the top.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+		onView(allOf(withText(R.string.login), withResourceName("alertTitle")))
+				.check(matches(isDisplayed()));
+	}
+
+	@Test
+	public void restrictedLoginDialogLinkTest() {
+		String link = Locale.getDefault().getLanguage().equals(Locale.GERMAN.getLanguage())
+				? Constants.RESTRICTED_LOGIN_LINK_GERMAN
+				: baseActivityTestRule.getActivity().getString(R.string.schoolreg_link);
```
84n4n4: please use getResourcesString from uitestutils

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+		intended(expectedIntent);
+		Intents.release();
+	}
+
+	@Test
+	public void restrictedLoginDialogNotShownAfterConfirmationTest() {
+		onView(withId(android.R.id.button1))
+				.perform(click());
+
+		relaunchActivity();
```
84n4n4: as already stated in one of the other PRs im not a huge fan of activity switching and restarting in tests.

is there an easier way to check this? set whatever flag or shared preference that remembers that the dialog has been discarded, and then launch activity to just check that the dialog is not displayed on launch?

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+		onView(withText(R.string.dialog_restricted_login_title))
+				.check(doesNotExist());
+	}
+
+	@Test
+	public void restrictedLoginDialogReappearsAfterLogoutTest() {
+		UiTestUtils.setFakeToken(baseActivityTestRule.getActivity(), true);
+		relaunchActivity();
```
84n4n4: same here

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+				.check(matches(isDisplayed()));
+	}
+
+	@Test
+	public void loginWithOtherThanCreateAtSchoolUserTest() {
+		ServerCalls.useTestUrl = true;
+		String testUserName = "testUser" + System.currentTimeMillis();
+		String testPassword = "password";
+		String testUserMail = testUserName + "@catrob.at";
+		UiTestUtils.createValidUserWithCredentials(baseActivityTestRule.getActivity(), testUserName, testPassword, testUserMail);
```
84n4n4: solo utility function, see my comment on top at the import statement

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+		onView(withId(android.R.id.button1))
+				.perform(click());
+
+		onView(withId(R.id.dialog_login_username))
+				.perform(clearText(), typeText(testUserName));
```
84n4n4: add closeSoftKeyboard() as additional action after typeText.
i found that tests become increasingly flake and some really hard to find errors happen because sometimes the softkeyboard stays open and covers the next uielement you want to access.

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -282,6 +282,10 @@
 	public static final String ZIP_FILE_NAME = START_PROJECT + ".zip";
 	public static final String STANDALONE_PROJECT_NAME = BuildConfig.PROJECT_NAME;
 
+	//Restricted login
+	public static final String RESTRICTED_LOGIN_ACCEPTED = "restrictedLoginAccepted";
+	public static final String RESTRICTED_LOGIN_LINK_GERMAN = "http://catrob.at/schulregistrierung";
```
84n4n4: why not move this to strings xml?

catroid/build.gradle
sjaindl: This is because the flavored tests wouldn't compile without that magic line.
Otherwise the .R files would be generated with package name org.catrobat.catroid.createatschool.test.R
However, the "main tests" are using org.catrobat.catroid.test.R. This leads to compile errors, as the .R file wouldn't exist.
This is the same as in the phiro-branch (had to do that again here, as the branch isn't merged yet)

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
sjaindl: I had some troubles executing the tests in bulk, when there's no project - so with an empty "pocket code uitests" folder. There have been some strange exceptions going on in MainMenuActivitiy's onResume-method (UtilFile.createStandardProjectIfRootDirectoryIsEmpty), which seemed to me a bit flaky due to the different folder. However, I couldn't reproduce it now. Neither on the phone nor on the emulator. So, maybe there was some other issue going on. I've tried it several times again, it's working fine, so I've removed that line again.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
sjaindl: good hint

catroid/build.gradle
84n4n4: ok, good to know.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
84n4n4: regarding strange exceptions in main menu activity.

this could also arise from the activity switching (as already mentioned a few times). i found that sometimes this causes tests to hang / crash, since for some f*cked up reason a new test starting the main menu activity ends up in another activity.

reproducable example:
start the script activity via test rule
add a new scene via the switch scene brick
automatically puts you into scene list
end test

-> next test starting the main menu activity will break for sure (can be a lot of tests later also), because for some reason it ends up int eh program activity

not so reproducable example:
starting an activity with the test rule that starts a background thread and close it quick again (eg. short test)
can lead to either activity not being killed while next test is starting and the times out because hes waiting for a activity that the rule is never going to start. while ive observed this only happens when running them in bulk, if running standalone activity is finished.

so a lot of weird stuff happens due to background threads and activity switches. a lot of those problems are probably due to us messing around with the back-stack of the activities, and with not-so-android-best-practice way of starting / finishing activities and doing stuff in activity lifecyle methods that shouldnt be done at those points in time.

anyways, what i ended up doing to locate tests that do cause problems with other tests:
1. write a testsuite with all your tests (so youve got control over what test classes are run and in which order)
2. check that the problem now occurs
3. check that the flaky test works when run alone
4. start commenting out blocks of testclasses and run testsuite to see which is affecting the test

ive done this now a few times, and its a pain. so this is why i recommend limiting switching inbetween activities as far as possible. (especially main menu activity and other that use background threads proofed to be problematic)

in core espresso test suite ive got 180 tests of which ~140 are run with the pullrequest trigger, and currently those are stable and not affecting each other, many use the main menu activity. that took my countless hours trying to figure out what causes flakyness, learn from my fails...
there are also certain ui elements that cause flakyness out of the box, like spinners, googling that there seems to be a LOT of people having problems with those, so this is why i introduced the @Flaky annotation that will run test up to 3 times (default, can be changed with eg @Flaky(5)) before showing it as failed test. (but if theres tests affecting each other in above described way, no avail with @Flaky...)

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
bernadettespieler: the link is only availble in Englisch and German and it should not be translated via crowdin

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
sjaindl: Well, alternatively I could check whether the RESTRICTED_LOGIN_ACCEPTED preference has been set to true. I've changed that.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
sjaindl: Ok, I understand .. this really sounds like a pain.
Finally, I could get rid of activity restarting in this test class, so it should be fine now.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	private void setFakeToken(Context context, boolean valid) {
+		String token = valid ? VALID_FAKE_TOKEN : INVALID_FAKE_TOKEN;
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+		sharedPreferences.edit().putString(Constants.TOKEN, token).apply();
+	}
+
+	private void createValidUserWithCredentials(Context context, String testUser, String testPassword, String testEmail) {
+		try {
```
84n4n4: let this function throw the webconnection exception to the test.
so if that breaks it at least shows the right exception and not the test failing due to some viewAssertion. this would make debugging easier / clear to spot what exactly is wrong.

catroid/src/androidTestCreateatschool/java/org/catrobat/catroid/createatschool/uiespresso/ui/activity/RestrictedLoginTest.java
```diff
@@ -0,0 +1,253 @@
...
+			boolean userRegistered = ServerCalls.getInstance().register(testUser, testPassword, testEmail, "de", "at", Constants.NO_TOKEN, context);
```
84n4n4: let this function get the context itself via InstrumentationRegistry.getTargetContext() instead of handing it in. so one parameter less and a bit clearer (since just reading the first call to this function before reading the function itself i was wondering why it got the activity handed over)

84n4n4:
ok, besides the few comments on the test this is fine for me.
@cheidenreich can you also have a look at this?

cheidenreich:
did that yesterday and tested it on my device. from functional code its fine for me

PR #2482, opened 2017-09-05
CAT-FIX for the Clone Broadcast bug that made clones ignore messages

84n4n4:


84n4n4:
This is a quick and dirty fix to deal with the problem that clones were not getting broadcast messages.

The way to fix this properly would be to refactor, or better reimplement the whole broadcasting system, because its such a convoluted mess at the moment that there is no proper way of "fixing" it as it is, also the last change bringing the broadcastSequenceMap to sprites opens huge memory leaks and brings significant performance issues. (just compare airfight on 9.28 and 9.29 and also have a look at the memory consumption)

description of what i did in this fix:
- cloned sprites use the same (same as in they have a referece on) broadcastSequenceMap as the sprite they originate from.
- the clones are not registered in the scene, but only in the clonedSprites list in the stage, where they are cleared from once stage is restarted.
- the fix suggested in https://github.com/comeclarity/Catroid/tree/broadcast-clone-bug would make the tests work, but break the broadcasting system in a way that i have no clue of whats going on. (so programs just stop working at some point since no broadcasts are either sent or received anymore)
- i have no idea what possible side effects could arise through my "fix" here. ive tested with airfight and some of my programs, but no guarantees whatsoever...

what has to be done:
- first off, make the defaultException handler depending on a build flag and do not enable it for debug builds. i also found a hard to reproduce NFC exception being thrown sometimes when stage is finished, etc...
- go back to the drawing board and make build an architecture and a plan for the broadcasting system with clear responsibilities and dependencies. (because right now everything does a little of everything and its cluttered over sooo many classes..)
- implement it from scratch.
- write tests
- write documentation, yes, documentation, we have to start to document the critical parts and what the architecture looks like.

cheidenreich:
please run espresso tests

PR #2485, opened 2017-09-06
CAT-2529 Dynamic icons for main and program menu

FlorianWeissensteiner:
#CAT-2529
https://jira.catrob.at/browse/CAT-2529

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
@@ -65,8 +65,8 @@ public final class IconsUtil {
 	private static Rect largeIconSizeBricks = new Rect();
 	private static Rect smallIconSizeCategory = new Rect();
 	private static Rect largeIconSizeCategory = new Rect();
-	private static Rect largeIconSizeMainMenu = new Rect();
-	private static Rect largeIconSizeProgramMenu = new Rect();
+	public static Rect largeIconSizeMainMenu = new Rect();
+	public static Rect largeIconSizeProgramMenu = new Rect();
 	private static Rect largeIconSizeStageDialog = new Rect();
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
```diff
@@ -292,48 +293,24 @@ public static void addIcons(ViewGroup viewGroup, String category) {
...
-	}
-
-	public static void setProgramMenuIconSize(Context context, View contentView) {
-		if (!isLargeSize()) {
-			return;
+		ViewGroup content = (ViewGroup) contentView;
+		for (int i = 0; i < content.getChildCount(); i++) {
+			View child = content.getChildAt(i);
+
+			if (child instanceof LinearLayout || child instanceof RelativeLayout || child instanceof ScrollView) {
```
sjaindl: Wouldn't it be better to check for instanceof ViewGroup here? Since LinearLayout, RelativeLayout etc. are all ViewGroups and I think they should all be called recursively? E.g. RecyclerView would be missing in this check.. but with ViewGroup we wouldn't have to worry about all other container views.

sjaindl:
please rebase

catroid/src/main/java/org/catrobat/catroid/utils/IconsUtil.java
FlorianWeissensteiner: You are absolutely right. Fixed.

FlorianWeissensteiner:
@sjaindl Thank you for the review!
Do you want me to implement a test (if possible) within this PR, or should this be a separate ticket?

mfuchs:
just trying jenkins integration
test this please

sjaindl:
IMO the test(s) should bei implement within this PR, as it is related to the code

PR #2326, opened 2017-05-30
CAT-2382 Numbering style for bidirectional languages

ndrnour:
**Numbering styles vary depending on the region. bidirectional languages numbering style may be implemented by either Hindi digits or Arabic digits**

PR #2487, opened 2017-09-06
CAT-1900 Amount of Bricks (Show details) is wrong 

ndrnour:
espresso Test has activity switching.
if you have another scenario for the test, please let me know


ndrnour:
## should this PR get merged before #2326

PR #2515, opened 2017-09-21
Move Phiro initialization code back to MainMenuActivity

sjaindl:


catroid/src/phiro/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -22,6 +22,26 @@
...
+	protected void onCreate(Bundle savedInstanceState) {
+		enablePhiro();
+		super.onCreate(savedInstanceState);
+	}
+
+	private void enablePhiro() {
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
+
+		if (!sharedPreferences.getBoolean(Constants.PHIRO_INITIALIZED, false)) {
+			BaseSettingsActivity.setPhiroSharedPreferenceEnabled(this, true);
```
cheidenreich: why is the same code still present in the SettingsActivity?

sjaindl: True, that code should be removed.
I've removed it and also moved the Espresso test to a new PhiroMainMenuActivityTest-class.

PR #2524, opened 2017-09-26
[CAT-2344] fix the missing resuming call of facedetection object afte…

manuels87:
…r a scene-change

after a scene transition the face-detection is always stopped. when the new scene is opened, the face-detection was not resumed, this caused the reported problem.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -253,6 +253,7 @@ public class StageListener implements ApplicationListener {
 			collisionPolygonDebugRenderer.setAutoShapeType(true);
 			collisionPolygonDebugRenderer.setColor(Color.MAGENTA);
 		}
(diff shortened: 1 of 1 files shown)


84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
how about tests that verify correct behaviour?

rschluesselbauer:
How would you suggest testing this @84n4n4 ?
An espresso stage test with two scenes where you check the face detection state in both scenes?

cheidenreich:
test this please

84n4n4:
@rschluesselbauer 
>> An espresso stage test with two scenes where you check the face detection state in both scenes?

yeah, pretty much exactly like that....
theres a few facedetection resource tests already in the code base, check them out, and then just write a test program that switches scenes on a tap on screen, and test that.

PR #2527, opened 2017-09-28
Revert "CAT-2400 Automatize the creation of independent APKs."

mfuchs:
This reverts commit f78114277b7d82a529bf2e1e0f4e481c9cde09fd.

Adding the JumpingSumo Drone Support collides with the initial commit.
Even worse: This commit broke the standalone APK creation.
Thus the most sensible way is to revert it and find a better solution.
More details are outlined at CAT-2400.

### DIFF
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -50,20 +50,6 @@ apply from: 'gradle/standalone_apk_tasks.gradle'
 check.dependsOn 'checkstyle'
 check.dependsOn 'pmd'
 
-def appId = 'org.catrobat.catroid'
-def appName = '@string/app_name'
-
-// When -Pindependent was provided on the gradle command the APP name is changed.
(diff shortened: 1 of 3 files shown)


catroid/build.gradle
```diff
@@ -50,20 +50,6 @@ apply from: 'gradle/standalone_apk_tasks.gradle'
 check.dependsOn 'checkstyle'
 check.dependsOn 'pmd'
 
-def appId = 'org.catrobat.catroid'
-def appName = '@string/app_name'
```
robertpainsi: I liked that a lot! :crying_cat_face:
Any chance to get this running with these variables?

mfuchs: I plan to add it back again later, see also CAT-2400 on jira.

Though right now it is broken, so I fear the changes have to go.
In the future more PR jobs will ensure that issues like these are found earlier.

PR #2537, opened 2017-10-05
Add PrivacyPolicyDialog to createatschool/restricted login

sjaindl:
Add PrivacyPolicyDialog to createatschool/restricted login and fix crash on logout

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/dialogs/RestrictedLoginDialog.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/dialogs/RestrictedLoginDialog.java
@@ -36,7 +36,6 @@ import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.TextView;
 
-import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.R;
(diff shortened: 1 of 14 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/dialogs/RestrictedLoginDialog.java
```diff
@@ -89,7 +88,7 @@ public void onClick(View view) {
 								.apply();
 
 						if (!Utils.isUserLoggedIn(getActivity()) || !Utils.isCreateAtSchoolUser(getActivity())) {
-							ProjectManager.getInstance().showLogInDialog(getActivity(), false);
+							Utils.showPrivacyPolicyDialogWithLoginDialogAction(getActivity());
```
cheidenreich: is this also shown when the user has already accepted the privacy policy?

sjaindl: No, then the LoginDialog is shown. The method name was a bit confusing - I've changed that.

sjaindl:
I've changed the dialog flow according to the new requirements now

PR #2538, opened 2017-10-05
Refactor templates + port tests to Espresso

sjaindl:
Ports Templates Robotium tests to Espresso.

I've refactored the Templates code using dependency injection (dagger) and the Retrofit Web Client/RxJava, too (much faster and no need AsyncTasks anymore!). 
This way the test class can easily be mocked using the MockWebComponent and altogether it reduces dependencies.


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/MockCatroidApplication.java
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.util;
+
+import org.catrobat.catroid.CatroidApplication;
+import org.catrobat.catroid.uiespresso.web.util.DaggerMockWebComponent;
+import org.catrobat.catroid.uiespresso.web.util.MockWebRequestModule;
+
+public class MockCatroidApplication extends CatroidApplication {
```
84n4n4: this is not a mock per definition (since its not replacing the CatroidApplication, its still there, fully blown, just the web component has been replaced), can you please find a more suitable name for this. this is a decorator, the only mock is the requestModule

84n4n4: ok, ive seen now that multiple dagger tutorials use exactly this nomenclature, althouth im a fan of naming-things-what-they-are, if that is what is used everywhere else im ok...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/CatroidTestRunner.java
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.util;
+
+import android.app.Application;
+import android.content.Context;
+import android.support.test.runner.AndroidJUnitRunner;
+
+public class CatroidTestRunner extends AndroidJUnitRunner {
```
84n4n4: ive found some tutorials where the injection is done in a custom ActivityTestRule instead of writing a custom runner.
since we can only use one runner for the whole test setup, but could use a seperate rule for each test class, i would prefer to go with the option that provides finer granularity.
also, if something blows up in MockCatroidApplication -> MockWebRequestModule -> ... like the json beeing not there all malformed, all espresso tests would blow up, while if this is handled via the rule, only the test here that actuallly makes use of this would be affected.

catroid/src/main/java/org/catrobat/catroid/ui/adapter/TemplateAdapter.java
```diff
@@ -34,48 +37,78 @@
...
+import io.reactivex.functions.Consumer;
+import io.reactivex.schedulers.Schedulers;
+
+public class TemplateAdapter extends ArrayAdapter<TemplateData> {
+
+	private static final String TAG = TemplateAdapter.class.getSimpleName();
 	private static LayoutInflater inflater;
 
-	private final Context context;
+	private final Activity context;
```
84n4n4: please rename "context" to "activity" also...

catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
```diff
@@ -51,6 +57,11 @@ public void onCreate() {
 		CatroidApplication.context = getApplicationContext();
 		BaseSettingsActivity.applyAccessibilitySettings(context);
 		registerActivityLifecycleCallbacks();
+
+		webComponent = DaggerWebComponent.builder()
```
84n4n4: if this is only used in TemplateAdapter, why is initialized here?

sjaindl:
Injection is now done with a custom dagger test rule using DaggerMock instead of a custom JUnitRunner. Injection of the web component is also done directly in TemplateAdapter instead of within the application.

84n4n4:
please run espresso tests

84n4n4:
running espresso tests with the jenkins hook here isnt working, develop would have to be merged over here, or at least the testsuites package updated to what it is in develop, and category annotations should be added to all NOLB specific tests aswell.
i dont know if this PR is the right one to do all that, but this should be done asap, to not only write espresso tests, but also have the means of running them...

sjaindl:
Well, I think we should do that in that PR partially. I suggest to do the following:
1. Merge of develop into feature/NOLB in a different branch
2. rebase of this branch to feature/NOLB
3. Adding annotations to make tests work in this branch

sjaindl:
@84n4n4 The branch is now up to date with the last release branch - I've added annotations to the tests

84n4n4:
ok for me!
id prefer having the mocks in utils package in tests in an additional package (since we now got "rules", "matchers", "actions", "mocks" would be nice aswell... but up to you if you want to do another push on this.
ill just start the espresso test run to get an idea of what the current status is.

84n4n4:
please run espresso tests

sjaindl:
ok, I did another push and changed the package of the mock classes to "..utils.mocks"

PR #2539, opened 2017-10-09
CAT-2558 Evaluate empty list to empty string ""

robertpainsi:
To be Scratch compatible, an empty list evaluates to an empty string "".
Previously was 0.0.


robertpainsi:
please run espresso tests

PR #2545, opened 2017-10-12
CAT-2573 Use Scratch random function

robertpainsi:
Fixes CAT-2573 by using the [Scratch random function implementation](https://github.com/LLK/scratch-vm/blob/c807f02e37b4746bc503929ae10059de5baa5105/src/blocks/scratch3_operators.js#L80-L91) to
generate random numbers within a specific range.
___
I manually tested it with `[0,1], [0,10], [1,10], [-1,1], [-1.5,1.5], [0, 10.0]`. Please test it as well.
JFI: If you use e.g. `10.0`, the random value will be a double with decimal places, whereas using `10` will result in integers.


robertpainsi:
please run espresso tests

PR #2548, opened 2017-10-16
CAT-2553 JENKINS-175 Fixes generation of standalone and independent A…

mfuchs:
…PKs.

Indepent/unique APks can be generated again.
For that the standalone APK task had to be refactored massively.

Before
------
Before the standalone APK generation needed two gradle calls.
The first would determine properties like the application ID and label
and store them in a file.
The second gradle call would then use these stored properties to actually
set the application ID, label etc. accordingly and then create the APK.

After
-----
Now the necessary standalone generation work is split into four steps
which can all performed in a single gradle call:

1. Configuration
2. Preparation
3. Assembling
4. Cleanup

The configuration step determines the application ID and label.
This is done before any other task is executed and thus happens early enough.
The preparation step adapts the app rights, the app icon and provides
the PocketCode app as zip file.
The APK assembling relies on default behavior of gradle.
Finally the cleanup step ensures that there are no modifications left
in the src/ directory.

Note: For compatibility with arsdk the application label of the top-level
AndroidManifest.xml is always used.

Advantages/Disadvantages
------------------------
+ A single call
  ./gradlew -Pdownload=$MY_URL -Psuffix=$MY_SUFFIX assembleStandaloneDebug
  is sufficient to create a standalone APK.
+ App names can contain special characters like quotes (") now.
+ Screenshots that aren't png files are ignored.
+ After the task is run the cleanup step ensures that there are no local changes.
+ The task can be executed multiple times without generating invalid
  app ids like "org.catrobat.catroid.generated123.generated123.generated123".
+ Apps are downloaded once and then cached in a local .apps directory.
- The assembleStandaloneDebug needs the -Pdownload and -Psuffix parameters
  to create a meaningful APK.
  There are no defaults anymore since the configuration step is always executed.
  No matter which task is called.
  Applying defaults there would mean to always change the app ID/label.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ sed_replace.txt
 *.apk
 *.ap_
 
+# cache for standalone apps
+.apps/
+
 # workspace settings
 *.metadata
 
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -39,6 +39,13 @@ repositories {
     maven { url 'https://maven.google.com' }
 }
 
+ext {
+    appId = 'org.catrobat.catroid'
+    appName = '@string/app_name'
+    manifestAppName = '@string/app_name'
+    manifestAppIcon = '@drawable/ic_launcher'
+}
+
 apply plugin: 'com.android.application'
 apply plugin: 'io.fabric'
 apply plugin: 'checkstyle'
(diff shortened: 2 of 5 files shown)


mfuchs:
test this please

84n4n4:
regarding last comment, can we merge this now?

84n4n4:
i see a problem with the caching of program files:
if a program is updated by a user and re-uploaded, the Id and program name stay the same. so if an apk has been generated once, it will always be generated from this snapshot of the program in the share while not considering updates there.

84n4n4:
ok, then id suggest to keep it, just in case someone has to locally debug it again to speed up the process. if infrastructure ever changes in a way so that apks are generated in the same workspace again, weve got to remember to deal with it.

84n4n4:
please run espresso tests

PR #2553, opened 2017-10-22
CAT-2582 Accessibility tests and refactor

FlorianWeissensteiner:
#CAT-2582
https://jira.catrob.at/browse/CAT-2582

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
@@ -1,0 +1,188 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.ui.activity;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.espresso.IdlingRegistry;
+import android.support.test.espresso.IdlingResource;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.ui.AccessibilityPreferencesActivity;
+import org.catrobat.catroid.ui.BaseSettingsActivity;
+import org.catrobat.catroid.ui.MainMenuActivity;
+import org.catrobat.catroid.uiespresso.ui.activity.utils.AccessibilityActivitiesUtils;
+import org.catrobat.catroid.uiespresso.util.UiTestUtils;
+import org.catrobat.catroid.uiespresso.util.rules.DontGenerateDefaultProjectActivityInstrumentationRule;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.action.ViewActions.scrollTo;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertTrue;
+
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_ADDITIONAL_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_DRAGNDROP_DELAY;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_FONTFACE;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_FONTFACE_VALUE_DYSLEXIC;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_HIGH_CONTRAST;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_HIGH_CONTRAST_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_LARGE_ELEMENT_SPACING;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_LARGE_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_LARGE_TEXT;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_ADDITIONAL_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_DRAGNDROP_DELAY;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_FONTFACE;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_HIGH_CONTRAST;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_HIGH_CONTRAST_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_LARGE_ELEMENT_SPACING;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_LARGE_ICONS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_LARGE_TEXT;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_MYPROFILE_STARTER_BRICKS;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_PROFILE_ACTIVE;
+import static org.catrobat.catroid.ui.BaseSettingsActivity.ACCESS_STARTER_BRICKS;
+
+@RunWith(AndroidJUnit4.class)
+public class AccessibilityPreferencesActivityTest {
+
+	private IdlingResource idlingResource;
+
+	private List<String> settings = new ArrayList<>(Arrays.asList(ACCESS_LARGE_TEXT, ACCESS_HIGH_CONTRAST,
+			ACCESS_ADDITIONAL_ICONS, ACCESS_LARGE_ICONS, ACCESS_HIGH_CONTRAST_ICONS, ACCESS_LARGE_ELEMENT_SPACING,
+			ACCESS_STARTER_BRICKS, ACCESS_DRAGNDROP_DELAY));
+	private List<String> myProfileSettings = new ArrayList<>(Arrays.asList(ACCESS_MYPROFILE_LARGE_TEXT, ACCESS_MYPROFILE_HIGH_CONTRAST,
+			ACCESS_MYPROFILE_ADDITIONAL_ICONS, ACCESS_MYPROFILE_LARGE_ICONS, ACCESS_MYPROFILE_HIGH_CONTRAST_ICONS, ACCESS_MYPROFILE_LARGE_ELEMENT_SPACING,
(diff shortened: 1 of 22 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
```diff
@@ -0,0 +1,156 @@
...
+	private List<String> settings = new ArrayList<>(Arrays.asList(ACCESS_LARGE_TEXT, ACCESS_HIGH_CONTRAST,
+			ACCESS_ADDITIONAL_ICONS, ACCESS_LARGE_ICONS, ACCESS_HIGH_CONTRAST_ICONS, ACCESS_LARGE_ELEMENT_SPACING,
+			ACCESS_STARTER_BRICKS, ACCESS_DRAGNDROP_DELAY));
+
+	@Rule
+	public BaseActivityInstrumentationRule<AccessibilityPreferencesActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(AccessibilityPreferencesActivity.class, true, false);
+
+	@Before
+	public void setUp() throws Exception {
```
84n4n4: first off i would consider it best practice to check what state the shared preferences are in, and leave them after the test in exactly this configuration again.
so read and save the values youre going to change, and put them back in teardown. see SettingsActivityTest for example

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
```diff
@@ -0,0 +1,156 @@
...
+		SharedPreferences sharedPreferences = PreferenceManager
+				.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+
+		onView(withText(R.string.preference_access_title_fontface)).perform(click());
+		onView(withText(R.string.preference_access_title_fontface_dyslexic)).perform(click());
+		assertEquals(sharedPreferences.getString(ACCESS_MYPROFILE_FONTFACE, ""), ACCESS_FONTFACE_VALUE_DYSLEXIC);
+
+		onView(withText(R.string.close)).perform(click());
+
+		onView(withText(R.string.preference_access_title_large_text)).perform(scrollTo(), click());
```
84n4n4: if youd use onData you could skip on the scrollTo() part, which in my experience is not always doing what you want. (at least check this test on a tablet and a very small phone)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
```diff
@@ -0,0 +1,156 @@
...
+		assertTrue(sharedPreferences.getBoolean(ACCESS_MYPROFILE_LARGE_ELEMENT_SPACING, false));
+
+		onView(withText(R.string.preference_access_title_starter_bricks)).perform(scrollTo(), click());
+		assertTrue(sharedPreferences.getBoolean(ACCESS_MYPROFILE_STARTER_BRICKS, false));
+
+		onView(withText(R.string.preference_access_title_dragndrop_delay)).perform(scrollTo(), click());
+		assertTrue(sharedPreferences.getBoolean(ACCESS_MYPROFILE_DRAGNDROP_DELAY, false));
+
+		onView(withId(R.id.scratch_project_apply_button)).perform(click());
+		onView(withText(R.string.ok)).perform(click());
```
84n4n4: dont do this. restarting the app and putting it back into main menu activity is going to be a bitch and break other tests. (due to tha background threads, etc.)

the one time i couldnt get around doing something like this (LanguageSwitchMainMenuTest) i had to:
- get the respawned MainMenuActivity and register its idling resource
- write a new testRule that prevents defaultProject beeing generated
to make this somehow work without breaking other tests.

please get updated with develop to enable git hooked jenkins runs for espresso, and run all espresso tests in a bunch to see if new tests break other tests.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+		for (String setting : settings) {
+			assertFalse(sharedPreferences.getBoolean(setting, false));
+		}
+		assertEquals(sharedPreferences.getString(BaseSettingsActivity.ACCESS_FONTFACE, ""),
+				BaseSettingsActivity.ACCESS_FONTFACE_VALUE_STANDARD);
+	}
+}
```
84n4n4: i would rewrite this test to:

- save current state of shared preferences in setup
- set all to false in setup
- test if all options are available, and if they write into shared preferences correctly (and NOT apply / restart)
- reset the shared preference values to what they were before the test and restart

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityProfilesActivityTest.java
```diff
@@ -0,0 +1,194 @@
...
+	public void switchProfileTest() {
+		onView(withText(R.string.preference_access_title_profile_1)).perform(click());
+
+		onView(withId(R.id.access_active_profile_title)).check(matches(TextViewMatchers.hasValueEqualTo(
+				resources.getString(R.string.preference_access_title_profile_1))));
+		onView(withId(R.id.access_active_profile_summary)).check(matches(TextViewMatchers.hasValueEqualTo(
+				resources.getString(R.string.preference_access_summary_profile_1))));
+
+		onView(withId(R.id.scratch_project_apply_button)).perform(click());
+		onView(withText(R.string.ok)).perform(click());
```
84n4n4: same here as for the other test. this WILL break stuff, trust me...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityProfilesActivityTest.java
```diff
@@ -0,0 +1,194 @@
...
+			if (expectedFontface.equals(ACCESS_FONTFACE_VALUE_STANDARD)) {
+				assertEquals(expectedFontface, ACCESS_FONTFACE_VALUE_STANDARD);
+			} else if (expectedFontface.equals(ACCESS_FONTFACE_VALUE_SERIF)) {
+				assertEquals(expectedFontface, ACCESS_FONTFACE_VALUE_SERIF);
+			} else if (expectedFontface.equals(ACCESS_FONTFACE_VALUE_DYSLEXIC)) {
+				assertEquals(expectedFontface, ACCESS_FONTFACE_VALUE_DYSLEXIC);
+			}
+		}
+	}
+}
```
84n4n4: grundsätzlich find ich den test overkill.
eigentlich würd kein UI gebraucht werden, also am besten wäre hier stattdessen ein integrationTest ohne UI.

wenns denn dann umbedingt espresso sein muss, würd ich hergehen, und die
- testRule auf AccessibilityPreferenceActivity
- activity erst im setup (oder noch später, erst im test) mit intent launchen, wobei im intent das jeweilige testProfile (das es gar net im xml geben muss) mitübergeben
- im test schaun obs richtig befüllt wurde

84n4n4: oder mit intending nachschaun dass die profile activity den richtigen intent schickt wenn ein profile geclickt wurde.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
FlorianWeissensteiner: Makes sense! Done.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
FlorianWeissensteiner: So I tried using onData with a custom Matcher instead, and racked my brain why it didn't work. It would check the first three (visible) checkboxes correctly, and then do absolutely nothing. The test just kept running indefinitely.
Turns out it doesn't work because the AdapterView is inside a ScrollVIew, and from what I've gathered on the internet, a ScrollView should be used in conjunction with the scrollTo() action.
Furthermore the AdapterView is completely collapsed, meaning that its bounding box is big enough to display all of its elements, and the AdapterView itself can't be scrolled anymore. This is necessary for the ScrollView to work as it does.

So this is a bit of a special case, and I don't think using scrollTo() can be avoided. What do you think?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
FlorianWeissensteiner: I would argue that the restart is an essential part of the Accessibility settings, and that it makes sense to test if it correctly quits to the MainMenu, and then check if the UI was correctly adjusted in the settings. Only when i apply the settings (and the app restarts as a result) do the settings get saved in the SharedPreferences. I guess I could call the function to save the settings manually in the test, but that I'd say that defeats the purpose of a UI Test a little bit.

I think it would be worth to go the extra mile, and do what you did to get around breaking it. Do you disagree?

84n4n4: ok, then, have a look at the LanguageSwitchMainMenuTest in develop,
- register the idling resource of MainMenuActivity right after you restart
- using the no-DefaultProject rule helps heaps

keep an eye on other tests when running them in conjunction with this one here, and if something else starts randomly breaking, try to remember that this test could be a reason for it.

i probably should add another category annotation for those kinds of tests so one can easily filter for them and see if everything else suddenly works again when not running those. so the set of tests one has to check if that happens would be highlighted...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityProfilesActivityTest.java
FlorianWeissensteiner: I axed the xml testing, and checked the intents instead. So this now always stays in its own activiy.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
FlorianWeissensteiner: Registered the resource, and applied the no-DefaultProject rule.

"please get updated with develop to enable git hooked jenkins runs for espresso, and run all espresso tests in a bunch to see if new tests break other tests."

I'm afraid I can't do that because this is on the feature/NOLB branch which has a bazillion conflicts with the develop branch, and I honestly don't think i can piece together a working version with my limited knowledge of what does what.

FlorianWeissensteiner:
Not sure why the Standalone APK test failed. The error doesn't seem to be related to my changes.

84n4n4:
looks ok.
gradle changes are similar to what has been done already in develop some weeks ago, upgrading to higher espresso versions and testrunner version.
my guess on the merge conflicts is the move of the baseActivityTestRule to rules package.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
84n4n4: "I'm afraid I can't do that because this is on the feature/NOLB branch which has a bazillion conflicts with the develop branch, and I honestly don't think i can piece together a working version with my limited knowledge of what does what."

ok, thats beyond this PR, but well have to sit together and find a way on how to achieve this, rather sooner than later, because its not going to become easier the more code goes in there.

84n4n4: i just saw that this seems to have just happened in  https://github.com/Catrobat/Catroid/pull/2564, hence the merge conflicts...

FlorianWeissensteiner: Oh great! I guess we will wait until it's merged, to rebase and test this PR?
Speaking of which, how do you run the espresso tests? I only found a "PullRequest-Espresso" Job on Jenkins but it says "This project is currently disabled".

FlorianWeissensteiner:
The 2 lint issues that fail the default test seem to be fixed here #2572

84n4n4:
test this please

84n4n4:
ok for me

84n4n4:
please run espresso tests

84n4n4:
that above is the trigger phrase for jenkins ....

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/AccessibilityPreferencesActivityTest.java
FlorianWeissensteiner: I just realized that I already rebased #2564 into this PR. Woops.

FlorianWeissensteiner:
I've ran all of the failing espresso tests individually locally on an emulator, and except for the first one (BroadcastForClonesRegressionTest.testIfClonesBroadcastReceiversAreRemovedOnRestart) which instead of crashing the process passes, the other 7 tests also fail with the same error messages.

84n4n4:
if those tests fail also individually its something that has become broken in NOLB branch, since theres only 4 failing tests in develop.
also if you find tests that only fail when running in batch with others, its something that become broken due to nolb tests affecting others.

in develop weve currently got 4 failing tests, and that is pretty much stable:
uiespresso.content.brick.app.BroadcastBricksTest.testRemoveUnusedMessagesBroadcastSend
uiespresso.content.brick.app.BroadcastBricksTest.testRemoveUnusedMessagesBroadcastReceive
uiespresso.content.brick.app.DeleteItemOfUserListBrickTest.testDeleteItemOfUserListBrickMultipleLists
uiespresso.content.brick.app.ReplaceItemInUserListTest.testCreateNewUserListAndDeletion

PR #2556, opened 2017-10-24
CAT-2581 fix showVariable Bricks to show non-latin String and number …

ndrnour:
…+ fix Swahili language tag
+ OCR Test, have a look at https://github.com/tesseract-ocr/tesseract


ndrnour:
@84n4n4  could you please have a look!
thanks a lot

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/rtl/ShowTextBrickSupportNonLatinStringAndNumberTest.java
```diff
@@ -0,0 +1,244 @@
...
+import static junit.framework.Assert.assertNotSame;
+
+import static org.catrobat.catroid.stage.StageActivity.stageListener;
+
+@RunWith(AndroidJUnit4.class)
+public class ShowTextBrickSupportNonLatinStringAndNumberTest {
+	@Rule
+	public BaseActivityInstrumentationRule<ScriptActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(ScriptActivity.class, true, false);
+	private float zeroToNine = 0123456789F;
```
84n4n4: change to 1023456789 and the zero will also be checked.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/rtl/ShowTextBrickSupportNonLatinStringAndNumberTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+	@Before
+	public void setUp() throws Exception {
+		createProjectAndGetStartScriptWithImages("showTextBrickTest1");
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@Test
```
84n4n4: do this via paramterized tests, since its essentially the same test, but with different input strings.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/rtl/ShowTextBrickSupportNonLatinStringAndNumberTest.java
```diff
@@ -0,0 +1,244 @@
...
+				.perform(click());
+
+		onView(withId(R.id.formula_editor_edit_field))
+				.check(matches(withText("'" + variableValue + "' ")));
+		onView(withId(R.id.button_play))
+				.perform(click());
+		onView(isRoot())
+				.perform(CustomActions.wait(1500));
+
+		stageListener.makeManualScreenshot();
```
84n4n4: perfect would be if you could solve this without write / read actions on the sd card, eg, by doing what makeManualScreenshot does, but without saving, so calling ScreenUtils.getFrameBufferPixels(screenshotX, screenshotY, screenshotWidth, screenshotHeight,	true); directly and using the contents. (if that is possible...)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/rtl/ShowTextBrickSupportNonLatinStringAndNumberTest.java
```diff
@@ -0,0 +1,244 @@
...
+		pressBack();
+		pressBack();
+
+		String filePath = Environment.getExternalStorageDirectory().getAbsolutePath();
```
84n4n4: dont do this on every test.

get the correct stage contents with the untranslated value once in startup and save it in a member variable.
then you can also directly start the stage activity in the testRule. then there would be no ui interactions necessary for the whole test and no activity switches between stage and script activity

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -124,6 +122,58 @@ private boolean isNumberAndInteger(String variableValue) {
...
+		GLUtils.texImage2D(GLES20.GL_TEXTURE_2D, 0, bitmap, 0);
+		GLES20.glBindTexture(GLES20.GL_TEXTURE_2D, 0);
+		bitmap.recycle();
+		// Draw and dispose
+		batch.draw(tex, posX, posY);
+		batch.flush();
+		tex.dispose();
+	}
+
+	private String convertToEnglishDigits(String value) {
```
84n4n4: please write a unit test that checks this.

then we could even think about not skipping on the ui test, since if the string here is correct its going to be fine. and also, if the string "1023456789" is correctly translated here in this function, but something else breaks, the ui test would not recognize this. 
so eg. "1023456789" is not correctly rendered out to the stage, its not done correctly no matter what language it was coming from, and since EN is compared to other locales, incorrect is compared to incorrect...

ndrnour: done, also I rewrote the test as OCR test, 
have a look at https://github.com/tesseract-ocr/tesseract

catroid/src/test/java/org/catrobat/catroid/test/code/ConvertToEnglishDigitsTest.java
```diff
@@ -0,0 +1,87 @@
...
+		// Gujarati
+		assertEquals(ZERO_TO_NINE, Integer.parseInt(convertToEnglishDigits(GUJARATI_NUMBER_ZERO_TO_NINE)));
+		assertEquals((double) ZERO_TO_NINE, Double.parseDouble(convertToEnglishDigits(GUJARATI_NUMBER_ZERO_TO_NINE)));
+
+		// Tamil
+		assertEquals(ZERO_TO_NINE, Integer.parseInt(convertToEnglishDigits(TAMIL_NUMBER_ZERO_TO_NINE)));
+		assertEquals((double) ZERO_TO_NINE, Double.parseDouble(convertToEnglishDigits(TAMIL_NUMBER_ZERO_TO_NINE)));
+	}
+
+	private String convertToEnglishDigits(String value) {
```
84n4n4: this test is always going to succeed, even if its become broken in pocket code.
please test the method in pocket code that does this instead of a function of this test.

since the class containing this method in prod code has lots of android dependencies, you will have to move the test over to androidTest.org.catrobat.catroid.test package.
then you can access the method (once you make it public, which would be ok for me to make this testable)

catroid/build.gradle
```diff
@@ -169,6 +169,7 @@ dependencies {
     compile 'com.koushikdutta.async:androidasync:2.+'
     compile 'com.sothree.slidinguppanel:library:3.3.0'
     compile 'com.squareup.picasso:picasso:2.5.2'
+    compile 'com.rmtheis:tess-two:8.0.0'
```
84n4n4: make this a testCompile so its not compiled into the debug and prod versions aswell.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/ocr/OcrShowTextBrickTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+		tessocr.onDestroy();
+	}
+
+	@Test
+	public void testShowHindiNumberFormatAsEngNumberFormat() throws Exception {
+		createProject(AR_ONE_TO_ZERO);
+		baseActivityTestRule.launchActivity(null);
+
+		onView(withId(R.id.button_play))
```
84n4n4: use StageActivity in the test rule, and skip the wait if possible. (if it fails with the wait, please have a look at eg. SayBubbleBrickStageTest, ive wrote a "ScriptEvaluationGateBrick" that helps you ensure that your brick under test has been executed in stage before the test continues.

catroid/src/test/java/org/catrobat/catroid/test/code/ConvertToEnglishDigitsTest.java
ndrnour: done.

84n4n4:
test this please

cheidenreich:
266k addtitions seems a freaking lot to me for this 😮

84n4n4:
i talked with michael regarding jenkins, and i think we all can agree that we do not want to have 100mb+ in form of binaries in git.
so proposed solution is that the ocr files should be removed from this PR (and squashed so that they never end up in git history), and that we should try to use the debian provided OCR packages instead, which would make everything a LOT easier, so no caching, no files in git, etc...

cheidenreich:
closing this for now, please reopen once properly implemented

PR #2557, opened 2017-10-25
[CAT-2571] Exclude native libs with arm64-v8a architecture 

rschluesselbauer:
If one of the native libs is loaded for arm64-v8a, the app expects all libs to be available for that architecture. Only some libs were available (from the parrot sdk) - the app crashed when looking for libgdx.so.
This way 32bit libs are used on 64 bit devices. 

This is a quick fix - to support 64 bit, we need to include the remaining native libs for 64 bit architecture.


84n4n4:
test this please

PR #2560, opened 2017-10-29
CAT-NOLB: Fix press of home/back in flavored WebView

sjaindl:


catroid/src/main/java/org/catrobat/catroid/ui/BaseMainMenuActivity.java
```diff
@@ -389,7 +389,7 @@ public void handleWebButton(View view) {
 		}
 		Utils.getTrackingUtilProxy().trackStartWebSessionExplore();
 
-		String url = Constants.FLAVORED_BASE_URL_HTTPS;
+		String url = Utils.getFlavoredUrl(this);
```
84n4n4: if this is only used in this very place, why put it in a global static utility function?

sjaindl: The function is also used in WebViewActivity (see diff in file below)

PR #2562, opened 2017-10-31
[CAT-2584] Add new splashscreen for google codein that is displayed once

rschluesselbauer:


84n4n4:
test this please

84n4n4:
please run espresso tests

84n4n4:
well, this leads to at least one failing UI test, -> the first test running MainMenuActivity will blow up.

my question is now is, if this is going to stay in the codebase, or if this is going to be deleted right after release, or if this same splash screen / its handling is going to be used more often from now on.

so i see multiple solutions here:
- using the test rule to set the shared preferences flag always to false.
- using a build flag, to activate this functionality only in release builds

rschluesselbauer:
Yeah that makes sense.
I modified the test for the MainActivity to set SharedPreferences correctly.
I am not sure if there are plans to leave this kind of screen in longer than Google Code-in is going on (January 17th, 2018).

I also added a clickListener to the LinearLayout so it will scroll down on click.

catroid/src/main/java/org/catrobat/catroid/ui/GoogleCodeInActivity.java
```diff
@@ -0,0 +1,58 @@
...
+		setContentView(R.layout.activity_google_code_in);
+		findViewById(R.id.gci_ok_button).setOnClickListener(new View.OnClickListener() {
+			@Override
+			public void onClick(View v) {
+				GoogleCodeInActivity.this.finish();
+			}
+		});
+		final ScrollView scrollView = (ScrollView) findViewById(R.id.gci_scroll_view);
+		findViewById(R.id.gci_linear_layout).setOnClickListener(new View.OnClickListener() {
+																	@Override
```
84n4n4: some intendation issue here

84n4n4:
>> I am not sure if there are plans to leave this kind of screen in longer than Google Code-in is going on

since this is quite a long time i dont want to have random failing test (since the first one involfing main menu activity is going to break) in the code base for that time, please resolve this:

- using the test rule to set the shared preferences flag always to false.
- using a build flag, to activate this functionality only in release builds
- any other way you find suitable

catroid/src/main/java/org/catrobat/catroid/ui/GoogleCodeInActivity.java
rschluesselbauer: Oh that looks different in the IDE ;-)
It's just a very large indent, because it indents the implementation of the anonymous class to the same indentation level as the new-keyword.
I'll move it to a new line.

rschluesselbauer:
I changed the MainMenuActivity test to set the SharedPreferences before testing. I'm not aware which other test cases might fail - I can't really see any Jenkins results. Could you give me a pointer to what you mean exactly?
For doing that with a rule I'd have to know which tests need that rule, right?

84n4n4:
>I changed the MainMenuActivity test to set the SharedPreferences before testing. 

thats an old robotium test, i dont care about those.

theres quite a number of espresso tests involving MainMenuActivity, some dialogs, some language, some settings, etc and they are scattered.
so dont go there and start modifying existing tests, since the next guy that implements a test like this maybe gets random failures because he didnt realize he has to do some shared preference magic to have this get working.
have it in a central place, so either in prod code with a switch based on a build flag, or put the shared preference magic in the test rule so it will affect all tests.

rschluesselbauer:
Ok, sorry, I misunderstood about simply adjusting the rule for all test cases.
I'll go with the build flag - seems like less overhead IMO.

84n4n4:
i dont know about permissions, but if you can see the leeroy tab, you should be able to see all jobs there aswell.

84n4n4:
please run espresso tests

84n4n4:
test this please

rschluesselbauer:
The espresso tests that are failing right now are supposed to fail, right?
I fixed the checkstyle error.

84n4n4:
please run espresso tests

84n4n4:
i did see the build flag check yesterday in this PR, now its gone again. did you remove it?
regarding espresso tests, there should be 4 failing as in develop, last run here is 6 with the smoketest failing first, this is because now the build flag again is gone and it runs into the splash screen instead of the mainmenu. the other one, AskBrickStageTest.testAskBrickEmptyAnswer is odd, ill start another run.

84n4n4:
please run espresso tests

rschluesselbauer:
Ugh.. Been working on multiple devices - didn't perform a pull before the force push.
Hopefully this change brings this abomination of a PR to an end :-D

84n4n4:
please run espresso tests

84n4n4:
test this please

84n4n4:
ok for me

PR #2565, opened 2017-11-01
CAT-2576 Fix crash when replacing sound from media library

Hydro8182:
Added check to SoundAdapter getItemId to see if the SoundInfo
is in the idMap.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/adapter/SoundAdapter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/adapter/SoundAdapter.java
@@ -61,6 +61,13 @@ public class SoundAdapter extends SoundBaseAdapter implements ActionModeActivity
 			return INVALID_ID;
 		}
 		SoundInfo item = getItem(position);
+
+		if (!idMap.containsKey(item)) {
+			idMap.clear();
+			for (int i = 0; i < getCount(); i++) {
+				idMap.put(getItem(i), i);
+			}
+		}
 		return idMap.get(item);
 	}
 


84n4n4:
test this please

84n4n4:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/ui/adapter/SoundAdapter.java
```diff
@@ -61,6 +61,13 @@ public long getItemId(int position) {
 			return INVALID_ID;
 		}
 		SoundInfo item = getItem(position);
+
+		if (!idMap.containsKey(item)) {
+			idMap.clear();
+			for (int i = 0; i < getCount(); i++) {
+				idMap.put(getItem(i), i);
+			}
+		}
```
84n4n4: sadly its not going to be that easy...
so this now breaks the copy function of sounds, (as described in the comment by nur in the original jira ticket, that bug now escalates to blow up the app)

the problem i see here is that this is architecture wise a MESS, so in short:
SoundAdapter has a hashmap of soundinfo, while beeing a soundBaseAdapter that has a list of soundinfo, while that beeing an arrayAdapter that has a list of soundinfo while behaving like a list.

so the soundInfos are stored in 2 lists and a hashmap, scattered amongst 3 classes of which again 2 behave also like list, and have to be kept consistent which for obvious reasons breaks.

so this piece of code fixes one small issue and opens a bigger one. the whole thing has to be refactored to get rid of the amount of lists involved.

Hydro8182: Alright, so should I just stop working on this for the time being?
I feel like trying to refactor this completely may be a bit much for my first pull request. :)

84n4n4: yes, feel free to pick another ticket

84n4n4: @thmq  is taking over this ticket

PR #2570, opened 2017-11-02
CAT-Release Activating drones in release version in gradle

84n4n4:


84n4n4:
please run espresso tests

PR #2571, opened 2017-11-02
CAT-2514 (NOLB) Refactor string generation

sjaindl:
Creates string entries using TranslationCommands instead of Util functions and shortens string keys to remove duplicates. 

New strings are already on crowdin.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/content/TranslationCommandsTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/content/TranslationCommandsTest.java
@@ -1,0 +1,164 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.test.content;
+
+import android.content.res.Configuration;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.ProjectManager;
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.common.LookData;
+import org.catrobat.catroid.content.Project;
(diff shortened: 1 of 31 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/content/TranslationCommandsTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.catrobat.catroid.uiespresso.util.UiTestUtils;
+import org.catrobat.catroid.utils.UtilFile;
+import org.catrobat.catroid.utils.Utils;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+public class TranslationCommandsTest extends AndroidTestCase {
```
84n4n4: AndroidTestCase has been deprecated, it would be nice to have new tests in Junit4 (there shouldnt be a TestRule required for this since it isnt run as activity test).
for this test conversion should be rather straight forward, remove the "extends AndrdoiTestCase" and make test class not inherited, add @RunsWith(AndroidJunit4.class), and add \@Before \@After \@Test  annotations.

if test crashed due to something like "cannot create Handler" or "cannot call looper"  add the following rule:
	@Rule
	public UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
and annotate the test with 	@UiThreadTest
this happens with tests that run some part of the project that do try to initialize a new handler, which crashes because with regular AndroidJunit4 tests the test, and everything it does, is not executed on the UIThread of the android device...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/util/UiTestUtils.java
```diff
@@ -90,4 +96,14 @@ public static boolean comparePixelRgbaArrays(byte[] firstArray, byte[] secondArr
 		}
 		return true;
 	}
+
+	public static void setLocale(Locale locale) {
```
84n4n4: just no.
this is a very very specific function now used in only 2 tests, this has no business beeing put in a global (!) utility function.

when NOT changing the language via language picker in setting activity, it is sufficient to just call  SettingsActivity.updateLocale instead (as should be sufficient in the other test) since the shared preferences are update independently of the updateLocale call in SettingsActivity.

also i dislike importing and utilizing uitest methods in a "unit" test. if id have to change something, i would not expect a unit test beeing affected by anything i do in espresso tests...

sjaindl:
The testcase is now changed to use Junit4 runner (it worked without adding UiThreadTestRule, thx for the hint) & SettingsActivity.updateLocale is called directly now

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/rtl/LanguageSwitchMainMenuTest.java
```diff
@@ -85,18 +82,9 @@ public void setUp() {
 
 	@After
 	public void tearDown() {
-		resetToDefaultLanguage();
-		IdlingRegistry.getInstance().unregister(idlingResource);
-	}
-
-	private void resetToDefaultLanguage() {
```
84n4n4: but this is necessary for this test!!!

sjaindl:
ok, I've changed it there because you commented that it should also be sufficient there.
Anyway, the LanguageSwitchMainMenuTest is restored now.

84n4n4:
please run espresso tests

84n4n4:
in the two NOLB meetings was decided to not do this. closing this PR accordingly

PR #2580, opened 2017-11-04
CAT-2195: Allow to search Scratch project using ID or URL

cfkh:
Allow to search also using only a Scratch project's ID number in the URL, or even the full URL to the project


rschluesselbauer:
This ticket has been rejected (https://jira.catrob.at/browse/CAT-2195). Could you close the PR or clarify?

cfkh:
Wolfgang wanted it to be implemented. I don't know why it got rejected. Thomas Hirsch wrote a comment: please move this to the catrobat share project, this is not a catroid issue

But actually this is a Catroid issue. As the the converter is also part of the Pocket Code app (otherwise I couldn't have made this PR). Shortly afterwards the issue went to rejected. I also wrote a comment that it is implemented. Nobody answered. Did it again, and again nobody answered. So no, I can't clarify, sorry.

cheidenreich:
We have to rethink, if this is actually a task that should be handled in the app or in the web. This applies for the Scratch2Catrobat in general

PR #2586, opened 2017-11-07
CAT-2609 Fix - Broadcast breaks when receiving clones are deleted

84n4n4:


84n4n4:
please run espresso tests

84n4n4:
this is really hard to test, since the use case is complex (clone that is deleted is a receiver for broadcasts), and that the original bug is that this blows up in a null pointer exception. so my test consists of just running a program that reproduces this, if it works for a few seconds without throwing an exception it succeeds, if it blows up though it also kills the whole instrumentation due to the severity of the exception...
im open for suggestions on how to better test this,...

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
test this please

PR #2592, opened 2017-11-09
CAT-2203 Fix crash when adding a sound from file system

sjaindl:
Setting the soundName in addSoundChooseFile() is unnecessary and may lead to a crash --> it will be set in onActivityResult correctly after a sound has been choosen.


84n4n4:
doesnt the same apply to the soundSource in the next line aswell? its also set in some branches of the callback.
as a side note, it just goes against the grain for me personally when i see stuff beeing done after a startActivityForResult...

sjaindl:
soundSource is necessary - there's no distinguishment e.g. between recording a sound via SoundRecorder and adding a sound from file system later.
Yeah true, but setting it after startActivityForResult doesn't really make sense - I've moved that lines.

PR #2593, opened 2017-11-10
add arabic typefaces and fix icon position with RTL

ndrnour:
add 2 Arabic typefaces "Sabaa, STC"
fix the icon position in RTL
>>>>>>>>>>>>>>>>>>>>>>
screenshot_20171110-031539
screenshot_20171110-031552
screenshot_20171110-031623
screenshot_20171110-031449
screenshot_20171110-031501
screenshot_20171110-031523


catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -658,127 +658,119 @@ public String getBrickCategory(Brick brick, Sprite sprite, Context context) {
 		List<Brick> categoryBricks = new LinkedList<>();
 		categoryBricks = setupControlCategoryList(context);
```
ndrnour: @84n4n4 after a looooooooong search :( here was the problem.

ndrnour:
@bernadettespieler please have a look, thanks

84n4n4:
does it make sense to have those available in languages other than arabic?
if i choose them in eg. english i get numbers in eastern arabic format, so do we really want to offer this option always?

PR #2598, opened 2017-11-12
Fail release build if Crashlytics credentials are missing

robertpainsi:
By removing the fabric.properties file, building the release version
will now result in a gradle error if the fabric.properties file is
missing. This guarantees that the Crashlytics credentials have been
added to the release version.

> FAILURE: Build failed with an exception.
> Crashlytics Developer Tools error

The debug version will work as normal (because of
ext.enableCrashlytics = false) and will just log an error during runtime
which can be ignored.

> java.lang.IllegalArgumentException: Fabric could not be initialized,
  API key missing from AndroidManifest.xml


robertpainsi:
Will not work on Jenkins because Jenkins tries to build the release version as well :crying_cat_face:

PR #2605, opened 2017-11-13
CAT-2362 Fix rename project crash and media library overwrite sound bug

84n4n4:


84n4n4:
fixing code by deleting code...

84n4n4:
please run espresso tests

PR #2614, opened 2017-11-16
CAT String release update

rschluesselbauer:


84n4n4:
test this please

84n4n4:
one error in values-sw strings.xml:

not referencing numbered arguments.
id suggest reverting this to the previous english value manually and somehow flag it for review in crowdin

PR #2623, opened 2017-11-17
CAT-FIX Show codeIn dialog only if translation available

84n4n4:


84n4n4:
test this please

rschluesselbauer:
Code is looking good.
Tested it and it works.

PR #2624, opened 2017-11-20
CAT-FIX Crash when opening project on first app startup

rschluesselbauer:
**Issue**
The app crashes when opening some projects like:
https://share.catrob.at/pocketcode/program/40654
https://share.catrob.at/pocketcode/program/26670
but only on first application startup.

**Explanation**
This happens because Crashlytics is disabled per default -> therefore it is not initialized.
However in onCreate of MainMenuActivity Crashlytics is enabled by writing R.xml.preferences to the SharedPreferences.

When a crash occurs, the app thinks Crashlytics is enabled (because the SharedPreferences say so) and tries to log with it. This results in a crash because it isn't initialized.

**Solution**
If no SharedPreference is set for Crashlytics yet, we assume true so that it will be initialized. (On first startup it  (= the shared preference) will be initialized to true by onCreate of MainMenuActivity anyways)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/CrashReporter.java
@@ -48,7 +48,8 @@ public final class CrashReporter {
 	}
 
 	private static boolean isReportingEnabled() {
-		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, false) && isCrashReportEnabled;
+		return preferences != null && preferences.getBoolean(SettingsActivity.SETTINGS_CRASH_REPORTS, true)
+				&& isCrashReportEnabled;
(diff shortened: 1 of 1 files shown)


robertpainsi:
test this please

robertpainsi:
Please fix Checkstyle warning, then I'll merge it asap :wink:

rschluesselbauer:
No matter how small the change - I will always get checkstyle errors - sorry about that...
Should be fine now.

robertpainsi:
test this please

IMO the problem is the Android Studio formatter. It's currently not possible to specify the (binary) operator location (same or next line) and enable keep line breaks. Some people already reported this bug years ago, still no fix :worried:
Rarely the gradle job `checkstyle` also doesn't print all violations. Then you need to run the gradle job `clean` first.

Btw: Nice finding!

PR #2625, opened 2017-11-20
CAT-FIX Crash when opening project on first app startup

rschluesselbauer:
Copy of https://github.com/Catrobat/Catroid/pull/2624 for NOLB


robertpainsi:
test this please

PR #2630, opened 2017-11-22
[CAT-2606] ARDrone2.0 PlayLEDAnimation brick

dkienreich:


dkienreich:
please run espresso tests

dkienreich:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/content/actions/DronePlayLedAnimationAction.java
```diff
@@ -26,16 +26,132 @@
...
+		if (dcs != null) {
+			ledAnimation();
+		}
+	}
+
+	protected void ledAnimation() {
 		if (dcs != null) {
-			dcs.playLedAnimation(5.0f, 3, ARDRONE_LED_ANIMATION.ARDRONE_LED_ANIMATION_BLINK_ORANGE.ordinal());
+			switch (ledAnimationType) {
+				case STANDARD:
```
84n4n4: since this is just mapping values of a LedAnimation enum to some ARDRONE_LED.ARDRONE_LED_ANIMATION_STANDARD wouldnt it be possible to use the enum from ardrone and not have our own?

or if above is not wished, create a mapping of those in a hashmap:
HashMap<LedAnimation, DroneProxy.ARDRONE_LED_ANIMATION> animationTypeMapping = new HashMap<>();
animationTypeMapping.put(LedAnimation.BLANK, DroneProxy.ARDRONE_LED_ANIMATION.ARDRONE_LED_ANIMATION_BLANK);

or at least put 
dcs.playLedAnimation(5.0f, (int) duration, ARDRONE_LED_ANIMATION.ARDRONE_LED_ANIMATION_BLANK.ordinal());
super.setDuration(duration);
in a function with the ARDRONE_LED_ANIMATION as parameter

with any of the above the size of this would shrink from 100 lines chunk of code to sub 50...

dkienreich: thanks for the helpful hints, i will implement a shorter and better solution for this different drone animation types

cheidenreich:
test this please

PR #2632, opened 2017-11-22
Only report exception if brick can't be found during loading program

robertpainsi:
It only make sense to report an exception, if the brick can't be found
in any of the packages containing bricks.


robertpainsi:
:tophat:
:frog: Thank you!
:necktie:

PR #2642, opened 2017-11-29
CAT-1333 Update PMD to 5.3+

M-Schiller:


catroid/src/androidTest/java/org/catrobat/catroid/uitest/content/brick/PhiroSeekBarColorBrickTest.java
```diff
@@ -96,7 +98,7 @@ private void testFormulaEditorInput(int phiroLedColorEditText, int phiroLedColor
 		solo.clickOnView(solo.getView(phiroLedColorEditText));
 		solo.clickOnView(solo.getView(phiroLedColorTextView));
 
-		UiTestUtils.insertDoubleIntoEditText(solo, 100.0);
+		insertDoubleIntoEditText(solo, 100.0);
```
robertpainsi: Hmm, did pmd really complain about this? If I undo your changes here, I still don't get any pmd warnings.
If this isn't something pmd complains, please undo :smirk_cat:

robertpainsi:
> Applying rule set filter for exclusions: The rule "BooleanInversion" in ruleset "controversial" has been removed from PMD and no longer exists. Please change your ruleset!

84n4n4:
test this please

M-Schiller:
@robertpainsi Can you think of an alternative to UiTestUtils.java:421 since this is the only place in code that requires  excluding the rule  `AvoidDecimalLiteralsInBigDecimalConstructor`?

robertpainsi:
@M-Schiller, not really. Using the _BigDecimal_ constructor seems to be the easiest way to convert a double to a string without an exponent. Since it's used in tests only, it's fine to just [suppress pmd](http://pmd.sourceforge.net/pmd-4.3/suppressing.html) at this line only.

M-Schiller:
I think all changes unrelated to pmd should be excluded from this commit. I think UiTestUtils should be consistent regarding its public static methods used inside the class itself:
should they _all_ be called like this:
`UiTestUtils.myPublicStaticMethod(...)`
or like this:
`myPublicStaticMethod(...)`?

robertpainsi:
Great! Looking good!
ok to test

> consistent regarding its public static methods used inside the class itself:
should they all be called like this:
`UiTestUtils.myPublicStaticMethod(...)`
or like this:
`myPublicStaticMethod(...)`?

Is there a PMD or Checkstyle rule to check for that?

robertpainsi:
please run espresso tests

M-Schiller:
No, at least not yet.

robertpainsi:
In general, I personally prefer the second one because the class reference is redundant.
The class `uitest.util.UiTestUtils` will be removed once we moved all robotium UI tests to espresso.
IMO it's not necessary to go throw other files and remove redundant class references.

If you want to, you can create a ticket on Checkstyle for such check.

PR #2643, opened 2017-12-01
[CAT-2630] Permission GET_TASK

dkienreich:


84n4n4:
please run espresso tests

PR #2655, opened 2017-12-06
CAT-2601 refactored FormulaEditorListFragmentTest

batmanandrobin:
Added the FormulaEditorListFragmentTest which is implemented as a parameterized test
Also added a new Matcher.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
@@ -1,0 +1,152 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.formulaeditor;
+
+import android.support.annotation.StringRes;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.bricks.ChangeSizeByNBrick;
+import org.catrobat.catroid.ui.ScriptActivity;
+import org.catrobat.catroid.uiespresso.content.brick.utils.BrickTestUtils;
+import org.catrobat.catroid.uiespresso.testsuites.Cat;
+import org.catrobat.catroid.uiespresso.testsuites.Level;
+import org.catrobat.catroid.uiespresso.util.UiTestUtils;
+import org.catrobat.catroid.uiespresso.util.matchers.FormulaEditorFunctionListMatchers;
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityInstrumentationRule;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Arrays;
+
+import static android.support.test.espresso.Espresso.onData;
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+import static org.catrobat.catroid.uiespresso.content.brick.utils.BrickDataInteractionWrapper.onBrickAtPosition;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.core.Is.is;
+import static org.junit.runners.Parameterized.Parameter;
+import static org.junit.runners.Parameterized.Parameters;
+
+@Category({Cat.AppUi.class, Level.Smoke.class})
+@RunWith(Parameterized.class)
+public class FormulaEditorFunctionListTest {
+	@Rule
+	public BaseActivityInstrumentationRule<ScriptActivity> baseActivityTestRule = new BaseActivityInstrumentationRule<>(ScriptActivity.class, true, false);
+
+	@Parameters(name = "{2}" + "-Test")
+	public static Iterable<Object[]> data() {
+		return Arrays.asList(new Object[][] {
+				{R.string.formula_editor_function_sin, R.string.formula_editor_function_sin_parameter, "sin"},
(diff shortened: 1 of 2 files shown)


84n4n4:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,180 @@
...
+		Script script = BrickTestUtils.createProjectAndGetStartScript("FormulaEditorFunctionListTest");
+		script.addBrick(new ChangeSizeByNBrick(0));
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@Test
+	public void testFunctionsListElements() {
+		onBrickAtPosition(0).checkShowsText(R.string.brick_when_started);
+		onBrickAtPosition(1).checkShowsText(R.string.brick_change_size_by);
+		onBrickAtPosition(1).perform(click());
```
84n4n4: this works just by accident with this very brick since the click location when clicking on the brick luckily is where the editText is located.
please click on the edit text, because else this could easily be broken with layout changes...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+		onView(withText(R.string.formula_editor_functions)).perform(click());
+
+		String formulaEditorFunctionString = InstrumentationRegistry.getTargetContext()
+				.getString(formulaEditorFunction);
+
+		String formulaEditorFunctionParameterString = InstrumentationRegistry.getTargetContext()
```
84n4n4: theres
UiTestUtils.getResourcesString() to shorten this a bit...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,180 @@
...
+	private String getSelectedFunctionString(String function, String functionParameter) {
+		String selectedFunctionString = function;
+		Boolean bracketsUsed = functionParameter.contains("(") | functionParameter.contains(")");
+
+		if (bracketsUsed) {
+			selectedFunctionString += "(";
+			functionParameter = functionParameter.substring(1, functionParameter.length() - 1);
+		}
+
+		String[] parameterArray = functionParameter.split(",");
```
84n4n4: please make this an iterable and then use a for/each to iterate...

84n4n4: ok, after reading this to the end, theres no loop or anything necessary.
all parameter strings start with "(" and end with ")", besides no_parameter ones.
the spacing can be easily done by someString.replace(","," , ")

batmanandrobin:
> ok, after reading this to the end, theres no loop or anything necessary.
> all parameter strings start with "(" and end with ")", besides no_parameter ones.
> the spacing can be easily done by someString.replace(","," , ")

Not all parameter strings start with "(" and end with ")"
--> Pi has no parameters... which makes my quick and dirty implementation even more dirty

batmanandrobin:
Updated the Test

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+		onView(withId(R.id.formula_editor_edit_field)).check(matches(withText(selectedFunctionString)));
+	}
+
+	private String getSelectedFunctionString(String functionString) {
+		functionString += " ";
+		functionString = functionString.replace("(", "( ");
+		functionString = functionString.replace(")", " )");
+		functionString = functionString.replace(",", " , ");
+
```
84n4n4: id prefer:

return functionString.replace("(", "( ")
				.replace(")", " )")
				.replace(",", " , ") + " ";

or
return functionString.replace("(", "( ")
				.replace(")", " )")
				.replace(",", " , ")
				.concat(" ");

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,150 @@
...
+		String formulaEditorFunctionParameterString = UiTestUtils.getResourcesString(formulaEditorFunctionParameter);
+
+		String editorFunction = formulaEditorFunctionString + formulaEditorFunctionParameterString;
+		String selectedFunctionString = getSelectedFunctionString(editorFunction);
+
+		onData(allOf(is(instanceOf(String.class)), is(editorFunction)))
+				.inAdapterView(FormulaEditorFunctionListMatchers.isFunctionListView())
+				.check(matches(isDisplayed()));
+
+		onView(withText(editorFunction)).perform(click());
```
84n4n4: dont use onView for clicking on this, you rely on call above onData to have scrolled you to the right place....

go with this instead, if you click on it, it also checks that it is there and visible (if not it will not be able to click on it...)

`onData(allOf(is(instanceOf(String.class)), is(editorFunction)))
				.inAdapterView(FormulaEditorFunctionListMatchers.isFunctionListView())
				.perform(click());`

batmanandrobin:
> dont use onView for clicking on this, you rely on call above onData to have scrolled you to the right place....
> 
> go with this instead, if you click on it, it also checks that it is there and visible (if not it will not be able to click on it...)
> 
> onData(allOf(is(instanceOf(String.class)), is(editorFunction))) .inAdapterView(FormulaEditorFunctionListMatchers.isFunctionListView()) .perform(click());

I had this first but this does not work very reliable in this list view. 
The first element of this view is getting ignored everytime.

batmanandrobin:
Fixed the issue with the onData(...).perform(click()) by using onChildView(...).

84n4n4:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorFunctionListTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+		onView(withText(R.string.formula_editor_functions)).perform(click());
+
+		String formulaEditorFunctionString = UiTestUtils.getResourcesString(formulaEditorFunction);
+		String formulaEditorFunctionParameterString = UiTestUtils.getResourcesString(formulaEditorFunctionParameter);
+
+		String editorFunction = formulaEditorFunctionString + formulaEditorFunctionParameterString;
+		String selectedFunctionString = getSelectedFunctionString(editorFunction);
+
+		onData(allOf(is(instanceOf(String.class)), is(editorFunction)))
```
84n4n4: if im not mistaken, we talked about this today. so is this also fixable / possible to do in one call now via just clicking on the childView with the string instead of the id?

batmanandrobin: But why do you want to click on the childView with the Text? 
The view element is selected with a string already:
`onData(allOf(is(instanceOf(String.class)), is(editorFunction)))`

In my opinion 	using `.onChildView(withId(R.id.fragment_formula_editor_list_item))` is more elegant than using withText(editorFunction)

Nevertheless i still did some optimizations on this test.

batmanandrobin: And yes, using `.onChildView(withText(editorFunction))` also works

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

84n4n4:
join and exp tests are failing.

i saw in another PR from another person that the strings xml with the default values have been changed accordingly. what exactly is going on here?

batmanandrobin:
Tests will be put into the bugfix branch then

batmanandrobin:
The tests are failing because the exp(0) will be added as exp(1) to the formula editor edit field.
exp(0) shall be added as exp(0) to the formula editor edit field
There is also something wrong with the join. 
These tests failing are bugs that need to be fixed.

84n4n4:
is there a ticket id for the bug please put it in the tests as comment.
if not, please add a ticket

batmanandrobin:
I created a ticket for this. 
Jira Ticket Number: CAT-2649

PR #2626, opened 2017-11-20
CAT-2594 reimplement SetVariableTest

AntiDog:
Should I also delete the old Testcase?

PR #2656, opened 2017-12-07
CAT-2587 Reimplement ChangeVariableTest in Espresso

Hydro8182:


84n4n4:
test this please

84n4n4:
test this please

84n4n4:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/ChangeVariableTest.java
```diff
@@ -0,0 +1,136 @@
...
+		String userVariablename = "testvariable1";
+		String userVariablename2 = "testvariable2";
+
+		onBrickAtPosition(1).onFormulaTextField(R.id.brick_change_variable_edit_text)
+				.performNewVariable(userVariablename);
+		onBrickAtPosition(1)
+				.performCopyBrick();
+		onBrickAtPosition(1).onFormulaTextField(R.id.brick_change_variable_edit_text)
+				.performNewVariable(userVariablename2);
+		onBrickAtPosition(1).performDragNDrop(BrickCoordinatesProvider.DOWN_ONE_POSITION);
```
84n4n4: what exactly is supposed to be happnening here? why is it necessary to move the brick?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickSpinnerDataInteractionWrapper.java
```diff
@@ -52,6 +55,16 @@ public BrickSpinnerDataInteractionWrapper checkShowsText(String text) {
 		return new BrickSpinnerDataInteractionWrapper(dataInteraction);
 	}
 
+	public BrickSpinnerDataInteractionWrapper checkHasVariable(String variableName) {
+		dataInteraction.perform(click());
+		onView(allOf(
+				withText(variableName),
+				hasSibling(withText(R.string.brick_variable_spinner_create_new_variable)))).
```
84n4n4: why is hasSibling necessary?
if this is actually checking if a string is there, and takes a string as parameter,  do not name is checkHas**Variable**
please just make a new version of checkValuesAvailable that takes an array of strings, or implement a new method just taking one string and call it "checkContains"

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickFormulaEditTextDataInteractionWrapper.java
```diff
@@ -102,4 +103,23 @@ public BrickFormulaEditTextDataInteractionWrapper performEnterString(int stringR
 				.perform(click());
 		return new BrickFormulaEditTextDataInteractionWrapper(dataInteraction);
 	}
+
+	public BrickFormulaEditTextDataInteractionWrapper performNewVariable(String variableName)
```
84n4n4: since this is kind of special for this kind of test (since everywhere else its sufficient to just use the "new" in spinner) i would prefer to have this as a function in the test class.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
```diff
@@ -84,4 +98,36 @@ public static void createUserListFromDataFragment(String userListName, boolean f
 		onView(withText(userListName))
 				.check(matches(isDisplayed()));
 	}
+
+	public static void addBrick(Class<?> brickHeaderClass, int brickCategoryId) {
```
84n4n4: as far as i can see this function isnt called anywhere in this PR.
is this necessary?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
```diff
@@ -84,4 +98,36 @@ public static void createUserListFromDataFragment(String userListName, boolean f
...
+				.perform(click());
+		onData(Is.is(instanceOf(brickHeaderClass))).inAdapterView(BrickPrototypeListMatchers.isBrickPrototypeView())
+				.perform(click());
+
+		onData(instanceOf(Brick.class))
+				.inAdapterView(ScriptListMatchers.isScriptListView()).atPosition(1)
+				.perform(click());
+	}
+
+	public static void deleteItemWithTextFromFormulaEditorDataList(String text) {
```
84n4n4: this is only used in one occasion in one single test. please move this function to the test class.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
```diff
@@ -84,4 +98,36 @@ public static void createUserListFromDataFragment(String userListName, boolean f
...
+		onView(withId(R.id.formula_editor_keyboard_data))
+				.perform(click());
+		onView(allOf(withText(text), withId(R.id
+				.fragment_formula_editor_datalist_item_name_text_view)))
+				.perform(longClick());
+		onView(withText(R.string.delete))
+				.perform(click());
+		try {
```
84n4n4: i really dislike try / catches in test code.
also this method leaves the user/tester in a different fragment than where he was when he started it.

AntiDog: I don't like try&catch eighter but I have no idea how I am supposed to do it without. In my last pull request, you told me that you want me to make a function of it which can also be used in another test case. However, depending on if that variable is in use there is an additional Dialog. Do you have an Idea how to check if the deleted variable is in use by the user/tester to make a function for this without try/catch?

84n4n4:
ok, ive seen that theres quite substantial overlap in those two PRs
https://github.com/Catrobat/Catroid/pull/2626
https://github.com/Catrobat/Catroid/pull/2656

please coordinate with each other, and DO NOT COPY CODE from other open pull requests. this isnt what git is good for... and it will lead to the latter PR having merge conflicts to be resolved.

also its misleading for me to see methods ending up in some utility classes while theres only one usage, etc. while its also being used by someone else in another PR.

regarding the utility function madness.
i want to keep the focus of such classes as small as possible, why you ask? please have a look at UiTestUtils in the robotium test package. nobody can maintain a mess like this...
anyways, since the methods for formula editor magic are accumulating, please feel free to add a new separate class for such methods.
take care though to:
- do stuff via UI only if really necessary, or if it is what youre actually testing. if neither is the case, just access the data structure directly to do your task. (so if youre not testing if "adding a brick" works via ui, just make a new brick and append it to the script)
- write proper functions with proper names. that means that theres no sideffects, so not to switch fragment and leave user there to go back themselves if this is not what the function name is saying.
- especially in test function, NO try / catch blocks. i want tests to fail if something breaks. and especially no try / catches for expected behaviour unless youre actually testing that exceptions are thrown by the app in certain circumstances.
- keep the flow as straight as possible, so minimal if / else parts where it does different things...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/ChangeVariableTest.java
Hydro8182: I don't really know to be honest. It was like this in the original test. Maybe there was a bug related to that at some point in the past.
Should I just remove it?

84n4n4: yes, please.
since we cant guess on what is a regression test or not (since this stuff was just put in the middle of yet another test instead of creating a separate test) i think were better of to throw away stuff like this and keep tests as small as possible to not end up having the same mess again...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/utils/BrickTestUtils.java
84n4n4: yes, you can get the variable list from the project and check if it is in there...
or you can have the dev that uses this function take care of this by adding a parameter to the function.

Hydro8182:
I made all the suggested changes. Could somebody take a look at this again?

84n4n4:
test this please

84n4n4:
please run espresso tests

84n4n4:
please squash commits....

Hydro8182:
@84n4n4 It's only one commit now.

84n4n4:
please fix the checkstyle warnings

Hydro8182:
@84n4n4 Should be done. Is there a way to run checkstyle locally?

robertpainsi:
you can run those checks yourself within android studio, there are gradle tasks "lint", "checkstyle" and "pmd".

or in the terminal, e.g.

`./gradlew checkstyle`

Also there is a gradle task called "check" which runs lint, checkstyle, pmd and I don't know what else in one go.

84n4n4:
test this please

Hydro8182:
@robertpainsi Got it, thanks.

PR #2662, opened 2017-12-13
CAT-2083 Fixed rename variables bug

batmanandrobin:
The problem was that the FormulaEditorEditText field is not getting updated after renaming variables or lists.
All tokens inside the InternFormula are changed which share the same name as the renamed variable (before renaming)
Inside the RenameVariableDialog the updateVariableReferences(...) or updateListReferences(...) is called, which updates all tokens inside the InternFormula object.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
@@ -762,6 +762,20 @@ public class FormulaEditorFragment extends Fragment implements OnKeyListener,
 		}
 	}
 
+	public void updateFormulaUserVariable(String oldName, String newName) {
+		if (formulaEditorEditText == null) {
+			return;
+		}
+		formulaEditorEditText.updateVariableReferences(oldName, newName);
+	}
+
+	public void updateFormulaUserList(String oldName, String newName) {
+		if (formulaEditorEditText == null) {
+			return;
+		}
+		formulaEditorEditText.updateListReferences(oldName, newName);
+	}
+
 	private void showFormularEditorCategorylistFragment(String tag, int actionbarResId) {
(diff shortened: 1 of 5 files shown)


rschluesselbauer:
This is a great start!
There is some more work to do though:
The ticket states that no matter if you press "Ok" or "back" after renaming the formula, the reference should be updated. (In the section **Observed behaviour**: Go back to the scripts activity. Note that there as well, the formula in the brick still contains the "abc" variable)

It also makes sense to update the references in all other bricks.

Espresso tests for this renaming behaviour of lists and variables in the formula editor would be very nice.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
```diff
@@ -888,4 +902,4 @@ public void updateButtonsOnKeyboardAndInvalidateOptionsMenu() {
 			backspaceOnKeyboard.setEnabled(true);
 		}
 	}
-}
+}
```
rschluesselbauer: Checkstyle warning: File does not end with a newline.

batmanandrobin:
I already created an espresso test for testing the formula editor. (Ticket Nr.: CAT-2643)
This only tests the formula editor after changing UserVariables and UserLists not the bricks.
If you can set this to "Ready for development" state i can create a pull request on this. (I wanted to do this tomorrow anyways)

batmanandrobin:
The variable which is renamed, is also changed in the current brick which is edited.

Actually it should be renamed in every brick in every script the variable, which shall be renamed, is located. This should be handled in a different ticket maybe, because the FormulaEditFragment will be refactored.

In the FormulaEditorEditText.java the new formula is pushed to history... check if this should be added to history or not... otherwise the undo has no effect.

84n4n4:
test this please

84n4n4:
please run espresso tests

PR #2663, opened 2017-12-14
CAT-2524 Pen bricks should also work for clones

erry00777:
A Sprite created through the CloneBrick action, is correctly created
but it is never added to the Sprite List in the ProjectManager Class.
When the draw function of the PenActor Class is called to perform the
Pen Brick actions it looks through the List of Sprites in the
ProjectManager Class, and performs the necessary actions according to
the values of the PenConfiguration Flags of every Sprite in the List.
Since the Cloned Sprite is not in the List, no Pen actions are
performed for it. To resolve this issue the Cloned sprite is added to
the Sprite List of the ProjectManager Class. This way when the draw
function is called, the the Cloned Sprite is also found in the List and
the necessary Pen Brick actions(PenUP/down/Stamp..) are Performed.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
@@ -377,6 +377,7 @@ public class Sprite implements Serializable, Cloneable {
 		cloneLook(cloneSprite);
 		setUserAndVariableBrickReferences(cloneSprite, userBricks);
 
+		ProjectManager.getInstance().addSprite(cloneSprite);
 		ProjectManager.getInstance().setCurrentSprite(originalSprite);
 
 		return cloneSprite;


catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -377,6 +377,7 @@ public Sprite cloneForCloneBrick() {
 		cloneLook(cloneSprite);
 		setUserAndVariableBrickReferences(cloneSprite, userBricks);
 
+		ProjectManager.getInstance().addSprite(cloneSprite);
 		ProjectManager.getInstance().setCurrentSprite(originalSprite);
```
84n4n4: just a quick thought here. weve already got massive memory leaks with clones and broadcasts (just check out the android monitor when running that airfight game). so if theres sprites added to yet another list somewhere, are they ever going to be removed again?

84n4n4:
please run espresso tests

84n4n4:
test this please

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
erry00777: Thanks for the suggestion, i have been trying to track this List but it doesn't seem that it is ever cleaned. This means that not only the elements that i add but also the elements previously in the list are never removed, just dropped when a new project is initialized.

robertpainsi: I'm not sure what _this_ list is, but if you're talking about the sprite list of scenes/project
https://github.com/Catrobat/Catroid/blob/b7ad30d730e041a2e1828834cda4d9d09dc7142b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java#L305-L309

erry00777: My bad, totally missed that, thanks for pointing it out. I tested it now and yes, after the execution of the program, when the dispose action is triggered the cloned Sprites that are now added in the Sprite class are removed from the List.

robertpainsi: No problem! :wink:

I'm not a fan of populating the **real** sprites list of every scene with just temporary objects. These objects in the list are the objects shown in the IDE and also are serialized. But hey, everything is already implemented that way…

robertpainsi: > But hey, everything is already implemented that way…

or not. Don't know :smile:  Just assuming because the method `disposeClonedSprites` exists.

84n4n4:
please run espresso tests

PR #2669, opened 2017-12-18
CAT-2458 Make collision events work with clones

rschluesselbauer:
- Make collision broadcast messages individual for each clone
- Give each clone its own BroadcastSequenceMap
- Call broadcast listeners for clones too

Edit:
- Add small test for collision broadcast messages of a clone

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/physics/PhysicsCollisionScriptInteractionTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/physics/PhysicsCollisionScriptInteractionTest.java
@@ -164,4 +164,16 @@ public class PhysicsCollisionScriptInteractionTest extends InstrumentationTestCa
 			Assert.fail("Unexpected Exception in Reflection");
 		}
 	}
+
+	public void testCollisionScriptMessageOfClone() {
+		Sprite clone = firstSprite.cloneForCloneBrick();
+
+		for (Script script : clone.getScriptList()) {
+			if (script instanceof CollisionScript) {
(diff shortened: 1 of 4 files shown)


84n4n4:
this breaks the airfight game, after a handful of cloned bullets have been shot the controls get broken   / unresponsive

84n4n4:
the broadcast system is fubar, the clone system also. we should discuss how to go from here, since fixing one bug opens 20 new ones now.

robertpainsi:
I'm not sure if you just refer to the physics collision event handling or the whole broadcast system.
I'm not familiar with the new broadcast implementation. However, the physics collision event handling has always been a mess and a hack. The clone feature has never been tested, and is also hacky in many places.

Maybe it's worth a GSoC project? To improve the whole event managing? Including restructuring, make it reusable, testable and much much more. So, we also start to improve the core and stage logic. What do you think?

I haven't researched yet how the event logic in scratch works, but since it's open sourced, maybe we can copy it somehow. https://github.com/LLK/scratch-vm

84n4n4:
@robertpainsi 
>>However, the physics collision event handling has always been a mess and a hack.

same same with the regular broadcast system

rschluesselbauer:
@robertpainsi @84n4n4 I have done some refactoring of the bc system over the holidays - I'll show you the concept next time I see you.

> I haven't researched yet how the event logic in scratch works, but since it's open sourced, maybe we 
> can copy it somehow

Good call, I'll have a look at the way scratch does it.

cheidenreich:
ok, I'll close this for now

PR #2671, opened 2017-12-18
CAT-2361 Fix wrong rotation for rotation style "left-right"

rschluesselbauer:


84n4n4:
test this please

84n4n4:
test this please

84n4n4:
on the left screen bounce it still does something odd. for the first two look switches or so, the orientation stays the same while direction of movement has changed already.
kind of looks similar to what it looked like before this fix, but now for the short period it did switch before, it does not switch now...

rschluesselbauer:
Seemed to happen with very few specific step sizes - should be fixed :)

84n4n4:
test this please

84n4n4:
test this please

84n4n4:
test this please

84n4n4:
test this please

PR #2677, opened 2017-12-19
CAT-2127 Write stamps to framebuffer instantly

rschluesselbauer:
- Write stamps to frame buffer with the stamp action synchronously - this way we don't rely on the right timing of the rendering method.
- Dispose frame buffer and batch when stage is terminated (should have been done before)


84n4n4:
test this please

84n4n4:
test this please

PR #2718, opened 2018-02-13
Scratch Converter wont open

ndrnour:
*because of missing toolbar layout
 *using SupportActionBar instead of ActionBar


thmq:
please run espresso test

thmq:
please run espresso tests

thmq:
please run espresso tests

PR #2724, opened 2018-02-15
CAT-2598 Refactor FormulaEditorEditTextTest

AntiDog:
removed ugly stuff

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
@@ -1,0 +1,136 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.formulaeditor;
+
+import android.support.annotation.StringRes;
+import android.view.View;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.bricks.ChangeSizeByNBrick;
+import org.catrobat.catroid.ui.SpriteActivity;
+import org.catrobat.catroid.uiespresso.content.brick.utils.BrickTestUtils;
+import org.catrobat.catroid.uiespresso.testsuites.Cat;
+import org.catrobat.catroid.uiespresso.testsuites.Level;
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityInstrumentationRule;
+import org.hamcrest.Matcher;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+
+import java.util.Arrays;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.BRACKETCLOSE;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.BRACKETOPEN;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.DIVIDE;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.MINUS;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.MULT;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.Math.PLUS;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.COMMA;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM0;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM1;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM2;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM3;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM4;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM5;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM6;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM7;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM8;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.NumPad.NUM9;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.onFormulaEditor;
+import static org.catrobat.catroid.uiespresso.util.UiTestUtils.getResources;
+
+@Category({Cat.AppUi.class, Level.Smoke.class})
+@RunWith(Parameterized.class)
+public class FormulaEditorEditTextGenericTests {
+
+	@Rule
+	public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(SpriteActivity.class, SpriteActivity.EXTRA_FRAGMENT_POSITION,
+			SpriteActivity.FRAGMENT_SCRIPTS);
+
+	@Before
+	public void setUp() throws Exception {
+		Script script = BrickTestUtils.createProjectAndGetStartScript("FormulaEditorEditTextTest");
+		script.addBrick(new ChangeSizeByNBrick());
+		baseActivityTestRule.launchActivity();
+		onView(withId(R.id.brick_change_size_by_edit_text))
+				.perform(click());
+	}
+
+	@Parameterized.Parameters(name = "{2}" + "-Test")
+	public static Iterable<Object[]> data() {
+		return Arrays.asList(new Object[][] {
+				{NUM0, "0", "button_0"},
+				{NUM1, "1", "button_1"},
+				{NUM2, "2", "button_2"},
+				{NUM3, "3", "button_3"},
+				{NUM4, "4", "button_4"},
+				{NUM5, "5", "button_5"},
+				{NUM6, "6", "button_6"},
+				{NUM7, "7", "button_7"},
+				{NUM8, "8", "button_8"},
+				{NUM9, "9", "button_9"},
+				{COMMA, "0" + getResources().getString(
+						R.string.formula_editor_decimal_mark), "button_decimal_mark"},
+				{PLUS, getResources().getString(
+						R.string.formula_editor_operator_plus), "button_plus"},
+				{MINUS, getResources().getString(
+						R.string.formula_editor_operator_minus), "button_minus"},
+				{MULT, getResources().getString(
+						R.string.formula_editor_operator_mult), "button_mult"},
+				{DIVIDE, getResources().getString(
+						R.string.formula_editor_operator_divide), "button_divide"},
+				{BRACKETOPEN, getResources().getString(
+						R.string.formula_editor_bracket_open), "button_bracket_open"},
+				{BRACKETCLOSE, getResources().getString(
+						R.string.formula_editor_bracket_close), "button_bracket_close"}
+		});
+	}
+
+	@Parameter
+	public @StringRes Matcher<View> formulaEditorKeyboardButton;
+
+	@Parameter(1)
+	public String shouldBe;
+
+	@Parameter(2)
+	public String testName;
+
+	@Category({Cat.CatrobatLanguage.class, Level.Smoke.class})
+	@Test
+	public void testCustomKeys() {
+		onView(formulaEditorKeyboardButton)
+				.perform(click());
+		onFormulaEditor().checkShows(shouldBe + " ");
+	}
+}
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
@@ -1,0 +1,247 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.formulaeditor;
+
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.bricks.ChangeSizeByNBrick;
+import org.catrobat.catroid.ui.SpriteActivity;
+import org.catrobat.catroid.uiespresso.content.brick.utils.BrickTestUtils;
+import org.catrobat.catroid.uiespresso.testsuites.Cat;
+import org.catrobat.catroid.uiespresso.testsuites.Level;
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityInstrumentationRule;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.Espresso.pressBack;
+import static android.support.test.espresso.action.ViewActions.click;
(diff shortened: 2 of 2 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+		onBrickAtPosition(1)
+				.checkShowsText("0 ");
+	}
+
+	@Category({Cat.CatrobatLanguage.class, Level.Smoke.class})
+	@Test
+	public void testDiscardDialog() {
+		onFormulaEditor()
+				.performEnterFormula("99");
+		pressBack();
```
84n4n4: please split this test into smaller tests. its really hard to follow whats going on.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+		onView(withText(getCurrentActivity().getString(R.string.formula_editor_discard_changes_dialog_title)))
```
84n4n4: "withText" also takes string ids:

onView(withText(R.string.formula_editor_discard_changes_dialog_title))

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+		onFormulaEditor()
+				.performEnterFormula(".9");
+		pressBack();
+		onView(withText(getCurrentActivity().getString(R.string.formula_editor_discard_changes_dialog_title)))
+				.check(matches(isDisplayed()));
+		onView(withText(yes))
+				.perform(click());
+		checkToastMessage(R.string.formula_editor_changes_saved);
+
+		onView(allOf(withId(R.id.brick_change_size_by_edit_text), withEffectiveVisibility(Visibility.VISIBLE)))
```
84n4n4: please access bricks via the wrapper methods. (eg. onBrickAtPosition... when in script view).
due to the pressBack() switching around its really hard to follow... 
the onBrickAtPosition would give the additional info that now were back in the scriptFragment

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+				.perform(click());
+		pressBack();
+
+		onView(allOf(withId(R.id.brick_change_size_by_edit_text), withEffectiveVisibility(Visibility.VISIBLE)))
+				.check(matches(withText("99" + decimalMark + "9 ")));
+
+		onView(allOf(withId(R.id.brick_change_size_by_edit_text), withEffectiveVisibility(Visibility.VISIBLE)))
+				.perform(click());
+		onView(withId(R.id.formula_editor_keyboard_delete))
```
84n4n4: please use only wrapper methods of the formula editor wrapper instead of using ids directly.

formula editor will be replaced / refactored, so if all the ids change it would be nice that we could only update a single wrapper instead of a s*load of tests that reference every single button on it...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+				.perform(click());
+		onView(withId(R.id.formula_editor_keyboard_compute))
+				.perform(click());
+		onView(withId(R.id.formula_editor_compute_dialog_textview))
+				.check(matches(withText("-8.111")));
+	}
+
+	private int getAbsoluteCursorPosition() {
```
84n4n4: it would be even nicer if youd put this in a custom matcher.
and then add this to the formula editor wrapper

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,407 @@
...
+		FormulaEditorEditText editText = (FormulaEditorEditText) getCurrentActivity().findViewById(R.id
+				.formula_editor_edit_field);
+		Object position = Reflection.getPrivateField(editText, "absoluteCursorPosition");
+		assertTrue("Cursor position not valid", position instanceof Number);
+		return ((Number) position).intValue();
+	}
+
+	private void setAbsoluteCursorPosition(int position) {
```
84n4n4: same as above, just a custom action instead of a custom matcher

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
AntiDog: do you mean I should add a function like this:
```
public FormulaEditorWrapper checkHasCursorPosition(int expected)
{
    FormulaEditorEditText editText = (FormulaEditorEditText) getCurrentActivity().findViewById(R.id.formula_editor_edit_field);
    Object position = Reflection.getPrivateField(editText, "absoluteCursorPosition");
    assertEquals(((Number) position).intValue(),expected);
    return new FormulaEditorWrapper();
} 
```
to the FormulaEditorWrapper? This seems somehow wrong to me since I directly access private fields of the FormulaEditorEditText and this isn't really Espresso stuff and I have no idea how to do this in Espresso or if this is even possible to do

AntiDog:
Tried to apply suggested changes but I'm unsure if you really meant I should change setAbsoluteCursorPosition & checkHasCursorPosition to this. But I have no idea how to do it otherwise?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/utils/FormulaEditorWrapper.java
```diff
@@ -109,6 +115,24 @@ public FormulaEditorWrapper checkShows(String expected) {
 		return new FormulaEditorWrapper();
 	}
 
+	public FormulaEditorWrapper checkHasCursorPosition(int expected) {
+		FormulaEditorEditText editText = (FormulaEditorEditText) getCurrentActivity()
```
84n4n4: just saw this now. getCurrentActivity is not safe to be used like this. this is what custom matchers are intended for, so you dont have to aquire the activity like this.

84n4n4:
checkHasCursorPosition and setAbsoluteCursorPosition should be changed, implement a new viewMatcher for the checkHasCursorPosition and call it something like cursorAt(int position), then it can be used as
.check(matches(cursorAt(5)))
and then put this functionality in the existing formula editor wrapper.
the setAbsoluteCursorPosition should be implemented as a custom action, and also moved to the formulaeditor wrapper.
there are a few actions already in the codebase you can have a look at, just search for "implements ViewAction" and "static ViewAction" in the uiespresso package, or have a look at examples online. same with the matchers, theres a lot of them around already in the uiespresso package, eg have a look at RecyclerViewItemMatcher to see how to properly aquire an ui object without getting the activity first...

AntiDog:
@84n4n4 are you sure this is possible by doing it with a custom action? I'm running into strange issues while trying to do so. I think I did it right because the action does set the cursor, however, it isn't "remembered" once the some of the keyboard buttons are pressed. Could you please take a look this album which is containing screenshots of the issue and is explaining the details: https://imgur.com/a/HYP83

It didn't happen with the old implementation and only setting the position seems to work, if I check it right after setting the position everything is as it should be (visually and via code).

Did I mess something up with the custom Action? I never did one before so it could very well be my fault... 

(All this doesn't happen if you click manually at the position)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,393 @@
...
+	@Rule
+	public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(SpriteActivity.class, true, false);
+
+	@Before
+	public void setUp() throws Exception {
+		Script script = BrickTestUtils.createProjectAndGetStartScript("FormulaEditorEditTextTest");
+		script.addBrick(new ChangeSizeByNBrick());
+		baseActivityTestRule.launchActivity(null);
+		sin = getCurrentActivity().getString(R.string.formula_editor_function_sin);
```
84n4n4: please do not use getCurrentActivity, this should be only used for asserting that one is in a certain activity and a few other rare cases.
either get the targetContext (proper) from the instrumentation, or get the activity from the rule (kind of proper), or just use the getResourcesString method from utils that uses the targetContext...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,393 @@
...
+
+	private String sin;
+	private String random;
+	private String decimalMark;
+	private String no;
+	private String yes;
+
+	@Rule
+	public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(SpriteActivity.class, true, false);
```
84n4n4: please use 
BaseActivityInstrumentationRule<>(SpriteActivity.class, SpriteActivity.EXTRA_FRAGMENT_POSITION,
			SpriteActivity.FRAGMENT_SCRIPTS)

and baseActivityTestRule.launchActivity()  (without the null)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,393 @@
...
+		yes = getCurrentActivity().getString(R.string.yes);
+		decimalMark = getCurrentActivity().getString(R.string.formula_editor_decimal_mark);
+		onView(withId(R.id.brick_change_size_by_edit_text))
+				.perform(click());
+	}
+
+	@Category({Cat.CatrobatLanguage.class, Level.Smoke.class})
+	@Test
+	public void testSingleTapOnFunctionName() {
+		onView(withId(R.id.formula_editor_keyboard_function))
```
84n4n4: theres wrapper methods for selectiong formulas, please use those, i do not want to have any ids of formula editor elements in the tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,393 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.content.brick.app;
```
84n4n4: please move this test to the formulaEditor package.

AntiDog: Ok sure. Will try to fix these issues ASAP.
Have you already taken a look at my question?
 (https://github.com/Catrobat/Catroid/pull/2724#issuecomment-368007494)
I already tried several things but none of them worked. I really need some help with that issue I tried everything I could come up with by now... Would you mind taking a look into this together some time next week in the panic room?

84n4n4:
ok, i had a closer look on whats going on in formulaEditorEditText to set the position.
now i wonder how this could ever work (and not why its not working anymore in the action).
i found a simpler way to match for the position, so you can just use  getSelectionStart on the editText, but the other way round for setting a specific selection is pretty much not possible the way this was implemented. (please go ahead and have a look at FormulaEditorEditText on how to not implement an edit text and fuck it up for everyone else.) this is fubar.
so i propose dropping the tests that set a cursor position.

AntiDog:
OK

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/utils/FormulaEditorWrapper.java
```diff
@@ -130,6 +132,16 @@ public void performClickOn(Matcher<View> matcher) {
 				.perform(click());
 	}
 
+	public void performLongClickOn(Matcher<View> matcher) {
```
84n4n4: since theres only one usage for each of those two methods, and they are rather specific for those tests they should not be in this wrapper (nobody else is going to need those, and theres little added readability)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,274 @@
...
+		no = getResourcesString(R.string.no);
+		yes = getResourcesString(R.string.yes);
+		decimalMark = getResourcesString(R.string.formula_editor_decimal_mark);
+		onBrickAtPosition(1)
+				.perform(click());
+	}
+
+	@Category({Cat.CatrobatLanguage.class, Level.Smoke.class})
+	@Test
+	public void testHighlightColor() {
```
84n4n4: please remove this test

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
```diff
@@ -0,0 +1,120 @@
...
+				{R.id.formula_editor_keyboard_1, "1 ", "button_1"},
+				{R.id.formula_editor_keyboard_2, "2 ", "button_2"},
+				{R.id.formula_editor_keyboard_3, "3 ", "button_3"},
+				{R.id.formula_editor_keyboard_4, "4 ", "button_4"},
+				{R.id.formula_editor_keyboard_5, "5 ", "button_5"},
+				{R.id.formula_editor_keyboard_6, "6 ", "button_6"},
+				{R.id.formula_editor_keyboard_7, "7 ", "button_7"},
+				{R.id.formula_editor_keyboard_8, "8 ", "button_8"},
+				{R.id.formula_editor_keyboard_9, "9 ", "button_9"},
+				{R.id.formula_editor_keyboard_decimal_mark, "0" + getResources().getString(R.string
```
84n4n4: theres also getResourcesString in uitestutils, which to further increase readability you could proxy through a new method in this test to eg. getString().
if you add the " " in the test where you check them it would clean this up significantly...

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
```diff
@@ -0,0 +1,120 @@
...
+		script.addBrick(new ChangeSizeByNBrick());
+		baseActivityTestRule.launchActivity(null);
+		onView(withId(R.id.brick_change_size_by_edit_text))
+				.perform(click());
+	}
+
+	@Parameterized.Parameters(name = "{2}" + "-Test")
+	public static Iterable<Object[]> data() {
+		return Arrays.asList(new Object[][] {
+				{R.id.formula_editor_keyboard_0, "0 ", "button_0"},
```
84n4n4: please use the matchers already provided in formulaEditorWrapper (NUM0, NUM1, etc..) or even go there and just have one parameter for the test, eg. "8"
and then enter it via performEnterFormula or performEnterNumber and then check if it is in the textfield with an added " " at the end.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.onFormulaEditor;
+import static org.catrobat.catroid.uiespresso.util.UiTestUtils.getResources;
+
+@Category({Cat.AppUi.class, Level.Smoke.class})
+@RunWith(Parameterized.class)
+public class FormulaEditorEditTextGenericTests {
+
+	@Rule
+	public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(SpriteActivity.class, true, false);
```
84n4n4: please state the fragment you want to launch in the rules constructor (already updated in all other tests in code base), and the launchActivity() (without the null)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextGenericTests.java
AntiDog: I don't really get what you mean by "proxying through a new method" and using getResourcesString doesn't improve anything and IMO it makes it even worse since android studio already has the option to only show the text "." instead of getResources().getString(...) while it shows getResourcesString(".") if I use the function. Please see the screenshot below if this explanation isn't clear :S
Difference between the two scenarios

For me, it doesn't make sense to use getResourcesString() over the getResources().getString() since it makes it harder to read because android studio doesn't know that function and hence it can't view it simplified. However, if you really want it this way I can, of course, change it to this I'm just unsure if I understand correctly what you mean...

84n4n4:
test this please

84n4n4:
please run espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,259 @@
...
+		onFormulaEditor()
+				.checkShows(" ");
+	}
+
+	@Category({Cat.CatrobatLanguage.class, Level.Smoke.class})
+	@Test
+	public void testFunctionDeletion() {
+		onFormulaEditor()
+				.performOpenCategory(FUNCTIONS);
+		onView(withText(containsString(random)))
```
84n4n4: please use:
onFormulaEditor()
				.performOpenCategory(FormulaEditorWrapper.Category.FUNCTIONS)
				.performSelect(random);

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,259 @@
...
+				.perform(click());
+		onFormulaEditor()
+				.performClickOn(BACKSPACE);
+		onFormulaEditor()
+				.performClickOn(BACKSPACE);
+
+		onFormulaEditor()
+				.performOpenCategory(FUNCTIONS);
+
+		onView(allOf(withText(containsString(random)), withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE)))
```
84n4n4: onFormulaEditor()
				.performOpenCategory(FormulaEditorWrapper.Category.FUNCTIONS)
				.performSelect(random);

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorEditTextTest.java
```diff
@@ -0,0 +1,259 @@
...
+		onFormulaEditor()
+				.performClickOn(BACKSPACE);
+		onFormulaEditor()
+				.performClickOn(COMPUTE);
+		onView(withId(R.id.formula_editor_compute_dialog_textview))
+				.check(matches(withText("-8.111")));
+	}
+
+	// https://stackoverflow.com/questions/28390574/checking-toast-message-in-android-espresso
+	public void checkToastMessage(int messageID) {
```
84n4n4: i really dislike "check..." methods, since it is never clear what it actually does.
either rename this to "assertToastMessageDisplayed", or even better, have this function return the view matcher and rename to onToastMessage:
ViewInteraction onToastMessage(ViewMatcher matcher) {
	return onView(someMatcher)
			.inRoot(withDecorView(Matchers.not(baseActivityTestRule.getActivity().getWindow()
			.getDecorView())));
}

and then use it as:
onToastMessage(withText(messageId))
	.check(matches(isDisplayed));

this would have the advantage that you can pass anything as a matcher, and since withText() works with Ids as well as strings you get both for free.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/utils/FormulaEditorWrapper.java
```diff
@@ -130,6 +131,11 @@ public void performClickOn(Matcher<View> matcher) {
 				.perform(click());
 	}
 
+	public void performDoubleClickOn(Matcher<View> matcher) {
```
84n4n4: this is so specific to a single test that it doesnt have any business in here. please move to the test, (but use the static matcher from the wrapper)

84n4n4:
test this please

84n4n4:
theres now a toast matcher:
	public static ViewInteraction onToast(Matcher<View> viewMatcher) {


to be used like this:
onToast(withText("some string or id")).check(matches(isDisplayed))

84n4n4:
test this please

84n4n4:
please run espresso tests

84n4n4:
please run espresso tests

PR #2728, opened 2018-02-19
CAT-2312 Add Audio Test for SensorBox

JoachimLesser:
+ Add test with Arduino SensorServerBox if audio is played from PlaySoundBrick.
 + Add SensorboxTestSuite.java for testruns on Jenkins
 + update SensorTestArduinoServerConnection.java with timeout (already) on Socket creation.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/PlaySoundBrickStageTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/PlaySoundBrickStageTest.java
@@ -1,0 +1,130 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2017 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/PlaySoundBrickStageTest.java
```diff
@@ -0,0 +1,129 @@
...
+	private int waitingTime = 2500;
+
+	@Rule
+	public BaseActivityInstrumentationRule<SpriteActivity> programMenuActivityRule = new
+			BaseActivityInstrumentationRule<>(SpriteActivity.class, SpriteActivity.EXTRA_FRAGMENT_POSITION, SpriteActivity.FRAGMENT_SCRIPTS);
+
+	@Before
+	public void setUp() throws Exception {
+		createProject();
+		programMenuActivityRule.launchActivity(null);
```
84n4n4: should be "launchActivity()", if you pass null, the fragment you set in the constructor will be ignored.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/PlaySoundBrickStageTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+		SensorTestArduinoServerConnection.checkAudioSensorValue(SensorTestArduinoServerConnection
+				.SET_AUDIO_ON_VALUE, waitingTime);
+	}
+
+	@Category({Cat.CatrobatLanguage.class, Level.Functional.class, Cat.Gadgets.class, Cat.SettingsAndPermissions
+			.class, Cat.SensorBox.class})
+	@Test
+	public void testNoSoundPlayed() {
+		if (soundFile != null && soundFile.exists()) {
```
84n4n4: is this really supposed to be working like this?
deleting the file, while still leaving the reference in soundInfoList and the brick.

if this is like the name suggests just supposed to just stay quite, use another create project that doenst add a sound in the first place and move the createProject and launchActivity in the respecive tests

JoachimLesser: TY for the fast review! I pushed the refactored version.

PR #2737, opened 2018-02-25
MUS-219 PlayAnimation for multiple tacts

oliewa92:
Change the animation when pressing the play button and make it possible to play multiple tacts.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
@@ -52,6 +52,7 @@ import java.util.Random;
 
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.action.ViewActions.swipeLeft;
 import static android.support.test.espresso.assertion.ViewAssertions.matches;
 import static android.support.test.espresso.contrib.RecyclerViewActions.actionOnItemAtPosition;
 import static android.support.test.espresso.contrib.RecyclerViewActions.scrollToPosition;
@@ -112,7 +113,7 @@ public class PocketMusicActivityTest {
 		return new RecyclerViewMatcher(recyclerViewId);
 	}
 
-	private static Matcher<View> withToggledNoteView(final int position) {
+	private static Matcher<View> isNoteViewToggled(final int position) {
 		return new BoundedMatcher<View, TrackView>(TrackView.class) {
 			@Override
 			public void describeTo(Description description) {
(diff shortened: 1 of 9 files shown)


catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/TactScrollRecyclerView.java
```diff
@@ -52,11 +52,14 @@
 	private static final int MINIMUM_TACT_COUNT = 2;
 	private static final int TACT_VIEW_TYPE = 0;
 	private static final int PLUS_VIEW_TYPE = 1;
+	private static final int EMPTY_VIEW_TYPE = 2;
```
rohmab06: I would assign TACTS_PER_SCREEN to it. Maybe a more specific name would be good: EMPTY_PLAY_ON_END_VIEW_TYPE?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
```diff
@@ -128,10 +129,40 @@ protected boolean matchesSafely(TrackView item) {
 		};
 	}
 
+	private static Matcher<View> withRecyclerViewMeasuredSize() {
```
84n4n4: please rename to something describing what this does, eg. withSizeGreaterZero, (without looking at this matcher code i would have no idea what it is for when reading withRecyclerViewMeasuredSize in test code)

84n4n4: or better **is**SizeGreaterZero, to follow espresso matcher naming conventions.
"with" is used mostly for stuff that takes a parameter that describes something to look for, eg. withText, withId, etc.

while "is" is used for stuff that just matches a condition without parameters, isDisplayed()

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/pocketmusic/PocketMusicActivityTest.java
```diff
@@ -128,10 +129,40 @@ protected boolean matchesSafely(TrackView item) {
...
+			}
+
+			@Override
+			protected boolean matchesSafely(TactScrollRecyclerView item) {
+				return item.getMeasuredHeight() > 0 && item.getMeasuredWidth() > 0;
+			}
+		};
+	}
+
+	private static Matcher<View> withRecyclerViewPlaying() {
```
84n4n4: same as above...

catroid/src/main/java/org/catrobat/catroid/pocketmusic/ui/TactScrollRecyclerView.java
oliewa92: Changed the name and the value of the types to keep the confusion to a minimum.

catroid/src/main/java/org/catrobat/catroid/pocketmusic/ScrollController.java
```diff
@@ -33,58 +33,75 @@
...
 import org.catrobat.catroid.pocketmusic.note.NoteLength;
+import org.catrobat.catroid.pocketmusic.ui.AnimatorUpdateCallback;
+import org.catrobat.catroid.pocketmusic.ui.PianoView;
 import org.catrobat.catroid.pocketmusic.ui.TactScrollRecyclerView;
+import org.catrobat.catroid.pocketmusic.ui.TrackRowView;
 
 public class ScrollController {
 
-	private static final int LAST_NOTE_IN_TRACK_VIEW = 8;
+	private static final int NUMBER_OF_NOTES_IN_TRACK_VIEW = 8;
```
rohmab06: shouldn't it be dependend on the TACTS_PER_SCREEN variable from TactScrollRecyclerView?

oliewa92: you are correct, made it now dependent on TACTS_PER_SCREEN and the MusicalBeat of the TrackGrid in the TactScrollRecyclerView

rohmab06: great!

PR #2758, opened 2018-03-13
CAT-2646 Crash when showing variable with empty string

ndrnour:


ndrnour:
please run espresso tests

thmq:
@robertpainsi do you think this can cause any problems?

robertpainsi:
I like your idea of showing a hint. Since we don't show variables like Scratch does (variable name and text in some kind of bubble), an empty string will show nothing and the user won't know if the show variable brick was executed at all.

imo words like "initialized" should be avoided because kids won't understand them. "variable x is empty" or just "empty" (alternatively nothing?) makes more sense imo. In grey and cursive to highlight that it isn't the string value itself.
However, I'm somehow more drawn to just show an empty string (empty texture or hide actor) as a quickfix.

There are various reasons why a variable evaluates to an empty string.
* The user has set the value to an empty string.
* element(i, list) where i is index out of bounce.
* letter(i, text) where i is index out of bounce.
* the list is empty
* look name when the sprite doesn't have any looks.
* …

I think the problem is, that show variable should always show the variable name as well.
To show text, we should provide it's own bricks and additional bricks like color, size, etc. I hope such feature will be developed.

Hmm, it looks like it has to be discussed first. It would be great if you could discuss that with the UX team and the POs. Maybe work out a temporary solution as well, if necessary. We can always change the text at any time later.

ndrnour:
OK, just let me know, how should be the structure of the hint.

ndrnour:
please run espresso tests

ndrnour:
test this please

ndrnour:
@thmq  could you please have a look!

robertpainsi:
I'm rarely in the panic room nowadays. So, did you discuss this topic with the POs and UX Team?

I'll add a few comments (localization, hard-coded x position, etc) later on once it has been approved.

ndrnour:
@robertpainsi 
I couldn't find a stable way to make the displayed error messages (No Variable Set, No Value Set ) on the stage translatable except using the applicationContext to access Strings.
for that reason i used again Constant error message.
if anyone have a better way. please do it
otherwise 
@thmq  could u please merge it to avoid the Bug in the next release.

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -116,6 +116,8 @@ private void drawVariables(List<UserVariable> variableList, Batch batch) {
 					if (variable.getVisible()) {
 						if (isNumberAndInteger(variableValue)) {
 							drawText(batch, variableValueWithoutDecimal, xPosition, yPosition);
+						} else if (variable.getValue().toString().isEmpty()) {
+							drawText(batch, Constants.NO_VALUE_SET, xPosition - 200, yPosition);
```
robertpainsi: I'm not sure about the `-200`.
If you list multiple variables (one below the other), they won't be arranged nicely. Also the positioning will be inconsistent with all the other options (number, text, no variable set)

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -116,6 +116,8 @@ private void drawVariables(List<UserVariable> variableList, Batch batch) {
 					if (variable.getVisible()) {
 						if (isNumberAndInteger(variableValue)) {
 							drawText(batch, variableValueWithoutDecimal, xPosition, yPosition);
+						} else if (variable.getValue().toString().isEmpty()) {
```
robertpainsi: `variableValue.isEmpty()` :wink:

robertpainsi:
I *guess* it has been discussed → Removing label [Under Discussion]

ndrnour:
please run espresso tests

robertpainsi:
please run espresso tests

PR #2761, opened 2018-03-14
[REFACTORING] Cleanup, adapted default values for scenes, sprites, …

thmq:
…added image preview for sprites.


thmq:
please run espresso tests

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -102,6 +115,26 @@ protected void onCreate(Bundle savedInstanceState) {
 		}
 	}
 
+	private void showContentView(@Content int content) {
+		switch (content) {
+			case FRAGMENT:
+				findViewById(R.id.main_menu_buttons_container).setVisibility(View.VISIBLE);
```
cheidenreich: we could put all the calls to `findViewById(...)` before the switch and store the views into variables, which would make that block more readable

PR #2763, opened 2018-03-14
CAT-2671 Fix Show/HideTextBrick not backward compatible

robertpainsi:
The unused member "userVariableName" has been removed from ShowTextBrick
and HideTextBrick in commit 20638ec4f. Loading older programs (Catrobat
language <= 0.991) which stored this member in the code.xml will crash
on load.

Omiting the member in XStream fixes the crash.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
+++ b/catroid/src/main/java/org/catrobat/catroid/io/StorageHandler.java
@@ -327,6 +327,9 @@ public final class StorageHandler {
 		xstream.omitField(FlashBrick.class, "spinnerValues");
 		xstream.omitField(StopScriptBrick.class, "spinnerValue");
 
+		xstream.omitField(ShowTextBrick.class, "userVariableName");
+		xstream.omitField(HideTextBrick.class, "userVariableName");
+
 		setProgramXstreamAliases();
 	}
 


robertpainsi:
please run espresso tests

robertpainsi:
Same as https://github.com/Catrobat/Catroid/pull/1998
Note that in this case it isn't a XML change and doesn't need a newer Catrobat language. The members were unused for quite some versions and therefore `null`. It looks like that XStream doesn't store `null` values in the XML anyway. That's why you can't reproduce the error with the current play store version. You have to use ~0.9.27 or earlier.

The [documentation](http://x-stream.github.io/javadoc/com/thoughtworks/xstream/XStream.html#omitField-java.lang.Class-java.lang.String-) only states that omit does only omit from serializing fields, however, it also works for deserializing. Here is the trace that shows why
* https://github.com/x-stream/xstream/blob/6acae3f1c9a6ab94e78d0bb0e39429c120fcd718/xstream/src/java/com/thoughtworks/xstream/converters/reflection/AbstractReflectionConverter.java#L351
* https://github.com/x-stream/xstream/blob/6acae3f1c9a6ab94e78d0bb0e39429c120fcd718/xstream/src/java/com/thoughtworks/xstream/converters/reflection/AbstractReflectionConverter.java#L490-L491
* https://github.com/x-stream/xstream/blob/6acae3f1c9a6ab94e78d0bb0e39429c120fcd718/xstream/src/java/com/thoughtworks/xstream/mapper/FieldAliasingMapper.java#L83-L84

robertpainsi:
:tophat:
:koala: (Thanks!)

PR #2764, opened 2018-03-15
Accessibility profiles sandbox

84n4n4:


84n4n4:
if we would skip on the "custom" profile, the whole thing would stay more compact and way easier to read/understand, see version here: https://github.com/84n4n4/Catroid/compare/accessibilityProfiles?expand=1

84n4n4:
ah, yeah, and do not merge yet, commits have to be squashed and i expect some lint/cmd errors and stuff...

PR #2765, opened 2018-03-15
Preparations for Flavoring

cheidenreich:
Skeleton for product flavors.
Just sets the app name and launcher icon for now.
All flavors still have the same appID in this branch


re-cheid:
I'll fix the errors later

robertpainsi:
So, which of you two is going to fix the Checkstyle errors? :rofl: 

You know what, instead of fixing them, just create a new PR and change the end date in the [license header template](https://github.com/Catrobat/Catroid/blob/develop/catroid/src/test/res/agpl_license_text.txt) to something  like `20[1-9][0-9]`. Newly created resources in Android Studio will always have the current year. Old resources can be updated over the year at any time without blocking development.

I'm for completely dropping the date range though. Date ranges with a current end date are bullshit anyway. Don't believe me? Take a look at [Google](https://github.com/google) or [Facebook](https://github.com/facebook).

84n4n4:
from https://www.gnu.org/licenses/gpl-howto.html 

The copyright notice

The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each past release; for example, “Copyright 1998, 1999 Terry Jones” if some releases were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.

For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

although im fully aware that the year ranges are a pain in the ass, since gnu agpl suggests this format we should use it. (since google and facebook links above use apache or bsd licences im not convinced that this is also fine in our case...)

84n4n4:
test this please

robertpainsi:
Thanks for your input ([CAT-2305](https://jira.catrob.at/browse/CAT-2305)). I hope that one day they will update the info to better suite resources that are updated frequently.
Btw, don't forget to retrigger testsruns for all PRs. Maybe there is a jenkins job that does that?

PR #2777, opened 2018-03-26
[GRADLE] Updated android support lib, multidex, play services

thmq:


thmq:
please run espresso tests

PR #2778, opened 2018-03-26
CAT-2377 Implement WhenTouchDownScript and WhenScript as an EventScript

rschluesselbauer:
**Issue**
WhenTouchDownScript does not restart a script, if you touch down two times. Instead the script will run twice in parallel.

**Solution**
Handle WhenTouchScript through the new event system.
To streamline the touch scripts, we also handle WhenScript (basically WhenTapped) through the new event system.
Refactor WhenScript a bit & fix backward compatibility.
Implement basic test for prior incorrect behaviour of WhenTouchDownScript.

**Suggestion**
Rename WhenScript to WhenTappedScript, because it has never been used for anything other than that. That means we have to do some more backwards compatibility stuff, but it would probably confuse people less.


thmq:
test this please

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/content/eventids/EventId.java
```diff
@@ -22,7 +22,48 @@
...
+	}
+
+	public static final int OTHER = 0;
+	public static final int TAP = 1;
+	public static final int TAP_BACKGROUND = 2;
+
+	@EventType
+	private final int type;
+
+	public EventId(int type) {
```
84n4n4: if theres only EventType ints allowed for this, please add @EventType to the parameter.

public EventId(@EventType int type)

84n4n4:
nice!
please fix the one pmd error and have a look at my comment about the EventType int parameter.

rschluesselbauer:
Thanks!
I made the changes + I moved some code from BackwardsCompatibleXstream to StorageHandler.
It makes more sense there and the functionality (omitField) is already part of the Xstream, so I could remove a redundant method from BackwardsCompatibleXstream.

rschluesselbauer:
Removed an old test called WhenActionTest (makes no sense as there is no WhenAction).
Implemented a new WhenScriptTest instead.

84n4n4:
test this please

PR #2780, opened 2018-03-27
[REFACTORING] Removed unused code from backpack

thmq:


thmq:
please run espresso tests

PR #2781, opened 2018-04-06
[TESTING] Added Sprite and Scene Controller Tests

thmq:
Also slightly changed controller behaviour. Copying and packing no longer has side effects i.e. objects are not added to respective lists inside the controller methods.


thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

PR #2785, opened 2018-04-12
Move build steps to scripts, which run on Jenkins like on the local machine

redeamer:
The developer needs to be easily able to run the same build steps on the dev machine like they are run on Jenkins. This requirement includes using the same emulator setting and the same build options like Jenkins.

Therefore the logic of the build steps from the Jenkinsfile is moved to separate script per step ( `buildScripts/build_*`-scripts). Those scripts are capable of running inside a Jenkins build and on a local developer machine. The script are intended to work on Linux/Mac/Windows.

The scripts take care of creating and starting the emulator. The AVD to use can be configured via the `buildScripts/emulator_config.ini`-file.

To avoid external dependencies the Jenkins-Android-Helper: https://github.com/redeamer/jenkins-android-helper (used to handle the SDK installation and the emulator handling) is imported to `buildScripts/jenkins-android-helper`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,5 +62,8 @@ Thumbs.db
 .gradle/
 build/
 
+# Cache for build scripts
+__pycache__
+
 #api-keys
 catroid/google-services.json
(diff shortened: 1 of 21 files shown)


84n4n4:
test this please

PR #2787, opened 2018-04-16
[REFACTORING] Reimplemented unzip without 3rd party tools, added test

thmq:


thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/io/ZipArchiver.java
```diff
@@ -0,0 +1,75 @@
...
+						while ((len = zis.read(buffer)) != -1) {
+							os.write(buffer, 0, len);
+						}
+					} finally {
+						os.close();
+					}
+				}
+			}
+		} finally {
+			zis.close();
```
robertpainsi: Nice way of preventing using null checks! :1st_place_medal:

catroid/src/main/java/org/catrobat/catroid/io/ZipArchiver.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+public class ZipArchiver {
+
+	public void unzip(InputStream is, String dstDir) throws IOException {
```
robertpainsi: It's not a pure function but I would still make it a utils class since no state is stored internally and having multiple instances currently makes no sense.
I'm fine with keeping it the way it is as well.

catroid/src/androidTest/java/org/catrobat/catroid/test/apkgenerator/UnzipProjectFromAssetsTest.java
```diff
@@ -0,0 +1,74 @@
...
+		new ZipArchiver().unzip(inputStream, TEST_PROJECT_PATH);
+
+		assertTrue(new File(TEST_PROJECT_PATH).exists());
+
+		InputStream is = InstrumentationRegistry.getContext().getAssets().open("generated965.zip");
+		ZipInputStream zis = new ZipInputStream(is);
+		ZipEntry entry;
+
+		while ((entry = zis.getNextEntry()) != null) {
+			assertTrue(new File(TEST_PROJECT_PATH + "/" + entry.getName()).exists());
```
robertpainsi:
```java
new File(TEST_PROJECT_PATH, entry.getName())
```

catroid/src/androidTest/java/org/catrobat/catroid/test/apkgenerator/UnzipProjectFromAssetsTest.java
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		zis.close();
+		is.close();
+	}
```
robertpainsi: Improving stuff and also adding tests! That's the way to go!

thmq:
@robertpainsi I have thought about providing methods that take File objects as parameters. However, the StorageHandler as well uses paths instead of files too. I guess the idea behind it was that all classes that do not directly work with storage operations do not have to bother with Files either. I'm not completely happy with it, but I decided to keep it consistent for now. I'm going to think about it a little more, but I think I'll leave it for now. I hope that's alright with you?

@everyone: I plan on adding more tests, so please do NOT merge yet.

catroid/src/androidTest/java/org/catrobat/catroid/test/apkgenerator/UnzipProjectFromAssetsTest.java
thmq: I actually developed it test-driven. 😼

robertpainsi:
So no path-`String`s nor `File`s outside the `StorageHandler` or io package. That definitely sounds good to me!

thmq:
please run espresso tests

PR #2789, opened 2018-04-16
MUS-220: Navigation though tacts

oliewa92:
Implement a fast scroller so you can navigate faster through the tacts.
Also show in which tact you are, when you are scrolling with the fastscroller.


hrieke:
Hi,
Quick question about the FastScoller found at
https://github.com/Catrobat/Catroid/tree/develop/catroid/src/main/java/org/catrobat/catroid/pocketmusic/fastscroller 
- was this copied from https://github.com/FutureMind/recycler-fast-scroll/
?
I'm seeing way too much code overlap, so I am confused to who might have been the original author.

Thank you

PR #2790, opened 2018-04-18
CAT Use internal storage for standalone builds

lwagner94:
This PR adds additional methods to StorageHandler, such as
getRootDirectory(), setRootDirectory() that allow setting the storage
location at runtime. This is necessary because we cannot obtain the path
to internal storage at compile time.
The application code was modified so that we do not use the root directory
defined in Constants.java directly any more, instead we call the
getRootDirectory() method to get the path.
The directory structure is now not created during initialization of the
StorageHandler, but lazily when we first call getRootDirectory().
This is necessary so that we can change the root directory before first use.

As a result of this change, there is no interference between
standalone apps and the PocketCode application.

NOTE: The test code still uses the Constant DEFAULT_ROOT as the tests basically only target "normal" app mode, not the standalone mode. If I should refactor the test code as well, simply drop a line here and I will update the PR.


thmq:
@lwagner94 please rebase this.

thmq:
Nothing since April 18. I commented 23 days ago -> still nothing. a lot of merge conflicts. does not build. -> hence: I'm closing the PR. please rebase and reopen if you still want the change merged.

PR #2796, opened 2018-04-25
CAT-2674 BUG solved the undo redo buttons with updating the options menu

PlantinumX:
- the onPrepareOptionsMenu() function isn't called after overwriting the recent value in the PhiroMotorMoveBackwardBrick. So I invalidate the options menu to update it.


84n4n4:
i just added a comment to the jira ticket, this happens for other bricks that have their own fragment with sliders instead of formula editor. (eg. pen color brick)

so this should be fixed for all of them. please modify the jira ticket also accordingly.

84n4n4:
test this please

84n4n4:
please run espresso tests

PR #2807, opened 2018-05-09
CAT-2043 Translate toast messages

ndrnour:


ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/utils/ToastUtil.java
```diff
@@ -39,18 +39,10 @@ public static void showError(Context context, String message) {
...
-	}
-
 	public static void showSuccess(Context context, String message) {
 		createToast(context, message, false);
 	}
 
-	public static void showSuccess(Context context, int messageId) {
-		createToast(context, context.getResources().getString(messageId), false);
-	}
-
```
robertpainsi: Ahm, I'm not sure about these changes. Why did you remove these methods?

ndrnour: just unused Methods.

robertpainsi: Yes :smile: My question _Why did you remove these methods?_ was meant to be much more general. So I also implied changes like
```diff
-ToastUtil.showError(…, R.string.foo);
+ToastUtil.showError(…, getString(R.string.foo))
```
in my statement. imo this helper method is definitely useful because you already pass in a context object (from which you get your resources).

___
Btw, please take a look at the hint about RTL at https://jira.catrob.at/browse/CAT-2043?focusedCommentId=18602&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18602

ndrnour:
please run espresso tests

thmq:
test this please

thmq:
please run espresso tests

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/RecyclerViewFragment.java
```diff
@@ -296,7 +296,7 @@ protected void prepareActionMode(@ActionModeType int type) {
 
 	private void startActionMode(@ActionModeType int type) {
 		if (adapter.getItems().isEmpty()) {
-			ToastUtil.showError(getActivity(), R.string.am_empty_list);
+			ToastUtil.showError(getActivity(), getString(R.string.am_empty_list));
```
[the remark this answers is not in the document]
ndrnour: cause we have 2 methods which they do the same purpose as far as i know
`public static void showError(Context context, String message) {
		createToast(context, message, true);
	}`

`public static void showError(Context context, int messageId) {
		createToast(context, context.getResources().getString(messageId), true);
	}`
so i just used the first method everywhere, the second one is now unused.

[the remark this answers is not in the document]
thmq: ok 👍 shouldn't we then remove the unused method? I don't know, what do you think?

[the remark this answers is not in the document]
robertpainsi: IMO both methods are useful. They are util methods to provide an easy way to create a Toast, either by a string (which can be used for dynamic data) or a string id. The string id method has the same purpose but provides a shorter, simpler way to create a Toast which is a perfect use case for a util method.

It simplifies lots of cases but mostly cases like
```diff
- ToastUtil.showError(currentActivity, R.string.error_scratch_program_conversion_failed);
+ ToastUtil.showError(currentActivity, currentActivity.getResources().getString(R.string.error_scratch_program_conversion_failed));
```

btw, [Googles Toast.makeText](https://developer.android.com/reference/android/widget/Toast.html) provides two ways to create a standard Toast as well (string id or string).
___
For this ticket, please only address the hard-coded strings and string concatenation (localization) for now. We can talk about removing (or improving) the util methods later :wink:
```java
// $ grep -ER 'ToastUtil.*"'
ToastUtil.showError(PreStageActivity.this, "Error: connecting to " + host + ":" + port + " failed");
ToastUtil.showError(this, getResources().getText(R.string.error_project_upload).toString() + " " + serverAnswer);
ToastUtil.showError(this, R.string.notification_low_battery + " " + Integer.toString(value));
ToastUtil.showSuccess(context, numberOfBackPackedItems + " " + textForScripts + " "
ToastUtil.showSuccess(activity, selectedScriptGroupBackPack + " " + textForUnPacking);
ToastUtil.showError(getActivity(), "There was a problem during Google+ Signin. Status: "
```

ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

ndrnour:
@robertpainsi  could you please review again

catroid/src/main/java/org/catrobat/catroid/drone/ardrone/DroneStageActivity.java
```diff
@@ -147,7 +147,7 @@ public void onDroneBatteryChanged(int value) {
 
 		DroneControlService dcs = DroneServiceWrapper.getInstance().getDroneService();
 		if (dcs != null && (value < DroneInitializer.DRONE_BATTERY_THRESHOLD) && dcs.getDroneNavData().flying && !droneBatteryMessageShown) {
-			ToastUtil.showError(this, R.string.notification_low_battery + " " + Integer.toString(value));
+			ToastUtil.showError(this, getString(R.string.notification_low_battery) + " " + value);
```
robertpainsi: If you want, you can also localize this resource. Up to you :wink:

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -199,7 +199,7 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
 					intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_SINGLE_TOP);
 					startActivity(intent);
 				} else {
-					ToastUtil.showError(getApplicationContext(), R.string.error_no_whatsapp);
+					ToastUtil.showError(getBaseContext(), R.string.error_no_whatsapp);
```
robertpainsi: I can't approve these changes. For the ToastUtil, I guess it doesn't matter which context you pass in?…

@thmq, since you're very experienced with Android, maybe you can take a look at the context changes?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/SignInDialog.java
```diff
@@ -342,7 +342,7 @@ private void triggerGPlusLogin(GoogleSignInResult result) {
 			GoogleSignInAccount account = result.getSignInAccount();
 			onGoogleLogInComplete(account);
 		} else {
-			ToastUtil.showError(getActivity(), "ERROR Google+ Signin. Status: " + result.getStatus());
+			ToastUtil.showError(getActivity(), getString(R.string.error_google_plus_sign_in, result.getStatus().getStatusMessage()));
```
robertpainsi: Even though the documentation for the Status class is non existent, let's go with that.

catroid/src/main/res/values/strings.xml
```diff
@@ -901,6 +901,8 @@
     <string name="show_password">Show Password</string>
     <string name="register_password_mismatch">Passwords do not match</string>
     <string name="sign_in_error">Error during Sign-In</string>
+    <string name="error_connecting_to">Error: connecting failed to %s : %d%%</string>
```
robertpainsi: Either use
* `Failed to connect to %s : %d`
* `Connection to %s : %d failed`
* `Couldn't connect to %s : %d`
* or something similar

I could swear that the [material design writing guidelines](https://material.io/design/usability/accessibility.html#writing) have been much more precise on such strings. From a consistency point of view it doesn't matter which one you choose, we use all styles :smile:

IMO it should be `Couldn't connect to Raspberry Pi` (optionally with host and port). What do you think?

(Adding/changing string resources is expensive. It takes a long time until strings are translated to all languages. So we should always choose strings carefully!)

robertpainsi: btw, is there a better place for that string? All other strings in this block deal with login user accounts (login, register) only.

catroid/src/main/res/values/strings.xml
```diff
@@ -901,6 +901,8 @@
...
+    <string name="error_google_plus_sign_in">ERROR Google+ Sign in. Status: %s</string>
```
robertpainsi: Maybe you have some example status messages (since you used `result.getStatus().getStatusMessage()`)?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/SignInDialog.java
ndrnour: `result.getStatus()` will return an object of type Status
but `result.getStatus().getStatusMessage()` will return a string therefor i used it.
did i do it wrong?

catroid/src/main/java/org/catrobat/catroid/drone/ardrone/DroneStageActivity.java
ndrnour: ok

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
ndrnour: i used `getBaseContext()` instead of `getApplicationContext()` to reduce using the _**static_Contetxt**_

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/SignInDialog.java
robertpainsi: > did i do it wrong?

Quite the opposite actually! :wink: I like your solution.

I was just mentioning that there is no [documentation for the getStatusMessage method](https://developers.google.com/android/reference/com/google/android/gms/common/api/Status#getStatusMessage()). I also couldn't find more information online. There are some code snippets at https://developers.google.com/games/services/android/signin which are using `getStatusMessage` as well. However, it still would've been great to see some example messages that `getStatusMessage` return.

I can't say for sure what `result.getStatus().toString()` returns but it looks like it's returning a String only meant to be read by developers. `getStatusMessage` seems better suited. So, let's go with your solution! :1st_place_medal:

catroid/src/main/res/values/strings.xml
```diff
@@ -1366,6 +1367,7 @@
     <!-- Raspi Settings -->
     <string name="raspi_settings_title_name">Raspberry Pi settings</string>
     <string name="raspi_settings_gpio_version">GPIO Version</string>
+    <string name="error_connecting_to">Error: Could not connect to %s : %d%%</string>
```
robertpainsi: Ahm, why the two `%` at the end?

ndrnour: i was getting a lint error without %% .
but now it's fine -_- i don't know how or why! just a magic happened :)))

ndrnour:
please run espresso tests

catroid/src/main/res/values/strings.xml
```diff
@@ -901,6 +901,7 @@
     <string name="show_password">Show Password</string>
     <string name="register_password_mismatch">Passwords do not match</string>
     <string name="sign_in_error">Error during Sign-In</string>
+    <string name="error_google_plus_sign_in">ERROR Google+ Sign in. Status: %s</string>
```
robertpainsi: After looking through a lot of string resources, I think there is no value in trying to improve these newly added strings :laughing: So I'll keep my suggestions at a minimum. Feel free to add feedback, especially if you disagree.

`Google Sign-In error. Status: %d`

ndrnour: why %d at  end?

robertpainsi: > why %d at end? string not Integer

Ah, yes! Nice catch!

PR #2818, opened 2018-05-22
CAT-2706 Refactor WhenConditionScript to use the new event system

rschluesselbauer:
- Add ConditionScriptTrigger that evaluates a condition and triggers
  actions (via event) if it evaluates to true
- Add Set of ConditionScriptTriggers to Sprite, initialize it at
  beginning of Stage, invalidate at end of Stage
- Evaluate ConditionScriptTriggers after every tick in Look.act
- Write test for WhenConditionScript
- Replace getAllBricks method from Sprite with getListWithAllBricks
  method (they behaved almost the same way and the difference in
  behaviour does not affect any of the callers)
- Remove old part of code in StageListener that was only necessary for
  the old code coverage tool EMMA (no longer in use)


rschluesselbauer:
test this please

84n4n4:
test this please

PR #2819, opened 2018-05-22
CAT-2673 none sense error message while registration 

ndrnour:
* Sets an error message that will be displayed below the TextInputLayouts 
(Username/E-mail/Password/confirm Password ) instead of showing AlertDialogs
* Check E-mail and Username availability in background
* Username should start with Latin character or number and contains no "@" - to block using email as username
* Check if the the provided Email_address matches Patterns.EMAIL_ADDRESS
* Password should be at least 6 characters 
* Check if Password and Confirm Password are matched
**have a look at the [video](https://drive.google.com/open?id=1rVOdzGVHA-_MGbm_fFbrdLzWaUw2OdqB
)**
* it still remains one problem which is : 
change the language of the App to Deutsch
try to register --> server error " Invalid country"
_need to be fixed with the web team " I've tried to arrange a meeting with them but No answer yet"_

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
@@ -28,6 +28,7 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
 import android.os.Environment;
 import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
@@ -95,7 +96,11 @@ public final class Utils {
 	public static boolean isNetworkAvailable(Context context) {
 		ConnectivityManager connectivityManager = (ConnectivityManager) context
 				.getSystemService(Context.CONNECTIVITY_SERVICE);
-		return connectivityManager != null && connectivityManager.getActiveNetworkInfo().isConnected();
+		NetworkInfo activeNetworkInfo = null;
+		if (connectivityManager != null) {
+			activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
+		}
+		return activeNetworkInfo != null && activeNetworkInfo.isConnected();
 	}
 
 	public static boolean checkForNetworkError(boolean success, WebconnectionException exception) {
--- a/catroid/src/main/res/values/strings.xml
+++ b/catroid/src/main/res/values/strings.xml
@@ -901,12 +901,22 @@
     <string name="signin_choose_username_hint">The username should not be your real name.</string>
     <string name="oauth_username_taken">Someone else already uses this name. Please choose a different one!</string>
     <string name="show_password">Show Password</string>
-    <string name="register_password_mismatch">Passwords do not match</string>
     <string name="sign_in_error">Error during Sign-In</string>
     <string name="error_google_plus_sign_in">Google Sign-In error. Status: %s</string>
     <string name="error_facebook_session_expired">Session expired, please log in.</string>
     <string name="sign_in_with_facebook">Sign in with Facebook</string>
     <string name="sign_in_with_google">Sign in with Google</string>
+    <string name="error_register_empty_username">Empty username</string>
+    <string name="error_register_username_as_email">E-mail address can not be used as username</string>
+    <string name="error_register_invalid_username">Username can contain only Latin–letters (a–z), numbers (0–9), dashes (-), underscores (_), and periods (.).</string>
+    <string name="error_register_username_start_with">Username can not start with dashes (-), underscores (_), and periods (.).</string>
+    <string name="error_register_invalid_email_format">Invalid E-mail format</string>
+    <string name="error_register_empty_password">Empty Password</string>
+    <string name="error_register_password_at_least_6_characters">Password should be at least 6 characters</string>
+    <string name="error_register_passwords_mismatch">Passwords mismatch</string>
+    <string name="error_register_empty_confirm_password">Empty confirm password</string>
+    <string name="error_register_username_already_exists">Username already exists</string>
+    <string name="error_register_email_exists">Email Address already exists</string>
     <!--  -->
 
 
--- a/catroid/src/main/java/org/catrobat/catroid/transfers/CheckEmailAvailableTaskThroughRegistration.java
+++ b/catroid/src/main/java/org/catrobat/catroid/transfers/CheckEmailAvailableTaskThroughRegistration.java
@@ -1,0 +1,73 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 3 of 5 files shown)


ndrnour:
please run espresso tests

ndrnour:
@robertpainsi  could you have a look! any improvements ?

ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

ndrnour:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/transfers/RegistrationTask.java
```diff
@@ -134,12 +137,47 @@ private void showDialog(int messageId) {
...
 	}
 
+	private void showDialogWithMessage(int message) {
+		new AlertDialog.Builder(context).setTitle(R.string.register_error).setMessage(message)
+				.setPositiveButton(R.string.ok, null).show();
+	}
+
+	private void openResetPasswordWebPage() {
+		Intent resetPassword = new Intent(Intent.ACTION_VIEW, Uri.parse(Constants.RESET_PASSWORD_URL));
+		context.startActivity(resetPassword);
```
robertpainsi: imo that's a great idea! I love it!

Can you please create a ticket for that so a ux member can review it and give some feedback/approve it.

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -165,6 +166,17 @@
...
+	// RegistrationTask Error-Messages
+	public static final String MISSING_PASSWORD = "The password is missing.";
+	public static final String NOT_BLANK_USERNAME = "Username must not be blank";
+	public static final String NOT_BLANK_EMAIL = "email must not be blank";
+	public static final String PASSWORD_LENGTH = "Your password must have at least 6 characters.";
+	public static final String INVALID_COUNTRY = "Invalid country";
+	public static final String USERNAME_EXISTS = "This username already exists.";
+	public static final String INVALID_USERNAME_VALUE = "This value is not valid.";
+	public static final String INVALID_EMAIL_ADDRESS = "Your email seems to be invalid";
+	public static final String EMAIL_ADDRESS_EXISTS = "This email address already exists.";
```
robertpainsi: These strings are actually localized strings from the web api. Checking for these values should be avoided imo. They are meant to be read by the user and can easily change any time.

We are passing a language value to the register web api. I don't know if it's the device language or the app language though but the api response can be a localized by using this value. So the web api could response with a localized string.

Currently it's hard for me to say how and where to fix it. Especially with having material design in mind where all errors should be reported on every input field :thinking: 

* Should the web team just respond with localized error messages?
* And/Or should the web team just respond with an error code or/and a description of the error in plain English (for developers only)
* And/Or should Catroid also implement some checking before sending a request to the server?

I'm not going into details for each point, because this it's getting confusing really quick. However, these are the three main points which should be discussed with the web and ux team first. As you know, I'm currently rarely in Graz nowadays. So it would be great if you could do it. Also pinging @thmq because he is working on material design.

The `This value is not valid.` message is definitely something the web team should fix!

catroid/src/main/java/org/catrobat/catroid/transfers/RegistrationTask.java
ndrnour: done! CAT-2733

catroid/src/main/res/values/strings.xml
```diff
@@ -889,6 +889,16 @@
     <string name="new_user_registered">New account created</string>
     <string name="user_logged_in">Login successful</string>
     <string name="register_error">Login invalid</string>
+    <string name="register_error_missing_password">The password is missing!</string>
+    <string name="register_error_blank_username">Username must not be blank!</string>
+    <string name="register_error_blank_email">Email must not be blank!</string>
+    <string name="register_error_password_length">Your password must have at least 6 characters!</string>
+    <string name="register_error_invalid_country">Invalid country!</string>
+    <string name="register_error_username_exists">This username already exists!</string>
+    <string name="register_username_invalid_input_value">Username can contain only Latin–letters (a–z), numbers (0–9), dashes (-), underscores (_), and periods (.)!</string>
```
robertpainsi: JFI: Checked it, not true. Used _Tüf الحُرُوف_ as username, so I think all sorts of characters are allowed (excluding the @ sign)

robertpainsi: Just saw your ticket about limiting the usable characters in a username. I'll drop a comment there later today.

ndrnour: non- Latin-letters are also excludes 
screenshot_20180603-225908

robertpainsi: You're totally right!

This seems to be inconsistent with the Web registration form (I only tested it there with username `Tüf الحُرُوف`). It even looks like there are three different user name validation implementations :worried: 

1. [CreateUserRequest.php#L20](https://github.com/Catrobat/Catroweb-Symfony/blob/b5cce0fd23ed0987c0a410fc275d9cb960148087/src/Catrobat/AppBundle/Requests/CreateUserRequest.php#L20)
2. [CreateOAuthUserRequest.php#L19](https://github.com/Catrobat/Catroweb-Symfony/blob/b5cce0fd23ed0987c0a410fc275d9cb960148087/src/Catrobat/AppBundle/Requests/CreateOAuthUserRequest.php#L19)
3. [RegisterController.php#L15](https://github.com/Catrobat/Catroweb-Symfony/blob/763f2ec79243e0cc5ebb66c01f1916088a271808/src/Catrobat/AppBundle/Controller/Web/RegisterController.php#L15)

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
ndrnour: since these strings are localized from the web api, then as i guess its hard to deal with ticket [CAT-2733](https://jira.catrob.at/browse/CAT-2733). what do you think about?

robertpainsi: Ahm, can you please be a little more specific?

ndrnour: to show the dialog with reset password depends on a specific answer from the server which is in this case "This email address already exists." if this answer was localized then i have to deal with 54 translation of this answer to show up the resetPassword dialog!
I hope you got my point?

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -118,7 +119,11 @@ public void onClick(DialogInterface dialog, int which) {
 	public static boolean isNetworkAvailable(Context context) {
 		ConnectivityManager connectivityManager = (ConnectivityManager) context
 				.getSystemService(Context.CONNECTIVITY_SERVICE);
-		return connectivityManager != null && connectivityManager.getActiveNetworkInfo().isConnected();
+		NetworkInfo activeNetworkInfo = null;
+		if (connectivityManager != null) {
+			activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
+		}
+		return activeNetworkInfo != null && activeNetworkInfo.isConnected();
```
ndrnour: i did these changes because App crush if there is no active internet connection while taping on register Btn.

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
robertpainsi: Ah, I got ya! Yes, an error code (or similar) would definitely be beneficial here. Especially if the error is displayed on the `EditText` itself.

PR #2821, opened 2018-05-23
CAT-2106 Do not allow email addresses as user names

ndrnour:
easy way to fix this by preventing user from using "@" in the username field
cause Z.B `example@gmail` is understandable AS E-Mail address as well

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/RegistrationDialogFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/RegistrationDialogFragment.java
@@ -125,16 +125,22 @@ public class RegistrationDialogFragment extends DialogFragment implements OnRegi
 		String passwordConfirmation = confirmPasswordInputLayout.getEditText().getText().toString();
 		String email = emailInputLayout.getEditText().getText().toString();
 
-		if (password.equals(passwordConfirmation)) {
+		if (username.contains("@")) {
+			showErrorAlertDialog(R.string.register_email_as_username_error);
+		} else if (!password.equals(passwordConfirmation)) {
+			showErrorAlertDialog(R.string.register_password_mismatch);
+		} else {
 			RegistrationTask registrationTask = new RegistrationTask(getActivity(), username, password, email);
 			registrationTask.setOnRegistrationCompleteListener(this);
 			registrationTask.execute();
-		} else {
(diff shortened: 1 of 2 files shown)


ndrnour:
@robertpainsi what's your opinion about what I did here?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/login/RegistrationDialogFragment.java
```diff
@@ -125,16 +125,22 @@ private void onRegisterButtonClick() {
...
 			RegistrationTask registrationTask = new RegistrationTask(getActivity(), username, password, email);
 			registrationTask.setOnRegistrationCompleteListener(this);
 			registrationTask.execute();
 		} else {
-			new AlertDialog.Builder(getActivity())
-					.setTitle(R.string.register_error)
-					.setMessage(R.string.register_password_mismatch)
-					.setPositiveButton(R.string.ok, null)
-					.show();
+			showErrorAlertDialog(R.string.register_password_mismatch);
```
robertpainsi: Please change the order of the if-else statements
```
if ("@" in username)
  show error
else if (not password equals) 
  show error
else
  execute registration task
```

ndrnour: done.

catroid/src/main/res/values/strings.xml
```diff
@@ -902,6 +902,7 @@
     <string name="oauth_username_taken">Someone else already uses this name. Please choose a different one!</string>
     <string name="show_password">Show Password</string>
     <string name="register_password_mismatch">Passwords do not match</string>
+    <string name="register_email_as_username_error">Email address can not used as username</string>
```
robertpainsi: Email address can not **be** used as username

robertpainsi:
The web team also just checks for the @ sign https://github.com/Catrobat/Catroweb-Symfony/pull/123/files#diff-7f7ef968389cbeb3ea2934cb53172debR21

catroid/src/main/res/values/strings.xml
ndrnour: done

ndrnour:
I've fixed this in other PR [2819](https://github.com/Catrobat/Catroid/pull/2819)

PR #2824, opened 2018-05-24
CAT-2729 Replace integer division with float division in PlaySoundAndWaitBrick

rschluesselbauer:
With the int division the calculated duration was very inaccurate.


robertpainsi:
test this please

robertpainsi:
please run espresso tests

PR #2825, opened 2018-05-24
CAT-2730 Refactor WhenStartedBrick, WhenClonedBrick, WhenNfcBrick and WhenGamepadButtonBrick to use the new event system

rschluesselbauer:
- Let StartScript, WhenClonedScript, WhenGamepadButtonScript and
WhenNfcScript implement EventScript and create corresponding EventIds
- Fire Start event at the beginning of the stage
- Fire Clone event when cloning of a sprite is done
- Fire Nfc event and GamepadEvent at the appropriate places
- Refactor WhenGamepadButtonBrick and CastManager (minor)
- Remove old methods for action creation from Sprite

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/eventids/NfcEventId.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/eventids/NfcEventId.java
@@ -20,34 +20,16 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+
 package org.catrobat.catroid.content.eventids;
 
-import android.support.annotation.IntDef;
+import com.google.common.base.Objects;
 
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
+public class NfcEventId extends EventId {
+	final String tag;
 
-public class EventId {
-	@Retention(RetentionPolicy.SOURCE)
-	@IntDef({TAP, TAP_BACKGROUND, START, START_AS_CLONE, OTHER})
-	public @interface EventType {
-	}
-
-	public static final int OTHER = 0;
-	public static final int TAP = 1;
-	public static final int TAP_BACKGROUND = 2;
-	public static final int START = 3;
-	public static final int START_AS_CLONE = 4;
-
-	@EventType
-	private final int type;
-
-	public EventId(@EventType int type) {
-		this.type = type;
-	}
-
-	protected EventId() {
-		this.type = OTHER;
+	public NfcEventId(String tag) {
+		this.tag = tag;
 	}
 
 	@Override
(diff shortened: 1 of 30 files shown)


rschluesselbauer:
If the tests run through, this PR is ready for review.

thmq:
test this please

thmq:
please run espresso tests

thmq:
It's funny, that the coordinator is not listed as a member on github an hence cannot even start test runs on jenkins 😆

thmq:
test this please

catroid/src/main/java/org/catrobat/catroid/content/eventids/NfcEventId.java
```diff
@@ -0,0 +1,54 @@
...
+	public NfcEventId(String tag) {
+		this.tag = tag;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
```
thmq: is there a reason you use ```getClass() != o.getClass() ``` instead of ``` !o instanceof NfcEventId``` ? I'm not saying you should change it, I would just like to hear your reasons 😉

rschluesselbauer: Actually I just let Android Studio autogenerate it.
I would probably also use instanceof, because it provides nullcheck and class check in one. The only difference I would expect is the behaviour when this class gets inherited.

rschluesselbauer:
Yeah I axed about the rights once, but apparently it didn't work ;)
Thanks for merging!

catroid/src/main/java/org/catrobat/catroid/content/eventids/NfcEventId.java
thmq: sure, makes sense. so do you want it to behave this way when it is inherited? -> anyway, only food for thought anyway. all fine for now as far as I'm concerned. I'll merge it.

thmq: once you rebased it of course 😉

rschluesselbauer:
I fixed the merge conflict in Sprite (due to formatting).
Please rerun tests and merge :-)

rschluesselbauer:
test this please

catroid/src/main/java/org/catrobat/catroid/content/eventids/NfcEventId.java
rschluesselbauer: Now that I think about it, it makes sense that two objects can only be equal if they have the exact same class. 
Ofc it depends on what you want to do, but intuitively this is exactly what I would expect equals to do.

PR #2826, opened 2018-05-25
[GSoC18] Simplified Scene Class, removed merge code relics from bricks

thmq:


thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
test this please

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -249,10 +250,10 @@ public void createAndAddActions(@CreateActionsMode int includeStartActions) {
...
 
 	private void createAndAddAction(Script script, @CreateActionsMode int includeStartActions) {
-		if (script.isCommentedOut()) {
-			return;
-		} else if (!isClone && script instanceof StartScript
-				|| isClone && script instanceof WhenClonedScript) {
+		if (!script.isCommentedOut() && !isClone
+				&& script instanceof StartScript
+				|| isClone
+				&& script instanceof WhenClonedScript) {
```
robertpainsi: Since`&&`>`||`, braces are missing `!script.isCommentedOut() && (a || b)`. btw, the formatting is a little off here :wink:

However, I think this is a behavior change:
Previously, if the script is commented out, the method exits;
Now, if the script is commented out, the if block evaluates to false, however the else if block (`else if (script instanceof EventScript)`) will still be executed (but shouldn't).

I personally love guard clauses, but ___imo___ (note that the imo is underlined and italic! :wink:) it should be
```java
if (script.isCommentedOut()) {
	return;
}

if (!isClone && script instanceof…
```

If the script is commented out, early exit and we're done here.
If not, do the "real" logic.

catroid/src/main/java/org/catrobat/catroid/content/Scene.java
```diff
@@ -75,56 +69,91 @@
 	@XStreamAlias("data")
 	private DataContainer dataContainer = null;
 	@XStreamAlias("originalWidth")
-	private int originalWidth = 0;
+	public int originalWidth = 0;
 	@XStreamAlias("originalHeight")
-	private int originalHeight = 0;
+	public int originalHeight = 0;
```
robertpainsi: Are we serializing unused fields here or do (will) they have any purpose? (I'm not up-to-date with the different screen size handling…)

Why the change though?

thmq: I have absolutely no idea what it does, but it is part of the code.xml and because I did not want to change the xml parsing, I left the variable and just removed the unused getters, which then caused a checkstyle or lint (I don't remember which one it was) error because you're not supposed to have unused private variables (obviously).

thmq:
test this please

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
thmq: sure :wink

thmq:
test this please

thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

catroid/src/main/java/org/catrobat/catroid/content/Scene.java
robertpainsi: Ah, now I get it, thanks!

thmq:
please run espresso tests

thmq:
Espresso run [here](https://jenkins.catrob.at/job/CatroidTriggerPullRequestEspressoTest/721/testReport/).

PR #2832, opened 2018-06-04

shellp:
- larger font-size for axes

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
@@ -87,6 +87,7 @@ public class StageActivity extends AndroidApplication {
 	public static final int ASK_MESSAGE = 0;
 	public static final int REGISTER_INTENT = 1;
 	private static final int PERFORM_INTENT = 2;
+	private static final int NUMBER_OF_BITS_PER_COLOR_CHANNEL = 8;
 
 	private StageAudioFocus stageAudioFocus;
 	private PendingIntent pendingIntent;
@@ -123,13 +124,12 @@ public class StageActivity extends AndroidApplication {
 
 		getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
 
-		stageListener = new StageListener();
+		stageListener = new StageListener(getContext().getResources().getColor(R.color.axis_font_color));
 		stageDialog = new StageDialog(this, stageListener, R.style.stage_dialog);
 		calculateScreenSizes();
 
-		// need we this here?
 		configuration = new AndroidApplicationConfiguration();
-		configuration.r = configuration.g = configuration.b = configuration.a = 8;
+		configuration.r = configuration.g = configuration.b = configuration.a = NUMBER_OF_BITS_PER_COLOR_CHANNEL;
 
 		if (ProjectManager.getInstance().getCurrentProject().isCastProject()) {
 			setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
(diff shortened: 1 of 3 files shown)


thmq:
test this please

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -195,6 +192,10 @@ public void create() {
 		virtualWidthHalf = virtualWidth / 2;
 		virtualHeightHalf = virtualHeight / 2;
 
+		font = new BitmapFont();
+		font.setColor(1f, 0f, 0.05f, 1f);
```
84n4n4: is there any reason why this is hard coded and not in colors.xml?

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -195,6 +192,10 @@ public void create() {
...
+		font.getData().setScale(1.2f);
```
84n4n4: since hard coded values should be avoided, and i guess its important that its the same value as in line 633, please put this in a member variable and give it a proper name.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -591,21 +592,45 @@ private void printPhysicsLabelOnScreen() {
...
+		layout.setText(font, String.valueOf((int) virtualWidthHalf));
+
+		float landscapeHeight = (virtualHeight < virtualWidth) ? virtualHeight : virtualWidth;
+		float fontPercentage = layout.height / landscapeHeight;
+		float scaleFactor = 0.03f / fontPercentage;
+
+		layout.height = layout.height * scaleFactor;
+		layout.width = layout.width * scaleFactor;
+
+		font.getData().setScale(scaleFactor);
```
84n4n4: method with side effect... so after one has called this, you have to reset the font scale factor to what it was before yourself...

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -195,6 +198,15 @@ public void create() {
 		virtualWidthHalf = virtualWidth / 2;
 		virtualHeightHalf = virtualHeight / 2;
 
+		font = new BitmapFont();
+
+		Context context = CatroidApplication.getAppContext();
```
84n4n4: this bothers me. i want to get rid of all the CatroidApplication.getAppContext calls, not introduce new ones.
id rather have this resource getting stuff somewhere else, since this seems to be the only point in this whole class that requires a context. one possibility would be that the stageActivity has a axesColor.

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -127,6 +128,8 @@ public void onCreate(Bundle savedInstanceState) {
 		stageDialog = new StageDialog(this, stageListener, R.style.stage_dialog);
 		calculateScreenSizes();
 
+		axesColor = getContext().getResources().getColor(R.color.axis_font_color);
```
84n4n4: so is this really working? since new StageListener is called in 127, while the color is set here.

add the color to the constructor of the stage listener, initilialize the color there, and dont even save it here in the stage activity since its never again used (besides switching the scene, which is again handled in the stage listener anyways.)

and since youre already touching this, please remove below comment, and make a new final int NUMBER_OF_BITS_PER_COLOR_CHANNEL = 8; and use it instead of the hardcoded 8...

wslany:
Tested on 360x480 up to 1440x2960 on various devices, from Android 4.3 to Android 8.0, and in landscape and portrait projects. Looked good on all of them.

wslany:
However, there seem to be no automatic tests. How would one go on to test this?

84n4n4:
please write unit tests, and then reopen.

PR #2843, opened 2018-06-08
CAT-2674 BUG regarding undo/redo buttons and discarding the changes.

PlantinumX:
CAT-2674 BUG regarding undo/redo buttons and discarding the changes.
Discarding the changes results in  storing false values.
To solve both bugs  changed structure in FormulaEditorHistory and made getBrickorCustomView public to access it from formulaEditorEditText.
In addition created new Class State for storing the State. A state consists of
InternformulaState and Brickfield.
Bug with discarding the changes-> solved it  with pushing the according values on top of stack and consider it as a save.
Changed in PenBricksTest the testPenColorBrick testcase because it is not
testing the right behaviour.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
+++ b/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
@@ -22,31 +22,42 @@
  */
 package org.catrobat.catroid.formulaeditor;
 
+
+import org.catrobat.catroid.content.bricks.Brick;
+
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Stack;
 
 public class FormulaEditorHistory {
 
 	private static final int MAXIMUM_HISTORY_LENGTH = 32;
-	private Stack<InternFormulaState> undoStack = null;
-	private Stack<InternFormulaState> redoStack = null;
-	private InternFormulaState current = null;
+	private Stack<State> undoStack;
+	private Stack<State>redoStack;
+	private State current;
 	private boolean hasUnsavedChanges = false;
+	Map<Brick.BrickField,InternFormulaState> initialStates;
 
-	public FormulaEditorHistory(InternFormulaState internFormulaState) {
-		current = internFormulaState;
-		undoStack = new Stack<InternFormulaState>();
-		redoStack = new Stack<InternFormulaState>();
+
+	public FormulaEditorHistory(State state) {
+		undoStack = new Stack<State>();
+		redoStack = new Stack<State>();
+		initialStates = new HashMap<Brick.BrickField,InternFormulaState>();
+		current = state;
 	}
 
-	public void push(InternFormulaState internFormulaState) {
+	public void push(State state) {
 
-		if (current != null && current.equals(internFormulaState)) {
+		if (current != null && current.equals(state)) {
 			return;
 		}
 		if (current != null) {
 			undoStack.push(current);
 		}
-		current = internFormulaState;
+		if(!initialStates.containsKey(current.brickField)) {
+			initialStates.put(current.brickField,current.internFormulaState);
+		}
+		current = state;
 		redoStack.clear();
 		hasUnsavedChanges = true;
 		if (undoStack.size() > MAXIMUM_HISTORY_LENGTH) {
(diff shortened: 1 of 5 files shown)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
```diff
@@ -22,51 +22,68 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
 import java.util.Stack;
 
 public class FormulaEditorHistory {
 
 	private static final int MAXIMUM_HISTORY_LENGTH = 32;
 	private Stack<InternFormulaState> undoStack = null;
 	private Stack<InternFormulaState> redoStack = null;
```
84n4n4: maintaining two collections where each entry is corresponding with one in the other collection should be avoided. (you have to always maintain integrity of both lists, if something goes wrong here and a value from one list is lost, or added too much, both lists are corrupted)

so instead you could implement a small class that contains the brick field and the formula state, and one collection (in this case stack) that contains for each undo step a single object.:

public class UndoState<InternFormulaState, BrickField> { 
  public final InternFormulaState internFormulaState; 
  public final BrickField brickField; 
  public Tuple(InternFormulaState internFormulaState, BrickField brickField) { 
    this.brickField = brickField; 
    this.internFormulaState = yinternFormulaState 
  } 
} 

private Stack<UndoState> undoStack;

as a side note, theres no point in the null initialization.

PlantinumX: Thank you for the hint!

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
```diff
@@ -857,4 +850,22 @@ public void updateButtonsOnKeyboardAndInvalidateOptionsMenu() {
 			backspaceOnKeyboard.setEnabled(true);
 		}
 	}
+
+	private boolean discardFormulaIfPossible() {
```
84n4n4: wouldnt it be easier to save the initial state of a formula instead of unrolling the complete undo stack?

also, dont call this discardFormulaIfPossible, because this it performs the undo if theres unsaved changes. therefore when it is "nessecary", not possible, and it performs this on FormulaS.
but anyways, dont rename this method, but split it, and have the check if changes are to be discarded where it is called, and only make a discardChanges method.

PlantinumX: no because in case of colorseekbar you have 3 states which should be saved so it is easier to iterate through the stack

catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
```diff
@@ -673,25 +664,23 @@ private void showToast(int resourceId, boolean error) {
...
 			builder.setTitle(R.string.formula_editor_discard_changes_dialog_title)
 					.setMessage(R.string.formula_editor_discard_changes_dialog_message)
 					.setNegativeButton(R.string.no, new OnClickListener() {
 
 						@Override
 						public void onClick(DialogInterface dialog, int which) {
-							ToastUtil.showError(getActivity(), R.string.formula_editor_changes_discarded);
-							currentFormula.setDisplayText(null);
-							onUserDismiss();
+							if (discardFormulaIfPossible()) {
```
84n4n4: if this returns false for whatever reason the dialog will not go away...

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
```diff
@@ -22,51 +22,68 @@
...
+	private Stack<Brick.BrickField> brickFieldStackUndo = null;
+	private Stack<Brick.BrickField> brickFieldStackRedo = null;
 	private InternFormulaState current = null;
+	private Brick.BrickField currentBrickField = null;
 	private boolean hasUnsavedChanges = false;
 
 	public FormulaEditorHistory(InternFormulaState internFormulaState) {
 		current = internFormulaState;
```
84n4n4: why not have the brickField also already passed here?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
```diff
@@ -103,6 +124,32 @@ public boolean hasUnsavedChanges() {
...
+
+	public Brick.BrickField getPeekBrickFieldStackUndo() {
+		return brickFieldStackUndo.peek();
+	}
+
+	public Brick.BrickField getPeekBrickFieldStackRedo() {
+		return brickFieldStackRedo.peek();
+	}
+
+	public Map<Brick.BrickField, InternFormulaState> getStartingValues() {
```
84n4n4: please rename this to "initial" instead of starting

84n4n4:
tests missing for undo bug (editing multiple brick fields).
are there tests for saving / dropping formula changes?
add tests for saving / dropping formula changes in multiple brick fields.
test if your undo is working as expected.

84n4n4:
please reopen once this is ready for review again

PR #2848, opened 2018-06-11
Update google-services-template.json

robertpainsi:
I'm currently investigating in some wrong Firebase user event logging stats (screen_view). This definitely will __not__ fix it, I just think it's time to update the google-services.json file during the process.

I did a testrun in the Firebase DebugView, works as expected.


robertpainsi:
:tophat:<br>:frog:<br>:necktie:

PR #2863, opened 2018-06-29
CAT-2663 Pen Action refactoring

shellp:
- buffer locking
- fix for quick movements of sprites

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/actions/ChangeXByNAction.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/actions/ChangeXByNAction.java
@@ -22,6 +22,7 @@
  */
 package org.catrobat.catroid.content.actions;
 
+import android.graphics.PointF;
 import android.util.Log;
 
 import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
@@ -40,6 +41,11 @@ public class ChangeXByNAction extends TemporalAction {
 		try {
 			Float newX = xMovement == null ? Float.valueOf(0f) : xMovement.interpretFloat(sprite);
 			sprite.look.changeXInUserInterfaceDimensionUnit(newX);
+
+			if (sprite.penConfiguration.penDown) {
+				sprite.penConfiguration.pointsToDraw.add(new PointF(sprite.look.getXInUserInterfaceDimensionUnit(),
+						sprite.look.getYInUserInterfaceDimensionUnit()));
+			}
 		} catch (InterpretationException interpretationException) {
 			Log.d(getClass().getSimpleName(), "Formula interpretation for this specific Brick failed.", interpretationException);
 		}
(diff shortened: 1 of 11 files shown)


catroid/src/main/java/org/catrobat/catroid/content/actions/ChangeXByNAction.java
```diff
@@ -40,6 +41,11 @@ protected void update(float percent) {
 		try {
 			Float newX = xMovement == null ? Float.valueOf(0f) : xMovement.interpretFloat(sprite);
 			sprite.look.changeXInUserInterfaceDimensionUnit(newX);
+
+			if (sprite.penConfiguration.penDown) {
+				sprite.penConfiguration.pointsToDraw.add(new PointF(sprite.look.getXInUserInterfaceDimensionUnit(),
+						sprite.look.getYInUserInterfaceDimensionUnit()));
+			}
```
robertpainsi: This approach won't work with the physics engine because position changes aren't triggered by actions.
Also actions are easily missed (e.g. IfOnEdgeBounceAction) and developers will forget to add it for new bricks/features.

catroid/src/main/java/org/catrobat/catroid/stage/PenActor.java
```diff
@@ -103,18 +111,20 @@ private void drawLinesForSprite(Sprite sprite) {
...
+			nextPoint = pen.pointsToDraw.poll();
 		}
 
 		renderer.end();
-		pen.previousPoint.x = x;
-		pen.previousPoint.y = y;
 	}
 
-	public void dispose() {
+	public synchronized void dispose() {
```
robertpainsi: I can't say for sure, but I think this is the only method causing problems with concurrency.
The `StageListener.finish` method (which calls `PenActor.dispose`) is also called from other Threads. Could the `StageListener.dispose` handle it? Would that solve all concurrency problems?

rschluesselbauer: I looked into it, it should also be done when reloading the project. Might also cause issues when "reset" is called.
This is what I came up with to fix it for the release: https://github.com/Catrobat/Catroid/pull/2872

robertpainsi: What I meant, is, instead of disposing the `PenActor` in `StageListener.finish`, dispose it in `StageListener.dispose`. No need for synrchonization because all ApplicationListener methods are called from the OpenGL Thread.

> All methods are called in a thread that has the OpenGL context current. You can thus safely create and manipulate graphics resources.

rschluesselbauer:
I am quite sure that dispose is not called when the project is restarted,
but I don't have the means to test it right now.

Robert Painsi <notifications@github.com> schrieb am Fr., 6. Juli 2018,
14:38:

robertpainsi:
I'll comment on the other PR to avoid any further misconception :wink:

84n4n4:
please rebase and have a look at comments above

PR #2864, opened 2018-07-01
[GSoC18] Refactor DataContainer and Fixed Copying Issues

thmq:


catroid/src/main/java/org/catrobat/catroid/content/Project.java
```diff
@@ -133,41 +131,14 @@ public Project(SupportProject supportProject, Context context) {
...
-			Iterator variablesIterator = dataContainer.spriteVariables.keySet().iterator();
-			while (variablesIterator.hasNext()) {
-				Sprite sprite = (Sprite) variablesIterator.next();
-				if (sprite == null) {
-					variablesIterator.remove();
-				}
-			}
-		}
-	}
-
```
robertpainsi: There are ~70 programs having `null` as key. This code has been added in a hotfix https://github.com/Catrobat/Catroid/commit/626e5fdb4ca60bf6f7835301defcbab475a64c53, maybe that's why the number is so low.

Should we keep it? Maybe use [Map.remove](https://developer.android.com/reference/java/util/Map.html#remove(java.lang.Object))?
This code is a good indicator that the DataContainer should prevent from storing null Sprites (future improvements :wink:)

thmq: sure, I can do that. btw: https://github.com/Catrobat/Catroid/pull/2864/files#diff-cb1b43b9b2d00fb9b796dba270e0093e prevents you from adding UserData to a null key.

robertpainsi: Awesome! :party:

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -92,14 +91,9 @@ public static String convertToEnglishDigits(String value) {
...
-		Map<Sprite, List<UserVariable>> spriteVariableMap = dataContainer.getSpriteVariableMap();
-		List<UserVariable> spriteVariableList = spriteVariableMap.get(sprite);
-		List<UserVariable> userBrickVariableList = dataContainer.getOrCreateVariableListForUserBrick(userBrick);
-
-		drawVariables(projectVariableList, batch);
-		drawVariables(spriteVariableList, batch);
-		drawVariables(userBrickVariableList, batch);
+		drawVariables(dataContainer.getProjectUserVariables(), batch);
+		drawVariables(dataContainer.getSpriteUserVariables(sprite), batch);
+		drawVariables(dataContainer.getUserBrickUserVariables(userBrick), batch);
```
robertpainsi: `java.lang.NullPointerException: This map does not allow null keys`

catroid/src/main/java/org/catrobat/catroid/content/Project.java
thmq: I have moved the null entry removal just below that [that] (I wasn't able to get the exact line - I'm sure I could have but it took me too long).

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
thmq: for now I went with option 2.

thmq:
please run espresso tests

thmq:
please run espresso tests

thmq:
please run espresso tests

robertpainsi:
@84n4n4, thanks for also having a look at this!

PR #2868, opened 2018-07-03
CAT-2674 BUG regarding undo/redo buttons and discarding the changes.

PlantinumX:
Discarding the changes in bricks results in  storing false values.
To solve both bugs  changed structure in FormulaEditorHistory and made getBrickorCustomView public to access it from formulaEditorEditText.
In addition created new Class State for storing the State. A state consists of
InternformulaState and Brickfield.
Bug with discarding the changes-> solved it  with pushing the according values on top of stack and consider it as a save.
Changed in PenBricksTest the testPenColorBrick testcase because it is not testing the right
behaviour. In addition wrote 2 TCs in PenBricksTest and 2 in PhiroMoveMotorBackwardBrickTest to test if the bugs are fixed.


84n4n4:
if you could just remove the for loops in the tests. (since if this ever breaks at say the 2nd or 3rd undo, this information is lost, and one again has to go there and either add log output or write individual calls instead of the loop to see when and what exactly is broken)

PR #2869, opened 2018-07-03
Clone commented bricks fix

84n4n4:


84n4n4:
ok, so the original issue:

if theres bricks commented out, a clone of the sprite will still execute these bricks.
this was because the "commentedOut" flag was never set for the clone.

anyways, this fixes this, not in a nice way, since it would be nice to have bricks update this themselves, but due to the interface, abstract, inheritance clusterfuck this is not feasible.

to be noted: the same thing is going on for scripts, so currently scripts that are commented out in the original sprite are going to be NOT commented out in the cloned version, but this currently doesnt have any effect since commenting out a script will comment out all contained bricks, which in turn is now updated in the clone...
i could also include this, but again, theres no nice way of doing this besides adding this comment flag setting magic in every single implementation of script.
yay...

test for this will be added tomorrow...

thmq:
we could also add this to the clone method of the brick instead of specifically calling it in the list.

rschluesselbauer:
@wslany It's because the SayBubbleBrick/ThinkBubbleBrick is a FormulaBrick & the clone method of the FormulaBrick calls the super.clone method -> members of super classes are shallow copied.

In many bricks we do not call the super.clone method, so as @thmq suggested, it is probably cleaner to call super.clone in all bricks (and then use the result of that). It's also more effort.

PR #2872, opened 2018-07-05
CAT-2738 Fix PenActor concurrency issues for release

rschluesselbauer:
Synchronize on StageListener instance during rendering/execution and for stopping/reloading a project.

This way the PenActor can only be removed before/after rendering/execution.
The "remove" libgdx lifecycle method removes it from the queue for rendering/execution.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/PenActor.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/PenActor.java
@@ -114,14 +114,17 @@ public class PenActor extends Actor {
 		pen.previousPoint.y = y;
 	}
 
-	public void dispose() {
-		if (buffer != null) {
-			buffer.dispose();
-			buffer = null;
+	@Override
+	public boolean remove() {
+		boolean hadParent = super.remove();
+		if (hadParent) {
+			if (buffer != null) {
+				buffer.dispose();
+			}
+			if (bufferBatch != null) {
+				bufferBatch.dispose();
+			}
 		}
-		if (bufferBatch != null) {
-			bufferBatch.dispose();
-			bufferBatch = null;
-		}
+		return hadParent;
 	}
 }
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -61,7 +61,6 @@ import org.catrobat.catroid.common.LookData;
 import org.catrobat.catroid.common.ScreenModes;
 import org.catrobat.catroid.common.ScreenValues;
 import org.catrobat.catroid.content.EventWrapper;
-import org.catrobat.catroid.content.Look;
 import org.catrobat.catroid.content.Project;
 import org.catrobat.catroid.content.Scene;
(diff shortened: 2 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/PenActor.java
```diff
@@ -114,14 +114,17 @@ private void drawLinesForSprite(Sprite sprite) {
...
-		if (buffer != null) {
-			buffer.dispose();
-			buffer = null;
-		}
-		if (bufferBatch != null) {
-			bufferBatch.dispose();
-			bufferBatch = null;
+	@Override
+	public boolean remove() {
+		boolean hadParent = super.remove();
```
rschluesselbauer: Here the actor is removed from render/execution queue of the stage.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -488,31 +476,32 @@ public void render() {
...
-			}
-			long executionTimeOfActionsUpdate = SystemClock.uptimeMillis() - timeBeforeActionsUpdate;
-			if (executionTimeOfActionsUpdate <= ACTIONS_COMPUTATION_TIME_MAXIMUM) {
-				deltaActionTimeDivisor += 1f;
-				deltaActionTimeDivisor = Math.min(DELTA_ACTIONS_DIVIDER_MAXIMUM, deltaActionTimeDivisor);
-			} else {
-				deltaActionTimeDivisor -= 1f;
-				deltaActionTimeDivisor = Math.max(1f, deltaActionTimeDivisor);
+		synchronized (this) {
+			if (!paused) {
```
rschluesselbauer: snychronize over rendering and execution

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -753,10 +745,6 @@ public Stage getStage() {
 		return stage;
 	}
 
-	public void removeActor(Look look) {
-		look.remove();
```
rschluesselbauer: unused method

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -488,31 +458,31 @@ public void render() {
...
-		if (!paused) {
-			float deltaTime = Gdx.graphics.getDeltaTime();
-
-			float optimizedDeltaTime = deltaTime / deltaActionTimeDivisor;
-			long timeBeforeActionsUpdate = SystemClock.uptimeMillis();
-			while (deltaTime > 0f) {
-				physicsWorld.step(optimizedDeltaTime);
-				stage.act(optimizedDeltaTime);
-				deltaTime -= optimizedDeltaTime;
+		synchronized (this) {
```
robertpainsi: From my understanding, concurrency problems with the PenActor occur because `PenActor.dispose` is called in `StageListener.finish` (which isn't called from the OpenGL thread).
Dispose the `PenActor` in the `StageListener.dispose` method and you don't need to synchronize the `PenActor`.

PR #2881, opened 2018-07-13
[GSoC18] Moved view inflation of Bricks to BrickBaseType.

thmq:
- all ~150 bricks are now inflated from a single point.
- brick now only need to provide LayoutRes through one method.
- prototype view and ScriptFragmentView are now inflated by the
base type, overrides still initialse the TextViews.
- bricks are still inflated newly on every getView(..) call,
anything else wouldn't have worked anyway (at least we know that now).
- many Bricks have become very minimal now.
- many things could still be improved -> e.g. alpha handling.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/BrickBaseType.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/BrickBaseType.java
@@ -23,8 +23,10 @@
 package org.catrobat.catroid.content.bricks;
 
 import android.content.Context;
+import android.support.annotation.CallSuper;
+import android.support.annotation.LayoutRes;
+import android.view.LayoutInflater;
 import android.view.View;
-import android.widget.BaseAdapter;
 import android.widget.CheckBox;
 
 import org.catrobat.catroid.R;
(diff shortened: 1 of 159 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/BrickBaseType.java
```diff
@@ -74,41 +98,26 @@ public void setCheckboxView() {
...
 		checkbox = view.findViewById(R.id.brick_checkbox);
 		checkbox.setChecked(isChecked);
 		checkbox.setVisibility(checkboxVisibility);
 		checkbox.setEnabled(enabled);
-
-		final Brick instance = this;
 		checkbox.setOnClickListener(new View.OnClickListener() {
 			@Override
 			public void onClick(View view) {
 				adapter.handleCheck(instance, ((CheckBox) view).isChecked());
```
robertpainsi: You can use `BrickBaseType.this` instead of creating a local variable.

thmq: thanks for the hint -> I just copied it from the old version. For now I'm not really changing too much, so there could definitely be more stuff like this :wink: .

thmq:
test this please

PR #2882, opened 2018-07-15
Only use StringRes for ToastUtils

robertpainsi:


robertpainsi:
:tophat:<br>:koala: (thank you!)

PR #2883, opened 2018-07-15
CAT-FIX Fix crash when not connected to a network

robertpainsi:
When the device is not connected to a network,
ConnectivityManager.getActiveNetworkInfo will return null which will
result in a NullPointerException. Added null checks to prevent this
crash.

There is a small behavior change when an error occurs in the WebView.
Previously, the additional info '…Check your Internet connection.' was
only shown if the device is not connected to a network. Now the info is
also shown if the device is connecting to a network. IMO and in this
case, the difference is almost non-existent. Also, if the device is
connecting to a network, it doesn't mean it will get a connection. So
showing the additional info in both cases is fine.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/SignInActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/SignInActivity.java
@@ -23,9 +23,7 @@
 
 package org.catrobat.catroid.ui;
 
-import android.content.Context;
 import android.content.Intent;
-import android.net.ConnectivityManager;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.text.Html;
@@ -46,6 +44,7 @@ import org.catrobat.catroid.ui.recyclerview.dialog.login.LoginDialogFragment;
 import org.catrobat.catroid.ui.recyclerview.dialog.login.RegistrationDialogFragment;
 import org.catrobat.catroid.ui.recyclerview.dialog.login.SignInCompleteListener;
 import org.catrobat.catroid.utils.ToastUtil;
+import org.catrobat.catroid.utils.Utils;
 
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -213,12 +211,8 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
 
 		@Override
 		public void onReceivedError(WebView view, int errorCode, String description, String failingUrl) {
-			ConnectivityManager cm = (ConnectivityManager) getBaseContext().getSystemService(Context.CONNECTIVITY_SERVICE);
-			NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
-			boolean isConnected = activeNetwork != null && activeNetwork.isConnectedOrConnecting();
```
robertpainsi: This is a small behavior change. Now it only depends on `isConnected` (not `isConnectedOrConnecting`), if the additional info `Check your Internet connection` is shown. IMO in both cases (not connected or connecting), it's fine to show this additional info, because in both cases you don't have a connection (to the internet).

Also [`isConnectedOrConnecting`](https://developer.android.com/reference/android/net/NetworkInfo.html#isConnectedOrConnecting())
> This method was deprecated in API level 28.

(but for other reasons)

catroid/src/main/java/org/catrobat/catroid/ui/SignInActivity.java
```diff
@@ -95,11 +94,7 @@ public void onStop() {
 	}
 
 	public void onButtonClick(final View view) {
-		ConnectivityManager connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
-		boolean isNetworkAvailable = connectivityManager != null
-				&& connectivityManager.getActiveNetworkInfo().isConnected();
-
-		if (!isNetworkAvailable) {
+		if (!Utils.isNetworkAvailable(this)) {
```
thmq: Actually I wanted to remove all these static utility classes, that get their context handed through. The reason why it still exists in `Utils` is because I have not had time to remove it everywhere. But if you think it is better this way, we can definitely find another solution.

robertpainsi:
Release branch it is then :D Thank you!

PR #2895, opened 2018-07-25
[GSoC18] Refactored Clones of Bricks/Scripts

thmq:
- added Tests to verify clone behaviour

 - Bricks now correctly override super.clone() if necessary which means:
   + commentOut state is correctly adapted on cloning
   + x.getClass() == x.clone().getClass is true.
   + all clone methods work consistantly (no more deep/shallow mixture)

 - ScriptBricks are now always linked to their scripts
   + no more duplicate creation of ScriptBricks or Scripts in the
     getBrick() / getScript() methods.

 - removed side magic where scripts where initialized through
   getScriptSafe(), which was not even implemented consistantly.


84n4n4:
please run espresso tests

PR #2897, opened 2018-07-25
[Under discussion] Hotfix for CollisionScript references not set correctly

robertpainsi:
How to reproduce the error

* Download https://share.catrob.at/pocketcode/program/28126
* Open and immediately press run (don't go in any script view)
* Press the ball. You can move the Players by touch down and move them.
--
* Instead of normally bouncing of, the ball moves around and behind the player.

The problem is, [`CollisionScript`](https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/content/CollisionScript.java) has two members
```java
private String spriteToCollideWithName;
private transient Sprite spriteToCollideWith;
```
where only the name is serialized. When deserializing, only the name is deserialized and the Sprite itself will remain null which is equal to 'collide with anything'. When opening the Sprites Script View (maybe scroll to the collision bricks), the program will work as expected though.

The problem will be fixed in a hotfix release. I haven't fully tested it nor added tests (yet, hopefully this testrun will be green :crossed_fingers:).
Feedback is highly appreciated, a better fix even more. Just keep in mind that it will be a hotfix release.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/CollisionScript.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/CollisionScript.java
@@ -60,19 +60,22 @@ public class CollisionScript extends Script {
 
 	public void setSpriteToCollideWithName(String spriteToCollideWithName) {
 		this.spriteToCollideWithName = spriteToCollideWithName;
-		updateSpriteToCollideWith();
+		updateSpriteToCollideWith(ProjectManager.getInstance().getCurrentlyEditedScene());
 	}
 
 	public Sprite getSpriteToCollideWith() {
-		updateSpriteToCollideWith();
+		updateSpriteToCollideWith(ProjectManager.getInstance().getCurrentlyEditedScene());
 		return spriteToCollideWith;
 	}
 
-	private void updateSpriteToCollideWith() {
-		if (spriteToCollideWithName != null
-				&& (spriteToCollideWith == null || !spriteToCollideWithName.equals(spriteToCollideWith.getName()))) {
-			Scene currentScene = ProjectManager.getInstance().getCurrentlyEditedScene();
-			spriteToCollideWith = currentScene.getSprite(spriteToCollideWithName);
+	public void updateSpriteToCollideWith(Scene scene) {
+		if (spriteToCollideWithName == null) {
+			spriteToCollideWith = null;
+		} else {
+			spriteToCollideWith = scene.getSprite(spriteToCollideWithName);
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/content/CollisionScript.java
```diff
@@ -60,19 +60,22 @@ public String getSpriteToCollideWithName() {
...
-			Scene currentScene = ProjectManager.getInstance().getCurrentlyEditedScene();
-			spriteToCollideWith = currentScene.getSprite(spriteToCollideWithName);
+	public void updateSpriteToCollideWith(Scene scene) {
+		if (spriteToCollideWithName == null) {
+			spriteToCollideWith = null;
+		} else {
+			spriteToCollideWith = scene.getSprite(spriteToCollideWithName);
+			if (spriteToCollideWith == null) {
+				spriteToCollideWithName = null;
+			}
```
robertpainsi: Will be called when the project has been loaded or a sprite has been renamed/deleted.
If a sprite has been renamed/deleted `scene.getSprite(name)` will return null and the name will be nulled as well.

thmq:
what if we restored the sprite from its name on de-serialisation? i.e. in the `readResolve()` method? I'll look into it.

robertpainsi:
Superseded by https://github.com/Catrobat/Catroid/pull/2898

robertpainsi:
Just reopen so the pr file changes will be synced with the branch

catroid/src/main/java/org/catrobat/catroid/content/CollisionScript.java
robertpainsi: Fixes following bug:
- Create a project containing two physics sprites.
- Go into the Scripts view of one of them. It should have a script like
```
When physical collision with
Other sprites name
```
- Change the collision sprite to `anything`
- Press back
- And go back into the Scripts view.

*What do you see?*
```
When physical collision with
Other sprites name
```

*What is the expected output?*
```
When physical collision with
anything
```

This is fixed now.
IMO there won't be any performance issues with my approach because it's called only for rendered bricks (and during loading) and the sprite list is (usually) very short. If really wanted, I sure can provide a more efficient way :wink:.
___
The inner if block is for deleted sprites. If a sprite is deleted, the `When physical collision with` changes to `anything`*

\* It changes to `anything` only if the project is loaded or the brick is rendered. So there are still problems, but this is the same behavior as in the last release.

PR #2898, opened 2018-07-26
[HOTFIX] Fixed deserialization of CollisionScript

thmq:
- removed transient Sprite field that was null
   after serialization.
 - Sprite is now updated every time instead of caching
   spriteToCollideWith in a private field.


rschluesselbauer:
Hi, I remember adding this because renaming a look was way easier that way.
What will happen now if you rename a referenced look?

thmq:
It will break, probably. :thinking:

PR #2901, opened 2018-07-27
JENKINS-223 Adds code coverage reports for the the set of PR tests.

mfuchs:
JaCoCo is used for code coverage.
[Here](https://jenkins.catrob.at/job/lab/job/Matthias_Fuchs/job/CatroidJacoco/2/cobertura/) is an example of what the results would look like when this PR is merged.


mfuchs:
I did some changes. The coverage is also generated when tests do not work.
Please review.

Jenkinsfile
```diff
@@ -91,16 +96,26 @@ pipeline {
```
redeamer: Is it intended, that the coverage is not generated for the unit-tests? Or will this be an additional story?

mfuchs: It is not implemented yet. I actually tried it but the unit tests had no coverage at all.
Nonetheless I will add it to the pr today.

PR #2906, opened 2018-07-31
[GSoC18] Refactored inflation for FormulaBricks

thmq:


catroid/src/main/java/org/catrobat/catroid/content/bricks/SpeakAndWaitBrick.java
```diff
@@ -23,48 +23,25 @@
...
-import java.io.File;
 import java.util.List;
 
 public class SpeakAndWaitBrick extends FormulaBrick {
 
 	private static final long serialVersionUID = 1L;
-	private transient View prototypeView;
-
-	File speechFile;
-	private float duration;
```
robertpainsi: Don't forget to omit these fields in the XstreamSerializer.

catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
```diff
@@ -101,48 +147,35 @@ public void onClick(View view) {
...
-				Utils.convertDoubleToPluralInteger(BrickValues.DRONE_MOVE_BRICK_DEFAULT_TIME_MILLISECONDS / 1000)));
-	}
-
-	protected <T> String formatNumberForPrototypeView(T value) {
-
-		String number = String.valueOf(value);
-
-		return FormatNumberUtil.cutTrailingZeros(number);
+		textView.setText(context.getResources()
+				.getQuantityString(R.plurals.second_plural, Utils.TRANSLATION_PLURAL_OTHER_INTEGER));
```
robertpainsi: Maybe if-else instead?

catroid/src/main/java/org/catrobat/catroid/content/bricks/VibrationBrick.java
```diff
@@ -71,54 +57,10 @@ public int getViewResource() {
 		return R.layout.brick_vibration;
 	}
 
-	@Override
-	public View getView(Context context) {
-		super.getView(context);
-		TextView editVibrate = (TextView) view.findViewById(R.id.brick_vibration_edit_text);
-		TextView secondTextVibrate = (TextView) view.findViewById(R.id.brick_vibration_second_label);
```
robertpainsi: Please also use your new `setSecondsLabel` here.

catroid/src/main/java/org/catrobat/catroid/content/bricks/DroneMoveBackwardBrick.java
```diff
@@ -40,81 +37,45 @@
 	private static final long serialVersionUID = 1L;
 
 	public DroneMoveBackwardBrick(int durationInMilliseconds, int powerInPercent) {
-		initializeBrickFields(new Formula(durationInMilliseconds / 1000.0), new Formula(powerInPercent));
+		this(new Formula(durationInMilliseconds / 1000.0), new Formula(powerInPercent));
 	}
```
robertpainsi: All bricks need to call `addAllowedBrickField` to set up the mapping correctly during deserialising now. So all Formula Bricks need to have an empty constructor which calls this(…).

I guess adding an empty constructor is the easiest solution for now. (readResolve won't do the trick because it sets an empty formula into the formulaMap)

catroid/src/main/java/org/catrobat/catroid/content/bricks/IfLogicEndBrick.java
```diff
@@ -37,17 +37,9 @@
...
-	}
-
-	public IfLogicElseBrick getIfElseBrick() {
-		return ifElseBrick;
-	}
-
-	public void setIfElseBrick(IfLogicElseBrick ifElseBrick) {
+	public IfLogicEndBrick(IfLogicBeginBrick ifLogicBeginBrick, IfLogicElseBrick ifElseBrick) {
 		this.ifElseBrick = ifElseBrick;
+		this.ifBeginBrick = ifLogicBeginBrick;
```
robertpainsi: Please also swap these two lines :laughing:

catroid/src/main/java/org/catrobat/catroid/content/bricks/SpeakAndWaitBrick.java
thmq: sure :wink I will also increase the language version.

catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
thmq: well, maybe.that depends on your beliefs I guess, but sure, I can change it.

catroid/src/main/java/org/catrobat/catroid/content/bricks/VibrationBrick.java
thmq: oh, well, interestingly this one did not come up on the `BrickValueParameterTest` :thinking:. someone should have a look at this test again anyway -> @84n4n4 maybe?

catroid/src/main/java/org/catrobat/catroid/content/bricks/DroneMoveBackwardBrick.java
thmq: oh, yes. that's an interesting one -> have been thinking about this yesterday. and I agree, for now this seems to be the best solution.

catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
robertpainsi: No, wait!

I guess the return is wrong here https://github.com/thmq/Catroid/blob/86ebc59f657cd91334a8476a18a44af3633ce2bd/catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java#L176

robertpainsi: Or at least that's what the log message says.

thmq: oh, yes! it should fallback to "other" -> :scream: I messed up something there.

robertpainsi: Kudos to your improved log message! :star:

thmq: thank you :kissing_closed_eyes:

thmq: so, found the issue: `return` should have been in the `try` block -> so that if an exception is thrown or the formula isn't a single number it falls back to "other".

however, I think this should be changed anyway as we cannot assume that just because it is correct to say `someVariableValue seconds` opposed to `someVariableValue second` in English it is also correct to say that in other languages.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SpeakAndWaitBrick.java
thmq: done.

PR #2912, opened 2018-08-07
[GSoC18] Remove MIPS

ThomasSchwengler:
* Exclude MIPS architecture binary files from builds
  - MIPS support has been removed from NDK >= 17, builds will fail if you have NDK installed as it tries to strip the MIPS binaries
---
* ~~Remove some unused files (eclipse configuration files, outdated gradle files in a wrong folder)~~
  - Moved to #2913
* ~~Migrate the build.gradle file to gradle >= 3.0.0 according to this guideline: https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration~~
  - Moved to #2914


ThomasSchwengler:
@redeamer That would be possible, I can build catroid with ndk 17 with this patch just fine now.

PR #2919, opened 2018-08-08
[GSoC18] Refactored Brick Spinners

thmq:
- introduced Nameable interface for all things
   getName(), setName().
 - added adaptable BrickSpinner that works with
   object references (Nameables), String options,
   and a "New..." option.
 - added setSelection(..) workflow to fix memory leaks in Bricks:
   whenever a selection that was buffered as object in the brick
   (e.g. a SoundInfo) is deleted from the parent Sprite's
   soundList, the Brick still had the reference. these nonsense
   references are now removed whenever the Spinner is set up.
 - added missing constructors form Brick Field PR.


thmq:
this probably replaces #2911.

PR #2921, opened 2018-08-09
CAT-2673 none sense error message when register

LPeteR90:
Comments from original PR

- removing unused Strings
- Sets an error message that will be displayed below the TextInputLayouts
- (Username/E-mail/Password/confirm Password ) instead of showing AlertDialogs
- Check E-mail and Username availability in background
- Username should start with Latin character or number and contains no "@" - to block using email as username
- Check if the the provided Email_address matches Patterns.EMAIL_ADDRESS
- Password should be at least 6 characters
- Check if Password and Confirm Password are matched


LPeteR90:
test this please

PR #2926, opened 2018-08-10
CATROID-37 crashes when restoring fragments after activity was destroyed

84n4n4:
Can be reproduced by setting Don't keep activities in developer options.
Problem is that some fragments have listeners or members that link back to
its activity or other views. This isnt recoverable when activity is killed
due to low memory or long inactivity. When app is left in this state it
will crash badly on reopening.
To fix this for now this PR prevents those fragments to be restored when
activity is recovering from such a killed state. (eg. formula editor open
- activity killed by OS - reopen activity is back in script view)

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaySoundBrick.java
@@ -32,6 +32,7 @@ import android.widget.Spinner;
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.R;
 import org.catrobat.catroid.common.SoundInfo;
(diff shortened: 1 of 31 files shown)


84n4n4:
DO NOT MERGE.

this is meant to be applied once hotfix-0.9.36 is merged back into master and develop. however this is already based off of hotfix-0.9.36 to prevent merge issues.

this is a patch to ease the pain for now, but
- most dialog fragments have to be refactored, well have to think about how to properly approach this, and if this is actually possible with our current user workflow. (spinner selections are a major issue here)
- formula editor is fubar, i dont think refactoring is a possibility there so id propose to keep this patch in place, and hopefully someday someone rewrites the whole formula editor. (since it knows and needs so much, maybe it would be a wise idea to make it its own activity)

PR #2930, opened 2018-08-13
JENKINS-231 Do not upload APKs from develop to files.catrob.at.

mfuchs:
The generated APKs are accessible from the Jenkins jobs, including PR jobs.


Jenkinsfile
```diff
@@ -146,19 +142,6 @@ pipeline {
...
 			}
 		}
-
-		stage('Upload to share') {
-			when {
-				branch "${env.CATROBAT_SHARE_UPLOAD_BRANCH}"
-			}
-
-			steps {
-				unstash "apk-independent"
```
redeamer: The 'stash' in line 145 is unneeded now and should be removed as well.

PR #2933, opened 2018-08-13
[GSoC18] Cleanup Brick Constructors/If-Else Hell

thmq:
- added interface for IfElseLogicBeginBrick, so
   that we can reuse the Else and End Bricks without the
   stupid inheritance. -> this is surely still not ideal,
   but fine for now.

 - added missing constructors, and empty constructors now
   add a formula with value "0", so you can easily see
   "uninitialized" Bricks.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoNxtMotorStopBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoNxtMotorStopBrick.java
@@ -25,7 +25,6 @@ package org.catrobat.catroid.content.bricks;
 import android.content.Context;
 import android.view.View;
 import android.widget.AdapterView;
-import android.widget.AdapterView.OnItemSelectedListener;
 import android.widget.ArrayAdapter;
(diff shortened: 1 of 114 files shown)


catroid/src/main/java/org/catrobat/catroid/content/bricks/LegoNxtMotorStopBrick.java
```diff
@@ -84,33 +65,36 @@ public int getViewResource() {
...
-		motorSpinner.setSelection(motorEnum.ordinal());
-		return view;
-	}
 
-	@Override
-	public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
-		motorEnum = Motor.values()[position];
-		motor = motorEnum.name();
+		ArrayAdapter<CharSequence> spinnerAdapter = ArrayAdapter
+				.createFromResource(context, R.array.ev3_stop_motor_chooser, android.R.layout.simple_spinner_item);
```
robertpainsi: `R.array.nxt_stop_motor_chooser`

PR #2935, opened 2018-08-14
[GSoC18] Clean up build.gradle files

ThomasSchwengler:
* Move `google-services` classpath to project
* Remove outdated, unused `gradle:2.3.3` from project buildscript dependencies
* Merge `android` section
* Move commonly used versions to `ext.*Version` variables
* Use specific version for `androidasync` dependency
* Group dependencies
* Explicitely supply versions for some `com.android.support.*` libraries to avoid different versions being used. (See `GradleCompatible` lint rule)
* Remove some lint ignores
  * `GradleDynamicVersion`: It should be necessary to supply a specific library version so library upgrades don't happen without anyones knowledge
  * `DefaultLocale`: This rule doesn't get violated anyways

Depends on #2934

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,17 +31,17 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
-        classpath 'com.google.gms:google-services:3.0.0'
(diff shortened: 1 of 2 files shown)


ThomasSchwengler:
Rebased

ThomasSchwengler:
Rebased

PR #2939, opened 2018-08-14
Hotfix 0.9.40

84n4n4:


wslany:
Checked on Galaxy Nexus with Android 4.3, this bug was fixed on it.

PR #2945, opened 2018-08-16
CATROID-43 HOTFIX Api level 19 crashes on brick spinners and formula …

84n4n4:
…fields


84n4n4:
successful test on API19: 
https://jenkins.catrob.at/job/Catroid-ManualEmulatorTest/4/
(this is ONLY 7 tests, if something else is broken on API19, we still dont know about it)

PR #2953, opened 2018-08-29
CATROID-68 API level increased to 19, Apk size reduction

84n4n4:


84n4n4:
for next release target api is supposed to be increased to 19.
in that case the armeabiv6,v5 libs could be also removed, saving us another 3Mb of apks size.

gitarrenfudler:
That's a cool PR! Impressing...

PR #2955, opened 2018-09-02
CATROID-54 Wrong title in action bar after exiting the formula editor

ndrnour:


84n4n4:
can you please add a test for this.

ndrnour:
yes, but the test will contain Fragments-switching. is it okay?
ScriptFragment--->> FormulaEditor then back again to ScriptFragment.
or have you another way to test?

PR #2958, opened 2018-09-04

shellp:
- larger font-size for axes
- simple test to check axes font size and color


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -575,20 +579,47 @@ private void printPhysicsLabelOnScreen() {
 		batch.end();
 	}
 
+	public BitmapFont getAxesFont() {
+		return font;
+	}
+
+	private GlyphLayout getScaledFont() {
```
84n4n4: unit test this.

84n4n4:
please reopen when done

PR #2959, opened 2018-09-04
CATROID-51 Missing 'Tap "+" to add objects' when there are no objects

ndrnour:


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/SpriteListFragmentExplanationTextProjectWithObjectTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+import static org.catrobat.catroid.uiespresso.ui.fragment.rvutils.RecyclerViewInteractionWrapper.onRecyclerView;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.not;
+
+@RunWith(AndroidJUnit4.class)
+public class SpriteListFragmentExplanationTextProjectWithObjectTest {
+	@Rule
+	public BaseActivityInstrumentationRule<ProjectActivity> baseActivityTestRule = new
```
LPeteR90: Please initialize the rule with the Fragment you want to test

LPeteR90: e.g. CopySoundTest 
`@Rule 
public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
			BaseActivityInstrumentationRule<>(SpriteActivity.class, SpriteActivity.EXTRA_FRAGMENT_POSITION,
			SpriteActivity.FRAGMENT_SOUNDS);`

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/SpriteListFragmentExplanationTextNoObjectsProjectTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+
+@RunWith(AndroidJUnit4.class)
+public class SpriteListFragmentExplanationTextNoObjectsProjectTest {
+	@Rule
+	public BaseActivityInstrumentationRule<ProjectActivity> baseActivityTestRule = new
```
LPeteR90: See the comment on the test Rule for the other test

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/SpriteListFragmentExplanationTextProjectWithObjectTest.java
ndrnour: screenshot_20180906-230307
I don't understand why? I'm testing this activity "see the attached screenshot" 
which is "ProjectActivity" or am I wrong?

84n4n4: No, the screenshot shows the spriteListFragment inside the projectActivity. Theres are more fragments.

Just starting the activity and relying on it to show the right fragment is something i want to always prevent in the espresso test suit.

Let me explain...
The fragment stack is not cleared when a test ends, that leads to the fact that the next test starting the same activity will continue in the same fragment as the test before that left it.
Interestingly, when a test crashes, that most of the times clears the fragments again.

So for example:
We have a lot of brick tests testing the script fragment in sprite activity. Its its default fragment, so if you dont specify any it will start there.
Someone writes a test that tests the sound fragment, also inside the sprite activity. Everything works fine, unless something changes the order in which tests are run, and now the sound fragment test runs inbetween. It will cause the following scriptlist brick to fail, and only this one, (and there can be other working tests inbetween as long as they test other activities) because it ends up in sound fragment. It does this not by giving any useful error message to make things worse, but a "could not click on location 537x366" or something similar useless espresso madness.
Now add to this that per default we dont really log in which order tests are running on jenkins, especially not working tests, while here the sound fragment test is green and working, but breaking another one.

This took me days and weeks to figure out, and ive got a few more grey hairs just because of this espresso "feature".
So i want to prevent this from ever happening again, by just always explicitely stating the fragment under test, because then the issue does not happen. (Exceptions are of course fragmentless activities like stage)

LPeteR90: Yes, that's right, you're testing the ProjectActivity, so the code sample with the SpriteActivity was probably misleading. However, the point is, that you can also pass the Fragment you're testing explicitly as a param to the Rule, so the line should be 

```public BaseActivityInstrumentationRule<ProjectActivity> baseActivityTestRule = new BaseActivityInstrumentationRule<>(ProjectActivity.class, ProjectActivity.EXTRA_FRAGMENT_POSITION, ProjectActivity.FRAGMENT_SPRITES);```

ndrnour: @LPeteR90  @84n4n4 
 **I got it. Thanks for helping me ^_***

PR #2960, opened 2018-09-04
CATROID-48 Crash when switching scene

LPeteR90:
- fix bug that crashes catroid when cloning sprite after scene transition
- add espresso test for that specific behavior

https://jira.catrob.at/browse/CATROID-48

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -245,15 +245,17 @@ public class StageListener implements ApplicationListener {
 	}
 
 	public void cloneSpriteAndAddToStage(Sprite cloneMe) {
-		Sprite copy = cloneMe.cloneForCloneBrick();
-		copy.look.createBrightnessContrastHueShader();
-		stage.getRoot().addActorBefore(cloneMe.look, copy.look);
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -245,15 +245,17 @@ public void create() {
...
-		Sprite copy = cloneMe.cloneForCloneBrick();
-		copy.look.createBrightnessContrastHueShader();
-		stage.getRoot().addActorBefore(cloneMe.look, copy.look);
-		sprites.add(copy);
-		if (!copy.getLookList().isEmpty()) {
-			copy.look.setLookData(copy.getLookList().get(0));
-		}
-		copy.initializeEventThreads(EventId.START_AS_CLONE);
-		copy.initConditionScriptTriggers();
+		if (scene.getSprite(cloneMe.getName()) != null) {
```
robertpainsi: Cloning a clone won't work anymore because clones aren't stored in the scene itself. Also if the transitioned scene (Scene 2) has an object with the same name, the app will still crash.

(I'll add another comment to the Jira ticket asap)

wslany:
Please see comments in jira ticket.

PR #2962, opened 2018-09-06
DONOTMERGE PAINT-326: Import pocket paint library

ThomasSchwengler:
# DONOTMERGE this is still in active development

* Add paintroid as dependency
* Refactor intents to use the internal version of pocket paint and avoid searching for it


ThomasSchwengler:
:sob:

PR #2963, opened 2018-09-06
JENKINS-236 Activate JaCoCo for tests only if -PenableCoverage is spe…

mfuchs:
…cified in the gradle call.

Unfortunately running JaCoCo leads to a worse experience debugging the code.
The debugger does not show the value of variables anymore.
This is a known issue, see also https://issuetracker.google.com/issues/37019591

The property -PenableCoverage is set by Jenkins builds automatically.
No debugger needed there. ;)
Manually running tests in Android Studio with code coverage still works
and is not affected by this change.

:exclamation: Before merging check that the code coverage is roughly the same as on develop. This would indicate that code coverage collecting still works with these changes.


mfuchs:
I just checked the build results.
The code coverage is correct.
Thus these changes work.

PR #2966, opened 2018-09-11
CATROID-36 Fix download dialog already existing project

gPathpp:


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/ReplaceExistingProjectDialogFragment.java
```diff
@@ -78,10 +78,12 @@ public void onCheckedChanged(RadioGroup group, @IdRes int checkedId) {
 				}
 			}
 		});
+		radioGroup.check(R.id.rename);
 
 		final AlertDialog alertDialog = new AlertDialog.Builder(getActivity())
-				.setTitle(R.string.overwrite_text)
 				.setView(view)
+				.setTitle(R.string.overwrite_title)
+				.setMessage(R.string.overwrite_text)
```
thmq: This `AlerDialog.Builder()` already uses a `.setView(..)` so `.setMessage()`  kind of collides with the custom view (a lot of fancy side magic starts to happen if you combine those two -> and especially through different API versions and on different device models this could be a problem).

If you want to add some textfield with a message, please just add a new `<TextField>...</TextField>` in the layout (`R.layout.dialog_overwrite_project`) and set the message there.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/ReplaceExistingProjectDialogFragment.java
```diff
@@ -78,10 +78,12 @@ public void onCheckedChanged(RadioGroup group, @IdRes int checkedId) {
 				}
 			}
 		});
+		radioGroup.check(R.id.rename);
```
thmq: The default selection for the `RadioGroup` can be changed via the layout resource [here](https://github.com/Catrobat/Catroid/blob/f1f8cfe6c80149bdae9f2244fad8cc5a0ccdd7cb/catroid/src/main/res/layout/dialog_overwrite_project.xml#L39). You can also rearrange the sequence (if you want rename before replace) and if you want to change the default visibility of the `inputLayout` there as well.

PR #2972, opened 2018-09-20
CATROID-59 + CATROID-56 move to internal storage

ndrnour:


catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.catrobat.catroid.common.SharedPreferenceKeys.SHOW_COPY_PROJECTS_FROM_EXTERNAL_STORAGE_DIALOG;
+import static org.catrobat.catroid.io.StorageOperations.transferData;
+
+public class ImportProjectsFromExternalStorage implements DialogInterface.OnClickListener {
+	public static final String TAG = ImportProjectsFromExternalStorage.class.getSimpleName();
+
+	private static final File POCKET_CODE_ROOT_DIRECTORY = new File(
```
84n4n4: please use DEFAULT_ROOT_DIRECTORY from flavored constants here instead of rebuilding it

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+			} else {
+				transferData(srcFile, new File(dstDir, srcFileName));
+			}
+		}
+
+		return dstDir;
+	}
+
+	public void importOldPocketCodePrograms() {
+		if (FlavoredConstants.DEFAULT_ROOT_DIRECTORY.equals(POCKET_CODE_ROOT_DIRECTORY) && pocketCodeExternalStorageDirectoryExists()) {
```
84n4n4: why is this necessary?

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+
+import static org.catrobat.catroid.common.SharedPreferenceKeys.SHOW_COPY_PROJECTS_FROM_EXTERNAL_STORAGE_DIALOG;
+import static org.catrobat.catroid.io.StorageOperations.transferData;
+
+public class ImportProjectsFromExternalStorage implements DialogInterface.OnClickListener {
+	public static final String TAG = ImportProjectsFromExternalStorage.class.getSimpleName();
+
+	private static final File POCKET_CODE_ROOT_DIRECTORY = new File(
+			CatroidApplication.getAppContext().getFilesDir().getAbsolutePath() + "/Programs");
+	private static final File EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
```
84n4n4: this should be also taken from flavored constants, or it will not work for any other flavor than pocket code

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+		if (!dstDir.isDirectory()) {
+			throw new IOException("Cannot create directory: " + dstDir.getAbsolutePath());
+		}
+
+		for (File srcFile : srcDir.listFiles()) {
+			String srcFileName = srcFile.getName();
+			if (srcFile.isDirectory()) {
+				boolean fileExists = false;
+				for (File dstFile : dstDir.listFiles()) {
+					if (dstFile.getName().equals(srcFileName) && dstFile.getParent().endsWith("Programs") && srcFile.getParent().endsWith("Pocket Code") && !srcFileName.equals("backpack")) {
```
84n4n4: this again contains hardcoded "pocket code" folder name. this will not work for any other flavor

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
84n4n4: also something i must have overlooked on previous commit, we currently do not have a Programs folder, so why introduce one?

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	public void importOldPocketCodePrograms() {
+		if (FlavoredConstants.DEFAULT_ROOT_DIRECTORY.equals(POCKET_CODE_ROOT_DIRECTORY) && pocketCodeExternalStorageDirectoryExists()) {
+			showImportProjectsDialog(activity);
+		}
+	}
+
+	private boolean pocketCodeExternalStorageDirectoryExists() {
+		return EXTERNAL_STORAGE_ROOT_DIRECTORY.exists();
```
84n4n4: unnecessary function

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+
+	private void showImportProjectsDialog(Activity activity) {
+		AlertDialog.Builder alertDialog = new AlertDialog.Builder(activity);
+		alertDialog.setTitle(R.string.import_dialog_title)
+				.setCancelable(false)
+				.setMessage(R.string.import_dialog_message)
+				.setPositiveButton(R.string.ok, this)
+				.setNeutralButton(R.string.import_dialog_copy_only, this);
+
+		boolean showDialog = PreferenceManager
```
84n4n4: move this out of this function to where showImportProjectsDialog is called, and only build a dialog if you want to show it

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+		if (showDialog) {
+			alertDialog.show();
+		}
+	}
+
+	@Override
+	public void onClick(DialogInterface dialog, int which) {
+		switch (which) {
+			case DialogInterface.BUTTON_POSITIVE:
+				copyProjects();
```
84n4n4: if there is an IOexception or any other problem copying the files the folder will be still deleted.
there also should be some information for the user if there is a problem importing certain projects, and they should not be deleted if import was not working.

84n4n4: maybe copy projects individually in copyProjects and delete them there only if copying was successful. also if copying was not successful, i guess it should be deleted from app internal storage again.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+	private boolean pocketCodeExternalStorageDirectoryExists() {
+		return EXTERNAL_STORAGE_ROOT_DIRECTORY.exists();
+	}
+
+	private void showImportProjectsDialog(Activity activity) {
+		AlertDialog.Builder alertDialog = new AlertDialog.Builder(activity);
+		alertDialog.setTitle(R.string.import_dialog_title)
+				.setCancelable(false)
+				.setMessage(R.string.import_dialog_message)
+				.setPositiveButton(R.string.ok, this)
```
84n4n4: give the user some information that "ok" means "move" and "import_dialog_copy_only" means "copy"

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,169 @@
...
+				copyProjects();
+				try {
+					StorageOperations.deleteDir(EXTERNAL_STORAGE_ROOT_DIRECTORY);
+				} catch (IOException e) {
+					Log.e(TAG, e.getMessage());
+				}
+				doNotShowDialogAgain();
+				break;
+			case DialogInterface.BUTTON_NEUTRAL:
+				copyProjects();
```
84n4n4: since this could take some time, it should not run on the main thread (hearing stories about 100mb+ sized projects out there).
android has this uithread watchdog timer thingy running, and after a rather short time of a busy/frozen ui thread will open this dialog "Pocket code isnt responding" that offers the user to kill the app. (which for obvious reasons would be bad to do whilst copying stuff around)

so this should be in an async task. (of course also applies to above call to copyProjects())

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: done and fixed the prev commit

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: if there is no Pocket Code Folder in ExternalStorage ==> no projects needed to be copied!

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: ok

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: nothing hardcoded anymore.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: ok removed.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: where should be the infos ? in the dialog message?

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: done.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: > there also should be some information for the user if there is a problem importing certain projects, and they should not be deleted if import was not working.
some information like what?

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: should ProgressDialog be shown ?

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.catrobat.catroid.common.FlavoredConstants.OLD_EXTERNAL_STORAGE_ROOT_DIRECTORY;
+import static org.catrobat.catroid.common.SharedPreferenceKeys.SHOW_COPY_PROJECTS_FROM_EXTERNAL_STORAGE_DIALOG;
+import static org.catrobat.catroid.io.StorageOperations.transferData;
+
+public class ImportProjectsFromExternalStorage implements DialogInterface.OnClickListener {
+	public static final String TAG = ImportProjectsFromExternalStorage.class.getSimpleName();
+	private static String pocketCodeEsFolderName = "Pocket Code";
```
84n4n4: the folder names should be located in flavored constants (where the folder paths were previously), because this here will not work with any other flavor than pocket code.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+	private static String pocketCodeIsFolderName = "files";
+	private static String backpackFolderName = "backpack";
```
84n4n4: those two should be final aswell and then of course in caps

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+	private Activity activity;
+
+	public ImportProjectsFromExternalStorage(Activity mActivity) {
+		activity = mActivity;
+	}
+
+	private static File copyProjectsFromExternalToInternalStorage(File srcDir, File dstDir) throws IOException {
```
84n4n4: this is actually copying folders, not only Projects, and it doenst care if its internal or external storage since its getting just Files as src and destination.
copyFolderRecursive would be more fitting name for this.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+
+		if (!dstDir.isDirectory()) {
+			throw new IOException("Cannot create directory: " + dstDir.getAbsolutePath());
+		}
+
+		for (File srcFile : srcDir.listFiles()) {
+			String srcFileName = srcFile.getName();
+			if (srcFile.isDirectory()) {
+				boolean fileExists = false;
+				for (File dstFile : dstDir.listFiles()) {
```
84n4n4: instead of using a boolean flag, put this for loop in a function, eg. directoryContainsFile(File directory, String filename)

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+				doNotShowDialogAgain();
+				break;
+		}
+	}
+
+	private void doNotShowDialogAgain() {
+		PreferenceManager.getDefaultSharedPreferences(activity).edit().putBoolean(SHOW_COPY_PROJECTS_FROM_EXTERNAL_STORAGE_DIALOG, false).apply();
+	}
+
+	private boolean isAppInInternalStorageInstalled(Context context) {
```
84n4n4: why is this necessary? since we do not offer the internal storage as an option, everyone having this released code will be saving projects in internal storage.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
```diff
@@ -0,0 +1,235 @@
...
+		ApplicationInfo io = context.getApplicationInfo();
+		return io.sourceDir.startsWith("/data/");
+	}
+
+	private class CopyProjectsTask extends AsyncTask<String, Boolean, Boolean> {
+		private Activity activity;
```
84n4n4: please use a WeakReference<Activity> here and in on postExecute where you have to access it make sure that the reference is still valid.

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: done

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: done

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: done

catroid/src/main/java/org/catrobat/catroid/utils/ImportProjectsFromExternalStorage.java
ndrnour: deleted

PR #2975, opened 2018-09-25
DO NOT MERGE CATROID-48 Crash when switching scene

LPeteR90:
PoC for fixing CATROID-48 by adding abstract scene-transition aware actions

when continuing scene in example program (see jira ticket), code not executed due to scene switch gets executed after ContinueScene Action


LPeteR90:
Hmmm... I'll look into the ThreadScheduler solution a little bit more, could work too...

Regarding the soft scene transition: Don't know of any possible side effects and I think it should work fine to simply not execute (i.e. return false on act() or don't call act()) actions that are scheduled after a SceneTransitionAction in one render cycle, but that doesn't mean that there won't be any complications. :/

LPeteR90:
closing this PR since @rschluesselbauer will fix this ticket with changes in the ThreadScheduler.

PR #2976, opened 2018-09-25
CATROID-70 remove unused broadcast message deletion on stage start

84n4n4:


wslany:
Where is it tested that unused messages are removed on project save / load, as described in CATROID-70?

84n4n4:
There is no list of messages serialized into the xml, only the bricks and their set message are serialized. On project load the list is populated by the messages set in bricks. A broadcast brick can have only one message.
So by design, broadcast messages that arent selected in any brick are not stored.

84n4n4:
unit test added that asserts that unused messages are not saved/loaded.

wslany:
This means that the removal of unused messages works implicitly with the current design, but this feature will not be documented anymore in any way. Please remember: Our tests also serve as maintenance documentation. I wouldn't be very surprised if unused messages will not be removed anymore after some refactoring in two or three years, the reasoning being "nobody told me so, and it was not documented". Also, if this stays as it is, from now on every PO will have to check this manually for each PO review. Unacceptable ;-) The reversal of this functionality might actually be quite likely if we decide to not use XStream anymore and switch to a json format for the serialization, and somebody gets the idea that it would be good to list all messages somewhere in the code, e.g., in order to list them all at one place per scene, and use ids of messages in the code itself instead of the strings, which makes a lot of sense, e.g., if a feature gets implemented that allows to rename messages. I can alsmot guarantee that then suddenly unused messages will not be removed anymore. And also: the same argument of no-need-for-tests could be said for all functionality that works fine.

84n4n4:
@thmq @LPeteR90 @robertpainsi  can somone of you please review and reapprove this PR.

thank you!

robertpainsi:
Can't someone else do it?

PR #2978, opened 2018-09-25
CATROID-55,56,59 private Storage tickets

ndrnour:
CATROID-59 Exposing file uris
 * implementing FileProvider

CATROID-56 Import old Pocket Code folder into app-private storage.
CATROID-55Use app-storage instead of folder on external memory

UiTest included.

### DIFF
--- a/catroid/src/catroid/java/org/catrobat/catroid/common/FlavoredConstants.java
+++ b/catroid/src/catroid/java/org/catrobat/catroid/common/FlavoredConstants.java
@@ -25,17 +25,22 @@ package org.catrobat.catroid.common;
 
 import android.os.Environment;
 
+import org.catrobat.catroid.CatroidApplication;
+
 import java.io.File;
 
 import static org.catrobat.catroid.common.Constants.MAIN_URL_HTTPS;
 
 public final class FlavoredConstants {
-
-	public static final File DEFAULT_ROOT_DIRECTORY = new File(
-			Environment.getExternalStorageDirectory().getAbsolutePath() + "/Pocket Code");
-
 	// Web:
 	public static final String BASE_URL_HTTPS = MAIN_URL_HTTPS + "/pocketcode/";
+
+	public static final String POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME = "Pocket Code";
+
+	public static final File DEFAULT_ROOT_DIRECTORY = CatroidApplication.getAppContext().getFilesDir();
+
+	public static final File EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
+			Environment.getExternalStorageDirectory().getAbsolutePath(), POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME);
 
 	// Media Library:
 	public static final String LIBRARY_BASE_URL = "https://share.catrob.at/pocketcode/download-media/";
--- a/catroid/src/lunaAndCat/java/org/catrobat/catroid/common/FlavoredConstants.java
+++ b/catroid/src/lunaAndCat/java/org/catrobat/catroid/common/FlavoredConstants.java
@@ -25,17 +25,22 @@ package org.catrobat.catroid.common;
 
 import android.os.Environment;
 
+import org.catrobat.catroid.CatroidApplication;
+
 import java.io.File;
 
 import static org.catrobat.catroid.common.Constants.MAIN_URL_HTTPS;
 
 public final class FlavoredConstants {
-
-	public static final File DEFAULT_ROOT_DIRECTORY = new File(
-			Environment.getExternalStorageDirectory().getAbsolutePath() + "/Luna&Cat");
-
 	// Web:
 	public static final String BASE_URL_HTTPS = MAIN_URL_HTTPS + "/luna/";
+
+	public static final String POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME = "Luna&Cat";
+
+	public static final File DEFAULT_ROOT_DIRECTORY = CatroidApplication.getAppContext().getFilesDir();
+
+	public static final File EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
+			Environment.getExternalStorageDirectory().getAbsolutePath(), POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME);
 
 	// Media Library:
 	public static final String LIBRARY_BASE_URL = "https://share.catrob.at/pocketcode/download-media/";
--- a/catroid/src/main/AndroidManifest.xml
+++ b/catroid/src/main/AndroidManifest.xml
@@ -260,5 +260,16 @@
         <service android:name=".utils.StatusBarNotificationManager$NotificationActionService" />
         <service android:name=".cast.CastService" android:exported="false"/>
 
+        <provider
+            android:authorities="${applicationId}.fileProvider"
+            android:name="android.support.v4.content.FileProvider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_paths" />
+        </provider>
+
     </application>
 </manifest>
--- a/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
@@ -64,6 +64,7 @@ import org.catrobat.catroid.ui.recyclerview.dialog.AboutDialogFragment;
 import org.catrobat.catroid.ui.recyclerview.dialog.PrivacyPolicyDialogFragment;
 import org.catrobat.catroid.ui.recyclerview.fragment.MainMenuFragment;
 import org.catrobat.catroid.ui.settingsfragments.SettingsFragment;
+import org.catrobat.catroid.utils.ImportProjectsFromExternalStorage;
 import org.catrobat.catroid.utils.PathBuilder;
 import org.catrobat.catroid.utils.ScreenValueHandler;
 import org.catrobat.catroid.utils.ToastUtil;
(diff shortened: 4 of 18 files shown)


ndrnour:
replacer with #2978

catroid/src/catroid/java/org/catrobat/catroid/common/FlavoredConstants.java
```diff
@@ -25,15 +25,21 @@
...
 import java.io.File;
 
 import static org.catrobat.catroid.common.Constants.MAIN_URL_HTTPS;
 
 public final class FlavoredConstants {
 
-	public static final File DEFAULT_ROOT_DIRECTORY = new File(
+	public static final File DEFAULT_ROOT_DIRECTORY = CatroidApplication.getAppContext().getFilesDir();
+
+	public static final File OLD_EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
```
thmq: why would you call it `OLD_EXTERNAL_STORAGE_DIRECTORY`? if there is no new one. I think it should be fine if you just called it `EXTERNAL_STORAGE_DIRECTORY`.

catroid/src/main/AndroidManifest.xml
```diff
@@ -260,5 +260,16 @@
         <service android:name=".utils.StatusBarNotificationManager$NotificationActionService" />
         <service android:name=".cast.CastService" android:exported="false"/>
 
+        <provider
+            android:authorities="org.catrobat.catroid.fileProvider"
```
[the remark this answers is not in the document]
ndrnour: done. thanks for the Infos

catroid/src/lunaAndCat/java/org/catrobat/catroid/common/FlavoredConstants.java
```diff
@@ -25,15 +25,21 @@
...
 
 public final class FlavoredConstants {
 
-	public static final File DEFAULT_ROOT_DIRECTORY = new File(
+	public static final File DEFAULT_ROOT_DIRECTORY = CatroidApplication.getAppContext().getFilesDir();
+
+	public static final File EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
 			Environment.getExternalStorageDirectory().getAbsolutePath() + "/Luna&Cat");
 
+	public static final String POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME = "Pocket Code";
```
84n4n4: this has to be "Luna&Cat" aswell!!!

catroid/src/lunaAndCat/java/org/catrobat/catroid/common/FlavoredConstants.java
```diff
@@ -25,15 +25,21 @@
...
 import java.io.File;
 
 import static org.catrobat.catroid.common.Constants.MAIN_URL_HTTPS;
 
 public final class FlavoredConstants {
 
-	public static final File DEFAULT_ROOT_DIRECTORY = new File(
+	public static final File DEFAULT_ROOT_DIRECTORY = CatroidApplication.getAppContext().getFilesDir();
+
+	public static final File EXTERNAL_STORAGE_ROOT_DIRECTORY = new File(
```
84n4n4: its probably best to define "Pocket Code", "Luna&Cat", etc once here in constance, and then use this constant for creating EXTERNAL_STORAGE_ROOT_DIRECTORY, or we will fall in the same trap as below again and again....

catroid/src/main/java/org/catrobat/catroid/ui/ProjectActivity.java
```diff
@@ -212,7 +211,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
 				addSpriteFromUri(uri);
 				break;
 			case SPRITE_CAMERA:
-				uri = Uri.fromFile(new File(DEFAULT_ROOT_DIRECTORY, getString(R.string.default_look_name) + ".jpg"));
+				uri = new ImportFromCameraLauncher(this).getCacheCameraUri();
```
84n4n4: this would be way nicer when the camera intent would return the uri as a result data instead of making a new importFromCamerLauncher to reproduce the same uri again.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -195,6 +196,7 @@ private void onPermissionsGranted() {
 		if (SettingsFragment.isCastSharedPreferenceEnabled(this)) {
 			CastManager.getInstance().initializeCast(this);
 		}
+		new ImportProjectsFromExternalStorage(this).showImportProjectsDialog();
```
84n4n4: please move the check if this should be displayed out in this class

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/CopyProjectsTask.java
```diff
@@ -0,0 +1,85 @@
...
+	protected Boolean doInBackground(String... strings) {
+		boolean success = utils.copyProjects();
+		if (success && deleteFromExStorage) {
+			success = utils.deleteOldExternalPocketCodeFolder();
+		}
+		return success;
+	}
+
+	@Override
+	protected void onPostExecute(Boolean aBoolean) {
```
84n4n4: please rename aBoolean to success

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/CopyProjectsTask.java
```diff
@@ -0,0 +1,85 @@
...
+import android.app.ProgressDialog;
+import android.os.AsyncTask;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.utils.ToastUtil;
+
+import java.lang.ref.WeakReference;
+
+public class CopyProjectsTask extends AsyncTask<String, Boolean, Boolean> {
+	private ProjectsStorageUtils utils = new ProjectsStorageUtils();
```
84n4n4: please remove the project storage utils and move those methods in this class since they are only used here

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/ProjectsStorageUtils.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.catrobat.catroid.io.StorageOperations.getUniqueFile;
+import static org.catrobat.catroid.io.StorageOperations.transferData;
+
+public class ProjectsStorageUtils {
+
+	public static final String TAG = ProjectsStorageUtils.class.getSimpleName();
+	private static final String POCKET_CODE_INTERNAL_STORAGE_FOLDER_NAME = "files";
+	private static final String BACKPACK_FOLDER_NAME = "backpack";
+
+	private void copyFolderRecursive(File srcDir, File dstDir) throws IOException {
```
84n4n4: please split, have one function that just deals with getting a list of projects that are to be copied and deal with unique name creation, and then call copyDir (or whatever suitable method there is in StorageHandling class) to do the actual copying for each project separately.

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/ProjectsStorageUtils.java
```diff
@@ -0,0 +1,119 @@
...
+				} else {
+					copyFolderRecursive(srcFile, new File(dstDir, srcFile.getName()));
+				}
+			} else {
+				transferData(srcFile, new File(dstDir, srcFile.getName()));
+			}
+		}
+	}
+
+	private boolean directoryContainsFile(File srcFile, File dstDir) {
```
84n4n4: this should be replaced with getUniqueFile from StorageOperations

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/ProjectsStorageUtils.java
```diff
@@ -0,0 +1,119 @@
...
+			dstDir.mkdir();
+		}
+
+		if (!dstDir.isDirectory()) {
+			throw new IOException("Cannot create directory: " + dstDir.getAbsolutePath());
+		}
+
+		for (File srcFile : srcDir.listFiles()) {
+			if (srcFile.isDirectory()) {
+				if (directoryContainsFile(srcFile, dstDir)) {
```
84n4n4: the check of directoryContainsFile is unnecessary,

copyFolderRecursive(srcFile, getUniqueFile(srcFile.getName(), dstDir));

will just do fine for all, since it will return srcFile.getName if there is no file that it collides with.

84n4n4: anyways, copyFolderRecursive should be replaced with copyDir from StorageOperations, then this method is not called recursively anymore. (then also a lot of the checks at the start of it could be omitted)

catroid/src/main/java/org/catrobat/catroid/utils/storagestuff/ProjectsStorageUtils.java
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}
+	}
+
+	private boolean directoryContainsFile(File srcFile, File dstDir) {
+		for (File dstFile : dstDir.listFiles()) {
+			if (dstFile.getName().equals(srcFile.getName())
+					&& dstFile.getParent().endsWith(POCKET_CODE_INTERNAL_STORAGE_FOLDER_NAME)
+					&& srcFile.getParent().endsWith(POCKET_CODE_EXTERNAL_STORAGE_FOLDER_NAME)
+					&& !srcFile.getName().equals(BACKPACK_FOLDER_NAME)) {
```
84n4n4: maybe we should just omit copying the backpack folder, that would make everything way easier...

wslany:
As discussed with Nur: backpack should also be copied, and tests needed.

ndrnour:
done. Ui test included 
@84n4n4  thanks for your help & could you review the test?

PR #2983, opened 2018-10-01
CATROID-44 User Upload Unresponsive

ndrnour:
* showing Error message if the user trying to upload the default program
 * checking if tmp Folder is created before start uploading
  P.S: tmp Folder is first created when something from share website downloaded


gitarrenfudler:
Some minor things missing -> feedback on Jira! :)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/ProjectUploadActivityTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext());
+		token = sharedPreferences.getString(Constants.TOKEN, Constants.NO_TOKEN);
+		boolean tokenOk = ServerCalls.getInstance().checkToken(token, testUser);
+		assertTrue(tokenOk);
+		Intent intent = new Intent(InstrumentationRegistry.getTargetContext(), ProjectUploadActivity.class)
+				.putExtra(ProjectUploadActivity.PROJECT_NAME, Utils.getCurrentProjectName(InstrumentationRegistry.getTargetContext()));
+
+		// to avoid showing RatePocketCodeDialogFragment during the test
+		sharedPreferences.edit()
```
thmq: you should reset this change in `tearDown()`. Because it is good practice to leave the state of the application (and its shared preferences as you found them). So you could add a member to the test where you buffer the previous value of the preference, such as:

```
private int sharedPreferenceBuffer;

@Before
private void setUp() {
    sharedPreferenceBuffer = sharedPreferences.getInt(NUMBER_OF_UPLOADED_PROJECTS);
    ...
}

@After
private void tearDown() {
    sharedPreferences.putInt(NUMBER_OF_UPLOADED_PROJECTS, sharedPreferenceBuffer)
                     .commit();
   ...

File: catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/activity/ProjectUploadActivity.java [hunk 2]
```diff
@@ -185,6 +187,7 @@ private void onNextButtonClick() {
 		ProjectManager projectManager = ProjectManager.getInstance();
 
 		if (Utils.isDefaultProject(projectManager.getCurrentProject(), this)) {
+			ToastUtil.showError(this, R.string.error_upload_default_project);
```
thmq: Didn't you want to check this before? i.e. in the `Activity` that started the `UploadActivity`? so e.g. in `MainMenuFragment` in `public void onItemClick(@ButtonId int id) {..}` in the `case UPLOAD:` and wherever else the activity should be started.
```
void someMethod() {
   if (Utils.isDefaultProject(projectManager.getCurrentProject(), this)) {
       ToastUtil.showError(this, R.string.error_upload_default_project);
   } else {
        Intent intent = new Intent(getActivity(), ProjectUploadActivity.class)
                 .putExtra(ProjectUploadActivity.PROJECT_NAME, Utils.getCurrentProjectName(getActivity()));
        startActivity(intent);
```

And please also add a test for that behaviour then. :wink:

thmq:
nicely done  @ndrnour thanks a lot :wink:

thmq:
@ndrnour -> some stuff regarding the requirements has changed, I'll get back to you about it once it's cleared up.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/UploadProgressDialogFragment.java
```diff
@@ -128,14 +128,14 @@ public void run() {
 			@Override
 			public void onClick(View v) {
 				startWebViewActivity(SHARE_PROGRAM_URL + ServerCalls.getInstance().getProjectId());
-				dismiss();
+				alertDialog.getOwnerActivity().finish();
```
thmq: according to Android dev documentation `.getOwnerActivity(..)` can in some cases return `null`. so you could just add a null check and call `dismiss()` if the Activity was `null`.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/MainMenuFragment.java
```diff
@@ -157,10 +164,21 @@ public void onItemClick(@ButtonId int id) {
...
-				Intent intent = new Intent(getActivity(), ProjectUploadActivity.class)
-						.putExtra(ProjectUploadActivity.PROJECT_NAME, Utils.getCurrentProjectName(getActivity()));
-				startActivity(intent);
+				Project currentProject = null;
+				try {
+					currentProject = XstreamSerializer.getInstance().loadProject(Utils.getCurrentProjectName(getActivity()), getActivity());
+				} catch (IOException | LoadingProjectException e) {
+					Log.e(TAG, e.getMessage());
+				}
+				if (Utils.isDefaultProject(currentProject, getActivity())) {
```
thmq: here the `currentProject` is still `null` if an exception occurred -> it might be best to just move this code into the `try` block.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/MainMenuFragment.java
```diff
@@ -157,10 +164,22 @@ public void onItemClick(@ButtonId int id) {
 				startActivity(new Intent(getActivity(), WebViewActivity.class));
 				break;
 			case UPLOAD:
-				setShowProgressBar(true);
-				Intent intent = new Intent(getActivity(), ProjectUploadActivity.class)
-						.putExtra(ProjectUploadActivity.PROJECT_NAME, Utils.getCurrentProjectName(getActivity()));
-				startActivity(intent);
+				Project currentProject;
```
thmq: please move to line `170` so that:
``` Project currentProject = XstreamSerializer.getInstance().loadProject(currentProjectName, getActivity());```

I know I'm especially annoying with this PR. :crying_cat_face:

File: catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/ProjectListFragment.java [hunk 6]
```diff
@@ -292,9 +296,20 @@ public void onClick(DialogInterface dialog, int which) {
 										.commit();
 								break;
 							case 4:
-								Intent intent = new Intent(getActivity(), ProjectUploadActivity.class)
-										.putExtra(ProjectUploadActivity.PROJECT_NAME, item.getName());
-								startActivity(intent);
+								Project chosenProject;
```
thmq: same as in the other class -> just move the variable into the `try` block as well.

catroid/src/main/res/values/strings.xml
```diff
@@ -1341,6 +1341,7 @@
     <string name="error_project_compatability">The program is not compatible with this version of the app.</string>
     <string name="error_outdated_version">This app version is outdated. Update now?</string>
     <string name="error_upload_project_with_default_name">You cannot upload a program with the name \'My first program\'. Please rename the program.</string>
+    <string name="error_upload_default_project">You cannot upload default program.</string>
```
gitarrenfudler: Please change this string, sounds a bit strange without article.
"You cannot upload the default program."

gitarrenfudler:
Hi @ndrnour,
Thanks for adding the tests! :)
Please make changes to the string - see comment.
Otherwise it looks good, but will test the apk intensively again when this change is done.

catroid/src/main/res/values/strings.xml
ndrnour: I was expecting such a comment :) 
done.

ndrnour:
> Hi @ndrnour,
> Thanks for adding the tests! :)
> Please make changes to the string - see comment.
> Otherwise it looks good, but will test the apk intensively again when this change is done.

Hi, I've a question which is, 
why we don't check if the uploaded program is a cast/drone/jumping sumo default program? 
does it mean that, it's allowed to upload them, but not allowed to upload the default program?
it doesn't make sense (my opinion)!

wslany:
@ndrnour In principle you are right, but in 99.999% of the cases, the problem is the usual default project, since the other sample projects of the extensions are currently (almost) never used, and never have been uploaded. But even then, uploading them would do little harm, if it is even ever done. So here, it makes sense to forbid the standard default project, which previously cluttered our share a lot, but not the others. This may possibly change later, once we have many users and separate flavors for these extensions. But we follow the YAGNI = https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it  principle, so don't care for it now. I.e., just ignore the other default projects of the extensions.

thmq:
I agree with @wslany on this one. Besides it was not part of the specification.

wslany:
@ndrnour @thmq @mrmattmueller @84n4n4 please clarify what to do with this --- it was only merged in a planned hotfix branch (see above / screenshot below) that was however never released, and never into develop. However, the Jira ticket was closed as merged. What to do?
image

ndrnour:
NICE CATCH!!
dear @wslany the answer that you are looking for is "**NO NEED TO DO ANYTHING WITH THIS**" it's fixed by @thmq  in this PR #3104 which approved by me and merged by you.
actually I'm upset and disappointed since it's not the first time my work didn't see the light.

wslany:
Ah, ok, and thanks. And I understand why you are upset. As you may have seen, I am recently trying to take care of avoiding such situations. One of the measures is that all contributors need to be much more proactive in doing themselves, but also actively requesting code reviews as early as possible, ideally synchronously by the developer together with the reviewer. I'm behind it, but this must become ingrained into *everyone's* way of working. And everyone needs to take responsibility for pushing one's work into being published, up to the app stores - into code review is not enough, and even into develop is not enough. Your work is not done until it reaches our users. I'm wondering what we can do to get everyone committed and working together to getting this into everyone's mind.

PR #2984, opened 2018-10-01
CATROID-32/33/34/35

ndrnour:
* 32-In Say and Think bubbles, no trailing zeroes (.0 etc) should be shown
 * 33-Compute" in the formula editor should show no trailing zeroes (.0 etc)
 * Strip any trailing 0 from numbers before converting them into text strings
 * Do not show any trailing 0 in the list of values of variables and lists

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/NumberFormatsTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/NumberFormatsTest.java
@@ -1,0 +1,67 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.utiltests;
+
+import android.support.annotation.IdRes;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Arrays;
+
+import static junit.framework.Assert.assertEquals;
+
+import static org.catrobat.catroid.utils.NumberFormats.stringWithoutTrailingZero;
+
+@RunWith(Parameterized.class)
+public class NumberFormatsTest {
+
+	@Parameterized.Parameter
+	public @IdRes String name;
+	@Parameterized.Parameter(1)
+	public @IdRes String stringWithoutTrailingZero;
+
+	@Parameterized.Parameters(name = "{0}")
+	public static Iterable<Object[]> data() {
+		return Arrays.asList(new Object[][]{
+				{"0", stringWithoutTrailingZero(String.valueOf(0))},
+				{"8", stringWithoutTrailingZero(String.valueOf(8))},
+				{"-120", stringWithoutTrailingZero(String.valueOf(-120))},
+				{"0", stringWithoutTrailingZero(String.valueOf(0.0))},
+				{"0.5", stringWithoutTrailingZero(String.valueOf(0.5))},
+				{"0.7", stringWithoutTrailingZero(String.valueOf(0.70))},
+				{"0.103", stringWithoutTrailingZero(String.valueOf(0.1030))},
+				{"15.05", stringWithoutTrailingZero(String.valueOf(15.050))},
+				{"string.1900", stringWithoutTrailingZero("string.1900")},
+				{"Pocket", stringWithoutTrailingZero("Pocket")}
+		});
+	}
+
+	@Test
+	public void testStringWithoutTrailingZero() {
+		assertEquals(stringWithoutTrailingZero, name);
+	}
+}
+
--- a/catroid/src/main/java/org/catrobat/catroid/formulaeditor/Formula.java
+++ b/catroid/src/main/java/org/catrobat/catroid/formulaeditor/Formula.java
@@ -34,6 +34,8 @@ import org.catrobat.catroid.formulaeditor.datacontainer.DataContainer;
 import java.io.Serializable;
 import java.util.Set;
 
+import static org.catrobat.catroid.utils.NumberFormats.stringWithoutTrailingZero;
+
 public class Formula implements Serializable {
 
 	private static final long serialVersionUID = 1L;
@@ -161,7 +163,8 @@ public class Formula implements Serializable {
 			throw new InterpretationException("NaN in interpretString()");
 		}
 
-		return String.valueOf(interpretation);
+		String value = String.valueOf(interpretation);
+		return stringWithoutTrailingZero(value);
 	}
 
 	public Object interpretObject(Sprite sprite) {
--- a/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
+++ b/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
@@ -50,6 +50,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 
+import static org.catrobat.catroid.utils.NumberFormats.stringWithoutTrailingZero;
+
 public class FormulaElement implements Serializable {
 
 	private static final long serialVersionUID = 1L;
@@ -95,7 +97,7 @@ public class FormulaElement implements Serializable {
 	}
 
 	public String getValue() {
-		return value;
+		return stringWithoutTrailingZero(value);
 	}
 
 	public List<InternToken> getInternTokenList() {
@@ -141,7 +143,7 @@ public class FormulaElement implements Serializable {
 				internTokenList.add(new InternToken(InternTokenType.USER_LIST, this.value));
 				break;
 			case NUMBER:
-				internTokenList.add(new InternToken(InternTokenType.NUMBER, this.value));
+				internTokenList.add(new InternToken(InternTokenType.NUMBER, stringWithoutTrailingZero(this.value)));
 				break;
 			case SENSOR:
 				internTokenList.add(new InternToken(InternTokenType.SENSOR, this.value));
@@ -315,12 +317,7 @@ public class FormulaElement implements Serializable {
 		if (userListValues.size() == 0) {
 			return EMPTY_USER_LIST_INTERPRETATION_VALUE;
 		} else if (userListValues.size() == 1) {
-			Object userListValue = userListValues.get(0);
-			if (userListValue instanceof String) {
-				return userListValue;
-			} else {
-				return userListValue;
-			}
+			return userListValues.get(0);
 		} else {
 			return interpretMultipleItemsUserList(userListValues);
 		}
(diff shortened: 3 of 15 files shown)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/Formula.java
```diff
@@ -161,7 +164,25 @@ public String interpretString(Sprite sprite) throws InterpretationException {
...
+				}
+				returnedValue = String.valueOf(new DecimalFormat(pattern).format(Double.parseDouble(value)));
+			} catch (Exception e) {
+				Log.e(TAG, e.getMessage());
+				returnedValue = value;
+			}
+		} else {
+			returnedValue = value;
+		}
+		return returnedValue;
```
robertpainsi: I can't say if I'll review your PR due to time constraints… It's very likely that this will be my only comment on this PR :scream_cat:
However, the trailing zero is a problem addressing multiple components (related tickets). It would be great if you can create a static function for that. It will also be really easy to test.

> Ideally, there should be a single implementation for formatting these numbers…

ndrnour: I agree with you, could you suggest the name of the methode :( I still have problem with naming variables and methodes ^_#

ndrnour: 👍

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -507,4 +508,25 @@ public static int getBit(int number, int index) {
...
 	}
+
+	public static String stringWithoutTrailingZero(String value) {
+		String returnedValue;
+		if (value.contains(".")) {
+			try {
+				String[] splitter = value.split("\\.");
+				int afterDecimalCount = splitter[1].length();
+				String pattern = "#.";
+				for (int i = 0; i < afterDecimalCount; i++) {
```
84n4n4: is there going to be any limit on how many decimal places there should be displayed?

84n4n4:
please add unit test for this.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -507,4 +508,25 @@ public static int getBit(int number, int index) {
...
+					pattern = pattern.concat("#");
```
84n4n4: instead of using a for loop, you could use String.format and create a string with a given number of "#", see https://stackoverflow.com/questions/13475388/generate-fixed-length-strings-filled-with-whitespaces

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -507,4 +508,25 @@ public static int getBit(int number, int index) {
...
+				}
+				returnedValue = String.valueOf(new DecimalFormat(pattern).format(Double.parseDouble(value)));
```
84n4n4: isnt this producing exactly the same as "value" already is?

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -507,4 +508,25 @@ public static int getBit(int number, int index) {
 		}
 		return 0;
 	}
+
+	public static String stringWithoutTrailingZero(String value) {
```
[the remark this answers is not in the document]
ndrnour: done

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
ndrnour: no need anymore. I've found another way to do it

wslany:
In https://jira.catrob.at/browse/CATROID-35 and https://jira.catrob.at/browse/CATROID-34, numbers in list entries are _not yet correctly_ stripped of their trailing zeroes after the dot. Consider the following program:

    When program starts
    Add 1.00 to to list Xyz
    Add 1.00 to to list Xyz
    Say join( element( 1, *Xyz* ) , element( 2 , *Xyz* ) ) for 2 seconds
    Say *Xyz* for 2 seconds
    Add 1.05 to to list Xyz
    Say *Xyz* for 2 seconds

1. This first `say` brick _correctly_ shows `11` in a speech bubble for 2 seconds. 
2. The second `say` brick then _incorrectly_ shows `1.00 1.00` for another 2 seconds, instead of the correct value `11` again, which is the flattened list _without_ spaces (_without_ spaces because it should only have _single_ characters or _single_ digit numbers in _each_ element). 
3. The third `say` brick _incorrectly_ shows `1.00 1.00 1.05`, instead of the correct value `1 1 1.05` (flattened list, but the third element is not a single character or digit, and thus spaces are inserted).

Note that there are already automatic tests that specify the flattening of lists in that way, i.e., with spaces (when _not_ all elements are single characters or single digits) and without spaces (when _all_ elements are single characters or single digits). This behavior of flattened lists is inherited from Scratch.

Additionally, if one inspects the entries of the list in the formula editor after executing the whole script (i.e., after 6 seconds), by pressing the "Data" at the bottom left and then the small triangle on the right of the list name, the values of Xyz shown in the pull down listing _incorrectly_ are 

    1.00
    1.00
    1.05

instead of the _correct_ values

    1
    1
    1.05

Please correct and add tests for these cases.

ndrnour:
done.

wslany:
Problem number 2 in the list still has spaces, even though there should be _no_ spaces in that case, because both numbers in the list are single character numbers; see my change request:

> 2. The second `say` brick then _incorrectly_ shows `1.00 1.00` for another 2 seconds, instead of the correct value `11` again, which is the flattened list _without_ spaces (_without_ spaces because it should only have _single_ characters or _single_ digit numbers in _each_ element).

The result currently still incorrectly showing is `1 1` (with a space), but there should be _no_ space, i.e., the correct result should be `11`.

Note that if both numbers added to the list are single character numbers, the result is already shown correctly without space. So somehow, 1.0 is handled in a different way in the flattening of the list than a 1, which should not be the case as long as 1.0 and 1 are the same number, and are treated as numbers here.

thmq:
> The result currently still incorrectly showing is `1 1` (with a space), but there should be _no_ space, i.e., the correct result should be `11`.

Are you sure though? the `1`s come from a list of `1` so in the list these are separate numbers (one, one and not eleven). @wslany do you not want to visually separate these list elements? because if you really want to get rid of those whitespaces you would lose them between **all** types of objects in lists (e.g. strings, sensor values, etc.).

thmq:
@wslany thank you for the clarification :wink:

wslany:
One more comment: actually, what should happen and what I want you, @ndrnour, to implement is that when the user types in a formula and one of the **numbers** in the formula is something like `3.0` (or `3.00`), these numbers should not be kept in this format. Instead, when exiting the formula editor, the number should already be replaced by `3` etc. --- the next time this formula is shown on the screen, either in a parameter field or in the formula editor, only the number `3` is shown, and the `.0` etc need to be already deleted. 

Note: For **text** in formulas, it is of course different: "003.0000" must be kept as this string, since it is not really a number but a text.

ndrnour:
> types

working on it

ndrnour:
Done. thanks @thmq for help

wslany:
I just tried the newest version, ahead of the PO review phase (sorry), and there unfortunately is still a problem, as follows:

    When program starts
    Add 1.0 to to list Xyz
    Add 1 to to list Xyz
    Say join( '11 ', *Xyz* )

The text shown is '11 1 1'. The strange thing is that, because of the last correction, the script looks correct, i.e., with the 1.0 in the second brick already replaced by a single digit "1". However, upon closer inspection of the parameter in the formula editor, I found that there somehow is a space character after the 1 inserted in the second brick. If that space character is deleted in the formula editor, then the correct expected result, '11 11' is shown. So somehow, the automatic stripping of the ".0" in the second brick appears to insert a space at the end of the 1, i.e., resulting in a "1 " in the parameter field, and that is somehow taken into account when the *Xyz* list is flattened.

ndrnour:
> I just tried the newest version, ahead of the PO review phase (sorry), and there unfortunately is still a problem, as follows:
> 
> ```
> When program starts
> Add 1.0 to to list Xyz
> Add 1 to to list Xyz
> Say join( '11 ', *Xyz* )
> ```
> The text shown is '11 1 1'. The strange thing is that, because of the last correction, the script looks correct, i.e., with the 1.0 in the second brick already replaced by a single digit "1". However, upon closer inspection of the parameter in the formula editor, I found that there somehow is a space character after the 1 inserted in the second brick. If that space character is deleted in the formula editor, then the correct expected result, '11 11' is shown. So somehow, the automatic stripping of the ".0" in the second brick appears to insert a space at the end of the 1, i.e., resulting in a "1 " in the parameter field, and that is somehow taken into account when the _Xyz_ list is flattened.

fixed

wslany:
Looks good from my side. Please, have someone do a code review of the last changes, then we are ready to merge.

catroid/src/androidTest/java/org/catrobat/catroid/test/utiltests/NumberFormatsTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+	@Before
+	public void setUp() {
+		stringContainsDot = "string.1900";
+		stringValue = "Pocket";
+	}
+
+	@Test
+	public void testStringWithoutTrailingZero() {
+		assertEquals("0", NumberFormats.stringWithoutTrailingZero(String.valueOf(0)));
```
84n4n4: please make this a parameterized test

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -315,62 +317,42 @@ private Object interpretUserList(Sprite sprite) {
...
-					userListStringValues.add(stringValueOfListItem);
-					concatenateWithoutWhitespace = false;
-				}
+				userListStringValues.add(stringWithoutTrailingZero(stringValueOfListItem));
 			}
 		}
 
 		StringBuilder stringBuilder = new StringBuilder(userListStringValues.size());
-		boolean isFirstListItem = true;
 		for (String userListStringValue : userListStringValues) {
```
84n4n4: to make this a little bit easier to read - 
if youd move this for loop in a method that is called "listConsistsOfSingleCharacters" or similar, and then instead of using the boolean flag "whiteSpaceNeeded" add a new variable named "seperator" and before concatinting the list you set the separator to either "" or " " based on the return of above method

catroid/src/main/java/org/catrobat/catroid/utils/NumberFormats.java
```diff
@@ -0,0 +1,64 @@
...
+
+	private NumberFormats() {
+		throw new AssertionError();
+	}
+	public static final String TAG = NumberFormats.class.getSimpleName();
+
+	public static String stringWithoutTrailingZero(String value) {
+		if (value.contains(".") && value.matches("[0-9.-]+")) {
+			String[] splitter = value.split("\\.");
+			String splitter1 = splitter[1];
```
84n4n4: please rename splitter1 to "decimalPlaces"

catroid/src/main/java/org/catrobat/catroid/utils/NumberFormats.java
```diff
@@ -0,0 +1,64 @@
...
+			for (int i = 0; i < splitter1.length(); i++) {
```
84n4n4: wouldnt do  
splitter1.replaceAll("0*$", "");
the same thing as the loop?

84n4n4: also have a look at https://stackoverflow.com/questions/14984664/remove-trailing-zero-in-java
s = s.indexOf(".") < 0 ? s : s.replaceAll("0*$", "").replaceAll("\\.$", "");
should work in place of this whole function

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
ndrnour: done.

catroid/src/main/java/org/catrobat/catroid/utils/NumberFormats.java
ndrnour: you're right. done

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -315,62 +317,44 @@ private Object interpretUserList(Sprite sprite) {
...
-					concatenateWithoutWhitespace = false;
-				}
+				userListStringValues.add(stringWithoutTrailingZero(stringValueOfListItem));
 			}
 		}
 
 		StringBuilder stringBuilder = new StringBuilder(userListStringValues.size());
-		boolean isFirstListItem = true;
+		String separator = listConsistsOfSingleCharacters(userListStringValues);
 		for (String userListStringValue : userListStringValues) {
```
84n4n4: clean code wise it would be nicer to have that method return true or false and then go there with

String separator = listConsistsOfSingleCharacters(userListStringValues) ? "" : " ";

because method name implies to me that it would return true or false, not a separator.

ndrnour: ok, done

PR #2988, opened 2018-10-03
CATROID-47-Fix SetBackgroundByIndexBrick interpret formula bug

gPathpp:


catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundBrick.java
```diff
@@ -44,6 +45,12 @@ public SetBackgroundBrick() {
 	}
 
 	@Override
+	public View getView(Context context) {
```
thmq: this is not necessary. please remove it -> see my other comment for more context.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookByIndexBrick.java
```diff
@@ -67,24 +66,12 @@ public int getViewResource() {
...
-			label.setText(R.string.brick_set_background_by_index);
-		}
-
 		return view;
 	}
 
 	@Override
 	public View getPrototypeView(Context context) {
 		View prototypeView = super.getPrototypeView(context);
-
```
thmq: It is not necessary to override a method that just calls `super` and returns its result. You can just remove the entire method `public View getPrototypeView(Context context) {..}`

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetLookByIndexBrick.java
```diff
@@ -67,24 +66,12 @@ public int getViewResource() {
 	@Override
 	public View getView(Context context) {
 		super.getView(context);
-
```
thmq: It is not necessary to override a method that just calls super and returns its result. You can just remove the entire method public View getView(Context context) {..}

PR #2990, opened 2018-10-06
CATROID-72 Fix ContinueSceneBrick

rschluesselbauer:
Restoring from StageBackup now removes all the actors from stage and
adds all actors of the current scene.


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -543,6 +528,23 @@ public void render() {
...
 
+	private void addAllActorsToStage(@ResetSpriteType int reset) {
+		boolean addPenActor = true;
+		for (Sprite sprite : sprites) {
+			if (reset == RESET_SPRITE) {
+				sprite.resetSprite();
+				sprite.look.createBrightnessContrastHueShader();
+			}
+			stage.addActor(sprite.look);
+			if (addPenActor) {
```
84n4n4: please move this out of the for loop, since this is done once and only once, and do this before or after the loop.

84n4n4:
please change base to release-0.9.42

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
rschluesselbauer: Yeah, sorry I left it in there. It was done this way, because order matters, but I've found a different way now.

PR #2992, opened 2018-10-08

shellp:
- larger font-size for axes
- simple test to check axes font size


84n4n4:
please rebase and reopen

PR #2993, opened 2018-10-09
CATROID-72 Fix ContinueSceneBrick

thmq:
- Changed object references -> backup now holds new list containing
   references instead of a reference to the original list.
 - Removed stage object from backup, because now only one stage is used,
   where state is restored from actors.
 - Removed and simplified some code.


thmq:
Basically the same approach as #2991. Except some of the object references are different -> might be possible to unify the solutions, or it might create a lot of new bugs.

thmq:
ping @robertpainsi

robertpainsi:
Sy, I won't be able to review this PR. I'll just open one PR I prepared yesterday and quickly write a Jira ticket. That's it for the next ~3 weeks…

thmq:
@robertpainsi: don't worry about it. We'll try to make do without you then :wink:

gitarrenfudler:
Checked it with the described case, various derivations and a couple of existing projects - looks good! Thanks! :)

PR #2994, opened 2018-10-10
CAT-2658 Log git commit info

robertpainsi:
Logging the commit hash will help identify some problems.
* Advanced users can identify the commit the app was build from.
* When often switching commits, sometimes the build is not updated
  correctly and you end up unknowingly testing an older version of the
  app.
___
Maybe I can has dis? :hamburger::cat2:
I somehow think it's useful. Just last Monday I experienced this problem when bisecting a bug, [CATROID-79](https://jira.catrob.at/browse/CATROID-79). Sometimes the app was build from the previous commit, it's really frustrating…


catroid/src/main/java/org/catrobat/catroid/CatroidApplication.java
```diff
@@ -52,6 +52,7 @@
 	public void onCreate() {
 		super.onCreate();
 		Log.d(TAG, "CatroidApplication onCreate");
+		Log.d(TAG, "git commit info: " + BuildConfig.GIT_COMMIT_INFO);
```
robertpainsi: The log looks nicer if the commit info is logged after `CatroidApplication onCreate`.
```
10-10 16:27:36.178 29269-29269/org.catrobat.catroid D/CatroidApplication: CatroidApplication onCreate
    git commit info: branch log-git-commit-info @cf0c45c0d
```

LPeteR90:
There's a checkstyle error at build.gradle:393 (unnecessary newline). And are you sure the GIT_CURRENT_BRANCH field is used nowhere (thinking of jenkins...)?

Other than that, I'll be happy to approve ;)

robertpainsi:
Thanks for the hint! I think these `buildConfigField`s are just for the Android code itself. To pass data from gradle to java. But according to the git log, these fields have been added without being used… (a very long time ago though). So I'm not so sure anymore. Ping @Catrobat/jenkins If they aren't used, I'll also remove the `GIT_DESCRIBE` as well.
I'll fix the Checkstyle errors asap!

mfuchs:
Jenkins does not use these fields.

robertpainsi:
Updated and fixed. I've created a ticket on the new board to follow the new workflow and updated the link references here and in the commit message.

Thank you very much, @LPeteR90, @mfuchs

PR #2998, opened 2018-10-12
CATROID-48 Suspend all scripts on scene transition... 

rschluesselbauer:
...and resume all previously running scripts in the new scene.

A POC for how we could handle it - suspend all schedulers on scene switch and resume schedulers when a scene is continued. This way we can continue scripts that were running during scene switch. Is this desired behaviour?
Tests are WIP.


wslany:
Yes.

thmq:
@rschluesselbauer  please have a look at the tests and reopen when done.

PR #3008, opened 2018-10-25
CATROID-84, JENKINS-256 Uses android-emulators-gradle plugin instead of the buildScripts

mfuchs:
* Emulators can be specified in the build.gradle file.
* Emulators can be created/started/stopped via gradle commands.
* Dependencies (Android SDK, sdkmanager modules) are installed by the plugin
* logcat.txt files are stored automatically.
* The Python build-scripts files are not needed anymore.
* Some of the Jenkinsfiles are also not needed anymore.
  The PRs are handled by the multibranch pipeline job now, thus the specific
  PR Jenkinsfiles are redundant.
* Uses gradle commands directly in the Jenkinsfiles.
  That makes clearer what actually happens.
* Jenkinsfile.ManualTests directly specifies its paramters.
  That makes it easier in the future to update.
  Now the Job DSL files do not need to be touched whenever different
  paramters are needed.

Note: When this PR is merged I also need to update the Job DSL for the ManualTests job.
That should not be needed in the future.
Thus during the timespan this PR gets merged and I have updated the Job DSL the ManualTests job will not work.


mfuchs:
The build job failed since I am not part of the Catroid-GitHub-Team. Thus the changes in the Jenkinsfile are not picked-up by Jenkins for security reasons.

I did many jobs related to try the changes of JENKINS-256, see https://jenkins.catrob.at/job/lab/job/Matthias_Fuchs/

amix73:
These tests are too flaky - but finally No. 13 made it. 
https://jenkins.catrob.at/view/Catroid/job/Catroid/job/PR-3008/13/

PR #3009, opened 2018-10-29
CATROID-61 Add NotificationChannel SDKTarget26

gPathpp:


catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -331,4 +339,23 @@ private void closeNotificationBar() {
...
+			CharSequence name = context.getResources().getString(R.string.channel_name);
+			String description = context.getResources().getString(R.string.channel_description);
+			int importance = NotificationManager.IMPORTANCE_DEFAULT;
+			NotificationChannel channel = new NotificationChannel(CHANNEL_ID, name, importance);
+			channel.setDescription(description);
+			channel.enableVibration(false);
+			channel.enableLights(false);
+			channel.setSound(null, null);
+			// Register the channel with the system; you can't change the importance
+			// or other notification behaviors after this
```
LPeteR90: please remove comments

catroid/src/main/res/values/strings.xml
```diff
@@ -1300,6 +1300,11 @@
     <string name="drone_set_tilt_outdoor">20 degrees </string>
     <string name="drone_set_tilt_max">30 degrees</string>
 
+    <!-- Notification Channel -->
+    <string name="channel_name">Pocket Code</string>
+    <string name="channel_description">Pocket Code notification channel</string>
```
wslany: While you are at eliminating the hardcoded "Pocket Code" from strings: if it is easy, please also correct the other hardcoded strings in https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/res/values/strings.xml#L208-L209 => 

`<string name="error_scratch_converter_outdated_app_version">Your version of Pocket Code is outdated. Our Scratch
        converter only works on the latest version. Please update it through Google Play</string>
`

should be

`<string name="error_scratch_converter_outdated_app_version">Your version of this app is outdated. Our Scratch
        converter only works on the latest version. Please update the app.</string>
`

Please let me know whether it is possible. Thank you.

There are more eliminations of hardcoded names that are needed in strings.xml, e.g., concerning Google Play, but that will come with another pull request, independent of this one, so don't concern yourself with them in the present pull request.

PR #3010, opened 2018-10-30
CATROID-79 Wrong layer property for background

ndrnour:


robertpainsi:
Hey @ndrnour , thanks for working on this ticket! :grinning::pray:
https://share.catrob.at/pocketcode/program/8622 is a good program to check if the layer property is still correct. It just lakes the background layer value though. So, if everything works as expected, the rectangles should be numbered 1-4.

In the current implementation, I think it's only testable reliably with a stage test (integration test). The stage sneaks in some objects (Pen) between the first object and the Background. The background z-index is `0` and is correct, the first object z-index is `2`. When the Stitch feature is integrated, there will be another object between the background and the first object which increases the first object z-index to `3`.
Covering it with unit tests would be cool, but I think isn't possible. So maybe you can add an integration test here?

PR #3018, opened 2018-11-02
CATROID-127 Reviewed English strings.

thmq:
- new src strings as proposed by @wslany.


catroid/src/main/res/values/strings.xml
```diff
@@ -320,7 +320,7 @@
     <string name="am_convert">Convert</string>
     <string name="dialog_confirm_delete">You can\'t undo this!</string>
 
-    <string name="list_headline_sprites">Objects</string>
+    <string name="list_headline_sprites">Actors and objects</string>
```
robertpainsi: Just curious about this change. Why `actors`?

thmq: this changes were proposed by @wslany -> so you'll have to ask him :smirk_cat:

wslany: @robertpainsi Thank you for asking about the proposed change in terminology, as it gives me the opportunity to explain it here, and also to address some more fundamental questions about terminology used in our project. 

The metaphor we are using in Catrobat, inherited from Scratch, stems from theaters or movies: in theaters or movies as well as in Catrobat, we have a stage with a background, the actors follow their individual scripts individually (objects such as cars etc also move around individually), where the individual pieces of the scripts are mostly formulated as orders to the actor ('Wait 2 seconds ", "Go to random position", ...) or describe changes in objects or backgrounds ("Set mass to 2 kilograms", ...), some of the actors speak and/or ask questions, make sounds, hide and show themselves, go to the back and to the front of the stage, are all initially seen in 2D from the spectators' point of view on the stage, and have several sequential scenes. 

Note that any metaphor never has perfectly identical characteristics as the original, see for instance the "desktop metaphor", but that as a general rule in language use, one should not mix metaphors as much as possible in order to avoid confusion. Metaphors should make it simpler to understand what is meant by words and what can be done with the things and actors, e.g., "folders", or the original "dustbin" on desktop GUIs, where one can take something out of a dustbin in reality as well as in the GUI. That's also why "using good metaphors" is one of the 12 basic practices of the original eXtreme Programming. 

Scratch uses the term "costume" instead of our "look", but look is clearly a closer fit and more understandable in the context of the metaphor because the visible parts of objects and living beings on the stage, e.g., animals or heads of humans or monsters, in theater pieces and movies are obviously better described by the more generic "look" than the much more constrained term "costume". 

Scratch uses the term "Sprite" that, while traditionally having been used among programmers for moving graphical elements in computer games from a developer's point of view, wouldn't be known among our users from any formal programming education, since they wouldn't have had such previous knowledge nor education. "Sprite" also breaks the theater/movie metaphor, as it normally means "elf" or "spirit" in English (in German Kobold, Elfe or Elementargeist), which also does not fit well to the metaphor, e.g., when used for a ball, a human, an alien, car, rocket, or orc monster. The words object and actor are much broader in their usage and thus are a more natural and immediately understandable fit. No "Hmmm, what could that be?" delay... 

Additionally, the name "object" was clearly coming from the programming terminology of our software developer background. It breaks the metaphor unnecessarily: In a theater or movie, besides the background, you simply _do_ have animated/living as well as inanimate entities, in real life *commonly* named "actors" and "objects", which thus make the most sense in our chosen metaphor. Note that they also fit in any translated versions of the words. And they are commonly used and known by our intended target group of teenagers around the world. Moreover, I always thought that using "object" for living beings such as humans somehow felt degrading and insulting, not to mention that ordering them around and letting them speak made little sense for mere inanimate "objects". 

Having longer and more precise while well understandable names and phrases is not only good practice in the Clean Code methodology because it makes a lot of sense for maintainability reasons, it also is already in use everywhere else in our GUI, e.g., when we allow long phrases with spaces and exclamation points or even graphical emojis in variable names, and long phrases in the names of bricks and formula elements. Also, the artificial restriction in most programming languages that a name of a command, variable, message etc needs to be connected by non-space characters, only stems from the need to be able to more easily parse the written code by compilers and interpreters. For human readers, with their more advanced language parsing capabilities, this is not necessary and actually degrades readability. 

In the same context, I also strongly believe that "When scene starts" is a better name than "When project starts" for that brick: Not only does it perfectly fit the metaphor, even in theater pieces or movies with a _single_ scene, it also emphasizes the new concept of scenes we have introduced. It thereby pushes the concept of our scenes to the forefront of awareness among our users, which is a good thing as scenes are a bit difficult to discover as of now. Nevertheless, we still want to keep scenes a bit more in the background for simpler projects with single scenes, just hinting at "something more" through this new name of the "When scene starts" of the old "When project starts" brick. Scenes can thus be discovered by more advanced users, but with a clear meaning nevertheless even for users that do not get any additional explanation about scenes. Intuitiveness at its best ;-). It also captures much better the intended meaning for any _later_ scenes than the currently "When project starts" brick, which is currently named incorrectly in these second, third, etc scenes.

So, altogether, actors and objects are better suited in Catrobat than the term "object" alone. Since I wanted to correct many other things in strings.xml anyway as well, and considering the need to correct many translations anyway because of all these changes, I thought that now was a good opportunity to correct this unfortunate naming inconsistency as well. The change of name from "program" to "project" goes along the same lines. 

If you spot any other formulations that should be corrected now, please suggest them --- I'm almost sure I didn't catch them all. Consider it a proposal for discussion instead of an already decided change request. However, this format of a pull request allows to easily see all changes precisely, allows to suggest alternatives, and allows to easily discuss them one by one.

catroid/src/main/res/values/strings.xml
```diff
@@ -1156,7 +1153,7 @@
 
 
     <!-- Control bricks -->
-    <string name="brick_when_started">When program starts</string>
+    <string name="brick_when_started">When project or scene starts</string>
```
wslany:
```suggestion
    <string name="brick_when_started">When scene starts</string>
```
As I wrote in my longer comment, I strongly believe that "When scene starts" is a better name than "When project starts" (and "When project or scene starts", which is clumsy and not precise) for that brick: Not only does it perfectly fit the metaphor, even in theater pieces or movies with a _single_ scene, it also emphasizes the new concept of scenes we have introduced. It thereby pushes the concept of our scenes to the forefront of awareness among our users, which is a good thing as scenes are a bit difficult to discover as of now. Nevertheless, we still want to keep scenes a bit more in the background for simpler projects with single scenes, just hinting at "something more" through this new name of the "When scene starts" of the old "When project starts" brick. Scenes can thus be discovered by more advanced users, but with a clear meaning nevertheless even for users that do not get any additional explanation about scenes. Intuitiveness at its best ;-). It also captures much better the intended meaning for any _later_ scenes than the currently "When project starts" brick, which is currently named incorrectly in these second, third, etc scenes.

catroid/src/main/res/values/strings.xml
```diff
@@ -1337,44 +1330,44 @@
...
-    <string name="error_no_writiable_external_storage_available">This app needs writable external storage like a SD
-        card in order to run, but none was found.\nYou can fix this by making the external storage writable or inserting
-        a SD card, then trying again.</string>
-    <string name="error_no_write_access">This app needs permission to write data to your device. Otherwise you
-        can\'t create, modify or run programs. To continue, open settings and enable write access for
-        this app.</string>
-    <string name="error_project_upload">An error occurred while uploading the program.</string>
-    <string name="error_project_download">An error occurred while downloading the program.</string>
-    <string name="error_new_project">An error occurred while creating a new program.</string>
+    <string name="error_no_writable_external_storage_available">This app needs writable external storage like an
```
wslany:
```suggestion
    <string name="error_no_writable_external_storage_available">This app needs writable external storage like an
```
This typo correction will most likely be done in https://jira.catrob.at/browse/CATROID-57 so it may or may not be done here.

catroid/src/main/res/values/strings.xml
```diff
@@ -157,84 +157,84 @@
...
-    <string name="preference_raspi_host">Raspberry Pi IP-/Host-Address</string>
-    <string name="preference_raspi_port">Raspberry Pi Port</string>
+    <string name="preference_raspi_host">Raspberry Pi IP-/host-address</string>
+    <string name="preference_raspi_port">Raspberry Pi port</string>
     <!--  -->
 
     <!-- Scratch Converter activity -->
     <string name="beta">BETA</string>
-    <string name="error_scratch_program_not_accessible_any_more">Whoops! The program seems to have disappeared
+    <string name="error_scratch_program_not_accessible_any_more">Whoops! The project seems to have disappeared
```
wslany: General rule: "project" makes more sense than "program", and is easier to understand for novices. It also emphasizes the "tinkering" aspect (a project focuses on the "doing" stuff instead of the "product"), or the working in teams on a project. For more on this, also see http://web.media.mit.edu/~mres/papers/IJCCI-seeds-seymour-sowed.pdf

catroid/src/main/res/values/strings.xml
```diff
@@ -1167,15 +1164,15 @@
...
+    <string name="brick_stop_other_scripts">other scripts of this actor or object</string>
 
 
     <plurals name="second_plural">
         <item quantity="one">"second "</item>
         <item quantity="other">"seconds "</item>
     </plurals>
 
-    <string name="brick_collision_receive">When physical collision with</string>
+    <string name="brick_collision_receive">When collision under gravity with</string>
```
wslany:
```suggestion
    <string name="brick_collision_receive">When you collide with</string>
```
This would be an even better change, *if* we can extend the functionality to actually also check collision among non physical objects, which should be just a few lines of code since it already works with various sensors and the generalizes "When .... becomes true" event brick. I think I will propose that as a "bug ticket" to get it handled quickly, as it really is not much, just a little bit of logic glue code is required, and the current "When physical collision" brick is very difficult to use and understand by our users, mainly by the unclear formulation and the limited functionality. Changing both the text and the code would allow to change the translations in one swoop, too. Important IMHO, and a low hanging fruit. The "you" in the text also fits well with the "When you receive" bricks etc.

catroid/src/main/res/values/strings.xml
```diff
@@ -1076,14 +1074,13 @@
...
 
     <!-- Physics -->
-    <string name="default_project_name_physics">My first physics program</string>
     <string name="brick_set_physics_object_type">Set motion type to</string>
-    <string name="brick_set_physics_object_type_dynamic">bouncing with gravity</string>
-    <string name="brick_set_physics_object_type_fixed">others bounce off it</string>
-    <string name="brick_set_physics_object_type_no">no bouncing</string>
+    <string name="brick_set_physics_object_type_dynamic">moving and bouncing under gravity</string>
+    <string name="brick_set_physics_object_type_fixed">not moving under gravity, but others bounce off you under gravity</string>
+    <string name="brick_set_physics_object_type_no">not moving or bouncing under gravity (default)</string>
```
wslany: These changes should also help in making it much clearer what to expect from these settings related to the physics engine.

catroid/src/main/res/values/strings.xml
```diff
@@ -1167,15 +1164,15 @@
...
 
 
     <plurals name="second_plural">
         <item quantity="one">"second "</item>
         <item quantity="other">"seconds "</item>
     </plurals>
 
-    <string name="brick_collision_receive">When physical collision with</string>
+    <string name="brick_collision_receive">When collision under gravity with</string>
     <string name="collision_with_anything">anything</string>
```
wslany:
```suggestion
    <string name="collision_with_anything">any edge or object of motion type \'bouncing\'</string>
```
Clearer and more precise than what we have now, which is understandably poorly understood by our users as it is.

wslany:
@robertpainsi good idea. Please use the suggest mode in these lines.

robertpainsi:
> @robertpainsi good idea. Please use the suggest mode in these lines.

Unfortunately the lines I can comment on is limited to the once shown in the [diff](https://github.com/Catrobat/Catroid/pull/3018/files). It's also not possible to comment on expanded lines :crying_cat_face: (or did I miss something?)

wslany:
OK, then just write into a comments, and we'll take it into the next iteration, alright?

catroid/src/main/res/values/strings.xml
```diff
@@ -61,16 +61,16 @@
```
robertpainsi: From https://jira.catrob.at/browse/CAT-2707

https://github.com/Catrobat/Catroid/blob/f2ba555abc296cb4f81e0a67cb7c355f30135531/catroid/src/main/res/values/strings.xml#L1571-L1574
```diff
-    <string name="brick_jumping_sumo_rotate_left">Rotate Jumping Sumo left</string>
-    <string name="brick_jumping_sumo_rotate_right">Rotate Jumping Sumo right</string>
-    <string name="brick_jumping_sumo_rotate_text">°</string>
-    <string name="brick_jumping_sumo_turn">Turn Jumping Sumo</string>
+    <string name="brick_jumping_sumo_rotate_left">Turn Jumping Sumo left</string>
+    <string name="brick_jumping_sumo_rotate_right">Turn Jumping Sumo right</string>
+    <string name="brick_jumping_sumo_rotate_text">degrees</string>
+    <string name="brick_jumping_sumo_turn">Flip Jumping Sumo</string>
```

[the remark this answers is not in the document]
wslany: Yes, exactly, that's it.

catroid/src/main/res/values/strings.xml
```diff
@@ -1078,12 +1076,12 @@
     <!-- Physics -->
     <string name="default_project_name_physics">My first physics program</string>
```
wslany: I think `default_project_name_physics` is actually not used anymore.

catroid/src/main/res/values/strings.xml
```diff
@@ -1078,12 +1076,12 @@
...
     <string name="brick_set_physics_object_type">Set motion type to</string>
```
wslany:
```suggestion
    <string name="brick_set_physics_object_type">Set your motion type to</string>
```
Makes it clear that it concerns this actor or object only, and more consistent with other brick names that order actors and objects to do certain things, e.g., `When you receive...` etc.

catroid/src/main/res/values/strings.xml
robertpainsi: Thanks a lot for your detailed explanation and documentation.

Any concerns that users will think that actors and objects are somehow different (different functionality, behavior)? → needing to explain that functionally and behaviorally speaking, they are not.

wslany: Hmm, I don't think that there need to be any concerns. Using both terms together as proposed is more inclusive, because of the limitations of the natural language and what users know. For computer science people with a background in object oriented programming, it is clear that a person can be an object, but that is not so in real life, and our users only have a background in real life, but not a formal education in computer science or object oriented programming. If there would be a good phrase for both actors and objects (which unfortunately "sprite" isn't), then we could use it instead. But there seems to be none. And the analogy from reality is quite clear. One aspect that may possibly be done differently at some time later (if ever) would be to have different categories in the media gallery, or to show them in different ways. But I don't think that it would warrant handling them in different ways. In many cases, it might be OK to have "objects" that are somehow personified, e.g., a "car" that at the same time is somehow "alive" and has its own "personality", like in the movie "Cars", can be seen both as an actor than as an "object". Also, since the order in the list of actors and objects implies the default layer, it is important to be able to mix actors and objects in the list, so they cannot be kept separate in different lists.

catroid/src/main/res/values/strings.xml
```diff
@@ -61,16 +61,16 @@
 
     <string name="backpack_add">Add to backpack</string>
 
-    <string name="share_website_text">www.pocketcode.org</string>
+    <string name="share_website_text">share.catrob.at</string>
```
wslany: Actually, I don't think that such urls should be in strings.xml, since these strings can all be translated for all human languages that we support. Not a good idea to have different "translations" for this url.

wslany: Actually, I don't think that such urls should be in strings.xml, since these strings can all be translated for all human languages that we support. Not a good idea to have different "translations" for this url.

catroid/src/main/res/values/strings.xml
```diff
@@ -712,19 +712,19 @@
...
+    <string name="global_vars_headline">Variables for all actors and objects</string>
+    <string name="local_vars_headline">Variables for this actor or object only</string>
+    <string name="global_lists_headline">Lists for all actors and objects</string>
+    <string name="local_lists_headline">Lists for this actor or object only</string>
 
 
     <!-- NewDataDialog -->
-    <string name="new_data_for_all_sprites">For all objects</string>
-    <string name="new_data_for_this_sprite_only">For this object only</string>
+    <string name="new_data_for_all_sprites">For all actors and objects</string>
```
wslany:
```suggestion
    <string name="new_data_for_all_sprites">For all actors, objects, and clones in all scenes</string>
```

catroid/src/main/res/values/strings.xml
```diff
@@ -712,19 +712,19 @@
...
+    <string name="new_data_for_this_sprite_only">For this actor or object only</string>
```
wslany:
```suggestion
    <string name="new_data_for_this_sprite_only">For this actor, object, or one of its clones only</string>
```

catroid/src/main/res/values/strings.xml
```diff
@@ -84,27 +84,27 @@
     <!--  -->
 
     <!-- Main Menu -->
-    <string name="main_menu_continue">Continue</string>
-    <string name="main_menu_new">New</string>
-    <string name="main_menu_programs">Programs</string>
+    <string name="main_menu_continue">Continue project</string>
+    <string name="main_menu_new">New project</string>
+    <string name="main_menu_programs">Projects on phone</string>
```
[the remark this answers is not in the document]
wslany:
```suggestion
    <string name="main_menu_programs">Projects on device</string>
```
After rethinking this, "device" after all is a better fitting, more general name here. There are more devices supporting Android already now (watches, TVs, cameras, cars, etc) --- and who knows what will come next. Also, the telephone aspect is only one of many even when considering only smartphones.

catroid/src/main/res/values/strings.xml
```diff
@@ -1678,57 +1676,57 @@
...
     <string name="formula_editor_function_ceil_parameter">(0.3)</string>
-    <string name="formula_editor_function_max">max</string>
-    <string name="formula_editor_function_max_parameter">(0,1)</string>
-    <string name="formula_editor_function_min">min</string>
-    <string name="formula_editor_function_min_parameter">(0,1)</string>
+    <string name="formula_editor_function_max">maximum of</string>
+    <string name="formula_editor_function_max_parameter">(5,4)</string>
+    <string name="formula_editor_function_min">minimum of</string>
+    <string name="formula_editor_function_min_parameter">(7,2)</string>
     <string name="formula_editor_function_true">TRUE</string>
```
wslany:
```suggestion
    <string name="formula_editor_function_true">true</string>
```

catroid/src/main/res/values/strings.xml
```diff
@@ -1678,57 +1676,57 @@
...
     <string name="formula_editor_function_false">FALSE</string>
```
wslany:
```suggestion
    <string name="formula_editor_function_false">false</string>
```
There's no need for non-standard capitalization.

Similarly, again because there's no need for non-standard capitalization, replace https://github.com/Catrobat/Catroid/blob/e538e83152435fb69e2f618876eee7eb4cfa23e0/catroid/src/main/res/values/strings.xml#L1870-L1872 by

     <string name="formula_editor_logic_and">and</string>
     <string name="formula_editor_logic_or">or</string>
     <string name="formula_editor_logic_not">not</string>

catroid/src/main/res/values/strings.xml
```diff
@@ -1306,18 +1304,18 @@
 
     <!-- Notifications -->
     <string name="notification_screenshot_ok">Preview image saved</string>
-    <string name="notification_download_project_cancel">Canceled program download</string>
-    <string name="notification_reconvert_download_program_cancel">Canceled downloading/reconverting program</string>
+    <string name="notification_download_project_cancel">Canceled project download</string>
```
wslany:
```suggestion
    <string name="notification_download_project_cancel">Cancel project download</string>
```
This is used as a tappable text in a dialog (overwrite or rename downloaded project) that allows to cancel the download.

catroid/src/main/res/values/strings.xml
wslany:
```suggestion
    <string name="brick_collision_receive">When you bounce off</string>
```
Even better and more precise. Please use this. The functionality also will be slightly adapted, but this string is already better than what we have had so far.

catroid/src/main/res/values/strings.xml
```diff
@@ -1737,36 +1735,38 @@
...
     <string name="formula_editor_function_contains">contains</string>
     <string name="formula_editor_function_contains_parameter">(*list name*,1)</string>
-    <string name="formula_editor_function_collision">touches_object</string>
-    <string name="formula_editor_function_collides_with_edge">touches_edge</string>
-    <string name="formula_editor_function_touched">touches_finger</string>
-    <string name="formula_editor_sensor_x_acceleration">acceleration_x</string>
-    <string name="formula_editor_sensor_y_acceleration">acceleration_y</string>
-    <string name="formula_editor_sensor_z_acceleration">acceleration_z</string>
-    <string name="formula_editor_sensor_compass_direction">compass_direction</string>
+    <string name="formula_editor_function_number_of_items">number of elements</string>
```
wslany:
```suggestion
    <string name="formula_editor_function_number_of_items">number of items</string>
```
Sorry for the back and forth, but I checked once more, and in Scratch it is called "items", and also we decided in 2014 to rename the "element" function to "item", see https://jira.catrob.at/browse/CAT-2374

catroid/src/main/res/values/strings.xml
```diff
@@ -1737,36 +1735,38 @@
...
+    <string name="formula_editor_function_finger_x">screen touch x</string>
+    <string name="formula_editor_function_finger_y">screen touch y</string>
+    <string name="formula_editor_function_is_finger_touching">screen is touched</string>
+    <string name="formula_editor_function_multi_finger_x">screen touch x</string>
+    <string name="formula_editor_function_multi_finger_y">screen touch y</string>
+    <string name="formula_editor_function_is_multi_finger_touching">screen is touched</string>
+    <string name="formula_editor_function_index_of_last_finger">last screen touch index</string>
     <string name="formula_editor_function_no_parameter"></string>
     <string name="formula_editor_function_touch_parameter">(1)</string>
     <string name="formula_editor_function_list_item">element</string>
```
wslany:
```suggestion
    <string name="formula_editor_function_list_item">item</string>
```
I checked once more, and in Scratch it is called "item", and also we decided in 2014 to rename it to "item", see https://jira.catrob.at/browse/CAT-2374

catroid/src/main/res/values/strings.xml
thmq: It actually is.

catroid/src/main/res/values/strings.xml
thmq: this is not the url, it is only the display text of the url, so it does make sense to have it as a string resource, although we could mark it as untranslatable in crowdin.

catroid/src/main/res/values/strings.xml
```diff
@@ -966,34 +964,34 @@
...
+    <string name="pocket_music_default_project_name">Untitled song</string>
     <string name="pocket_music_default_track_name">Track 1</string>
     <!--  -->
 
 
     <!-- PreStage activity -->
     <string name="prestage_text_to_speech_engine_not_installed">Text-to-Speech engine not installed!\nDo you want to install it?</string>
     <string name="prestage_resource_not_available_title">Component not available</string>
-    <string name="prestage_resource_not_available_text">Unfortunately this program does not work on your device.
+    <string name="prestage_resource_not_available_text">Unfortunately this project does not work on your phone.
```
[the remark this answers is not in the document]
thmq: In accordance with your proposals here I'm going to change all instances of phone to device.

catroid/src/main/res/values/strings.xml
thmq: yes, I'll look into it.

catroid/src/main/res/values/strings.xml
```diff
@@ -679,14 +679,14 @@
     <!--  -->
 
     <!-- SpriteListFragment -->
-    <string name="fragment_sprite_text_description">Tap \"+\" to add objects</string>
+    <string name="fragment_sprite_text_description">Tap \"+\" to add actors or objects</string>
     <string name="default_sprite_name">My object</string>
```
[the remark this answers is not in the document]
thmq: that's not your fault :wink: it wasn't there before the rebase.

catroid/src/main/res/values/strings.xml
```diff
@@ -1195,7 +1193,7 @@
     <string name="brick_forever">Forever</string>
     <string name="brick_repeat">Repeat</string>
     <string name="brick_repeat_until">Repeat until</string>
-    <string name="brick_when_touched">When screen is touched</string>
+    <string name="brick_when_touched">When screen is tapped</string>
```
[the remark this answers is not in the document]
thmq: yes :smirk_cat:

wslany:
One more capitalization to lower case missed, sorry:

    <string name="formula_editor_list_dialog_title">New List</string>
to
    <string name="formula_editor_list_dialog_title">New list</string>

wslany:
Regarding 

    <string name="brick_set_physics_object_type_fixed">not moving under gravity, but others bounce off you under gravity</string>
the string is to long to fit into one line on all screens, but it's not really possible to shorten it without making it difficult to fully understand. Would there be a way to break the line where it appears? 

See screenshot below:
screenshot_20190117-051017

wslany:
Some more missed capitalizations to lower case:

    <string name="brick_go_to_touch_position">Touch position</string>
    <string name="brick_go_to_random_position">Random position</string>
to

    <string name="brick_go_to_touch_position">touch position</string>
    <string name="brick_go_to_random_position">random position</string>

and 

    <string name="color_red">Red</string>
    <string name="color_green">Green</string>
    <string name="color_blue">Blue</string>
    <string name="color_preview">Color Preview</string>
to

    <string name="color_red">red</string>
    <string name="color_green">green</string>
    <string name="color_blue">blue</string>
    <string name="color_preview">Color preview</string>

One more thing in the line preceding that block:

    
This concerns also the pen bricks, so better change to:

    

All the pull down / spinner menus in bricks for extensions (Lego etc) should also start with lower case. I can't list them all here (I'm on the phone only 😉), so please search for them, there are many, but here's an example:

    <string name="phiro_motor_left">Left</string>
to

    <string name="phiro_motor_left">left</string>

Also, in all bricks where "X" or "Y" appear in upper case letters, please change them to lower case, for example (there are several others):

    <string name="x_label">"X: "</string>
    <string name="y_label">"Y: "</string>
    <string name="brick_set_x">"Set X to "</string>
to

    <string name="x_label">"x: "</string>
    <string name="y_label">"y: "</string>
    <string name="brick_set_x">"Set x to "</string>

wslany:
One more:

    <string name="new_option">New…</string>
to

    <string name="new_option">new…</string>

I _believe_ I caught everything. At any rate, I'll stop now. It's early/late... 😂

wslany:
> > Regarding
> > ```
> > <string name="brick_set_physics_object_type_fixed">not moving under gravity, but others bounce off you under gravity</string>
> > ```
> > the string is too long to fit into one line on all screens, but it's not really possible to shorten it without making it difficult to fully understand. Would there be a way to break the line where it appears?
> > See screenshot below:
> > screenshot_20190117-051017
> 
> we cannot modify how a drop down text view (aka spinner) displays text in Android.

How about splitting that text line into two consecutive spinner entries, the first with ... at the end and the second with ... at the start, and when the user selects any one of them, internally it's the same choice, and in the scripts only the first string is shown?

thmq:
> > > Regarding
> > > ```
> > > <string name="brick_set_physics_object_type_fixed">not moving under gravity, but others bounce off you under gravity</string>
> > > ```
> > > the string is too long to fit into one line on all screens, but it's not really possible to shorten it without making it difficult to fully understand. Would there be a way to break the line where it appears?
> > > See screenshot below:
> > > screenshot_20190117-051017
> > 
> > 
> > we cannot modify how a drop down text view (aka spinner) displays text in Android.
> 
> How about splitting that text line into two consecutive spinner entries, the first with ... at the end and the second with ... at the start, and when the user selects any one of them, internally it's the same choice, and in the scripts only the first string is shown?

please don't make me (or anyone else) do this. this is going to cause so many problems, starting from how to handle it with regard to translations (explaining it to the translators, finding a solution for languages where this string is even longer, or shorter) to issues with testing.

if you really insist on the option to contain that much information, then we should perhaps try to find a different way of displaying it (in some other PR though), such as a dialog or any other means that allows for longer texts. fixing it this way is going to cost us later, there are many bug tickets that started just with a situation like this.

wslany:
> > > > Regarding
> > > > ```
> > > > <string name="brick_set_physics_object_type_fixed">not moving under gravity, but others bounce off you under gravity</string>
> > > > ```
> > > > the string is too long to fit into one line on all screens, but it's not really possible to shorten it without making it difficult to fully understand. Would there be a way to break the line where it appears?
> > > > See screenshot below:
> > > > [...] 
> > > 
> > > 
> > > we cannot modify how a drop down text view (aka spinner) displays text in Android.
> > 
> > 
> > How about splitting that text line into two consecutive spinner entries, the first with ... at the end and the second with ... at the start, and when the user selects any one of them, internally it's the same choice, and in the scripts only the first string is shown?
> 
> please don't make me (or anyone else) do this. this is going to cause so many problems, starting from how to handle it with regard to translations (explaining it to the translators, finding a solution for languages where this string is even longer, or shorter) to issues with testing.
> 
> if you really insist on the option to contain that much information, then we should perhaps try to find a different way of displaying it (in some other PR though), such as a dialog or any other means that allows for longer texts. fixing it this way is going to cost us later, there are many bug tickets that started just with a situation like this.

It's OK, so let's then relay that to a new ticket. Probably it's fine to fix it at the same time as CATROID-130, too which it is connected.

thmq:
> Just found this (missing space, and still capital letters) --- I assume that part of it has something to do with the issue that we use the same strings and partly the same code for Phiro LEDs and for the pen colors, see https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java#L114-L117 --- that's for the pen color, but it takes the strings from the Phiro resources. Historical reasons ... Anyway, at least the space must be corrected.
> 
> spc
> 
> In the Lego EV3 extension, there also seems to be a space missing here (in Catroid 0.9.44 there seems to have been a newline at that place?):
> 
> l

I'm pretty sure that this issue was not caused by this PR, but more likely by my brick refactoring PR. I must have missed it because this only occurs with devices that have a very high horizontal resolution.

PR #3023, opened 2018-11-06
 CAT-49  BUG Discard Changes

PlantinumX:
New Class named UndoState waas created for saving the initalStates(first values of the bricks) in map. The map stores all initial values of brickfields. If the  user discards the changes the old ones are used which are taken from the initialstates map, so discarding the changes over many bricks will result in an expected behaviour.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
@@ -1,0 +1,161 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.formulaeditor;
+
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.bricks.PlaceAtBrick;
+import org.catrobat.catroid.ui.SpriteActivity;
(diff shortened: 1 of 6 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
```diff
@@ -0,0 +1,161 @@
...
+	public void testGoBackToDiscardXChanges() {
+		onBrickAtPosition(brickPosition).checkShowsText(R.string.brick_place_at);
+		onBrickAtPosition(brickPosition).onFormulaTextField(R.id.brick_place_at_edit_text_x)
+				.perform(click());
+		onFormulaEditor()
+				.performEnterFormula("1234");
+		pressBack();
+		onView(withText(R.string.formula_editor_discard_changes_dialog_title))
+				.check(matches(isDisplayed()));
+		onView(withText(no))
```
dpedratscher: it irritates me, that the tests clicks 'no' = 'do not discard' and the positions are nevertheless set back to default zero. Or do I get this wrong?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorEditText.java
```diff
@@ -289,10 +292,10 @@ public boolean undo() {
 		if (!history.undoIsPossible()) {
 			return false;
 		}
-		InternFormulaState lastStep = history.backward();
+		UndoState lastStep = history.backward();
 		if (lastStep != null) {
 
-			internFormula = lastStep.createInternFormulaFromState();
+			internFormula = lastStep.internFormulaState.createInternFormulaFromState();
```
dpedratscher: Is there a better way to call this function? What if lastStep.interFormulaState == null? Maybe it's possible to provide a function in the lastStep itself, so we don't have to dig into two layers.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaEditorHistory.java
```diff
@@ -73,11 +82,11 @@ public InternFormulaState forward() {
 	}
 
 	public void updateCurrentSelection(InternFormulaTokenSelection internFormulaTokenSelection) {
-		current.setSelection(internFormulaTokenSelection);
+		current.internFormulaState.setSelection(internFormulaTokenSelection);
```
dpedratscher: Also here: What if current.internFormulaState == null? Is there a way to provide a function to directly set the selection?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
84n4n4: "no" means do not save changes.
thats ok, the ticket is about bugs when theres formulas not correctly set back to initial state, etc...

robertpainsi: https://github.com/Catrobat/Catroid/blob/b7cdf2126c624e71b8aa543288775cc9b0b24358/catroid/src/main/res/values/strings.xml#L1885
See string text/value (vs. name).

dpedratscher: Ömer and me agreed on renaming the variable 'R.string....discard...' in 'R.string....save...' because this is also the title of the dialog.

84n4n4:
please rebase and then reopen.
please also move your ticket in jira to the according state.

PR #3032, opened 2018-12-10
CATROID-63 The formula when selecting Edit Formula in the menu for bricks with multiple formulas isn't properly defined

ndrnour:
Tap on a brick with multiple formulas to open the menu and select Edit Formula. The Formula Editor opens and the "default formula" will be selected. This formula was once explicitly defined in the code, now isn't (kinda random now, e.g. the default formula of the PlaceAtBrick is Y now).

Simply use the first formula parameter, from the left.


84n4n4:
as discussed, please have following changes in FormulaBrick base class:

	public void showFormulaEditorToEditFormula(View view) {
		if (brickFieldToTextViewIdMap.inverse().containsKey(view.getId())) {
			FormulaEditorFragment.showFragment(view.getContext(), this, brickFieldToTextViewIdMap.inverse().get(view.getId()));
		} else {
			FormulaEditorFragment.showFragment(view.getContext(), this, getDefaultBrickField());
		}
	}

	protected BrickField getDefaultBrickField() {
		return formulaMap.keys().nextElement();
	}

and then just override getDefaultBrickField() in the bricks with more than one formula.

wslany:
Test this please.

wslany:
test this please

wslany:

PR #3033, opened 2018-12-11
CATROID-47-Fix SetBackgroundByIndexBrick interpret formula bug

gPathpp:


catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/SetBackgroundByIndexBrickTest.java
```diff
@@ -0,0 +1,101 @@
...
+		lookData2.setFile(imageFile2);
+		lookData2.setName("background2");
+		lookDataList.add(lookData2);
+
+		project.getDefaultScene().getBackgroundSprite().look.setLookData(lookData);
+	}
+
+	@Test
+	public void testSetBackgroundByIndex() {
+		ActionFactory actionFactory = new ActionFactory();
```
84n4n4: since this doesnt interact with the ui in any way, can this be done as headless integration test in test package?

catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/SetBackgroundByIndexBrickTest.java
```diff
@@ -0,0 +1,101 @@
...
+		LookData backgroundLookData = ProjectManager.getInstance().getCurrentlyPlayingScene().getBackgroundSprite().look.getLookData();
+		List<LookData> backgroundLookDataList = ProjectManager.getInstance().getCurrentlyPlayingScene().getBackgroundSprite().getLookList();
+
+		assertEquals(backgroundLookData, backgroundLookDataList.get((int) backgroundIndex - 1));
+	}
+
+	private Formula createFormulaWithVariable(String variableName) {
+		FormulaElement formulaElement = new FormulaElement(FormulaElement.ElementType.USER_VARIABLE, variableName, null);
+		InternToken it = formulaElement.getInternTokenList().get(0);
+		Log.e("bullshit", "Is UserVariable: " + it.isUserVariable());
```
84n4n4: no

catroid/src/main/java/org/catrobat/catroid/content/actions/SetBackgroundLookAction.java
```diff
@@ -0,0 +1,60 @@
...
+
+package org.catrobat.catroid.content.actions;
+
+import org.catrobat.catroid.ProjectManager;
+import org.catrobat.catroid.common.LookData;
+import org.catrobat.catroid.content.EventWrapper;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.content.eventids.SetBackgroundEventId;
+
+public class SetBackgroundLookAction extends EventAction{
```
84n4n4: i have no idea why lint isnt complaining, but "EventAction{" should have a space inbetween EventAction and the bracket

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundByIndexBrick.java
```diff
@@ -23,38 +23,47 @@
...
 
-public class SetBackgroundByIndexBrick extends SetLookByIndexBrick {
+import static org.catrobat.catroid.content.EventWrapper.NO_WAIT;
+
+public class SetBackgroundByIndexBrick extends FormulaBrick {
+
+	private static final long serialVersionUID = 1L;
+
+	@EventWrapper.WaitMode
+	protected transient int wait = NO_WAIT;
```
thmq: this is not necessary. you only use the member `wait` once in the entire class, it is not changed anywhere. just use the value directly:

```@Override
	public List<ScriptSequenceAction> addActionToSequence(Sprite sprite, ScriptSequenceAction sequence) {
		sequence.addAction(sprite.getActionFactory()
				.createSetLookByIndexAction(getSprite(), getFormulaWithBrickField(BrickField.LOOK_INDEX), wait));
				.createSetBackgroundLookByIndexAction(sprite, getFormulaWithBrickField(BrickField.LOOK_INDEX), NO_WAIT));

		return Collections.emptyList();
	}
```

and please have a look if it is used the same way in another brick and remove it there as well. otherwise this is just cargo cult.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundByIndexAndWaitBrick.java
```diff
@@ -32,6 +33,11 @@ public SetBackgroundByIndexAndWaitBrick() {
 		wait = EventWrapper.WAIT;
```
thmq: see my comment in  `SetBackgroundByIndexBrick.java`, but in short -> I think it is preferable to override the `addActionToSequence(..)` method here instead of using the member variable. it just obfuscates the code.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundAndWaitBrick.java
```diff
@@ -36,17 +37,18 @@
 
 public class SetBackgroundAndWaitBrick extends SetBackgroundBrick {
 
-	public SetBackgroundAndWaitBrick() {
-	}
+	public SetBackgroundAndWaitBrick() { }
```
thmq: this is not correctly formatted. I wonder what's going on with our static checks here, how did they miss it?

thmq:
@gPathpp please rebase this. Although it has no merge conflicts with the develop branch, it does not combine the code correctly (I tried it locally). a lot of the changes to the bricks in #3021 do not go well with what you have here. so please be sure to not only rebase, but have a look if everything works correctly and if your changes look the way you want them to. 

and don't forget to also consider the comments by @84n4n4. 

closing this pr for now, please reopen when you're done.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetBackgroundLookAction.java
robertpainsi: Me neither!
However, looking through the documentation, there is no rule or setting to force this :flushed:

…actually, there is one. By adding left and right curly braces to [WhitespaceAround](http://checkstyle.sourceforge.net/config_whitespace.html#WhitespaceAround). This will also result in a violation for e.g.
```java
public @interface RotationStyle {}
```

which however is already a violation according to the Android Studio formatter. I like this style and tried to adapt the AS formatter a long time ago without success. The changes always effected other code formats as well…

So, btt. I'll do some additional research here. Maybe write a GitHub issue.

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetBackgroundAndWaitBrick.java
robertpainsi: Looks like a bug to me → reported

robertpainsi:
@84n4n4, @thmq, @Catrobat/catroid, @__all__, please ping me if there are false positive or false negative Checkstyle violations :smiley:

PR #3039, opened 2018-12-19
CATROID-108 Fix "When becomes true" looses formula

robertpainsi:
WhenConditionBrick has a ConcurrentFormulaHashMap because it's extended
from Formula. WhenConditionScript has a ConcurrentFormulaHashMap because
the map is serialized. To work correctly, both members have to point to
the same instance.

When cloning a WhenConditionBrick, these two members pointed to
different ConcurrentFormulaHashMap thus causing CATROID-108.
___
- You can find the same info but a little more detailed (or phrased differently) at [CATROID-108](https://jira.catrob.at/browse/CATROID-108).
- Feel free to comment(criticize) on the test implementation :wink: 

Please consider a fix for [CATROID-108](https://jira.catrob.at/browse/CATROID-108) for the next release :grimacing:


robertpainsi:
Ah, wait. I didn't think about writing an UI test for that instead… Hmm… @84n4n4 activity switching would be a problem here, right?

robertpainsi:
And just retested it. You have to go back to the MainMenuActivity. Switching to another Sprite won't work to see the effects of the bug.

wslany:

PR #3041, opened 2018-12-20
 [CAT-49] BUG discarding changes

PlantinumX:
BUG created UndoState class for saving the initalStates in map.The stores all inital values of brickfields so if user discards the changes the old ones are used which are taken from the initialstates map.


84n4n4:
please rename commit message according to commit guidelines
please move ticket in jira to correct state

PR #3043, opened 2018-12-20
CATROID-119 use applicationIdSuffix to distinguish flavors in gradle

84n4n4:


84n4n4:
@mfuchs @LPeteR90 @thmq can you please have a look at this?

PR #3051, opened 2019-01-04
CATROID-115 - fix import from external storage

84n4n4:


wslany:
test this please

wslany:
test this please

wslany:
I just tried it with the apk from jenkins run #10 of the PR, on a device with an existing "My first program" project in a "Pocket Code" folder on the external memory. The new name was correct, and there was only one "My first program", but the looks of the clouds were missing, see the screenshot on https://jira.catrob.at/secure/attachment/15728/1.png

84n4n4:
this is now technically working - BUT - the workflow / user workflow for the import should be revised with further pocket code versions and rollouts in mind.

the import dialog is currently not cancellable - this means that if some user chooses to "copy" and then clears the shared preferences somehow, the dialog will be shown again, not beeing dismissable, and both copy and move overwriting any changes he made to his projects in apps storage in the meantime.
is the "copy" option really necessary? theres another story already in development for export, and leaving those files in place, probably puts the user in a spot where a future update of pocketcode resets the shared preferences again and shows him an indismissable dialog that has a high chance of deleting some of his progress on any selection.

i would suggest,
- "move" moves projects over deleting the pocket code folder in storage.
- "cancel" dismisses the dialog.
- "never show again" checkbox or popup when cancelled.

also either the check in the import functionality that checks if there are projects to import just looks for the "Pocket Code" folder on the sd card and if present shows the dialog.
this should either check that the "PocketCode" folder is there AND that it contains some more folders (therefore projects)
- or CATROID-57 changes the export directory to something else than "Pocket Code".

PO's please check above and suggest what user work flow should be implemented.

84n4n4:
superseeded by #3060

PR #3059, opened 2019-01-15
CATROID-79 Fixed wrong layer property for background

trinitydonohugh:


catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -625,7 +625,7 @@ public void setConvertToGroupItemSprite(boolean convertToGroupItemSprite) {
 	}
 
 	public boolean isBackgroundSprite() {
-		return look.getZIndex() == Constants.Z_INDEX_BACKGROUND;
+		return this == ProjectManager.getInstance().getCurrentlyEditedScene().getSpriteList().get(0);
```
thmq: this is potentially going to cause problems. e.g. when in the stage the scene that is running now might not be the currently edited scene and consequently this could change the behaviour there. besides it is very likely that there are going to be situations where the `ProjectManager` might not have the currently edited scene set at all which again would lead to incorrect behaviour. 

I suggest we have another look at it.

84n4n4:
please dont merge.
we just started working on CATROID-89 and found that its essentially the same ticket, but covering more.

84n4n4:
closed because superseded by #3065

PR #3061, opened 2019-01-16
CATROID-134 background name not translated

Fatmangh:


wslany:
test this please

wslany:
"Background" now is renamed, but only in the list of background and objects. However, as soon as you enter the background itself, in its overview with the buttons for Scripts, Looks, and Sounds, the top navigation bar again has the name of the background in the language from the original project source. Same also inside the scripts, looks, and sounds: The top navigation bar still contains the original language name. 

See images with German set as interface language but a project made by a (I guess) Russian kid: The displayed name should instead have been "Hintergrund" in the top navigation bar.

bg

bg2

PR #3068, opened 2019-01-19
CATROID-107 No keyboard for dialogs with text-input field.

ndrnour:
* since all of the dialogs with editText inherit TextInputDialog.java (thanks to @thmq ), one line of code is enough to fix this.
* I couldn't find a way to test if the keyboard is shown! if you have one, please share with me.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/TextInputDialog.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/dialog/TextInputDialog.java
@@ -31,6 +31,7 @@ import android.support.annotation.Nullable;
 import android.support.design.widget.TextInputLayout;
 import android.support.v7.app.AlertDialog;
 import android.text.Editable;
+import android.view.WindowManager;
 
 import org.catrobat.catroid.R;
 
@@ -84,6 +85,7 @@ public final class TextInputDialog extends AlertDialog {
 		@Override
 		public AlertDialog create() {
 			final AlertDialog alertDialog = super.create();
+			alertDialog.getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE);
 
 			alertDialog.setOnShowListener(new OnShowListener() {
 				@Override


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/ShowSoftKeyboardForNewDialogTest.java
```diff
@@ -0,0 +1,92 @@
...
+				.commit();
+	}
+
+	@Test
+	public void testSoftKeyboardIsShownWhenDialogIsShown() {
+		PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getTargetContext())
+				.edit()
+				.putBoolean(AGREED_TO_PRIVACY_POLICY_SETTINGS_KEY, true)
+				.commit();
+		baseActivityTestRule.launchActivity();
```
84n4n4: since there is no way of asserting anything on the keyboard with espresso, either assert it via the UI automator, or not test at all.
i can see this test causing a lot of hard to detect problems if anything goes wrong.

84n4n4:
this causes some layout issues with dialogs that have bigger layouts, some may be cut off in such a way that it is not clear to the user that there are more options / things to do on the dialog then initially visible.

scrolling in dialogs seems to only kick in if there is a certain amount of the view covered, leading to some beeing parially cut but not scrollable yet.

this method also automatically scrolls the dialog to the text input field, bigger dialogs like the program download rename dialog will then not have the explanation of whats going on and the other options visible on screen initially.

we should look into those issues more carefully, since afaik android suggests not to automatically open soft keyboards initially for dialogs.
screenshot_1548563120
capture _2019-01-26-19-51-25
capture _2019-01-26-20-03-15

84n4n4:
@thmq  please have a loot at this too.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/activity/ShowSoftKeyboardForNewDialogTest.java
ndrnour: deleted

ndrnour:
could you have a look again, I've adjusted the paddings of the Dialog's elements
screenshot_20190131-003305
screenshot_20190131-005043

ndrnour:
53 failed tests are shown after 13 builds! interesting *_*

jddeep:
@84n4n4 I am willing to contribute on Android projects for Catrobat. How to proceed?

PR #3069, opened 2019-01-22
CATROID-65 Support iOS and other platform version formats in XmlHeader

robertpainsi:
The Android platform version is an integer and has been stored in the
XmlHeader as double. However, the iOS platform version has a format like
MAJOR.MINOR.PATCH[1]. Programs with such a format can't be loaded by
Catroid.

XStream doesn't store any information about the type of the platform
version in the XmlHeader, e.g. <platformVersion>26.0</platformVersion>.
Simply changing the type from double to String is sufficient to allow
any version format.

[1] https://en.wikipedia.org/wiki/IOS_version_history
___
As you can see, no test. Once you verified everything works as expected (I verified a few random programs and also one with `<platformVersion>26.0.1 beta</platformVersion>`) , testing this is like testing XStream, imo.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/content/project/ProjectTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/content/project/ProjectTest.java
@@ -23,6 +23,7 @@
 package org.catrobat.catroid.test.content.project;
 
 import android.content.pm.PackageInfo;
+import android.os.Build;
 import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
 
@@ -46,6 +47,7 @@ public class ProjectTest {
 	private static final float OLD_LANGUAGE_VERSION = 0.8f;
 	private static final String OLD_APPLICATION_NAME = "catty";
 	private static final String OLD_PLATFORM = "iOS";
+	private static final String OLD_PLATFORM_VERSION = "1.0.0 beta";
 
 	@Test
 	public void testVersionName() throws Exception {
@@ -109,11 +111,13 @@ public class ProjectTest {
 		header.setCatrobatLanguageVersion(OLD_LANGUAGE_VERSION);
 		header.setApplicationName(OLD_APPLICATION_NAME);
 		header.setPlatform(OLD_PLATFORM);
+		header.setPlatformVersion(OLD_PLATFORM_VERSION);
 
 		project.setDeviceData(InstrumentationRegistry.getTargetContext());
 
 		assertEquals(Constants.CURRENT_CATROBAT_LANGUAGE_VERSION, header.getCatrobatLanguageVersion());
 		assertEquals(InstrumentationRegistry.getTargetContext().getString(R.string.app_name), header.getApplicationName());
 		assertEquals(Constants.PLATFORM_NAME, header.getPlatform());
+		assertEquals(String.valueOf(Build.VERSION.SDK_INT), header.getPlatformVersion());
 	}
 }
--- a/catroid/src/main/java/org/catrobat/catroid/content/Project.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Project.java
@@ -277,7 +277,7 @@ public class Project implements Serializable {
 	public void setDeviceData(Context context) {
 		// TODO add other header values
 		xmlHeader.setPlatform(Constants.PLATFORM_NAME);
-		xmlHeader.setPlatformVersion((double) Build.VERSION.SDK_INT);
+		xmlHeader.setPlatformVersion(String.valueOf(Build.VERSION.SDK_INT));
 		xmlHeader.setDeviceName(Build.MODEL);
 
 		xmlHeader.setCatrobatLanguageVersion(Constants.CURRENT_CATROBAT_LANGUAGE_VERSION);
--- a/catroid/src/main/java/org/catrobat/catroid/content/XmlHeader.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/XmlHeader.java
@@ -60,7 +60,7 @@ public class XmlHeader implements Serializable {
 	private String applicationVersion = "";
 	private String deviceName = "";
 	private String platform = "";
-	private double platformVersion = 0;
+	private String platformVersion = "";
 	@SuppressWarnings("unused")
 	private String tags = "";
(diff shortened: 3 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/content/Project.java
```diff
@@ -277,7 +277,7 @@ public void setCatrobatLanguageVersion(float catrobatLanguageVersion) {
 	public void setDeviceData(Context context) {
 		// TODO add other header values
 		xmlHeader.setPlatform(Constants.PLATFORM_NAME);
-		xmlHeader.setPlatformVersion((double) Build.VERSION.SDK_INT);
+		xmlHeader.setPlatformVersion(String.valueOf(Build.VERSION.SDK_INT));
```
robertpainsi: Since the platform version changed from double to String, Android version 26 will be stored as `<platformVersion>26</platformVersion>`, not as previously `<platformVersion>26.0</platformVersion>` (trailing zero!)

catroid/src/androidTest/java/org/catrobat/catroid/test/content/project/ProjectTest.java
```diff
@@ -109,11 +111,13 @@ public void testSetDeviceData() {
...
 		header.setApplicationName(OLD_APPLICATION_NAME);
 		header.setPlatform(OLD_PLATFORM);
+		header.setPlatformVersion(OLD_PLATFORM_VERSION);
 
 		project.setDeviceData(InstrumentationRegistry.getTargetContext());
 
 		assertEquals(Constants.CURRENT_CATROBAT_LANGUAGE_VERSION, header.getCatrobatLanguageVersion());
 		assertEquals(InstrumentationRegistry.getTargetContext().getString(R.string.app_name), header.getApplicationName());
 		assertEquals(Constants.PLATFORM_NAME, header.getPlatform());
+		assertEquals(String.valueOf(Build.VERSION.SDK_INT), header.getPlatformVersion());
```
robertpainsi: This test doesn't really test the change, but it was still missing imo. So i sneaked it in here :scream:

catroid/src/main/java/org/catrobat/catroid/content/Project.java
wslany: Will that make any problems with old projects, or new ones?

robertpainsi: > Will that make any problems with old projects, or new ones?

No, old Pocket Code apps will just store `26.0` whereas new ones will store `26`.
All Pocket Code apps are able to parse `26.0` and `26` since both can be parsed as Double or String.
___
More detailed explanation: The error message mentioned in CATROID-65
```
org.catrobat.catroid E/BackwardCompatibleCatrobatLanguageXStream: Conversion error For input string: "25a" : For input string: "25a"
    ---- Debugging information ----
    message             : For input string: "25a"
    cause-exception     : java.lang.NumberFormatException
    cause-message       : For input string: "25a"
    class               : java.lang.Double
    required-type       : java.lang.Double
    converter-type      : com.thoughtworks.xstream.converters.SingleValueConverterWrapper
    wrapped-converter   : com.thoughtworks.xstream.converters.basic.DoubleConverter
    path                : /program/header/platformVersion
    line number         : 16
    class[1]            : org.catrobat.catroid.content.XmlHeader
    converter-type[1]   : com.thoughtworks.xstream.converters.reflection.ReflectionConverter
    class[2]            : org.catrobat.catroid.content.Project
    version             : 0.0
    -------------------------------
```

So XStream uses [com.thoughtworks.xstream.converters.basic.DoubleConverter.java#L26-L29](https://github.com/x-stream/xstream/blob/a925ad0642f7bd6e188e27d61c6e7123dff119ad/xstream/src/java/com/thoughtworks/xstream/converters/basic/DoubleConverter.java#L26-L29) to parse the `platformVersion` as double
```java
@Override
public Object fromString(final String str) {
    return Double.valueOf(str);
}
```

When changing the `platformVersion` type from double to String, the `DoubleConverter` isn't used anymore, of course, and the parsed String is directly stored.

robertpainsi:
> do we have to do anything regarding backwards compatibility?

Good call! 
I somehow think we have to increase the `catrobatLanguageVersion` because for now (with the current catrobat language version), iOS has to use numbers. If the `catrobatLanguageVersion` has been increased, Catroid is able to parse Strings as `platformVersion`  and therefore Catty can store Strings as `platformVersion` .

Will the `catrobatLanguageVersion` be increase for the next release anyways?

catroid/src/main/java/org/catrobat/catroid/content/Project.java
robertpainsi: Just to Highlight: Once the `platformVersion` content isn't a parseable double anymore, e.g. Catty changes the format to MAJOR.MINOR.PATCH, old Pocket Code apps won't be able to open such programs.

When Catty should use a non numeric format like MAJOR.MINOR.PATCH all depends on how many and quickly users update Catroid to the latest version but also how many Catty programs are uploaded to the share. Since I don't have this data, I can't say when Catty should change their `platformVersion` format. However, the sooner this pull request is released, the sooner Catty can adapt their `platformVersion` to a non numeric value.

Summarized:
* Old and new Catroid apps have no problems loading programs created by an old or new Catroid app.
* Old Catroid apps will have problems once the `platformVersion` isn't a parseable double anymore.

wslany:
> > do we have to do anything regarding backwards compatibility?
> 
> Good call!
> I somehow think we have to increase the `catrobatLanguageVersion` because for now (with the current catrobat language version), iOS has to use numbers. If the `catrobatLanguageVersion` has been increased, Catroid is able to parse Strings as `platformVersion` and therefore Catty can store Strings as `platformVersion` .
> 
> Will the `catrobatLanguageVersion` be increase for the next release anyways?

The Embroidery / Stitching extension is new, and at least the Stitch brick is already new in develop, even though we may not yet set it to visible in the next release, so the language version will be increased. Depending on what other stories will be merged, e.g., https://jira.catrob.at/browse/CATROID-23 or https://jira.catrob.at/browse/CATROID-24 or https://jira.catrob.at/browse/CATROID-25 etc, there may be even more new bricks or sensors. So yes, @thmq please increase the `catrobatLanguageVersion` for the next release.

wslany:
test this please

PR #3074, opened 2019-01-30
TEST Changed scratch converter server test

thmq:
- test now no longer fails when scratch server does not provide
   projectData.
 - removed unnecessary code from server calls.
 - optimized exception handling for server calls.


wslany:
test this please

PR #3075, opened 2019-01-30
CATROID-139 Fix second brick not being visible

aniket9465:
After first brick was added, if second brick was of ScriptBrick type then adapter.updateItems(sprite) was not being called which was fixed .


thmq:
@aniket9465 clean fix, thank you! :+1:

PR #3076, opened 2019-01-31
CATROID-92 wrong Behavior of creating new Background in WhenBackgroundChangesToBrick

ndrnour:


thmq:
already fixed with #3052 :confused:

PR #3077, opened 2019-02-02
CATROID-49 bug regarding discarding changes

PlantinumX:
Cancelling Formula Editor changes doesn't revert all formulas or spinner
selections. To solve the bug a class named UndoState was implemented for
saving the initalStates in a map.
The map stores all inital values of brickfields so if user discards
changes the old ones are used which are taken from the initialstates
map.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorDiscardChangesTest.java
@@ -1,0 +1,161 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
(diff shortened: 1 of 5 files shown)


catroid/src/main/res/values/strings.xml
```diff
@@ -1905,6 +1905,7 @@ needs read and write access to it. You can always change permissions through you
     <string name="formula_editor_list_dialog_title">New list</string>
     <string name="formula_editor_discard_changes_dialog_message">Do you want to save your changes?</string>
     <string name="formula_editor_discard_changes_dialog_title">Save changes?</string>
+    <string name="formula_editor_save_changes_dialog_title">Save changes?</string>
```
ndrnour: please just use the old string which is 
<string name="formula_editor_discard_changes_dialog_title">Save changes?</string>
, no need to have a new one.

ndrnour: please just use the old string which is 
<string name="formula_editor_discard_changes_dialog_title">Save changes?</string>
, no need to have a new one.

ndrnour:
LGTM!

PR #3079, opened 2019-02-03
Flavored library urls and url updates

wslany:


thmq:
changes look good, however, it's not good if we merge 11 commits, because, besides adding untested (all immediate commits) and potentially unstable/un-build-able (potentially any of the untested) commits. so what we need to do is squash the 11 commits into one.

PR #3081, opened 2019-02-04
STRINGS Changed import messages, escaped quotes, fixed brick strings

thmq:
- changed project import message.
 - escaped special characters.
 - removed trailing whitespaces in strings.
 - replaced trailing whitespaces with padding.


wslany:
@thmq please check my change request at https://github.com/Catrobat/Catroid/pull/3081#pullrequestreview-200197439 as discussed face to face last week. I still can't find this change in the latest version. And sorry for the long response time!

PR #3083, opened 2019-02-05
CATROID-146 Fix multiple radio button selection in profile settings

aniket9465:
when we click on the radiobutton the onClick of the parent view is not called and so to fix this I made the radio button unclickable.


84n4n4:
hi!

i added a test that does exactly this, click on the checkbox to see what happens.
please have a look at it.

now this is fine from my side, lets see how the jenkins run is doing.

wslany:
Nice catch!

PR #3087, opened 2019-02-10
CATROID-22 Highlight brick parts

shellp:
- added option "highlight brick parts" to context menu (second from the top)
- added highlighting of all brick parts of controll structure bricks 
- added small test

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
@@ -220,12 +220,17 @@ public class SpriteActivity extends BaseActivity {
 
 	@Override
 	public void onBackPressed() {
-		boolean isDragAndDropActiveInFragment = getCurrentFragment() instanceof DragAndDropInterface
-				&& ((DragAndDropInterface) getCurrentFragment()).isCurrentlyMoving();
 
-		if (isDragAndDropActiveInFragment) {
-			((DragAndDropInterface) getCurrentFragment()).cancelMove();
-			return;
+		Fragment currentFragment = getCurrentFragment();
+		if (currentFragment instanceof ScriptFragment) {
+			if (((DragAndDropInterface) currentFragment).isCurrentlyMoving()) {
+				((DragAndDropInterface) currentFragment).cancelMove();
+				return;
+			}
+			if (((ScriptFragment) currentFragment).isCurrentlyHighlighted()) {
+				((ScriptFragment) currentFragment).cancelHighlighting();
+				return;
+			}
 		}
 		if (getCurrentFragment() instanceof FormulaEditorFragment) {
 			((FormulaEditorFragment) getCurrentFragment()).promptSave();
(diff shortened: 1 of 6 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
```diff
@@ -235,6 +235,12 @@ public void onBackPressed() {
 			getSupportFragmentManager().popBackStack();
 			return;
 		}
+		if (getCurrentFragment() instanceof ScriptFragment
```
thmq: I'm pretty sure this should be before `if (getSupportFragmentManager().getBackStackEntryCount() > 0) { ..` because if there is something on the back stack pressing the back button will not cancel the highlighting. you could probably combine it with the `isCurrentlyMoving(..)` call above, because for both states you will have to be in the `ScriptFragment`.

I'd suggest something like this:
```
Fragment currentFragment = getCurrentFragment();

if (currentFragment instanceof ScriptFragment) {
    if (((DragAndDropInterface) getCurrentFragment()).isCurrentlyMoving()) {
        ((DragAndDropInterface) getCurrentFragment()).cancelMove();
        return;
    }
    if (((ScriptFragment) getCurrentFragment()).isCurrentlyHighlighted())) {
        ((ScriptFragment) getCurrentFragment()).cancelHighlighting();
        return;
    }
```

although I'm not 100 % sure if that is the correct behaviour now. can highlighting happen at the same time as drag and drop?

catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
```diff
@@ -783,6 +789,11 @@ public void onCheckedChanged(CompoundButton compoundButton, boolean checked) {
 	}
 
 	public void handlePlayButton(View view) {
+		if (getCurrentFragment() instanceof ScriptFragment
```
thmq: as with the other suggestion I think you could unify something here. i.e. move both cases into one if block that checks if the user is in the `ScriptFragment` and then go from there. this would make it easier to read the block.

alternatively you could also just introduce a `boolean` variable for the condition `getCurrentFragment() instanceof ScriptFragment && ((ScriptFragment) getCurrentFragment()).isCurrentlyHighlighted()` so that both conditions stay separate, but look similar. 

you can do whichever you prefer :wink: 

```
if (isScriptFragment) {
    if (areBricksHighlighted) {
        ...
    }
    if (dragAndDropActive) {
        ...
    }
}
```
or 
```
if (areBricksHighlightedInFragment) {
    ...
}
if (dragAndDropActiveInFragment) {
    ...
}
```

catroid/src/main/java/org/catrobat/catroid/ui/fragment/ScriptFragment.java
```diff
@@ -498,6 +517,9 @@ public void onClick(DialogInterface dialog, int which) {
 	}
 
 	private void handleContextMenuItemClick(int itemId, BrickBaseType brick, int position) {
+		if (listView.isCurrentlyHighlighted()) {
```
thmq: this should not be necessary right? because once you open the context menu the highlighting should have been removed already. I understand that it sometimes makes sense to be careful and add extra checks for certain states that you think should not be possible. however, I think in this case you should be sure if this can happen or not.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/adapter/BrickAdapter.java
```diff
@@ -273,6 +273,15 @@ public boolean removeItems(List<BrickBaseType> items) {
 		return false;
 	}
 
+	public List<Integer> getPositionsOfItems(List<Brick> bricks) {
```
thmq: I'm not entirely sure if this is the best way to do it. I'll have to look into how you implemented the whole drawing methods in more detail. 

I'm not saying this is wrong, I just want to check if there are any performance issues with the fragment overdraw. I'll get back to you about it.

catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
shellp: highlighting and drag and drop at the same time should not be possible since tapping on something (when highlighting is active) stops the highlighting of the brick/bricks and executes the tapped.

thmq: ok, nice.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/adapter/BrickAdapter.java
thmq: seems ok.

84n4n4:
+1

wslany:
Nice!

PR #3088, opened 2019-02-11
CATROID-62 Offer to open .catrobat files in installed Catroid flavors

rschluesselbauer:
* Remove http & https intent filters that presumably did nothing
* Add working intent filters for file & content scheme and handle them in MainActivity
* Create ProjectImportFromArchiveTask (unfortunately it was not possible to reuse the ImportTask, because we need to parse *.catrobat files first to get the real project name & in case the project name changes (if it is not unique), we need to parse the project again, set a new name and save it. If we do not parse it again we lose looks, sounds, ...)
* Espresso Test testing the behaviour with an intent


rschluesselbauer:
About opening the imported project, I did not implement this.
If we get to the ProjectActivity through the intent filter, this means that MainActivity is not on the stack and by pressing back we would leave the app instead of going to main menu.

If this behaviour is desired I can implement it that way.
If this is not desired, I need some input on how to do it properly.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -362,6 +371,38 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+							if (success) {
+								ToastUtil.showSuccess(MainMenuActivity.this, R.string.project_successful_imported_toast);
+								Log.d("IMPORT_PROJECT", "Project imported successfully");
+							} else {
+								ToastUtil.showError(MainMenuActivity.this, R.string.error_during_importing_projects_toast);
+								Log.e("IMPORT_PROJECT", "Project import failed");
+							}
+						}
+					};
+					new ProjectImportFromArchiveTask(MainMenuActivity.this, importListener).execute(projectStream).get();
```
84n4n4: get() is blocking and will lead this to wait for the task to complete. so if i understand this correctly, that makes that async task sync to the main thread, so while its executed in another thread, the main thread is stuck here until its finished, therefore defying the async in asynctask.
https://developer.android.com/reference/android/os/AsyncTask.html#get()

catroid/src/main/java/org/catrobat/catroid/io/asynctask/ProjectImportFromArchiveTask.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.catrobat.catroid.common.Constants.CODE_XML_FILE_NAME;
+import static org.catrobat.catroid.common.FlavoredConstants.DEFAULT_ROOT_DIRECTORY;
+
+public class ProjectImportFromArchiveTask extends AsyncTask<InputStream, Boolean, Boolean> {
```
84n4n4: please have a look at ProjectImportTask

catroid/src/main/AndroidManifest.xml
```diff
@@ -124,34 +126,28 @@
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data
                     android:host="*"
+                    android:mimeType="application/octet-stream"
```
84n4n4: application/zip  would be more accurate

catroid/src/main/AndroidManifest.xml
```diff
@@ -124,34 +126,28 @@
...
                     android:pathPattern="@string/catroid_extension_pathPattern"
-                    android:scheme="http" />
+                    android:scheme="file" />
```
84n4n4: you can combine both schemes (file and content) in one intent filter this way:

            <intent-filter>
                <action android:name="android.intent.action.VIEW" />
                <category android:name="android.intent.category.CATEGORY_OPENABLE" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:scheme="content" />
                <data android:scheme="file" />
                <data android:mimeType="application/zip" />
                <data android:host="*" />
                <data android:pathPattern="/.*\\.catrobat" />
            </intent-filter>

catroid/src/main/AndroidManifest.xml
```diff
@@ -124,34 +126,28 @@
...
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data
                     android:host="*"
+                    android:mimeType="application/octet-stream"
                     android:pathPattern="@string/catroid_extension_pathPattern"
```
84n4n4: this is not working at all on my phone, reading up a bit on android:pathPattern i found that:
1. the filename isnt necessarily available to you when you get an URI, so it also doesnt match depending on which application you use to open (download and filemanager on android 26 is not working at all)

2. pathPattern is not performing well with subdirectories, so this will probably only match *.catrobat files in the root directory, this leads to a lot of people doing weird stuff like:
                <data android:pathPattern="/.*\\.catrobat" />
                <data android:pathPattern="/.*\\..*\\.catrobat" />
                <data android:pathPattern="/.*\\..*\\..*\\.catrobat" />
                <data android:pathPattern="/.*\\..*\\..*\\..*\\.catrobat" />
                <data android:pathPattern="/.*\\..*\\..*\\..*\\..*\\.catrobat" />

84n4n4: github is mangling the regex...

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
rschluesselbauer: I think it's probably unnecessary, that's true. But I highly doubt it's blocking the ui thread because it's executed from another async task.

catroid/src/main/java/org/catrobat/catroid/io/asynctask/ProjectImportFromArchiveTask.java
rschluesselbauer: Please give more detail, I have looked at the ProjectImportTask and thought it differs enough to justify another async task.

catroid/src/main/AndroidManifest.xml
rschluesselbauer: application/zip does not work on my Android 7.0 device

catroid/src/main/AndroidManifest.xml
rschluesselbauer: Hm so what can we do here?
If option 2 doesn't work with android 26 apps, do we leave it out completely?

catroid/src/main/AndroidManifest.xml
rschluesselbauer: Thanks will update

rschluesselbauer:
Issue was adopted by @84n4n4

PR #3089, opened 2019-02-12
CATROID-4 Create DST File

shellp:
- introduced conversion table to generate DST file
- added simple tests


catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,123 @@
...
+		context = InstrumentationRegistry.getTargetContext().getApplicationContext();
+		Project project = new Project(InstrumentationRegistry.getTargetContext(), projectName);
+		ProjectManager.getInstance().setCurrentProject(project);
+		fileGenerator = new DSTFileGenerator(context.getCacheDir());
+		file = new File(context.getCacheDir() + "/" + projectName + ".dst");
+	}
+
+	@Test
+	public void testConversionTable() {
+		for (int element = 0; element < DSTFileGenerator.CONVERSIONTABLE.length; element++) {
```
84n4n4: where is the CONVERSIONTABLE coming from in the first place? is there a reason to verify it like this over having it in a test res file and just making sure nobody changed it in the class?

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,123 @@
...
+		stitchpoints.add(new PointF(0, 0));
+		stitchpoints.add(new PointF(250, 0));
+		stitchpoints.add(new PointF(250, 250));
+		stitchpoints.add(new PointF(0, 250));
+		stitchpoints.add(new PointF(0, 0));
+
+		fileGenerator.createDSTFile(stitchpoints);
+		long dstFileLength = file.length();
+
+		assertTrue(dstFileLength >= minExpectedFileLength);
```
84n4n4: why ">="?
also asserting size is really minimalistic approach. have a comparison file in test res and compare it to the generated one would compare size and contents.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+	private float maxX;
+	private float minY;
+	private float maxY;
+	private File dstFile;
+	private String filename;
+	private FileOutputStream fileStream;
+
+	ArrayList<PointF> stitchPoints = null;
+
+	public DSTFileGenerator(File cache) {
```
84n4n4: if this always operates in cache there is no point in handing it in.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+		try {
+			dstFile.createNewFile();
+		} catch (IOException e) {
+			Log.e(TAG, Log.getStackTraceString(e));
+		}
+
+		stitchPoints = new ArrayList<>();
+	}
+
+	public File createDSTFile(ArrayList<PointF> stitchPoints) {
```
84n4n4: since this is called "createDSTFile" this should create the file and not the constructor above.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+		if (stitchPoints.size() > 1) {
+			prepareFileWriting(stitchPoints);
```
84n4n4: im generally suspicious with classes consisting only of methods without return values that have to be called in precisely the correct order while operating on its members.

eg. createDSTFile creates the file and then returns it, the file is never saved as a member, and is given as parameter to writeDSTFile which could be renamed to writeToDSTFile(File...)

why is reset necessary?

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+			writeStitches();
+			writeEndFile();
+		} catch (IllegalArgumentException ex) {
+			Log.e(TAG, "Remaining stitch distance should be zero!");
+		} catch (IOException e) {
+			Log.e(TAG, Log.getStackTraceString(e));
+		}
+	}
+
+	private void convertStitchPointsUnit(ArrayList<PointF> stitchPoints) {
```
84n4n4: when calling the method convert it  would be fitting to have it return something

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+			yDistance = (int) (nextPoint.y - currentPoint.y);
+
+			if ((xDistance < -MAXDISTANCE) || (xDistance > MAXDISTANCE)
+					|| (yDistance < -MAXDISTANCE) || (yDistance > MAXDISTANCE)) {
+				splitStitchAtIndex(index);
+			}
+		}
+	}
+
+	private void splitStitchAtIndex(int index) {
```
84n4n4: since this doesnt split the stitch path but interpolates additional stitches inbetween, please rename accordingly.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,244 @@
...
+		maxX = point.x;
+		minY = point.y;
+		maxY = point.y;
+	}
+
+	private void writeHeader() throws IOException {
+		String header = createHeaderString();
+
+		String headerFill = "";
+		for (int index = header.getBytes().length; index < HEADERMAXBYTE; index++) {
```
thmq: you could use something like

`repeated = new String(new char[n]).replace("\0", s);`

from [stackoverflow](https://stackoverflow.com/questions/1235179/simple-way-to-repeat-a-string-in-java) (obviously). :sweat_smile:

robertpainsi: Guava ftw! [Strings.repeat(String string, int count)](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Strings.html#repeat-java.lang.String-int-)

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+		InputStream inputStream = InstrumentationRegistry.getContext().getResources().openRawResource(org.catrobat
+				.catroid.test.R.raw.sample_dst_file);
+		File compareFile = StorageOperations.copyStreamToDir(inputStream, Constants.CACHE_DIR, "sample_dst_file.dst");
+
+		assertEquals(compareFile.length(), dstFile.length());
+
+		byte[] compareFileBytes = FileUtils.readFile(compareFile);
+		byte[] dstFileBytes = FileUtils.readFile(dstFile);
+
+		boolean fileBytesEquality = Arrays.equals(compareFileBytes, dstFileBytes);
```
84n4n4: please use assertArrayEquals

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+		final int expectedAmountJumpPoints = 2;
+		EmbroideryList points = new EmbroideryList();
+		points.add(new PointF(0, 0));
+		points.add(new PointF(1000, 1000));
+		ArrayList<DSTFileGenerator.StitchPoint> stitchPoints = fileGenerator.prepareStitchPoints(points);
+
+		assertEquals(points.size() + expectedAmountJumpPoints, stitchPoints.size());
+
+		int index = 0;
+		for (int i = 0; i < stitchPoints.size(); i++) {
```
84n4n4: loops and branching statements should be avoided in tests, this double index loop construction is extremely hard to read and comprehend.
why not construct a EmbroideryList points in a way that it will end up having a reasonable size after prepareStitchPoints and have an expected ArrayList<DSTFileGenerator.StitchPoint> that you compare it to.

or construct a EmbroideryListPoints points so small that you make sure that it interpolates just one point that you can assert against. and the same with jump points.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+		points.add(new PointF(100, 100));
+		points.add(new PointF(0, 0));
+
+		ArrayList<DSTFileGenerator.StitchPoint> stitchPoints = fileGenerator.prepareStitchPoints(points);
+		byte[] header = fileGenerator.getHeaderBytes(stitchPoints);
+		byte[] expectedHeader = expectedHeaderString.getBytes();
+
+		assertEquals(DSTFileGenerator.HEADERMAXBYTE, header.length);
+
+		for (int i = 0; i < DSTFileGenerator.HEADERMAXBYTE; i++) {
```
84n4n4: please no loops and branching statements in tests.
you could make the expected header also include the padding, or use substring on header and assert that first substring up to lenght matches expected and the remaining substring only consists of padding.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+		stitchpoints.add(new DSTFileGenerator.StitchPoint(new PointF(0, 0)));
+		stitchpoints.add(new DSTFileGenerator.StitchPoint(new PointF(20, 0)));
+		stitchpoints.add(new DSTFileGenerator.StitchPoint(new PointF(20, 20)));
+		stitchpoints.add(new DSTFileGenerator.StitchPoint(new PointF(0, 20)));
+		stitchpoints.add(new DSTFileGenerator.StitchPoint(new PointF(0, 0)));
+		byte[] stitchBytes = fileGenerator.getStitchesBytes(stitchpoints);
+
+		PointF lastPoint = stitchpoints.get(0).point;
+		int index = 0;
+		for (int i = 0; i < stitchBytes.length; i += 3) {
```
84n4n4: please no loops in tests, especially no duplicate indices.
you could also make this an parameterized test that just gets 2 points and does exactly the same.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,214 @@
...
+			index++;
+		}
+		assertEquals(index, stitchpoints.size());
+	}
+
+	@Test
+	public void testEndBytes() {
+		byte[] expectedEndBytes = new byte[] {0, 0, (byte) 0xF3};
+		byte[] endBytes = fileGenerator.getFileEndBytes();
+		boolean fileBytesEquality = Arrays.equals(expectedEndBytes, endBytes);
```
84n4n4: please use assertArrayEquals

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+		return jumpPoints;
+	}
+
+	private float interpolate(float endValue, float startValue, float percentage) {
+		return Math.round(endValue + percentage * (startValue - endValue));
+	}
+
+	public static class StitchPoint {
+		public PointF point;
+		public boolean jump = false;
```
rschluesselbauer: These two attributes should be final because you never write to them anyway.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+		StitchPoint currentPoint = stitchPoints.get(0);
+		ByteArray stitchBytes = new ByteArray();
+
+		for (int i = 0; i < stitchPoints.size(); i++) {
+			StitchPoint nextPoint = stitchPoints.get(i);
+			xChange = (int) (nextPoint.point.x - currentPoint.point.x);
+			yChange = (int) (nextPoint.point.y - currentPoint.point.y);
+			int xValue = (xChange < 0 ? CONVERSIONTABLE[(xChange * (-1)) + 121] : CONVERSIONTABLE[xChange]);
+			int yValue = (yChange < 0 ? CONVERSIONTABLE[(yChange * (-1)) + 121] : CONVERSIONTABLE[yChange]);
+			stitchBytes.addAll(getBytesForStitchPoint(xValue, yValue, currentPoint.jump));
```
rschluesselbauer: I think it would make sense to move getBytesForStitchPoint into the StitchPoint class and rename it accordingly (getBytes).

The calculation of a stitchpoint with xValue, yValue and currentPoint.jump could be a method in the StitchPoint class as well. Then you could have something pleasant like currentPoint.delta(nextPoint).getBytes() (I would use a name that describes what is happening here better than delta, but I don't fully understand what is going).

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+		}
+		return value;
+	}
+
+	@Test
+	public void testConversionTable() {
+		for (int element = 0; element < DSTFileGenerator.CONVERSIONTABLE.length; element++) {
+			int value = getValueForByte(DSTFileGenerator.CONVERSIONTABLE[element]);
+			if (element > 121) {
+				assertEquals((element - 121) * (-1), value);
```
84n4n4: it would be easier to understand, and way easier to see whats going wrong if something breaks, if youd also compile a expected and actual list and then compare them with arrayEquals

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+	DSTFileGenerator fileGenerator;
+	File dstFile;
+	final int[] conversion = {-81, 81, -27, 27, -9, 9, -3, 3, -1, 1};
+
+	@Before
+	public void setUp() {
+		Project project = new Project(InstrumentationRegistry.getTargetContext(), projectName);
+		ProjectManager.getInstance().setCurrentProject(project);
+		fileGenerator = new DSTFileGenerator();
+		dstFile = new File(Constants.CACHE_DIR + "/" + projectName + ".dst");
```
84n4n4: please use dstFile = new File(Constants.CACHE_DIR, projectName + ".dst") instead

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.ArrayList;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+
+@RunWith(AndroidJUnit4.class)
+public class DSTFileGeneratorTest {
+	final String projectName = "testProject";
```
84n4n4: please use DSTFileGeneratorTest.getClass().getSimpleName()
theres already 100s of tests that create "testProject" and its a mess if something goes wrong with file operations to locate the test that causes trouble.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+		stitchpoints.add(new PointF(-10, 0));
+		stitchpoints.add(new PointF(10, 0));
+		stitchpoints.add(new PointF(10, 10));
+		stitchpoints.add(new PointF(0, 15));
+		stitchpoints.add(new PointF(-10, 10));
+		stitchpoints.add(new PointF(-10, 0));
+		stitchpoints.add(new PointF(10, 10));
+		stitchpoints.add(new PointF(-10, 10));
+		stitchpoints.add(new PointF(10, 0));
+		if (dstFile.exists()) {
```
84n4n4: browsing over this testclass, it seems that this is the only test using this file, please create it in this test.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+				new DSTFileGenerator.StitchPoint(new PointF(0, 0), true),
+				new DSTFileGenerator.StitchPoint(new PointF(100, 100), true),
+				new DSTFileGenerator.StitchPoint(new PointF(200, 200))};
+
+		EmbroideryList points = new EmbroideryList();
+		points.add(new PointF(0, 0));
+		points.add(new PointF(100, 100));
+		ArrayList<DSTFileGenerator.StitchPoint> stitchPoints = fileGenerator.prepareStitchPoints(points);
+
+		assertEquals(points.size() + expectedAmountJumpPoints, stitchPoints.size());
```
84n4n4: please remove this, if this isnt true, the next assertion will fail anyways and also give you a pointer on why the length is wrong.

84n4n4:
this was already approved but was forcepushed again and it seems the jira ticket has been also changed.

jira ticket has unfulfilled acceptance critera:
* It should store in the .dst file only the patterns which were created during the last execution of the program, before entering the pause menu (if the user choses to "continue" the run of the program in the pause menu, the pattern that was constructed up to this point is further extended by the continued run). 

* For scenes: One .dst files should be created per scene.

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+			0x110, 0x111, 0x116, 0x114, 0x115, 0x16a, 0x168, 0x169, 0x162, 0x160, 0x161, 0x166, 0x164, 0x165, 0x14a,
+			0x148, 0x149, 0x142, 0x140, 0x141, 0x146, 0x144, 0x145, 0x15a, 0x158, 0x159, 0x152, 0x150, 0x151, 0x156,
+			0x154, 0x155, 0x2, 0x9, 0x8, 0xa, 0x25, 0x24, 0x26, 0x21, 0x20, 0x22, 0x29, 0x28, 0x2a, 0x95, 0x94, 0x96,
+			0x91, 0x90, 0x92, 0x99, 0x98, 0x9a, 0x85, 0x84, 0x86, 0x81, 0x80, 0x82, 0x89, 0x88, 0x8a, 0xa5, 0xa4,
+			0xa6, 0xa1, 0xa0, 0xa2, 0xa9, 0xa8, 0xaa, 0x255, 0x254, 0x256, 0x251, 0x250, 0x252, 0x259, 0x258, 0x25a,
+			0x245, 0x244, 0x246, 0x241, 0x240, 0x242, 0x249, 0x248, 0x24a, 0x265, 0x264, 0x266, 0x261, 0x260, 0x262,
+			0x269, 0x268, 0x26a, 0x215, 0x214, 0x216, 0x211, 0x210, 0x212, 0x219, 0x218, 0x21a, 0x205, 0x204, 0x206,
+			0x201, 0x200, 0x202, 0x209, 0x208, 0x20a, 0x225, 0x224, 0x226, 0x221, 0x220, 0x222, 0x229, 0x228, 0x22a,
+			0x295, 0x294, 0x296, 0x291, 0x290, 0x292, 0x299, 0x298, 0x29a, 0x285, 0x284, 0x286, 0x281, 0x280, 0x282,
+			0x289, 0x288, 0x28a, 0x2a5, 0x2a4, 0x2a6, 0x2a1, 0x2a0, 0x2a2, 0x2a9, 0x2a8, 0x2aa};
```
robertpainsi: Hi @shellp! These constant names are a little hard to read. I think adding a few `_` will help.

`MAXDISTANCE` → `MAX_DISTANCE`
`STITCHPOINTUNITFACTOR` → `STITCH_POINT_UNIT_FACTOR`
`HEADERMAXBYTE` → `HEADER_MAX_BYTE`
`CONVERSIONTABLE` → `CONVERSION_TABLE`

robertpainsi: Should have commented on line 65 :relaxed:

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -0,0 +1,295 @@
...
+			}
+		}
+	}
+
+	private ArrayList<PointF> convertStitchPointsUnit(ArrayList<PointF> stitchPoints) {
+		ArrayList<PointF> convertedStitchPoints = new ArrayList<>();
+
+		for (PointF stitch: stitchPoints) {
+			PointF stitchUnitMm = new PointF();
+			stitchUnitMm.x = Math.round(stitch.x * STITCHPOINTUNITFACTOR);
```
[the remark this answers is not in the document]
shellp: this has to do with the specification of the dst file. 
the dst file does not work with mm as its unit but with the next smaller unit (0.1mm). 
Therefore 1 step = 0.2 mm = 2 "unit of dst file"

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -0,0 +1,203 @@
...
+	@Test
+	public void testConversionTable() {
+		int[] expectedValueArray = new int[DSTFileGenerator.CONVERSION_TABLE.length];
+		int[] valueArray = new int[DSTFileGenerator.CONVERSION_TABLE.length];
+
+		for (int element = 0; element < DSTFileGenerator.CONVERSION_TABLE.length; element++) {
+			int value = getValueForByte(DSTFileGenerator.CONVERSION_TABLE[element]);
+			valueArray[element] = value;
+			expectedValueArray[element] = value;
+			if (element > 121) {
```
84n4n4: so before index 121 this isnt testing anything because expected and actual are set to the very same value.

84n4n4:
still open the two acceptance criteria from jira ticket:



    It should store in the .dst file only the patterns which were created during the last execution of the program, before entering the pause menu (if the user choses to "continue" the run of the program in the pause menu, the pattern that was constructed up to this point is further extended by the continued run).

    For scenes: One .dst files should be created per scene.


if this PRs content has been aligned with POs to not include this, then please remove it also from the ticket.

robertpainsi:
(Approved requested changes only)

wslany:
Bravo!! Thanks a lot :-)

PR #3091, opened 2019-02-13
CATROID-150 Fix wrong order of projects

aniket9465:
Files are sorted according to the lastUpdate time of their Constants File and that time won't change unless some change is made . So I updated that time when project was opened .


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
```diff
@@ -0,0 +1,170 @@
...
+		createProject(ProjectOneName);
+		createProject(ProjectTwoName);
+
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@Category({Cat.AppUi.class, Level.Smoke.class})
+	@Test
+	public void changeProjectOrderTest() {
+		try{
```
84n4n4: sleep is not necessary here, also no Thread.sleep in espresso tests

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
```diff
@@ -0,0 +1,170 @@
...
+			Thread.sleep(2000);
+		}
+		catch (Exception e){}
+		onRecyclerView().performOnItemWithText(ProjectTwoName,click());
```
84n4n4: please no switching activities in espresso tests if it is not really necessary / testable in another way.
alternatively you could just directly change the "lastModified" timestamp in a project and trigger a refresh of the adapter.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
```diff
@@ -0,0 +1,170 @@
...
+		Collections.sort(items, new Comparator<ProjectData>() {
+			@Override
+			public int compare(ProjectData project1, ProjectData project2) {
+				return Long.valueOf(project2.lastUsed).compareTo(project1.lastUsed);
+			}
+		});
+
+		if(!items.get(0).projectName.equals(ProjectTwoName))
+		{
+			fail("order not updated");
```
84n4n4: no.
no if -> fail constructs, this is what assert is for, in this case:
assertEquals(ProjectTwoName, items.get(0).projectName); will replace the if block.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
```diff
@@ -0,0 +1,170 @@
...
+import static org.junit.Assert.fail;
+
+@RunWith(AndroidJUnit4.class)
+public class changeProjectOrderTest {
+
+	@Rule
+	public BaseActivityInstrumentationRule<ProjectListActivity> baseActivityTestRule = new
+			BaseActivityInstrumentationRule<>(ProjectListActivity.class, true, false);
+
+	private String ProjectOneName = "ProjectOneName";
```
84n4n4: variables and members start with small letters, like projectOneName.
Classes start with capital letters.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+import static org.catrobat.catroid.common.FlavoredConstants.DEFAULT_ROOT_DIRECTORY;
+import static org.catrobat.catroid.uiespresso.ui.fragment.rvutils.RecyclerViewInteractionWrapper.onRecyclerView;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.fail;
+
+@RunWith(AndroidJUnit4.class)
+public class changeProjectOrderTest {
```
84n4n4: see comment below, classes start with capital letters

84n4n4:
please have a look at the jenkins run by clicking on the "Details" on the right side where the checks are above.
https://jenkins.catrob.at/job/Catroid/job/PR-3091/4/
please fix all checkstyle errors

84n4n4:
please leave the copyright header with 2018 for now, weve got static checks that will fail now, and we are going to update all copyright headers of all files batched in a separate PR.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
aniket9465: Purpose of the test is to check that the transition from list to a project changes the projects last modified .
If I change the lastModified timestamp in the test itself and someone deletes the code that changes lastModified on opening a project, the test would still pass.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/changeProjectOrderTest.java
aniket9465: Sleep was used because the precision of lastModified field is of a second and during testing the delay was not of one second between creating and opening of project, but now I handled it by changing the time of projects (as if they were modified 2 seconds before their actual modified time) before opening them.

wslany:
Still not doing what it should, e.g., when you:
1. download two projects, in addition to the default project.
2. open the third project, but exit it immediately gain.
3. observe that the project is still the third project.

Also note that there is https://github.com/Catrobat/Catroid/pull/3133 attempting to fix the same problem, but also failing so far.

wslany:
@aniket9465 should we code review this again already?

wslany:
Already solved by https://jira.catrob.at/browse/CATROID-180 --- thanks anyway!

PR #3095, opened 2019-02-15
CATROID-116 Compute touches_edge crashes in formula editor

ndrnour:
Test is included.
Bug happens cause stageListener is null if the stage is never started yet.


ndrnour:
the failed test "testSpriteFromGalleryIntentDismissTest" has nothing to do with my PR!

PR #3096, opened 2019-02-15
CATROID-90 Crash when aborting/canceling/dismissing the PlaySceneDialog and evaluating a local variable

ndrnour:


PlantinumX:
I reviewed it. Looks fine to me

ndrnour:
> please add a test and unify the start stage logic as described in the ticket

done. could you review it again?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -989,7 +989,7 @@ public boolean containsElement(ElementType elementType) {
 
 	public boolean isUserVariableWithTypeString(Sprite sprite) {
 		if (type == ElementType.USER_VARIABLE) {
-			DataContainer userVariableContainer = ProjectManager.getInstance().getCurrentlyPlayingScene()
+			DataContainer userVariableContainer = ProjectManager.getInstance().getCurrentlyEditedScene()
```
thmq: why did you change this? are you sure we should use the `currentlyEditedScene` here? this is going to change how this behaves.

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -415,4 +419,31 @@ public void addToRequiresPermissionTaskList(RequiresPermissionTask task) {
 	public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
 		permissionRequestActivityExtension.onRequestPermissionsResult(this, requestCode, permissions, grantResults);
 	}
+
+	public static void handlePlayButton(View view, final Activity activity) {
```
thmq: the parameter `view` is not used in this method, please remove it. the `View view` parameter in the methods in the other Activities is just necessary because it is called from the `.xml`, so it's fine there, but unnecessary here.

But as a suggestion: you could add the `ProjectManager` as a parameter to the method, because then the `ProjectManager.getInstance()` call could happen from the calling activity (e.g. `SpriteActivity`), so that one day, we might be able to refactor the God Object `ProjectManager` so that e.g. testing becomes easier.

ndrnour: done

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
ndrnour: since this method "isUserVariableWithTypeString(Sprite sprite)" is just only called from the method "getResultForComputeDialog(Context context)" , so nothing related to stage is changed. so I'm sure this is right.

thmq: alright, thank you for clearing that up. :wink:

PR #3098, opened 2019-02-17
CATROID-23/24 Implemented new ShowTextAndColorBrick and Action.

PlantinumX:
Add size parameter for Show variable brick. Wrote one TestCase
for new ShowTextColorAndSizeBrick.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
@@ -25,6 +25,7 @@ package org.catrobat.catroid.stage;
 
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
+import android.graphics.Color;
 import android.graphics.Paint;
 import android.opengl.GLES20;
 import android.opengl.GLUtils;
@@ -42,12 +43,15 @@ import org.catrobat.catroid.formulaeditor.UserVariable;
 import org.catrobat.catroid.formulaeditor.datacontainer.DataContainer;
 
 import java.util.List;
+import java.util.Locale;
 
 public class ShowTextActor extends Actor {
 
-	private final float textSize = 45;
+	private static final int DEFAULT_TEXT_SIZE = 45;
+	private float textSize;
 	private int xPosition;
 	private int yPosition;
+	private String color;
 	private UserVariable variableToShow;
 	private String variableNameToCompare;
 	private String variableValueWithoutDecimal;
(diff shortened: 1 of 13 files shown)


84n4n4:
ok, one last thing, to have the new brick available in the add brick fragment (to be able to add it via the UI) it has to be added to CategoryBricksFactory  in setupDataCategoryList.
thenk language version has to be incremented (since theres a new brick thats a change to the language) in Constants.CURRENT_CATROBAT_LANGUAGE_VERSION
and this has then to be updated in ProjectManager.loadProject (add another if clause there)

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -42,25 +43,31 @@
...
 
-	public ShowTextActor(UserVariable userVariable, int xPosition, int yPosition, Sprite sprite, UserBrick userBrick) {
+	public ShowTextActor(UserVariable userVariable, int xPosition, int yPosition, float relativeSize, String color,
+			Sprite sprite, UserBrick userBrick) {
 		this.variableToShow = userVariable;
 		this.variableNameToCompare = variableToShow.getName();
 		this.variableValueWithoutDecimal = null;
 		this.xPosition = xPosition;
 		this.yPosition = yPosition;
+		this.textSize = DEFAULT_TEXT_SIZE * relativeSize;
```
wslany: The new ShowTextAndColorBrick brick crashes Catroid for small and for very large size values. It also behaves strangely for negative size values (large negative size values move the shown value up the screen, for whatever reason). E.g, the app crashes when the brick is executed with a size value of 1%, or when the size is 100000%. 

It therefore is necessary to sanitize the size value to a legal range allowed by its implementation, and deal with all values above and below as having that highest, respectively lowest technically possible values. As an example, in case the technically feasible range would be from 5% up to 10000%: use 5 instead of 1 or -100, and use 10000 instead of 10001 and above.

wslany: The new ShowTextAndColorBrick brick crashes Catroid for small and for very large size values. It also behaves strangely for negative size values (large negative size values move the shown value up the screen, for whatever reason). E.g, the app crashes when the brick is executed with a size value of 1%, or when the size is 100000%. 

It therefore is necessary to sanitize the size value to a legal range allowed by its implementation, and deal with all values above and below as having that highest, respectively lowest technically possible values. As an example, in case the technically feasible range would be from 5% up to 10000%: use 5 instead of 1 or -100, and use 10000 instead of 10001 and above.

PR #3099, opened 2019-02-18
CATROID-144 white stage or crash when playing a renamed Scene

ndrnour:
just reloading the project after renaming the Scene fixed the bug.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
@@ -34,6 +34,7 @@ import org.catrobat.catroid.R;
 import org.catrobat.catroid.content.Project;
 import org.catrobat.catroid.content.Scene;
 import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.io.asynctask.ProjectLoadTask;
 import org.catrobat.catroid.ui.controller.BackpackListManager;
 import org.catrobat.catroid.ui.recyclerview.adapter.SceneAdapter;
 import org.catrobat.catroid.ui.recyclerview.backpack.BackpackActivity;
@@ -46,7 +47,7 @@ import java.util.List;
 import static org.catrobat.catroid.common.Constants.Z_INDEX_BACKGROUND;
 import static org.catrobat.catroid.common.SharedPreferenceKeys.SHOW_DETAILS_SCENES_PREFERENCE_KEY;
 
-public class SceneListFragment extends RecyclerViewFragment<Scene> {
+public class SceneListFragment extends RecyclerViewFragment<Scene> implements ProjectLoadTask.ProjectLoadListener {
 
 	public static final String TAG = SceneListFragment.class.getSimpleName();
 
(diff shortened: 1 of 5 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
```diff
@@ -208,6 +210,8 @@ public void renameItem(Scene item, String name) {
 		if (!item.getName().equals(name)) {
 			if (sceneController.rename(item, name)) {
 				ProjectManager.getInstance().saveProject(getActivity());
+				ProjectLoadTask projectLoadTask = new ProjectLoadTask(getContext(), this);
+				projectLoadTask.execute(Utils.getCurrentProjectName(getActivity()));
```
84n4n4: please use ProjectManager.getInstance().getCurrentProject().getName() instead

84n4n4:
please add a test

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
ndrnour: done.

ndrnour:
> please add a test

please review it again.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
```diff
@@ -208,6 +209,8 @@ public void renameItem(Scene item, String name) {
 		if (!item.getName().equals(name)) {
 			if (sceneController.rename(item, name)) {
 				ProjectManager.getInstance().saveProject(getActivity());
```
thmq: the `ProjectManager.saveProject(..)` starts an async task, so I'm not sure it is the best way to immediately start another task, that works on the same `.xml` file right after that.

ndrnour: any suggestions?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
```diff
@@ -174,8 +176,9 @@ protected void deleteItems(List<Scene> selectedItems) {
 			createEmptySceneWithDefaultName();
 		}
 
-		if (ProjectManager.getInstance().getCurrentProject().getSceneList().size() < 2) {
-			switchToSpriteListFragment();
+		Project currentProject = ProjectManager.getInstance().getCurrentProject();
+		if (currentProject.getSceneList().size() < 2) {
+			switchToSpriteListFragment(currentProject);
```
thmq: You added the parameter so that whenever a scene was removed you make sure that the `currentlyEditedScene` is updated correctly, which absolutely makes sense. However, how about you set it directly instead of going the extra way around it by using `setCurrentProject(..)`, as it would make it easier to understand what you do and why it is necessary.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
thmq: well, yes and no. I have provided a async task in #3104, with which you could either wait for the task with the listener provided there, or use it sync.

wslany:
@ndrnour @thmq What's the reason for the failed and aborted tests, see https://jenkins.catrob.at/job/Catroid/job/PR-3099/test_results_analyzer/ at the bottom?

ndrnour:
> @ndrnour @thmq What's the reason for the failed and aborted tests, see https://jenkins.catrob.at/job/Catroid/job/PR-3099/test_results_analyzer/ at the bottom?

I really have no Idea, this test "testSpriteFromGalleryIntentDismissTest" is now broken in all my pr and in #3104

PR #3100, opened 2019-02-18
JENKINS-248: Add Playground-App support for release pipeline testing

rainer-lankmayr:
- add the option to upload the Playground-App to org.catrobat.catroid.test on Google Play
- add flavor selection in Jenkinsfiles for APK-Release and Metadata

### DIFF
--- a/Jenkinsfile.buildMetadata
+++ b/Jenkinsfile.buildMetadata
@@ -27,17 +27,26 @@ pipeline {
         stage('Prepare build') {
             steps {
                 script {
-                    currentBuild.displayName = "#${env.BUILD_NUMBER}"
+                    currentBuild.displayName = "#${env.BUILD_NUMBER} | ${env.flavor} | ${env.gitBranch}"
                 }
             }
         }
 
         stage('Setup Translations') {
             steps {
-                sh '''
-                    set +x
-                    ./gradlew generateCrowdinMetadataCatroid -PcrowdinKey=$crowdinKey
-                '''
+                script {
+                    if (env.flavor == 'Playground') {
+                        sh '''
+                            set +x
+                            ./gradlew generateCrowdinMetadataCatroid -PcrowdinKey=$crowdinKey
+                        '''
+                    } else {
+                        sh '''
+                            set +x
+                            ./gradlew generateCrowdinMetadata${flavor} -PcrowdinKey=$crowdinKey
+                        '''
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


Jenkinsfile.buildMetadata
```diff
@@ -97,12 +114,24 @@ pipeline {
                 environment name: 'APPROVE_DEPLOY', value: 'yes'
             }
             steps {
-                sh "fastlane android promote_Catroid"
+                script {
+                    if (env.flavor == 'Playground') {
+                        echo 'Playground cannot be promoted to production!'
+                    } else {
+                        // deactivated for testing to not publish by accident
+                        //sh 'fastlane android promote_${flavor}'
```
jampo: still deactivated for commit on purpose? as it was active before

catroid/gradle/release_fastlane_tasks.gradle
```diff
@@ -65,7 +65,7 @@ android.productFlavors.all { flavor ->
 
         String fastlaneCommand = "fastlane screengrab" +
                 " --app_package_name 'org.catrobat.catroid'" +
-                " --use_tests_in_packages 'org.catrobat.catroid.screenshots.${flavor.name}'" +
+                " --use_tests_in_packages 'org.catrobat.catroid.screenshots.${flavor.name.toLowerCase()}'" +
```
jampo: why is to lowerCase just needed for this argument and not the others?

fastlane/Fastfile
```diff
@@ -105,4 +149,96 @@ platform :android do
...
+      track:                      'alpha',
+      json_key_data:              ENV['googlePlayJsonKey'],
+      apk_paths:                  './catroid/build/outputs/apk/catroid/signedRelease/catroid-catroid-signedRelease.apk',
+      skip_upload_apk:            'false',
+      skip_upload_metadata:       'true',
+      skip_upload_images:         'true',
+      skip_upload_screenshots:    'true',
+      validate_only:              'false',
+      check_superseded_tracks:    'false',
+    )
```
jampo: guess there is no Promotion for Playground APK right? - so this step can't be tested with this...

Jenkinsfile.buildMetadata
rainer-lankmayr: Yes this is still deactivated on purpose. This is probably not the cleanest way, but I still kept it in as a comment to keep it documented that this option is still planned in the original workflow and that is is supposed to be right at this step of the process.

fastlane/Fastfile
rainer-lankmayr: Exactly! The Playground APK is not supposed to be published to the public (production track) anyway. It is only published to the alpha track.

catroid/gradle/release_fastlane_tasks.gradle
rainer-lankmayr: This refers to a package name, which are all lowercase by convention. All the other arguments here refer to files and paths, which are camelCase. 
'flavor.name' itself is in camelCase as well. So it is only need for the argument containing the package name.

PR #3101, opened 2019-02-18
CATROID-25 Create WriteVariableOnDeviceBrick

gPathpp:


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
```diff
@@ -52,7 +54,7 @@ public Sprite convert(Sprite spriteToConvert, Scene parentScene) {
 		return convertedSprite;
 	}
 
-	public Sprite copy(Sprite spriteToCopy, Scene srcScene, Scene dstScene) throws IOException {
+	public Sprite copy(Sprite spriteToCopy, Scene srcScene, Scene dstScene, boolean copyDeviceFileName) throws IOException {
```
ThomasSchwengler: Please try to avoid boolean flags. This is very much unreadable.

wslany:
Sorry @LPeteR90 @ThomasSchwengler --- I only saw that the **merge conflicts** have to be resolved first by @gPathpp after I pressed on the re-review request --- my mistake! Please disregard the request until then.

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableManager.java
```diff
@@ -0,0 +1,181 @@
...
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.UUID;
+
+import static org.catrobat.catroid.common.Constants.DEVICE_VARIABLE_DIR;
+
+public final class DeviceVariableManager {
```
ThomasSchwengler: About this - it's yet another manager class and yet another singleton.
`Manager` most of the time means that it does way more than it should.

From what I'm seeing here this actually just manages a global lock via the `synchronized` keyword.
Other than that it just hides the fact that everything in here are public static helper methods.

There should be a more object oriented way to solve this.

catroid/src/main/java/org/catrobat/catroid/io/asynctask/ProjectCopyTask.java
```diff
@@ -51,7 +53,8 @@ public ProjectCopyTask(Context context, ProjectCopyListener listener) {
 	}
 
 	@Override
-	protected Boolean doInBackground(String... strings) {
+	@VisibleForTesting
```
ThomasSchwengler: the default for `otherwise` is `private`. I'm not sure if this task gets overloaded or not. If it does, then it should be `@VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)`.

PR #3106, opened 2019-02-23
CATROID-161 extent feature for show variable brick. 

PlantinumX:
Refactored some parts in Actionfactory. Wrote one new testcase in ShowTextColorAndSizeBrickTest


catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -223,9 +233,20 @@ private float sanitizeTextSize(float textSize) {
...
-		}
 		return textSize;
 	}
+
+	private int calculateAlignmentValuesForText(Paint paint, String text) {
+		switch (alignment) {
+			case ShowTextColorAndSizeBrick.ALIGNEMNT_STYLE_LEFT:
+				paint.setTextAlign(Align.LEFT);
+				return 0;
+			case ShowTextColorAndSizeBrick.ALIGNEMNT_STYLE_RIGHT:
```
wslany: same here...

catroid/src/main/res/layout/brick_show_variable_text_color_and_size.xml
```diff
@@ -121,6 +121,21 @@
             style="@style/BrickEditText"
             app:layout_inputType="number"
             app:layout_textField="true" />
+
+        <TextView
+            android:id="@+id/brick_show_variable_color_and_size_spinner"
```
wslany:
```suggestion
            android:id="@+id/brick_show_variable_color_size_align_spinner"
```
and other places too in accordance...

ndrnour:
since the new brick has multiple Brick fields, please modify the test "BricksWithMultipleBrickFieldsEditFormulaTest.java". just add the new brick to the project.
don't forget to override the method "getDefaultBrickField()" with return of the first brick field from lift " in your case is "BrickField.X_POSITION"

catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextColorSizeAlignmentBrick.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.List;
+
+public class ShowTextColorSizeAlignmentBrick extends UserVariableBrick {
+	private static final long serialVersionUID = 1L;
+	public static final int ALIGNMENT_STYLE_LEFT = 0;
+	public static final int ALIGNMENT_STYLE_CENTERED = 1;
+	public static final int ALIGNMENT_STYLE_RIGHT = 2;
+	public int alignmentSelection = ALIGNMENT_STYLE_CENTERED;
+
+	class AlignmentStyle implements Nameable, BrickSpinner.OnItemSelectedListener<AlignmentStyle> {
```
thmq: Actually the `ShowTextColorSizeAlignmentBrick` should implement `BrickSpinner.OnItemSelectedListener<AlignmentStyle>` so that the brick itself can react to spinner selection changes. then you can also remove the reference to the brick (`ShowTextColorSizeAlignmentBrick object;`) from the subclass. and besides if you want to access a member variable of the parent class from a subclass you should just use `ShowTextColorSizeAlignmentBrick.this` instead.

catroid/src/main/java/org/catrobat/catroid/stage/ShowTextActor.java
```diff
@@ -162,12 +175,13 @@ private void drawText(Batch batch, String text, float posX, float posY, String c
...
-		paint.setAntiAlias(true);
-		paint.setTextAlign(Paint.Align.LEFT);
 		float baseline = -paint.ascent();
-		int canvasWidth = 0;
-		int bitmapWidth = (int) (paint.measureText(text));
-		int height = (int) (baseline + paint.descent());
+		paint.setAntiAlias(true);
+
+		int canvasWidth = calculateAlignmentValuesForText(paint, text);
+		int bitmapWidth = (int) (paint.measureText(text)); // values[1]
```
thmq: please remove the comments.

catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextColorSizeAlignmentBrick.java
PlantinumX: Your suggestion was my first approach but cannot `ShowTextColorSizeAlignmentBrick` cannot implement BrickSpinner.OnItemSelectedListener<AlignmentStyle> because it extents UserVariable which implements BrickSpinner.OnItemSelectedListener.  But thx for the second advice :)

thmq: Oh yes. I missed that it there are going to be two different spinners in this brick. still it does not make sense to have the objects that are part of the spinner and therefore in its list implement the listener.  I suggest you solve it by adding another class that implements the listener to the brick. 

```
        ...
	public static final int ALIGNMENT_STYLE_CENTERED = 1;
	public static final int ALIGNMENT_STYLE_RIGHT = 2;
	public int alignmentSelection = ALIGNMENT_STYLE_CENTERED;     
  
        private transient SpinnerListener spinnerListener = new SpinnerListener();
	
	class SpinnerListener implements BrickSpinner.OnItemSelectedListener<AlignmentStyle> {
		
		@Override
		public void onStringOptionSelected(String string) {
		}

		@Override
		public void onItemSelected(@Nullable AlignmentStyle item) {
			ShowTextColorSizeAlignmentBrick.this.alignmentSelection = item.alignmentStyle;
		}

		@Override
		public void onNewOptionSelected() {
		}
	}
	class AlignmentStyle implements Nameable, BrickSpinner.OnItemSelectedListener<AlignmentStyle> {

		AlignmentStyle(String name, int alignmentStyle) {
			this.name = name;
			this.alignmentStyle = alignmentStyle;
		}
        ...
```

catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextColorSizeAlignmentBrick.java
```diff
@@ -56,12 +102,29 @@ public ShowTextColorAndSizeBrick(Formula xPosition, Formula yPosition, Formula s
...
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_left),
+				ALIGNMENT_STYLE_LEFT));
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_centered),
+				ALIGNMENT_STYLE_CENTERED));
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_right),
+				ALIGNMENT_STYLE_RIGHT));
+		BrickSpinner<AlignmentStyle> spinner =
+				new BrickSpinner<>(R.id.brick_show_variable_color_size_align_spinner, view, items);
+		spinner.setSelection(alignmentSelection);
+		spinner.setOnItemSelectedListener((AlignmentStyle) items.get(0));
```
thmq: and then just set it here like: 
`spinner.setOnItemSelectedListener(spinnerListener);

catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextColorSizeAlignmentBrick.java
thmq: **Do not forget** to then override the `clone()` method and set the member that holds the `spinnerListener` to null, otherwise every newly cloned brick that is created from this one has the same listener and consequently gets called whenever something is changed in the original brick.

```
	@Override
	public BrickBaseType clone() throws CloneNotSupportedException {
		ShowTextColorSizeAlignmentBrick clone = (ShowTextColorSizeAlignmentBrick) super.clone();
		clone.spinnerListener = null;
		return clone;
	}

```

thmq: Alternatively you could try and don't add a member variable for the spinner at all. I'm not sure if this works, because you can never be sure about what the garbage collector does in java. but perhaps you just have to add the class and then just do:

```
...
spinner.setOnItemSelectedListener(new SpinnerListener);
...
```
in `getView()`. or even move the whole class definition here, like this:

```
	@Override
	public View getView(Context context) {
		super.getView(context);
		List<Nameable> items = new ArrayList<>();
		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_left),
				ALIGNMENT_STYLE_LEFT));
		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_centered),
				ALIGNMENT_STYLE_CENTERED));
		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_right),
				ALIGNMENT_STYLE_RIGHT));
		BrickSpinner<AlignmentStyle> spinner =
				new BrickSpinner<>(R.id.brick_show_variable_color_size_align_spinner, view, items);
		spinner.setSelection(alignmentSelection);

		spinner.setOnItemSelectedListener(new BrickSpinner.OnItemSelectedListener<AlignmentStyle> (){

			@Override
			public void onStringOptionSelected(String string) {
			}

			@Override
			public void onItemSelected(@Nullable AlignmentStyle item) {
				ShowTextColorSizeAlignmentBrick.this.alignmentSelection = item.alignmentStyle;
			}

			@Override
			public void onNewOptionSelected() {
			}
		});
		return view;
	}
```

so many options :confused: but now that I think about it again the last two might be the best -> here you wouldn't even have to override the `clone(..)` method.

thmq:
:+1: thanks for implementing the changes so quickly.

thmq:
please just get rid of the checkstyle warning [see](https://jenkins.catrob.at/job/Catroid/job/PR-3106/19/checkstyleResult/) and it's good to merge.

catroid/src/main/java/org/catrobat/catroid/content/bricks/ShowTextColorSizeAlignmentBrick.java
```diff
@@ -56,12 +89,43 @@ public ShowTextColorAndSizeBrick(Formula xPosition, Formula yPosition, Formula s
...
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_left),
+				ALIGNMENT_STYLE_LEFT));
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_centered),
+				ALIGNMENT_STYLE_CENTERED));
+		items.add(new AlignmentStyle(context.getString(R.string.brick_show_variable_aligned_right),
+				ALIGNMENT_STYLE_RIGHT));
+		BrickSpinner<AlignmentStyle> spinner =
+				new BrickSpinner<>(R.id.brick_show_variable_color_size_align_spinner, view, items);
+		spinner.setSelection(alignmentSelection);
+		spinner.setOnItemSelectedListener(new BrickSpinner.OnItemSelectedListener<AlignmentStyle> () {
```
thmq: please remove the whitespace before `(` 

`<AlignmentStyle>()` instead of `<AlignmentStyle> ()`

[Checkstyle warning on Jenkins](https://jenkins.catrob.at/job/Catroid/job/PR-3106/19/checkstyleResult/)

PR #3108, opened 2019-02-25
CATROID-48 Suspend/resume all threads on scene transition

rschluesselbauer:
Add scheduler state [running / paused] to pause / resume all schedulers of a scene during scene transition.

Change idToEventThreadMap to type LinkedHashMultiMap (from HashMultiMap) to ensure consistent execution order (before the order of multiple threads started at the same time was not deterministic).


catroid/src/main/java/org/catrobat/catroid/common/ThreadScheduler.java
```diff
@@ -60,7 +73,7 @@ private void startThreadsInStartQueue() {
 	private void runThreadsForOneTick(Array<Action> actions, float delta) {
 		for (int i = 0; i < actions.size; i++) {
 			Action action = actions.get(i);
-			if (action.act(delta)) {
+			if (state == RUNNING && action.act(delta)) {
```
robertpainsi: I haven't looked up how unit-testable the `ThreadScheduler` is. Maybe you can test this line here with a unit test?

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
```diff
@@ -86,7 +86,7 @@
 	public transient PenConfiguration penConfiguration = new PenConfiguration();
 	private transient boolean convertToSingleSprite = false;
 	private transient boolean convertToGroupItemSprite = false;
-	private transient Multimap<EventId, EventThread> idToEventThreadMap = HashMultimap.create();
+	private transient Multimap<EventId, EventThread> idToEventThreadMap = LinkedHashMultimap.create();
```
robertpainsi: Please let me know why you switched the `HashMultimap` with a `LinkedHashMultimap` :wink:

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/SceneTransitionTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityInstrumentationRule;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import static org.catrobat.catroid.uiespresso.util.UserVariableAssertions.assertUserVariableEqualsWithTimeout;
+import static org.catrobat.catroid.uiespresso.util.UserVariableAssertions.assertUserVariableNotEqualsForTimeMs;
+
+public class SceneTransitionTest {
```
robertpainsi: I'll look into this in the next few days.
I think https://github.com/Catrobat/Catroid/pull/3134 could be useful to test this.

catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
rschluesselbauer: When an event is received, then all eventThreads are started in the order given by idToEventThreadMap. However the HashMultimap type didn't ensure consistent order.

This is especially problematic for testing behaviour of threads during a single tick.
If one thread should pause all other threads, but sometimes it is executed before & sometimes it is executed after the other threads, it's a pain to test.

In short it makes the execution order of the programs deterministic which is nice :-)

robertpainsi: Got it! Great explanation! Thanks a lot! :bowing_man:

catroid/src/main/java/org/catrobat/catroid/common/ThreadScheduler.java
rschluesselbauer: Great idea, should be doable 👍 
I'll look into it!

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/SceneTransitionTest.java
rschluesselbauer: I added the whole test as CatrobatTestrunnerTests -if this is the prefered way I can remove this test.

catroid/src/main/java/org/catrobat/catroid/common/ThreadScheduler.java
rschluesselbauer: Done

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/stage/SceneTransitionTest.java
robertpainsi: Ah, stupid me! I misunderstood that PR entirely. I didn't thought that projects have to be in `.catrobat`s. D'oh!

However, I still think that the _CatrobatTestrunner_ is suited here. It verifies that, when a scene switch brick is executed, the current scene stops (and doesn't execute any further bricks) and the new scene starts.

@84n4n4, just want to make sure that the _CatrobatTestrunner_ should be used here. And mostly that the _CatrobatTestrunner_ should be used in the future for such tests.
(I'll comment on this tests later)

rschluesselbauer: I checked with @84n4n4 and CatrobatTestrunner based tests are the way to go here, so I'll get rid of this espresso test.

robertpainsi: :kissing_cat:

robertpainsi: ~(I'll comment on this tests later)~ → LGTM!

rschluesselbauer: One more thing: In my tests I assume unset variables to be equal to 0 and it works, but I suspect that could cause issues.
Should I use a scene for setting up variable values first in this case?

robertpainsi: > …but I suspect that could cause issues.

Are you worrying that the default value `0` may change in the future or do you have other issues in mind?

rschluesselbauer: I retested and my assumption was wrong (I thought that initial values could change based on the value of the variable during the last run). The tests are ok then 👍

robertpainsi: :+1:

> so I'll get rid of this espresso test

…, then move the ticket to _Ready For Code Review and I'll approve it :tada:

robertpainsi: ~_then move the ticket to Ready For Code Review_~

wslany:
Cool, thanks a lot for this fix!

PR #3117, opened 2019-03-02
CATROID-171 Visual placement 2

ndrnour:
dear reviewer:
I wrote some comments to make the review easy for you, please read them before
@84n4n4 could you please provide me a way to test the visual placement without activity switching
thanks!

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -431,10 +431,6 @@ public class Look extends Image {
 		}
 	}
 
-	public float getRealRotation() {
-		return realRotation;
-	}
-
 	public boolean isFlipped() {
 		return (lookData != null && lookData.getTextureRegion().isFlipX());
 	}
--- a/catroid/src/main/java/org/catrobat/catroid/content/Project.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Project.java
@@ -43,11 +43,9 @@ import org.catrobat.catroid.io.XStreamFieldKeyOrder;
 import org.catrobat.catroid.physics.content.ActionPhysicsFactory;
 import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.ui.settingsfragments.SettingsFragment;
-import org.catrobat.catroid.utils.PathBuilder;
 import org.catrobat.catroid.utils.ScreenValueHandler;
 import org.catrobat.catroid.utils.Utils;
 
-import java.io.File;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
@@ -308,16 +306,6 @@ public class Project implements Serializable {
 			}
 		}
 		return null;
-	}
-
-	public boolean manualScreenshotExists(String manualScreenshotName) {
-
-		String path = PathBuilder.buildProjectPath(getName()) + "/" + manualScreenshotName;
-		File manualScreenShot = new File(path);
-		if (manualScreenShot.exists()) {
-			return false;
-		}
-		return true;
 	}
 
 	public void saveLegoNXTSettingsToProject(Context context) {
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
@@ -233,6 +233,7 @@ public class StageActivity extends AndroidApplication implements PermissionHandl
 			finish();
 		} else {
 			StageLifeCycleController.stagePause(this);
+			stageListener.makeScreenshot(false);
 			stageDialog.show();
 		}
 	}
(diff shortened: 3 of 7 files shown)


catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -431,10 +431,6 @@ public void setDirectionInUserInterfaceDimensionUnit(float degrees) {
 		}
 	}
 
-	public float getRealRotation() {
```
ndrnour: it's a not used and misleading method which has the same return value as getDirectionInUserInterfaceDimensionUnit(). so it's duplicated code therefore I deleted it.

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -233,6 +233,7 @@ public void onBackPressed() {
 			finish();
 		} else {
 			StageLifeCycleController.stagePause(this);
+			stageListener.makeScreenshot(false);
```
ndrnour: take an automatic screenshot after pausing the stage before showing the stage dialog. I guess this is the best place for it.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -106,18 +106,13 @@
 	private boolean reloadProject = false;
 	public boolean firstFrameDrawn = false;
 
-	private static boolean checkIfAutomaticScreenshotShouldBeTaken = true;
-	private boolean makeAutomaticScreenshot = false;
```
ndrnour: no need anymore for these flags.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -617,12 +593,13 @@ public void dispose() {
 		disposeClonedSprites();
 	}
 
-	public boolean makeManualScreenshot() {
+	public boolean makeScreenshot(boolean manual) {
```
ndrnour: since we need an updated automatic screenshot everytime the stage finished/paused. so I modified this method and reused for taking the automatic- and the manual- screenshot.

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -173,10 +178,10 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
 		try {
 			String backgroundBitmapPath;
-			if (manualScreenshot.exists()) {
-				backgroundBitmapPath = manualScreenshot.getPath();
-			} else if (automaticScreenshot.exists()) {
+			if (automaticScreenshot.exists()) {
```
ndrnour: switch round, see the updated description of CAT-14
"The ground image below the object should be the automatic screenshot, in case the project has been executed in the session and if the user is still in the same scene. In case the scene has not yet been executed, use the latest manual screenshot of the scene. If the scene has never been executed before, i.e., it is new, and thus no automatic or manual screenshot is available, and if the object also is not the background, then use the first look of the background of that scene as the ground image. However, if the background has no looks, no ground image shall exist and a white image is to be used for the ground. If the current object is itself the background, the ground image behind the look of the background is also white."

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -214,10 +220,23 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+				imageView.setRotation(0f);
+				break;
+			case ROTATION_STYLE_ALL_AROUND:
+				if (rotation != 90) {
+					imageView.setRotation(rotation - DEGREE_UI_OFFSET);
+				}
+				break;
+			case ROTATION_STYLE_LEFT_RIGHT_ONLY:
+				if (rotation < 0) {
+					scaledSpriteBitmap = createFlippedBitmap(scaledSpriteBitmap);
```
ndrnour: the best way to flip an ImageView by flipping its bitmap.

catroid/src/main/java/org/catrobat/catroid/content/Project.java
```diff
@@ -310,16 +308,6 @@ public Scene getSceneByName(String name) {
 		return null;
 	}
 
-	public boolean manualScreenshotExists(String manualScreenshotName) {
-
-		String path = PathBuilder.buildProjectPath(getName()) + "/" + manualScreenshotName;
```
ndrnour: since the screenshots are saved inside the Scene folder, this method will return always false, also it's not used anymore, therefore i deleted it

thmq:
:+1: very nice work!

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -159,6 +164,9 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 				scale = scaleWidthRatio / scaleHeightRatio;
 				maximizeLayoutHeight = (int) (currentScreenHeight * scale);
 				maximizeLayoutWidth = currentScreenWidth;
+			} else {
```
ndrnour: this missing else statement was causing the black screen.

ndrnour:
since @wslany need it in the release, closed see #3119

ndrnour:
> 👍 very nice work! and it's very good practice to add comments to help the reviewer understand what happening in this PR.

@thmq could you please approve it

PR #3119, opened 2019-03-03
CATROID-171 Visual placement 2

ndrnour:


ndrnour:
release PR same as #3117

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -173,10 +181,10 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
 		try {
 			String backgroundBitmapPath;
-			if (manualScreenshot.exists()) {
-				backgroundBitmapPath = manualScreenshot.getPath();
-			} else if (automaticScreenshot.exists()) {
+			if (automaticScreenshot.exists()) {
 				backgroundBitmapPath = automaticScreenshot.getPath();
+			} else if (manualScreenshot.exists()) {
+				backgroundBitmapPath = manualScreenshot.getPath();
```
thmq: this turns around priority:

previously: manual screenshot exists -> use manual screenshot for backgroundBitmapPath *ELSE* use automatic screenshot.

now: automatic screenshot exists -> use automatc screenshot for backgroundBitmapPath *ELSE* use manual screenshot. so whenever the user takes a manual screenshot it's going to be "overwritten" by the automatic one.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -617,12 +593,13 @@ public void dispose() {
 		disposeClonedSprites();
 	}
 
-	public boolean makeManualScreenshot() {
+	public boolean makeScreenshot(boolean manual) {
```
thmq: sorry to nitpick here, but would you mind calling it `takeScreenshot(..` instead of `makeScreenshot(..)`, please?

and another idea: how about instead of having a boolean flag you use the sceenshot name as a parameter? then it would be easier to read calls to it `takeScreenshot(ture)` vs. `takeScreenshot(SCREENSHOT_MANUAL_FILE_NAME)`. 

and do you really need the `this` in front of `screenshot`?

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
ndrnour: exactly the manual screenshot is always going to be overwritten, unless the automatic screenshot is not there or deleted somehow. according to the tickets description.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
ndrnour: done.
regarding"this" you're absolutely right, it's not my code, but it's my bad since I forget to delete it.

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
thmq: oh, yes. I'm sorry, I completely misunderstood that. I did not realise that this is only used for the visual placement here.

PR #3122, opened 2019-03-04
STCC-182 Fetching Meta Data in Pocket Code fails

AntiDog:
"tags" are no longer available in the scratch metadata, it doesn't seem there is any replacement for them. Removing them solves the "Whoops Server did not reply" and missing metadata.


thmq:
please adapt the test accordingly, then everything should be fine.

PR #3127, opened 2019-03-05
CATROID-137 Update broadcast messages and scenes after adding new ones from brick

ndrnour:
update Robolectric to version 4.2.1, because version 4.2 cause a problem on windows when running the test.
new tests:
 * BroadcastBricksTest.java
 * PointToBrickAdditionalTest.java
 * SceneBricksTest.java
new Wrapper:
 * BroadcastBrickDataInteractionWrapper.java
 * BroadcastBrickSpinnerDataInteractionWrapper.java

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
@@ -21,7 +21,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package org.catrobat.catroid.uiespresso.content.messagecontainer;
+package org.catrobat.catroid.uiespresso.content.brick.app;
 
 import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
@@ -32,38 +32,26 @@ import org.catrobat.catroid.content.BroadcastScript;
 import org.catrobat.catroid.content.Project;
 import org.catrobat.catroid.content.Script;
 import org.catrobat.catroid.content.Sprite;
-import org.catrobat.catroid.content.StartScript;
-import org.catrobat.catroid.content.bricks.BroadcastReceiverBrick;
-import org.catrobat.catroid.io.asynctask.ProjectLoadTask;
-import org.catrobat.catroid.io.asynctask.ProjectSaveTask;
+import org.catrobat.catroid.content.bricks.BroadcastBrick;
+import org.catrobat.catroid.content.bricks.BroadcastWaitBrick;
 import org.catrobat.catroid.test.utils.TestUtils;
 import org.catrobat.catroid.ui.SpriteActivity;
-import org.catrobat.catroid.uiespresso.annotations.Flaky;
-import org.catrobat.catroid.uiespresso.testsuites.Cat;
-import org.catrobat.catroid.uiespresso.testsuites.Level;
 import org.catrobat.catroid.uiespresso.util.rules.BaseActivityInstrumentationRule;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
-import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
-import static android.support.test.espresso.Espresso.onView;
-import static android.support.test.espresso.action.ViewActions.click;
-import static android.support.test.espresso.assertion.ViewAssertions.doesNotExist;
-import static android.support.test.espresso.matcher.ViewMatchers.withText;
-
-import static junit.framework.TestCase.assertTrue;
+import java.util.ArrayList;
+import java.util.List;
 
 import static org.catrobat.catroid.uiespresso.content.brick.utils.BroadcastBrickDataInteractionWrapper.onBroadcastBrickAtPosition;
 
 @RunWith(AndroidJUnit4.class)
-public class BroadcastReceiveBrickMessageContainerTest {
-
+public class BroadcastBricksTest {
 	private String defaultMessage = "defaultMessage";
-	private Project project;
-	private Sprite sprite;
+	String message = "Oida!";
 
 	@Rule
 	public BaseActivityInstrumentationRule<SpriteActivity> baseActivityTestRule = new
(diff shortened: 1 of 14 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastAndWaitBrickTest.java
```diff
@@ -61,33 +63,46 @@
...
+		TestUtils.deleteProjects(this.getClass().getName());
+	}
+
 	@Category({Cat.AppUi.class, Level.Functional.class})
 	@Test
 	@Flaky
 	public void testBroadcastAndWaitBrick() {
 		String uselessMessage = "useless";
-		createNewMessageOnSpinner(R.id.brick_broadcast_spinner, broadcastAndWaitPosition, uselessMessage);
+		createNewMessageOnSpinner(R.id.brick_broadcast_spinner, firstBroadcastAndWaitPosition, uselessMessage);
```
84n4n4: please create separate tests for issues like this. do not extend existing tests to additionally test some other stuff on the way.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/API19SpinnerAndFormulaFieldCrashesRegressionTest.java
```diff
@@ -26,13 +26,15 @@
 import android.support.test.runner.AndroidJUnit4;
 
 import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Script;
```
84n4n4: do not modify this test class, this is very specifically testing some api level 19 issues and is also not part of the regular jenkins PR testruns

84n4n4:
please have a look at LookFromGalleryIntentTest for how to mock a result from another activity to not switch activities

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastAndWaitBrickTest.java
ndrnour: done

wslany:
@84n4n4 could you code review this again please?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+		onView(withId(android.R.id.button1))
+				.perform(click());
+
+		onBrickAtPosition(position)
+				.onSpinner(R.id.brick_broadcast_spinner)
+				.checkShowsText(message);
+	}
+
+	private void checkBricksSpinnerHasAndShowsText(int brickPosition, String message) {
```
84n4n4: is it necessary to select all messages? spinner handling in espresso is flaky by design and doing this will cause problems.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
```diff
@@ -0,0 +1,143 @@
...
+				.checkShowsText(message);
+
+		checkBricksSpinnerHasAndShowsText(1, message);
+		checkBricksSpinnerHasAndShowsText(2, message);
+		checkBricksSpinnerHasAndShowsText(3, message);
+		checkBricksSpinnerHasAndShowsText(4, message);
+		checkBricksSpinnerHasAndShowsText(5, message);
+	}
+
+	private void createNewMessageOnSpinner(int position, String message) {
```
84n4n4: theres already 3 occurences of pretty much exactly this method in BroadcastAndWaitBrickTest, BroadcastReceiverTest, and BroadcastSendBrickTest.
this would be a nice time to literally wrap this up (in a eg. BroadcastSpinnerWrapper)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
```diff
@@ -0,0 +1,190 @@
...
+				.check(matches(isDisplayed()));
+		onView(withText(R.string.ok))
+				.perform(click());
+
+		String newObjectName = lookFileName.replace(".png", "");
+		checkBricksSpinnerHasAndShowsText(R.id.brick_point_to_spinner, 2, newObjectName);
+		checkBricksSpinnerHasAndShowsText(R.id.brick_clone_spinner, 3, newObjectName);
+	}
+
+	private void checkBricksSpinnerHasAndShowsText(int spinnerId, int brickPosition, String newSceneName) {
```
84n4n4: again, shouldnt it be enough to check if the test is set and contained in the dropdown without clicking on it.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+		intended(expectedChooserIntent);
+
+		onView(withText(R.string.new_sprite_dialog_title))
+				.check(matches(isDisplayed()));
+		onView(withText(R.string.ok))
+				.perform(click());
+
+		String newObjectName = lookFileName.replace(".png", "");
+		checkBricksSpinnerHasAndShowsText(R.id.brick_point_to_spinner, 2, newObjectName);
```
84n4n4: this doesnt check all point to bricks as the name of the testMethod suggests, but only the second.
(also if this is really a bug scenario it should be in a separate test)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
```diff
@@ -0,0 +1,190 @@
...
+		checkBricksSpinnerHasAndShowsText(R.id.brick_clone_spinner, 3, newObjectName);
```
84n4n4: the clone brick should be in its own separate test.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
```diff
@@ -0,0 +1,190 @@
...
+		Intents.release();
+		baseActivityTestRule.finishActivity();
+		StorageOperations.deleteDir(tmpPath);
+		try {
+			StorageOperations.deleteDir(new File(PathBuilder.buildProjectPath(projectName)));
+		} catch (IOException e) {
+			Log.d(getClass().getSimpleName(), "couldnt clean up project");
+		}
+
+		TestUtils.deleteProjects(this.getClass().getSimpleName());
```
84n4n4: doesnt this do the same thing as the try catch block above?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/SceneBricksTest.java
```diff
@@ -0,0 +1,115 @@
...
+		onBrickAtPosition(1)
+				.onSpinner(R.id.brick_scene_start_spinner)
+				.checkShowsText(newSceneName);
+
+		checkBricksSpinnerHasAndShowsText(R.id.brick_scene_start_spinner, 2, newSceneName);
+		checkBricksSpinnerHasAndShowsText(R.id.brick_scene_transition_spinner, 3, newSceneName);
+		checkBricksSpinnerHasAndShowsText(R.id.brick_scene_transition_spinner, 4, newSceneName);
+	}
+
+	private void checkBricksSpinnerHasAndShowsText(int spinnerId, int brickPosition, String newSceneName) {
```
84n4n4: same as already stated above

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
ndrnour: not really. test modified

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/BroadcastBricksTest.java
ndrnour: done

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
ndrnour: name of the method is changed

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PointToBrickAdditionalTest.java
ndrnour: I disagree, since I'm testing if the values of the spinner get updated after adding a new look using another brick "spinner of clone brick has no 'New..' option"

84n4n4:
tests are ok for me now, see tschranzes comments on the prod portion.

catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
```diff
@@ -825,4 +827,11 @@ public void handlePlayButton(View view) {
 		}
 		StageActivity.handlePlayButton(ProjectManager.getInstance(), this);
 	}
+
+	void notifyDataSetChanged() {
+		ScriptFragment parentFragment = (ScriptFragment) getSupportFragmentManager().findFragmentByTag(ScriptFragment.TAG);
```
thmq: you can use `getCurrentFragment()` here.

thmq: and maybe don't call it `parentFragment` because that would IMO imply that it's the parent of the current activity, which it isn't. perhaps something like `scriptFragment` because it has to be the `ScriptFragment` anyway. 

which besides, makes me wonder if you should really call the function `notifyDataSetChanged()` because it's more specific than that. it just notifies the fragment if it is the `ScriptFragment`. so maybe use something like `refeshScriptFragment()`.

catroid/src/main/java/org/catrobat/catroid/content/bricks/BroadcastMessageBrick.java
```diff
@@ -95,6 +95,7 @@ public void onNewOptionSelected() {
 					@Override
 					public void onPositiveButtonClick(DialogInterface dialog, String textInput) {
 						addItem(textInput);
+						notifyDataSetChanged(activity);
```
thmq: if you already have the activity here and if it has to be the `SpriteActivity`, which it has to be for the `ScriptFragment` to be present in it you might as well cast the activity to the `SpriteActivity` and then call the method you added to it below the `notifyDataSetChanged()` you added. would make sense to reuse it wouldn't it?

ndrnour:
@thmq @84n4n4  could you please review it again. 
thanks.

catroid/src/main/java/org/catrobat/catroid/content/bricks/BroadcastMessageBrick.java
```diff
@@ -83,15 +83,15 @@ public View getView(Context context) {
 
 	@Override
 	public void onNewOptionSelected() {
-		AppCompatActivity activity = UiUtils.getActivityFromView(view);
+		final SpriteActivity activity = (SpriteActivity) UiUtils.getActivityFromView(view);
```
thmq: please change to 
```
AppCompatActivity activity = UiUtils.getActivityFromView(view);
if (!(activity instanceof SpriteActivity)) {
    retrun;
}
```
because otherwise this is going to crash with a `ClassCastException` if the activity is not a `SpriteActivity`. and `instanceof` checks if it is `null` as well.

thmq: since you now rely on the parent activity being the `SpriteActivity` we could probably find another way to notify the other bricks to update their spinners. namely: we could just move this dialog to the `SpriteActivity` as a whole and then register this brick as listener, then we could call `refreshScriptFragment()` before or immediately after we call back to this brick from the `SpriteActivity`.

then you could get rid of `void notifyDataSetChanged(AppCompatActivity activity)` in the `BrickBaseType`, which I would absolutely prefer. plus, then it would be the same as for adding new looks or sprites etc.

ndrnour: > please change to
> 
> ```
> AppCompatActivity activity = UiUtils.getActivityFromView(view);
> if (!(activity instanceof SpriteActivity)) {
>     retrun;
> }
> ```
> because otherwise this is going to crash with a `ClassCastException` if the activity is not a `SpriteActivity`. and `instanceof` checks if it is `null` as well.

done.

ndrnour: > since you now rely on the parent activity being the `SpriteActivity` we could probably find another way to notify the other bricks to update their spinners. namely: we could just move this dialog to the `SpriteActivity` as a whole and then register this brick as listener, then we could call `refreshScriptFragment()` before or immediately after we call back to this brick from the `SpriteActivity`.
> 
> then you could get rid of `void notifyDataSetChanged(AppCompatActivity activity)` in the `BrickBaseType`, which I would absolutely prefer. plus, then it would be the same as for adding new looks or sprites etc.

since `void notifyDataSetChanged(AppCompatActivity activity)` has another 2 usages in `(SceneStartBrick.java, SceneTransitionBrick.java)`, I disagree

PR #3133, opened 2019-03-06
CATROID-180 Last access bug fixed

DinosaurierRex:


wslany:
Still not doing what it should, e.g., when you:
1. download two projects, in addition to the default project.
2. open the third project, but exit it immediately gain.
3. observe that the project is still the third project.

Also note that there is https://github.com/Catrobat/Catroid/pull/3091 attempting to fix the same problem, but also failing so far.

wslany:
Great work!

PR #3134, opened 2019-03-06
CATROID-179 catrobat test runner and assertion bricks

84n4n4:


84n4n4:
test run of this package (not yet included in the PR trigger runs of course)
https://jenkins.catrob.at/job/Catroid-ManualEmulatorTest/36/

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -160,6 +160,9 @@ public void loadProject(String projectName, Context context) throws LoadingProje
 		if (project.getCatrobatLanguageVersion() == 0.998f) {
 			project.setCatrobatLanguageVersion(0.999f);
 		}
+		if (project.getCatrobatLanguageVersion() == 0.998f) {
+			project.setCatrobatLanguageVersion(0.999f);
+		}
```
robertpainsi: This `if` is identical to the above one and will always evaluate to false :smile_cat:

84n4n4: whoopsie.. ill fix that right away

84n4n4:
new testrun:
https://jenkins.catrob.at/job/Catroid-ManualEmulatorTest/37/

robertpainsi:
Ah, I just saw that the new brick is serialized with the full package name in the code.xml
```<org.catrobat.catroid.content.bricks.AssertEqualsBrick type="AssertEqualsBrick">```

Just want to make sure that it was on purpose or if it should be added [here](https://github.com/Catrobat/Catroid/blob/da234c60aa42970cd69a7a3d97a5d8c8dbc2489f/catroid/src/main/java/org/catrobat/catroid/io/XstreamSerializer.java#L475), ping @84n4n4

PR #3136, opened 2019-03-06
CATROID-175 Fix concurrency issues in StageListener

robertpainsi:
The app crashes many times when leaving the stage due to concurrency
issues during releasing resources.

When leaving the Stage, StageListener.finish is called from the Android
UI thread, ApplicationListener methods (e.g. render or dispose) are
called from the OpenGL thread. Since finish will free up resources that
may be still used in render/dispose, the app will crash.

The proper way to release resources in libgdx is by using the
overwritten ApplicationListener.dispose method. Since all
ApplicationListener methods are called synchronously by the OpenGL
thread, moving most of the code from finish to dispose will fix all
concurrency issues.

I'm not sure if also moving the CameraManager, which is used by the
FaceDetection and maybe other features, from finish to dispose will
create new problems. So I kept it in the finish method.
___
Please, please, please don't make me write tests here :pray::sweat_smile: I know you can call `create` and `dispose` directly, wrap the dispose stuff in classes or add some getters and mocks, but still nah!…

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -419,20 +419,6 @@ public class StageListener implements ApplicationListener {
 		}
 	}
 
-	public void finish() {
-		SoundManager.getInstance().clear();
-		PhysicsShapeBuilder.getInstance().reset();
-		if (CameraManager.getInstance() != null) {
-			CameraManager.getInstance().setToDefaultCamera();
-		}
-		if (penActor != null) {
-			penActor.dispose();
-		}
-
-		embroideryList = null;
-		finished = true;
-	}
-
 	@Override
 	public void render() {
 		if (CameraManager.getInstance() != null && CameraManager.getInstance().getState() == CameraManager.CameraState.previewRunning) {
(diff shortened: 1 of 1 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -589,8 +575,24 @@ public void dispose() {
...
+		embroideryList = null;
+		if (penActor != null) {
+			penActor.dispose();
+		}
+	}
+
+	public void finish() {
+		if (CameraManager.getInstance() != null) {
+			CameraManager.getInstance().setToDefaultCamera();
+		}
```
robertpainsi: I'm not sure if there could be problems if I move the CameraManager code to the dispose method. Maybe with the FaceDetection feature?
To be honest, I haven't really looked into that. I guess it doesn't matter where the Camera is reset…

robertpainsi:
btw, if someone is interested. When leaving the stage, following methods are called:
```
render by GLThread 355
pause by AsyncTask #1
finish by AsyncTask #1
render by GLThread 355
pause by GLThread 355
dispose by GLThread 355
finish by main
pause by main
finish by main
```
So we really want to make sure everything finished :laughing: Oh, and paused

84n4n4:
we should think about putting this out as a hotfix, since right now this is by far our highest ranking crash...

84n4n4:
out of curiosity, i know about GLThread and Main, but where is AsyncTask #1 coming from?

robertpainsi:
> out of curiosity, i know about GLThread and Main, but where is AsyncTask #1 coming from?

https://github.com/Catrobat/Catroid/blob/73604c8410fdeaff1b1c5066dc9bb44eceb0bb88/catroid/src/main/java/org/catrobat/catroid/ui/dialogs/StageDialog.java#L191-L197

robertpainsi:
Moved to https://github.com/Catrobat/Catroid/pull/3146 and https://github.com/Catrobat/Catroid/pull/3147

PR #3138, opened 2019-03-07
JENKINS-297 Add a suffix for the APKs that contains the branch name a…

mfuchs:
…nd the build number.

This makes it easier to track where an APK came from.
Especially when you store multiple APKs locally.


mfuchs:
Squashed into one commit + extracted the functionality so that it can be shared between Catroid and Paintroid.

PR #3144, opened 2019-03-10
CATROID-184 Add a regular expression text extraction function

wslany:
Adds a new text function for formulas named "regular expression". In the formula editor the new function can be found at the bottom of the text functions.


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/CategoryListFragment.java
```diff
@@ -105,9 +105,11 @@
 			R.string.formula_editor_function_ceil_parameter, R.string.formula_editor_function_max_parameter,
 			R.string.formula_editor_function_min_parameter);
 	private static final List<Integer> STRING_FUNCTIONS = Arrays.asList(R.string.formula_editor_function_length,
-			R.string.formula_editor_function_letter, R.string.formula_editor_function_join);
+			R.string.formula_editor_function_letter, R.string.formula_editor_function_join,
+			R.string.formula_editor_function_regex);
```
rschluesselbauer: The function text takes two lines in the preview and it messes with the lower padding.
This is also an issue when using big font size accessability setting, so it should probably be done in another ticket.

wslany: Yeah, I saw this too, but it depends on the size of the screen --- on hull HD screen it is in one line, and on some smaller screens not just this but also several of our other functions need more than one line in other languages. So, a more general refactoring might be indicated at some point. Same also for many other places in the app. Already now there are many places where something that looks good in English does not look good in German and other languages, simply because the strings are longer (or shorter) in other languages. Sometimes the text is partly hidden and cannot be read. This problem needs a more complex and general refactoring than can be done in this ticket.

thmq:
nice :+1: one thing though: if we change what Catroid can do we also need to increase the language version. we should not forget that before the release, so it does not have to be in this PR, *BUT* I just wanted to write it down somewhere.

PR #3148, opened 2019-03-11
CATROID-185 - add project folders on sd card to filepicker

84n4n4:


catroid/src/androidTest/java/org/catrobat/catroid/test/io/filepicker/ListProjectFilesTest.java
```diff
@@ -48,23 +51,31 @@
 	private File tmpFolder;
 
 	@Before
-	public void setUp() {
+	public void setUp() throws IOException {
 		tmpFolder = new File(InstrumentationRegistry.getTargetContext().getCacheDir(), "ListProjectFilesTestTmp");
 		if (tmpFolder.exists()) {
```
thmq: good catch below! **BUT** you should be aware that you should 
`if (tmpFolder.isDirectory()) { ... }` instead of `if (tmpFolder.exists()) { ... }`
if you want to be sure it is a directory you're trying to delete (deleting a file with `StorageOperations.deleteDir(` is going to cause an IOException), if the exception is what you want you should probably think about changing it some way that it becomes more obvious that it is what you want.

catroid/src/androidTest/java/org/catrobat/catroid/test/io/filepicker/ListProjectFilesTest.java
```diff
@@ -90,4 +101,35 @@ public void testListFiles() throws IOException {
...
 		assertThat(projectFiles, not(hasItem(file4)));
 	}
+
+	@Test
+	public void testListProjectsOnExternalStorage() throws IOException {
+
+		File backpackFolder = new File(EXTERNAL_STORAGE_ROOT_DIRECTORY, "backpack");
+		assertTrue(backpackFolder.mkdirs());
+
+		File tempFolder = new File(EXTERNAL_STORAGE_ROOT_DIRECTORY, Constants.TMP_DIR_NAME);
```
thmq: it would be nice if you could change it to static imports for the constants. 

>"So when should you use static import? [..] In other words, use it when you require frequent access to static members >from one or two classes.  [..] "

from [Oracle doc](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html). besides, if want to move the constants (e.g. to `FlavoredConstants`) then you don't have to make sure that every occurrence is changed correctly.

catroid/src/androidTest/java/org/catrobat/catroid/test/io/filepicker/ListProjectFilesTest.java
```diff
@@ -90,4 +101,35 @@ public void testListFiles() throws IOException {
...
+		assertTrue(tempFolder.mkdirs());
+
+		File projectFolder1 = new File(EXTERNAL_STORAGE_ROOT_DIRECTORY, "projectFolder1");
+		assertTrue(projectFolder1.mkdirs());
+		assertTrue(new File(projectFolder1, Constants.CODE_XML_FILE_NAME).createNewFile());
```
thmq: see comment about `Constants.TMP_DIR_NAME`.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/ProjectListFragment.java
```diff
@@ -87,6 +87,30 @@
...
+	private ProjectUnzipAndImportTask.ProjectUnzipAndImportListener projectUnzipAndImportListener =
+			new ProjectUnzipAndImportTask.ProjectUnzipAndImportListener() {
+				@Override
+				public void onImportFinished(boolean success) {
+						adapter.setItems(getItemList());
+						if (!success) {
+							ToastUtil.showError(getContext(), R.string.error_import_project);
+						}
+						setShowProgressBar(false);
+					}
```
thmq: indentation is weird here :confused:. (compare it to the override below). strange that no static checks complain about it. Can you have a look at it please?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/ProjectListFragment.java
```diff
@@ -162,11 +186,29 @@ private void importProject(Intent data) {
 			return;
 		}
 
+		if (!data.getData().getScheme().equals("file")) {
+			throw new RuntimeException("importProject has to be called with a file uri. (not a content uri");
```
thmq: how about using an `IllegalArgumentException` here instead?
> "Thrown to indicate that a method has been passed an illegal or inappropriate argument."

from [Oracle doc](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html).

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/ProjectListFragment.java
```diff
@@ -162,11 +186,29 @@ private void importProject(Intent data) {
...
+		}
+
 		try {
-			File projectZip = StorageOperations
-					.copyUriToDir(getContext().getContentResolver(), data.getData(), CACHE_DIR, CACHED_PROJECT_ZIP_FILE_NAME);
-			new ProjectUnzipAndImportTask(this)
-					.execute(projectZip);
+			File cacheFile = new File(CACHE_DIR, CACHED_PROJECT_ZIP_FILE_NAME);
```
thmq: just out of curiosity: how does the file at `CACHED_PROJECT_ZIP_FILE_NAME` become a directory? and is it then a directory that ends with ".zip"? 

from what I've seen below you just directly copy the project if `src` is a directory and just create a file if it's a zip. am I missing something here?

84n4n4:
going to be released as hotfix, https://github.com/Catrobat/Catroid/pull/3151 https://github.com/Catrobat/Catroid/pull/3152

PR #3149, opened 2019-03-11
CATROID-187 Fixed issue where renaming legacy projects does not work

thmq:
- this only happend when importing projects that have
     language version 0.999. because Xstream was not
     prepared correctly.
   - now changed the rename method so that it no longer relies
     on xstream validation.


wslany:
Problems... as discussed privately.

PR #3159, opened 2019-03-13
CATROID-188 Removed legacy UserBrick code

thmq:


thmq:
> @thmq What will happen with projects on share that already contain user defined bricks inserted by our Scratch converter? Also, this must be coordinated with the Scratch converter team, since the converter still creates projects that contain such bricks. Please clarify and then let me know.

Well, that is a question for the PO board actually. do you want these to still be compatible with the new user bricks? and if yes, we should still find a way where the scratch converter converts the scratch projects to something that is consistent with the Catrobat language. AFIK you already planned on not having the same variable scope as in Scratch's user defined bricks, so this definitely is an issue that requires more discussion. 

But: I did not intend to fix anything with it, I just removed things that did not really do anything anyway, or things that have to be changed anyway. TBH I do not know how we should handle it, but just leaving 1000+ lines of legacy code that are going to be in the way of re-implementing this in the code base, cannot be the solution, can it?

wslany:
As discussed earlier today, its not necessary to support thew currently 88 projects on share that have user bricks somehow inherited from Scratch converter.

PR #3160, opened 2019-03-13
CATROID-190 Removed incorrectly sanitized string project paths.

thmq:
- we do not want to unnecessarily "sanitize" project and scene paths


robertpainsi:
- Create a project named `:/a/foo%`
- Go back to the main menu
- Tap on _Continue project_

```
java.io.FileNotFoundException: /data/user/0/org.catrobat.catroid/files/:/a/foo%/code.xml does not exist.
```

thmq:
> * Create a project named `:/a/foo%`
> * Go back to the main menu
> * Tap on _Continue project_
> 
> ```
> java.io.FileNotFoundException: /data/user/0/org.catrobat.catroid/files/:/a/foo%/code.xml does not exist.
> ```
> Tested on Emulator, Android 8.0

yes, thank you for the comment but same on develop and on master naming a progject `:/a/foo%` at least breaks renaming. so sanitizing the paths the same way as before is in fact unnecessary, right?

what I wanted in this PR was to unify the behavior, and what we would then need is a policy on how we should deal with this. 

but: if you think it would better to fix it in this PR I can close it for now and re-open once we have found and agreed on a policy. @robertpainsi what do you think?

robertpainsi:
Ah, thanks for the updated and detailed information! :+1: I only tested it against the current play store version :relaxed: 

imo this bug has the highest priority and we should address this asap. Especially before the next develop release.

> @robertpainsi what do you think?

What's easier for you :wink: I'm withdrawing my request :smile: 

> either sanitize or disallow `/` in the name

I know that sanitizing user input is a pita but imo `/` is a fundamental character and we should support it. Then there is no difference in just sanitizing a single character or a bunch of them. Also we can stick to the current sanitizing and no additional code is needed to support "legacy" projects.

So I'm voting for keeping the name sanitizing.

thmq:
> What's easier for you :wink: I'm withdrawing my request :smile:

well, it's not about what's easier for me :wink: it's also about what is better for the project. as something to consider: if we sanitize the project names internally, then we have to do it consistently, which isn't too hard if we have a java object reference of the project. we can just use a method e.g. the `getDirectory()` of the project to make sure that we get the correctly sanitized path. however, if we don't have the project, so every time we just have the project name we would have to reintroduce some static utility method that everybody would have to use then. so the safer approach would be to just disallow the usage of `/` in project and scene names.

robertpainsi:
Sanitizing is broken. So to me, this pull request is just an intermediate one where you introduced the `getDirectory` method to ease further improvements, still keeping all options open. We have to discuss and fix sanitizing before the next develop release anyway.

thmq:
another thing: `"name"` does not work either. because apparently that's a problem with the xml. so I'm starting to think we should really go back to sanitizing everything.

PR #3164, opened 2019-03-17
CATROID-168 refactoring MediaDownloadService.

PlantinumX:
-MediaDownloadService is now a foreground service if > 26
	-ToastMessage is displayed in Activity
	-Deleted unused code in some parts.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/transfers/MediaDownloadService.java
+++ b/catroid/src/main/java/org/catrobat/catroid/transfers/MediaDownloadService.java
@@ -24,85 +24,55 @@ package org.catrobat.catroid.transfers;
 
 import android.app.IntentService;
 import android.content.Intent;
-import android.os.Handler;
+import android.os.Bundle;
 import android.os.ResultReceiver;
 import android.util.Log;
 
-import org.catrobat.catroid.R;
-import org.catrobat.catroid.common.Constants;
-import org.catrobat.catroid.utils.ToastUtil;
+import org.catrobat.catroid.utils.StatusBarNotificationManager;
+import org.catrobat.catroid.web.ProgressResponseBody;
 import org.catrobat.catroid.web.ServerCalls;
-import org.catrobat.catroid.web.WebconnectionException;
 
-import java.io.IOException;
+import static org.catrobat.catroid.common.Constants.UPDATE_DOWNLOAD_ERROR;
+import static org.catrobat.catroid.common.Constants.UPDATE_DOWNLOAD_SUCCESS;
 
 public class MediaDownloadService extends IntentService {
 
 	public static final String TAG = MediaDownloadService.class.getSimpleName();
 
-	public static final String RECEIVER_TAG = "receiver";
 	public static final String URL_TAG = "url";
 	public static final String MEDIA_FILE_PATH = "path";
-
-	public ResultReceiver receiver;
-	private Handler handler;
+	public static final String PROGRAM_NAME = "programname";
+	public static final String RECEIVER = "receiver";
 
 	public MediaDownloadService() {
 		super(MediaDownloadService.class.getSimpleName());
 	}
 
 	@Override
-	public void onCreate() {
-		super.onCreate();
-		handler = new Handler();
-	}
-
-	@Override
 	protected void onHandleIntent(Intent intent) {
-		boolean result = true;
 		String url = intent.getStringExtra(URL_TAG);
 		String fileString = intent.getStringExtra(MEDIA_FILE_PATH);
-		int errorMessage = R.string.error_unknown_error;
+		String programName = intent.getStringExtra(PROGRAM_NAME);
+		final ResultReceiver receiver = intent.getParcelableExtra(RECEIVER);
+		StatusBarNotificationManager manager = StatusBarNotificationManager.getInstance();
+		final int notificationId = manager.createDownloadNotification(getApplicationContext(), programName);
+		startForeground(notificationId,
+				StatusBarNotificationManager.getInstance().showOrUpdateNotification(notificationId, 0));
+		ServerCalls.getInstance().downloadMedia(url, fileString, receiver, notificationId,
+				new ServerCalls.DownloadCallback() {
+					@Override
+					public void onSuccess() {
+						Bundle bundle = new Bundle();
+						bundle.putLong(ProgressResponseBody.TAG_PROGRESS, 100);
+						receiver.send(UPDATE_DOWNLOAD_SUCCESS, bundle);
+					}
 
-		receiver = intent.getParcelableExtra(RECEIVER_TAG);
-		try {
-			ServerCalls.getInstance().downloadMedia(url, fileString, receiver);
-		} catch (IOException ioException) {
-			Log.e(TAG, Log.getStackTraceString(ioException));
-			result = false;
-			receiver.send(Constants.UPDATE_DOWNLOAD_ERROR, null);
-		} catch (WebconnectionException webconnectionException) {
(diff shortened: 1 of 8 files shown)


catroid/src/main/java/org/catrobat/catroid/transfers/project/ResultReceiverWrapper.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.os.Handler
+import android.os.ResultReceiver
+
+class ResultReceiverWrapper(handler: Handler): ResultReceiver(handler) {
+    var resultReceiverWrapperInterface: ResultReceiverWrapperInterface? = null
+
+    override fun onReceiveResult(resultCode: Int, resultData: Bundle?) {
+        resultReceiverWrapperInterface?.onReceiveResult(resultCode, resultData)
+    }
+
```
thmq: unnecessary newline. now that some of us are really contributing kotlin code, we should really introduce some lint, checkstyle and pmd checks to find stuff like this.

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -313,8 +318,26 @@ private boolean isWhatsappInstalled() {
...
+	@Override
+	public void onReceiveResult(int resultCode, Bundle resultData) {
+		long progress;
+		if (resultCode == Constants.UPDATE_DOWNLOAD_PROGRESS) {
+			progress = resultData.getLong(ProgressResponseBody.TAG_PROGRESS);
+			updateProgressDialog(progress);
+		} else if (resultCode == Constants.UPDATE_DOWNLOAD_SUCCESS) {
+			progress = resultData.getLong(ProgressResponseBody.TAG_PROGRESS);
+			updateProgressDialog(progress);
+			ToastUtil.showSuccess(WebViewActivity.this, "Download successful");
```
thmq: There is a hardcoded string `"Download successful"`. you really must not do that, please have a look at how to correctly provide strings that are visible on the UI.

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -313,8 +318,26 @@ private boolean isWhatsappInstalled() {
...
+		} else {
+			dismissProgressDialog();
+			ToastUtil.showError(WebViewActivity.this, "Download failed");
```
thmq: same as for `"Download successful"`.

catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -313,8 +318,26 @@ private boolean isWhatsappInstalled() {
 		}
 	}
 
+	@Override
+	public void onReceiveResult(int resultCode, Bundle resultData) {
+		long progress;
+		if (resultCode == Constants.UPDATE_DOWNLOAD_PROGRESS) {
```
thmq: you could use a `switch (resultCode) { case Constants.UPDATE_DOWNLOAD_PROGRESS: ...}`. and please use static imports instead of quantified ones for imports from `Constants`. (more about static imports [here](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html). -> but in short just remove the `Constants` quantifier and press alt + enter select static import and let android studio do its magic).

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -57,7 +57,7 @@
 	private static final StatusBarNotificationManager INSTANCE = new StatusBarNotificationManager();
 	public static final String CHANNEL_ID = "pocket_code_notification_channel_id";
 
-	private int notificationId;
+	private int notificationId  = Constants.FOREGROUND_SERVICE;
```
thmq: please use a static import (see other comment about it).

catroid/src/main/java/org/catrobat/catroid/utils/StatusBarNotificationManager.java
```diff
@@ -56,8 +57,7 @@
 
 	private static final StatusBarNotificationManager INSTANCE = new StatusBarNotificationManager();
 	public static final String CHANNEL_ID = "pocket_code_notification_channel_id";
-
-	private int notificationId;
+	private int notificationId = FOREGROUND_SERVICE;
```
LPeteR90: Why do we initialize notificationId with `FOREGROUND_SERVICE` here? What is the reason for doing that? In case this is done to avoid some unwanted behavior connected to the incrementing of notificationId when creating a new notifciation, we probably should consider a clearer approach or at least a clearer naming for this constant.

catroid/src/main/java/org/catrobat/catroid/transfers/MediaDownloadService.java
```diff
@@ -24,85 +24,55 @@
...
-				public void run() {
-					ToastUtil.showSuccess(getBaseContext(), messageId);
-				}
-			});
-		}
+		String programName = intent.getStringExtra(PROGRAM_NAME);
+		final ResultReceiver receiver = intent.getParcelableExtra(RECEIVER);
+		StatusBarNotificationManager manager = StatusBarNotificationManager.getInstance();
+		final int notificationId = manager.createDownloadNotification(getApplicationContext(), programName);
+		startForeground(notificationId,
```
LPeteR90: Regarding user experience (which we briefly talked about some time in the past, I believe): We do already have a download dialog. Showing an additional notification does not really provide a better experience. 

So basically, the question is, if we need a foreground service at all here (I know, [https://jira.catrob.at/browse/CATROID-168](CATROID-168) is about turning this service into a foreground service, but since a dialog is shown to the user anyway, we might just keep this service as a background service. 

So basically this is a question for a PO (@wslany) about how we want the media download to behave. 

Related to this question: How does the ui behave in case a user navigates back from the WebViewActivity (is this even possible?) and the media download has not yet finished? This might only be testable on an emulator when setting the connection to a very low speed.

LPeteR90: Another note: I'd be ok with this not being a foreground service, since the implications would not be problematic here.

PlantinumX: First of all I agree with you about changing/(closing and creating ) a  ticket for backgorund service for the MediaDownloadService as we talked about it weeks ago. But why now??? We talked weeks before, we could have asked @wslany before opening the PR. So the work/time I invested would be less than now.... 
We talked in the retroperspective about the lack of  communication in the team -> so pls next time we should talk thorugh it before anyone in the team does an PR and afterwards the conclusion  is "hmm yeah   pls implement it different and not as it is written in the ticket".

LPeteR90: First off, sorry for the late response, I was on vacation until today. 

Of course you are right, tickets should not be changed drastically when already being in CR. This ticket was not very well defined, so this situation is on me. 

However, changing the implementation from a foreground service back to a background service is only removing one LOC (and about five LOC that are unused after that), so this should not be much work. So basically the decision if we want a foreground service (with the implied progress notification additional to the progress dialog) or a background service (ui stays the same, but service may be stopped in case of app is moved to background) can be made during PO review. 

Note that I really do like the rest of the implementation and hope it can be merged soon, but there has to be a decision (which not I, but a PO should make) on this issue first.

PR #3166, opened 2019-03-18
CATROID-186 Visual placement interface ("Place at" brick) size glitch

ndrnour:


catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -132,50 +135,66 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
-		final float scaleWidthRatio = (float) currentScreenWidth / (float) virtualScreenWidth;
-		reversedScaleWidthRatio = (float) virtualScreenWidth / (float) currentScreenWidth;
 
+		float aspectRatio = (float) virtualScreenWidth / (float) virtualScreenHeight;
+		float screenAspectRatio = ScreenValues.getAspectRatio();
+
+		float scale;
+		float ratioHeight = (float) screenHeight / (float) virtualScreenHeight;
+		float ratioWidth = (float) screenWidth / (float) virtualScreenWidth;
 		ScreenModes projectScreenMood = currentProject.getXmlHeader().getScreenMode();
```
thmq: `projectScreenMood` should be `projectScreenMode` right?

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
```diff
@@ -132,50 +135,66 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
 		if (projectScreenMood == MAXIMIZE) {
```
thmq: you could use a `switch (projectScreenMode) { case MAXIMIZE: ... } ` instead of the `if else`.

ndrnour: as you wish.

catroid/src/main/java/org/catrobat/catroid/stage/VisualPlacementActivity.java
ndrnour: yes

wslany:
Thanks!!!!

PR #3167, opened 2019-03-18
CATROID-205 - move tests that do not require an emulator to run on JVM

84n4n4:


84n4n4:
this moves a good portion of our unit tests from src/androidTest to src/test. (about half of them)
those 600+ tests are now able to run on a JVM instead of an emulator or device, and taking about 3 sec (and no overhead).

this is the first step towards a fast local executable test base enabling devs to really do a tdd approach and run the tests just inbetween writing lines of code.

PR #3176, opened 2019-03-27
CATROID-88 Devices without Flash should ignore FlashBrick

DinosaurierRex:
When camera device has no flash available, the flash brick action should simply be ignored.
e.g. emulators or tablets
Added tests for FlashUtil so CameraManager and Camera are PowerMocks for simulating environment.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
@@ -27,6 +27,7 @@ import android.graphics.SurfaceTexture;
 import android.graphics.YuvImage;
 import android.hardware.Camera;
 import android.hardware.Camera.Parameters;
+import android.support.annotation.VisibleForTesting;
 import android.util.Log;
 import android.view.ViewGroup;
 import android.view.ViewParent;
@@ -96,6 +97,11 @@ public final class CameraManager implements DeviceCameraControl, Camera.PreviewC
 		return instance;
 	}
 
+	@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+	public static void setInstance(CameraManager cameraManager) {
+		instance = cameraManager;
+	}
+
 	public static void makeInstance() {
 		instance = new CameraManager();
 	}
@@ -120,17 +126,11 @@ public final class CameraManager implements DeviceCameraControl, Camera.PreviewC
 	}
 
 	public boolean hasBackCamera() {
-		if (backCameraInformation == null) {
-			return false;
-		}
-		return true;
+		return backCameraInformation != null;
 	}
 
 	public boolean hasFrontCamera() {
-		if (frontCameraInformation == null) {
-			return false;
-		}
-		return true;
+		return frontCameraInformation != null;
 	}
 
 	private boolean hasCameraFlash(int cameraId) {
@@ -308,6 +308,14 @@ public final class CameraManager implements DeviceCameraControl, Camera.PreviewC
 
 	public boolean hasCurrentCameraFlash() {
 		return currentCameraInformation.flashAvailable;
+	}
+
+	public boolean hasBackCameraFlash() {
+		return (hasBackCamera() && backCameraInformation.flashAvailable);
+	}
+
+	public boolean hasFrontCameraFlash() {
+		return (hasFrontCamera() && frontCameraInformation.flashAvailable);
 	}
 
 	public CameraState getState() {
@@ -543,18 +551,19 @@ public final class CameraManager implements DeviceCameraControl, Camera.PreviewC
 		}
 	}
 
-	public boolean switchToCameraWithFlash() {
+	public boolean isCameraFlashAvailable() {
+		return (hasBackCameraFlash() || hasFrontCameraFlash());
+	}
+
+	public void switchToCameraWithFlash() {
 		if (hasCurrentCameraFlash()) {
-			return true;
+			return;
 		}
-		if (frontCameraInformation.flashAvailable) {
+
+		if (hasFrontCameraFlash()) {
 			updateCamera(frontCameraInformation);
-			return true;
-		}
-		if (backCameraInformation.flashAvailable) {
+		} else if (hasBackCameraFlash()) {
 			updateCamera(backCameraInformation);
-			return true;
 		}
-		return false;
 	}
 }
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
@@ -301,12 +301,8 @@ public class StageResourceHolder implements GatherCollisionInformationTask.OnPol
 
 		if (requiredResourcesSet.contains(Brick.CAMERA_FLASH)) {
 			CameraManager.makeInstance();
-			if (CameraManager.getInstance().switchToCameraWithFlash()) {
-				FlashUtil.initializeFlash();
-				resourceInitialized();
-			} else {
-				resourceFailed(Brick.CAMERA_FLASH);
-			}
+			FlashUtil.initializeFlash();
+			resourceInitialized();
 		}
 
 		if (requiredResourcesSet.contains(Brick.VIBRATOR)) {
@@ -488,10 +484,6 @@ public class StageResourceHolder implements GatherCollisionInformationTask.OnPol
 				case Brick.CAMERA_FRONT:
 					failedResourcesMessage = failedResourcesMessage + stageActivity.getString(R.string
 							.prestage_no_front_camera_available);
-					break;
-				case Brick.CAMERA_FLASH:
-					failedResourcesMessage = failedResourcesMessage + stageActivity.getString(R.string
-							.prestage_no_flash_available);
 					break;
 				case Brick.VIBRATOR:
 					failedResourcesMessage = failedResourcesMessage + stageActivity.getString(R.string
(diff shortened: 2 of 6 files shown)


catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -96,8 +96,14 @@ public static CameraManager getInstance() {
 		return instance;
 	}
 
+	public static void setInstance(CameraManager cameraManager) {
+		instance = cameraManager;
+	}
+
 	public static void makeInstance() {
-		instance = new CameraManager();
+		if (instance == null) {
```
rschluesselbauer: I agree with this, because the constructor does some work & it's unnecessary to do it for each camera permission.
There are tests that possibly rely on makeInstance() resetting the CameraManager.
I have not had time to look into it further, but it's possible that the Stage too relies on CameraManager having a clean state. Check this out and see if it's worth it!

catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
```diff
@@ -489,10 +486,11 @@ private void showResourceFailedErrorDialog() {
 					failedResourcesMessage = failedResourcesMessage + stageActivity.getString(R.string
 							.prestage_no_front_camera_available);
 					break;
-				case Brick.CAMERA_FLASH:
-					failedResourcesMessage = failedResourcesMessage + stageActivity.getString(R.string
-							.prestage_no_flash_available);
-					break;
+				// CATROID-88: Flash brick should be ignored
```
rschluesselbauer: Please remove commented out code.

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -112,6 +124,11 @@ public static void initializeFlash() {
 
 	public static void flashOn() {
 		Log.d(TAG, "flashOn()");
+
+		if (!isAvailable()) {
+			return;
+		}
+
 		if (!CameraManager.getInstance().hasCurrentCameraFlash()) {
```
rschluesselbauer: Why was if (!CameraManager.getInstance().hasCurrentCameraFlash()) not enough?

catroid/src/test/java/org/catrobat/catroid/test/utiltests/FlashUtilTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+	@Before
+	public void setUp() {
+		oldCameraManagerInstance = CameraManager.getInstance();
+	}
+
+	@After
+	public void resetInstance() {
+		CameraManager.setInstance(oldCameraManagerInstance);
+	}
```
rschluesselbauer: Setup & cleanup methods are unnecessary here. If other tests rely on CameraManager having a specific state, then they need to make sure it is in that state in their own tests.

catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -543,18 +543,19 @@ public void destroyStage() {
 		}
 	}
 
-	public boolean switchToCameraWithFlash() {
+	public boolean isCameraFlashAvailable() {
+		return (frontCameraInformation.flashAvailable || backCameraInformation.flashAvailable);
```
rschluesselbauer: If you have a device with less than 2 cameras you run into a NPE here, because frontCameraInformation / backCameraInformation would not have been initialized in the CameraManager constructor. (Happened to me on an emulated device without camera)

catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -543,18 +543,19 @@ public void destroyStage() {
...
+	}
+
+	public void switchToCameraWithFlash() {
 		if (hasCurrentCameraFlash()) {
-			return true;
+			return;
 		}
+
 		if (frontCameraInformation.flashAvailable) {
```
rschluesselbauer: Same thing goes for this if-else-construct -> NPE if frontCameraInformation is not initialized (because no front camera exists).

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
rschluesselbauer: At a second glance - this is far from ignoring the flash brick!
If the currently active camera does not have a flash, it results in us destroying the stage - IMO the ...destroyStage() part needs to go.

catroid/src/test/java/org/catrobat/catroid/test/utiltests/FlashUtilTest.java
84n4n4: NoNoNoNo!
Always reset everything your tests changed in any kind of persistant state.

this can have so many side effects, not because a follow up test could need the camera manager directly, but the camera manager has 122 usages across the project, and if any of those its getting really tricky to debug and locate the problem. (eg. the compute dialog in formula editor accesses cameraManager)

i just spent 2 hours last week with a bug like that. someone mocked stage listener in a test and didnt have a teardown, leading to another test 100 tests apart in the test suite to fail due to a null pointer exception.

rschluesselbauer: Good point, didn't think about it that way...
Sorry for misinformation then!

catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -310,6 +310,20 @@ public boolean hasCurrentCameraFlash() {
 		return currentCameraInformation.flashAvailable;
 	}
 
+	public boolean hasBackCameraFlash() {
+		if (hasBackCamera()) {
```
thmq: you can use "early return" here:

```
return hasBackCamera() && backCameraInformation.flashAvailable:
```

Java evaluates the right hand side of the condition first, so even no null pointer exception can occur here.

catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -310,6 +310,20 @@ public boolean hasCurrentCameraFlash() {
...
+			return backCameraInformation.flashAvailable;
+		}
+		return false;
+	}
+
+	public boolean hasFrontCameraFlash() {
+		if (hasFrontCamera()) {
```
thmq: see comment above.

catroid/src/main/java/org/catrobat/catroid/camera/CameraManager.java
```diff
@@ -96,6 +97,11 @@ public static CameraManager getInstance() {
 		return instance;
 	}
 
+	@VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
thmq: :+1: nice one

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -51,8 +51,13 @@ public static boolean isOn() {
 		return currentFlashValue;
 	}
 
+	public static boolean isPaused() {
+		return paused;
+	}
+
 	public static void pauseFlash() {
-		Log.d(TAG, "pauseFlash");
+		Log.d(TAG, "pauseFlash()");
```
thmq: just remove the log altogether, as this was clearly added for debugging purposes. 

I think it is just bad practice to use log for debugging instead of Android Studio's debugger. breakpoints ftw! :smile:

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -64,7 +69,8 @@ public static void pauseFlash() {
 
 	public static void resumeFlash() {
 		Log.d(TAG, "resumeFlash()");
```
thmq: please remove the log (see other comment above).

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -94,7 +100,13 @@ public static void reset() {
 	public static void initializeFlash() {
 		Log.d(TAG, "initializeFlash()");
```
thmq: please remove (as with other logs).

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -112,6 +124,7 @@ public static void initializeFlash() {
 
 	public static void flashOn() {
 		Log.d(TAG, "flashOn()");
```
thmq: please remove log here too, while you're at it.

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
```diff
@@ -123,6 +136,11 @@ public static void flashOn() {
 
 	public static void flashOff() {
 		Log.d(TAG, "flashOff()");
```
thmq: another log that can be removed (I know it's not part of your commit, but please do it anyway).

catroid/src/main/java/org/catrobat/catroid/utils/FlashUtil.java
DinosaurierRex: As in the Ticket Description says
> Expected Behavior: The Flash brick should simply be ignored on devices without a flash.
i think we shouldn't destroy the stage. There should just nothing happen if you try to activate flash.

PR #3177, opened 2019-03-27
CATROID-202 example project object renamed to bird

DinosaurierRex:


DinosaurierRex:
Took the "default_cast_project_sprites_bird_name" string from resources. Seems to be the only string which contains bird, and translations should be present.

thmq:
we need to consider think about this again, I have to talk to the PO board and check with crowdin sources, I'll close for now.

PR #3180, opened 2019-03-28
CATROID-25 Create WriteVariableOnDeviceBrick

gPathpp:


catroid/src/main/java/org/catrobat/catroid/formulaeditor/UserVariable.java
```diff
@@ -105,4 +106,18 @@ public boolean equals(Object obj) {
 	public int hashCode() {
 		return name.hashCode();
 	}
+
+	public String getDeviceValueFileName() {
+		return deviceValueFileName;
+	}
+
+	public UserVariable copyDeviceValueFileName(UserVariable variable) {
```
thmq: what is this? what are you trying to do here? 

```
varA.copyDeviceValueFileName(varB);

```
this returns the variable that you called the method on and sets `varB`'s `deviceValueFileName` on `varA.

so all you can do with it is:

```
varA.setDeviceValueFileName(varB.getDeviceValueFileName())
```

so basically this should have been a setter.

catroid/src/main/java/org/catrobat/catroid/content/bricks/WriteVariableOnDeviceBrick.java
```diff
@@ -0,0 +1,60 @@
...
+	@Override
+	protected int getSpinnerId() {
+		return R.id.write_variable_spinner;
+	}
+
+	@Override
+	public List<ScriptSequenceAction> addActionToSequence(Sprite sprite, ScriptSequenceAction sequence) {
+		if (userVariable == null || userVariable.getName() == null) {
+			userVariable = new UserVariable("NoVariableSet", Constants.NO_VARIABLE_SELECTED);
+			userVariable.setDummy(true);
```
thmq: from what I see here I would expect that this is going to write `"NoVariableSet", Constants.NO_VARIABLE_SELECTED` to your file but that is not what it does. It has almost the same check in the action to prevent that 

```
		if (userVariable == null || userVariable.isDummy()) {
			return;
		}
```
why even bother setting `.setDummy()` here? you might as well either don't add the action if `userVariable == null || userVariable.getName() == null` is true or just move the check to the action and don't do it here. 

this way it's just unnecessary code obfuscation.

thmq:
this is the third time this PR was rejected:
see #3101 and #3169

gPathpp:
@84n4n4 and @thmq,
thank you for your honest review.
I'll take everything into account and reopen the PR with a better solution.

PR #3181, opened 2019-03-28
CATROID-178 Add a function to set up a web request

stefanreichenauer:
Adds a new text function for formulas named "web". In the formula editor the new function can be found at the bottom of the text functions.


84n4n4:
- webcall done in UI thread will produce ANRs and android killing our process
- calling ".get()" on an async task waits for it to finish, so theres nothing async anymore
- invalid urls lead to instant crash
- no language version increase
- no tests
- the difficult part with this ticket is that formulas are currently atomic in pocketcode, and are evaluated from libgdx thread. so currently theres no treaded path on how to deal with a formula that suddenly takes up to a minute to compute.
- this probably would be better implemented as a brick & action, instead of a formula to utilize the scheduler and not have everything stop.
- rebase! not merge! "pull develop" is "fetch" & "merge". why are there two commits with the same message?

PR #3182, opened 2019-04-02
CATROID-222 - Catrobat tests for clone bricks

84n4n4:


84n4n4:
this contains also some bugfixes for the assertionBricks (variables were added multiple times when cloned)

wslany:
@thmq please always also code review the Catrobat test projects in Catrobat language tests, see my comments above. They should abide to the similar standards as test code written in other programming languages, e.g., regarding clean code.

thmq:
> @thmq please always also code review the Catrobat test projects in Catrobat language tests, see my comments above. They should abide to similar standards as test code written in other programming languages, e.g., regarding clean code.

sure, I will do so from now on. I did not think to do so.

84n4n4:
the suggested approach with "wait for" and other blocking mechanisms poses one major problem, that whats to be asserted is pulled into a "wait for" condition, and the assertion block is never reached if the test is failing.
eg. the test in the screenshot, if the clone creation is broken, the test would fail due to timeout after 10sec, giving the dev no useful information of what went wrong. so dev has to start debugging the test to see what is going on.
if this is solved via the regular wait, it probably creates a small runtime overhead, but if something gets broken, it fails in less than a second, and gives the feedback that instead of 3 clones, theres 0.
since were testing a event trigger based language, writing tests doesnt work 1:1 as with JUnit, were essentially blackbox testing a multithreaded application, and care has to be taken to not produce racing conditions in tests, or get the tests stuck and have them terminate by timeout when theyre actually supposed to fail with a useful assertion message.

i tried multiple versions now on an emulator, bottom line is this current PR, then in the middle the same but with all looks and screenshots removed, and top is with waits replaced with "wait for" when possible.

Screenshot from 2019-04-04 11-42-52

the most significant speedup can be found when images were removed. the .catrobat files are now roughly 4-7kb in size.

84n4n4:
ive implemented an additional "Wait till actors idle"-brick, that does exactly what one would expect from the name. its only possible to add one per program, it would be theoretically possible to add multiple to the very same script, but as one can imagine it is very very easy to create deadlocks when using more than 1 of them. (so this is only pickable once in add bricks fragment). anyways, using clones its of course easy to still create deadlocks.
this has the disadvantage of adding another brick that is not trivial to implement (as the assertion brick was) for other catrobat implementations before they can make any use of the tests.

this of course cleans up the test cases and makes sure that no other actions will happen, and in contrast to above suggested wait construct can catch issues as for example the "When clone starts" beeing called too often. (this however was also the idea in the first place with the hard 0.2 wait, to give it enough time to have the intended calls go be performed and add some slack to also make sure that nothing happens beyond that)

Notable downside of the new brick or other gating mechanisms compared to the hard waits is that we will not have any visibility via the tests if theres some slowdown in production code. So eg. if cloning suddenly takes 0.9 seconds, tests will still be green.
this limited performance guarantees could only be done via hard waits, eg. 0.1 sec in tests.
so although it might seem like were future proofing this by adding gating mechanism, *id strongly recommend to add performance assurance tests with hard wait times in them*.

regarding the test duplication:
testWhenStartAsCloneSingle is there to test that WhenStartAsClone is called.
if something breaks in this mechanism, i get quite a bit of information out of:
eg.
testWhenStartAsCloneSingle: success
testWhenStartAsCloneMultiple: fails with expected 3 actual 2
vs.
testWhenStartAsCloneSingle: fails with expected 1 actual 0
testWhenStartAsCloneMultiple: fails with expected 3 actual 2

i personally prefer to have tests for the simplest testable assertion, and then add more complex ones as separate tests. weve had bugs before where broadcast listeners were called too often or some where dropped or other weird stuff happened. these tests here are intended to catch problems like this and give the dev a better idea of what exactly is the problem.

wslany:
Very well done!!

PR #3183, opened 2019-04-03
CATROID-88 Devices without Flash should ignore FlashBrick

DinosaurierRex:
When camera device has no flash available, the flash brick action should simply be ignored.
e.g. emulators or tablets
Added tests for FlashUtil so CameraManager and Camera are PowerMocks for simulating environment.

### DIFF
--- a/catroid/build.gradle
+++ b/catroid/build.gradle
@@ -462,9 +462,9 @@ dependencies {
     testImplementation "org.mockito:mockito-core:$mockitoVersion"
     testImplementation 'org.hamcrest:hamcrest-library:1.3'
 
-    testImplementation 'org.powermock:powermock:1.6.5'
-    testImplementation 'org.powermock:powermock-module-junit4:1.7.0'
-    testImplementation 'org.powermock:powermock-api-mockito2:1.7.0RC2'
+    testImplementation 'org.powermock:powermock:1.6.6'
+    testImplementation 'org.powermock:powermock-module-junit4:2.0.0'
+    testImplementation 'org.powermock:powermock-api-mockito2:2.0.0'
 
     testImplementation 'org.robolectric:robolectric:4.2'
     testImplementation "org.robolectric:shadows-multidex:4.2"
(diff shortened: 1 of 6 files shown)


DinosaurierRex:
Reopened because:


 

>    Create a new project and add a "Turn flashlight on" brick to the background, then execute the project. The flash is correctly turned on. Exit the stage using the Android back button. The flash stops. Then use the Android back button again to exit the pause menu. Observe that, incorrectly and unexpectedly, the flash is turned on again for approximately one second, while the script with the "When scene starts" and "Turn flashlight on" bricks is already visible. This does not happen in the Play store version (Pocket Code version 0.9.56).

Now fixed.

>     Now insert a "Use camera front" between the "When scene starts" and "Turn flashlight on" bricks, then execute the project. In the version of this PR, the flash on the rear (main camera side, i.e., not the screen side) incorrectly and unexpectedly is turned on, even though the other camera was chosen. If the phone had a selfie-style front flash (see https://www.gsmarena.com/results.php3?chkFrontflash=selected&sOSes=2 for a list of the currently 477 Android phones having a front side flash) the front side flash should have been turned on. I was so far not able to test this with this new version, but I know that it worked fine on an older version as I have tested it at some point on such a device. The correct behavior on phones without a front side flash should be to simply ignore the flash brick while the front side camera is selected. The rear side flash should not have been turned on.

I changed that because i thought about, if a user wants to activate flash, it would be nice to automatically switch to that camera with built-in flash. 
However i changed that to following behaviour:

- Init
If a program contains a flash brick, the available cameras are checked in StageResourceholder. Before starting the program, the current Camera will be selected as the one with the built in flash. 
- After Starting the program
The FlashUtil is now checking if the currently selected camera has flash available. So if e.g. front camera (without flash) gets selected while program flow, the flash brick will be ignored. On the Playstore Version i also had problems with building programs like the one described above (Select Front Camera -> Activate Flash). Tried it on three different phones.

**Suggestion:**
For the future it would be nice if pull requests like this one doesn't get closed, because we will loose the whole comment/review history, which is also nice for other people to read, so they can get an insight what was discussed and changed. A solution could be to label the pull request with something like "Under discussion" so we won't loose the history and we don't have to create a completely new pull request.

wslany:
> For the future it would be nice if pull requests like this one doesn't get closed, because we will loose the whole comment/review history, which is also nice for other people to read, so they can get an insight what was discussed and changed. A solution could be to label the pull request with something like "Under discussion" so we won't loose the history and we don't have to create a completely new pull request.

You are right, sorry. I will refrain from doing so (unless warranted for another reason).

PR #3186, opened 2019-04-05
Jenkins 301 build script add curl redirect

jampo:


jampo:
Closed PR as it is not hotfix critical anymore Standalone builds are working with current share again

PR #3187, opened 2019-04-06
CATROID-134 background name not translated

Fatmangh:


wslany:
Fixed by https://jira.catrob.at/browse/CATROID-225

PR #3188, opened 2019-04-08
CATROID-178 Add a "Send web request _ and store answer in _" brick

stefanreichenauer:
Adds a new brick to set up a web request named "Send web request to _
and store answer in _"


thmq:
@stefanreichenauer why do you open a PR if we should not review it?

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,137 @@
...
+	private final OkHttpClient okHttpClient;
+	private static final String USER_AGENT = "Mozilla/5.0 (compatible; Catrobatbot/1.0; +https://catrob.at/bot)";
+
+	private Call call;
+	private String url;
+	private WeakReference<WebRequestListener> weakListenerReference;
+
+	public WebConnection() {
+		okHttpClient = new OkHttpClient();
+		okHttpClient.setConnectionSpecs(Collections.singletonList(ConnectionSpec.MODERN_TLS));
```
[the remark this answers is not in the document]
wslany: Thanks!

stefanreichenauer:
I tried **setConnectionTimeout**, **setReadTimeout** and **setWriteTimeout** together (inclusive **setRetryOnConnectionFailure(false)**) and it didn't work as expected.
Using them would require further analysis on how they should be used.

The documentation of **setFollowSslRedirects** and **setFollowRedirects** states:
_If unset, protocol redirects will be followed_
which should already be the best choice for this story.

wslany:
Understood, then it's fine. Thanks for investigating! What's with the failing tests?

stefanreichenauer:
I don't know yet what causes the tests to fail.

The stage "Legacy Tests" fails with_
_Task :catroid:startEmulator FAILED_

_FAILURE: Build failed with an exception._

Everything else looks fine now.

84n4n4:
- this crashes when no variable is set. (create new project, add brick, start -> crash)
- language version not increased
- xstream alias missing for new brick
- i strongly recommend NOT using a foreign service as default (the first joke i got started with "i bought some shoes from a drug dealer...")
- its possible to load binary files, this is a potential crash for the user variable list fragment and of course text actors
please also see inline comments

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+
+public class WebRequestAction extends Action implements WebConnection.WebRequestListener {
+
+	private static final float TIMEOUT_DURATION = 60f;
+	private static final Double ERROR_TIMEOUT = 504d;
+	private static final Double ERROR_TOO_MANY_REQUESTS = 429d;
+	private static final Double ERROR_BAD_REQUEST = 400d;
+
+	private Sprite sprite;
+	private Formula changeVariable;
```
84n4n4: this is a formula, dont call it variable

84n4n4: same goes of course also for the setter method

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	public void setUserVariable(UserVariable userVariable) {
+		if (userVariable == null) {
+			return;
+		}
+		this.userVariable = userVariable;
+	}
+
+	public void setUrl(String url) {
```
84n4n4: according to androidStudio this does not have any usages

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+	private Sprite sprite;
+	private Formula changeVariable;
+	private String url;
+	private UserVariable userVariable;
+	private WebConnectionFactory webConnectionFactory;
+
+	private boolean firstStart = true;
+	private float time = 0f;
+	private WebConnection webConnection = null;
+	private boolean isRequestFinished = false;
```
84n4n4: please use intdefs instead of boolean flags wherever possible. (eg. this could be called RequestStatus.FINISHED, RequestStatus.WAITING, RequestStatus.NOT_SENT, etc...)

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+	private static final Double ERROR_BAD_REQUEST = 400d;
+
+	private Sprite sprite;
+	private Formula changeVariable;
+	private String url;
+	private UserVariable userVariable;
+	private WebConnectionFactory webConnectionFactory;
+
+	private boolean firstStart = true;
+	private float time = 0f;
```
84n4n4: please rename to elapsedTime

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+			Log.d(getClass().getSimpleName(), "Couldn't interpret formula", interpretationException);
+		}
+	}
+
+	@Override
+	public boolean act(float delta) {
+		if (firstStart) {
+			firstStart = false;
+			webConnection = webConnectionFactory.getWebConnection(url, this);
+			boolean successfullyAddedConnection = StageActivity.webConnectionHolder.addConnectionToList(webConnection);
```
84n4n4: the webconnection doesnt care about the implementational details in the webconnection holder. if thats a list, set, or single reference makes no difference here.
why share this information in the method name?
addConnection is sufficient, and easier to read.

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+			if (!successfullyAddedConnection) {
+				userVariable.setValue(ERROR_TOO_MANY_REQUESTS);
+				return true;
+			}
+
+			boolean successfullySentWebRequest = webConnection.sendWebRequest();
```
84n4n4: this boolean variables dont add any value for readability, for me they harm readability, because the fact that both start with "succefully..." doesnt help at all.
remove them and just have the webConnection.sendWebRequest() in the if clause.

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+	private boolean firstStart = true;
+	private float time = 0f;
+	private WebConnection webConnection = null;
+	private boolean isRequestFinished = false;
+	private String response = null;
+
+	public WebRequestAction() {
+	}
+
+	public void setUserVariable(UserVariable userVariable) {
```
84n4n4: when creating a new project and just adding this brick and pressing play this crashes because userVariable will stay uninitialized.
changeVariableAction for example checks if userVariable is null at the start of act() and just returns there.

catroid/src/main/java/org/catrobat/catroid/content/actions/WebRequestAction.java
```diff
@@ -0,0 +1,156 @@
...
+		this.sprite = sprite;
+	}
+
+	public void setWebConnectionFactory(WebConnectionFactory webConnectionFactory) {
+		this.webConnectionFactory = webConnectionFactory;
+	}
+
+	public void interpretUrl() {
+		try {
+			this.url = changeVariable.interpretString(sprite);
```
84n4n4: our rule of thumb has always been to use this. only when necessary because it makes code verbose and harder to read when used excessively.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+		this.weakListenerReference = new WeakReference<>(listener);
+	}
+
+	public void setUrl(String url) {
+		this.url = url;
+	}
+
+	private synchronized WebRequestListener getListenerAndSetReferenceToNull() {
+		if (this.weakListenerReference != null) {
+			WebRequestListener listener = this.weakListenerReference.get();
```
84n4n4: our rule of thumb has always been to use this. only when necessary because it makes code verbose and harder to read when used excessively.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+				}
+			});
+		}
+	}
+
+	private void checkIfUrlIsValid() {
+		try {
+			new URL(url);
+		} catch (MalformedURLException e) {
+			throw new RuntimeException("Malformed URL: " + url, e);
```
84n4n4: why the heck a RuntimeException?
dont write check() methods that dont have a return value and throw a runtime exception instead that quite possibly nobody knows that its throwing (because its a runtimeException you dont have to add it to the signature and nobody else will be warned about this by the IDE)

either remove the method and do the try catch block on the malformedURLExecption in sendWebRequest, or have this return true or false. (and if so, rename to something meaningful like "urlValid()" and add the url to the parameter and not operate on the member directly.

catroid/src/main/java/org/catrobat/catroid/web/WebConnectionHolder.java
```diff
@@ -0,0 +1,60 @@
...
+	public synchronized boolean addConnectionToList(WebConnection connection) {
+		if (connections.size() >= MAX_CONNECTIONS) {
+			return false;
+		}
+		connections.add(connection);
+		return true;
+	}
+
+	public synchronized void removeConnectionFromList(WebConnection connection) {
+		if (connection != null) {
```
84n4n4: why the null check? all three calls to this that i can locate will not be null.

catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -495,6 +500,10 @@
 				BrickValues.REPLACE_ITEM_IN_USERLIST_INDEX));
 		dataBrickList.add(new AskBrick(context.getString(R.string.brick_ask_default_question)));
 		dataBrickList.add(new AskSpeechBrick(context.getString(R.string.brick_ask_speech_default_question)));
+
+		if (BuildConfig.DEBUG) {
```
84n4n4: we always use feature toggles for this kind of thing. because then we can enable / disable features like this from the gradle file instead of digging through the code and locating this block here and the one above.

catroid/src/main/java/org/catrobat/catroid/stage/StageLifeCycleController.java
```diff
@@ -112,6 +113,8 @@ static void stageCreate(final StageActivity stageActivity) {
 		stageActivity.stageAudioFocus = new StageAudioFocus(stageActivity);
 		stageActivity.stageResourceHolder = new StageResourceHolder(stageActivity);
 
+		StageActivity.webConnectionHolder = new WebConnectionHolder();
```
84n4n4: please dont separate the initialization of a static field into another class...

and probably the WebConnectionHolder would be placed best in the stageListener, and not static.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+
+	public void setListener(WebRequestListener listener) {
+		this.weakListenerReference = new WeakReference<>(listener);
+	}
+
+	public void setUrl(String url) {
+		this.url = url;
+	}
+
+	private synchronized WebRequestListener getListenerAndSetReferenceToNull() {
```
84n4n4: please rename to popListener or getAndDetatchListener.

but why reset to null in the first place?

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+			return listener;
+		}
+		return null;
+	}
+
+	public synchronized boolean sendWebRequest() {
+		try {
+			checkIfUrlIsValid();
+		} catch (RuntimeException e) {
+			return false;
```
84n4n4: now that i see this. why not just call the listener with the error code and have him remove the request as usual instead of the 
boolean successfullySentWebRequest = webConnection.sendWebRequest();
if clause going on in act()?

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+				.url(url)
+				.header("User-Agent", USER_AGENT)
+				.build();
+
+		call = okHttpClient.newCall(request);
+		call.enqueue(getCallback());
+		return true;
+	}
+
+	private Callback getCallback() {
```
84n4n4: getter and setter methods are for methods and stuff that existst.
youre creating something new here, so please call this "createCallback"

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+		return new Callback() {
+			@Override
+			public void onFailure(Request request, IOException e) {
```
84n4n4: why isnt this doing anything?

if theres no network connection, this will take 60s to have the action time out, while the request is already long gone, and even if network is reestablished, the connection is gone through on failure and the action will wait 60s.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+public class WebConnection {
+	private final OkHttpClient okHttpClient;
+	private static final String USER_AGENT = "Mozilla/5.0 (compatible; Catrobatbot/1.0; +https://catrob.at/bot)";
+
+	private Call call;
+	private String url;
+	private WeakReference<WebRequestListener> weakListenerReference;
+
+	public WebConnection() {
+		okHttpClient = new OkHttpClient();
```
84n4n4: okhttpclient is quite a heavy object, and it isnt meant to be initialized for every single call...

so this would be best to be moved over to the webConnectionHolder.

also after looking at the okhttpClient i found that this supports stuff like limiting amount of concurrent requests, etc. so probably worth inspecting if the whole webConncectionHolder could be refactored into just wrapping a single okHttpClient and doing the rest via configuration on OkHttpClient.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
```diff
@@ -0,0 +1,140 @@
...
+						listener.onRequestFinished(String.valueOf(response.code()));
+					}
+				}
+			}
+		};
+	}
+
+	public void cancelCall() {
+		if (call != null) {
+			okHttpClient.getDispatcher().getExecutorService().execute(new Runnable() {
```
84n4n4: why does this have to be run as a thread on some okHttpClient provided stack?

catroid/src/main/java/org/catrobat/catroid/web/WebConnectionHolder.java
```diff
@@ -0,0 +1,60 @@
...
+
+package org.catrobat.catroid.web;
+
+import android.support.annotation.VisibleForTesting;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class WebConnectionHolder {
+	private static final int MAX_CONNECTIONS = 10;
```
84n4n4: as far as i stumbled into okHttp during this review, it seems like all behaviour (custom timeout, max number connections, max number connections towards same host, etc) is provided by okHttp already

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -0,0 +1,193 @@
...
+	private Project project;
+	private UserVariable userVariable;
+	private WebConnectionFactory webConnectionFactory;
+	private WebConnection webConnection;
+
+	@Before
+	public void setUp() throws Exception {
+		PowerMockito.mockStatic(GdxNativesLoader.class);
+		testSprite = new SingleSprite("testSprite");
+		project = new Project(MockUtil.mockContextForProject(), "testProject");
```
84n4n4: this test does not require a project

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -0,0 +1,193 @@
...
+		ProjectManager.getInstance().setCurrentProject(project);
+		userVariable = new UserVariable(TEST_USERVARIABLE);
+		project.addUserVariable(userVariable);
+
+		StageActivity.webConnectionHolder = Mockito.mock(WebConnectionHolder.class);
```
84n4n4: when mocking static fields you should really at least set them to null in teardown after the test. (or this mock stays there for the remaining test run, including all tests coming later...)

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -0,0 +1,193 @@
...
+	public void testTooManyRequests() {
+		ActionFactory factory = new ActionFactory();
+		WebRequestAction action =
+				(WebRequestAction) factory.createWebRequestAction(testSprite, new Formula(TEST_URL), userVariable);
+		action.setWebConnectionFactory(webConnectionFactory);
+		Mockito.when(webConnectionFactory.getWebConnection(Mockito.anyString(), Mockito.any())).thenReturn(webConnection);
+		Mockito.when(StageActivity.webConnectionHolder.addConnectionToList(Mockito.any(WebConnection.class)))
+			.thenReturn(false);
+
+		boolean isActionFinished = action.act(0f);
```
84n4n4: remove this variable and call inline

84n4n4: same for all occurences below

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -0,0 +1,193 @@
...
+		ActionFactory factory = new ActionFactory();
+		WebRequestAction action =
+				(WebRequestAction) factory.createWebRequestAction(testSprite, new Formula(TEST_URL), userVariable);
+		action.setWebConnectionFactory(webConnectionFactory);
+		Mockito.when(webConnectionFactory.getWebConnection(Mockito.anyString(), Mockito.any())).thenReturn(webConnection);
+		Mockito.when(StageActivity.webConnectionHolder.addConnectionToList(Mockito.any(WebConnection.class)))
+			.thenReturn(true);
+		Mockito.when(webConnection.sendWebRequest()).thenReturn(true);
+
+		boolean isActionFinished = action.act(0f);
```
84n4n4: please remove boolean variable.

catroid/src/test/java/org/catrobat/catroid/test/web/WebConnectionHolderTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+		assertTrue(isConnectionSuccessfullyAddedToList);
+		assertEquals(1, webConnectionHolder.getConnections().size());
+	}
+
+	@Test
+	public void testAddElevenConnectionsToList() {
+		WebConnection webConnection = Mockito.mock(WebConnection.class);
+		WebConnectionHolder webConnectionHolder = new WebConnectionHolder();
+
```
84n4n4: please dont...
just create a new list from array with 10 mocks of webConnection and set it in the connection holder and then add one to see that the 11th returns false.

tests should have a single purpose and test case, which should be easily spottable when reading the test. this is not at all the case with this one.

catroid/src/test/java/org/catrobat/catroid/test/web/WebConnectionHolderTest.java
```diff
@@ -0,0 +1,132 @@
...
+		isConnectionSuccessfullyAddedToList = webConnectionHolder.addConnectionToList(webConnection);
+		assertTrue(isConnectionSuccessfullyAddedToList);
+		assertEquals(2, webConnectionHolder.getConnections().size());
+
+		isConnectionSuccessfullyAddedToList = webConnectionHolder.addConnectionToList(webConnection);
+		assertTrue(isConnectionSuccessfullyAddedToList);
+		assertEquals(3, webConnectionHolder.getConnections().size());
+
+		webConnectionHolder.onPause();
```
84n4n4: also check that cancelCall is called on every connection. (and test for one connection is probably sufficient and easier to read and comprehend)

catroid/src/test/java/org/catrobat/catroid/test/web/WebConnectionTest.java
```diff
@@ -0,0 +1,67 @@
...
+	@Test
+	public void testSendRequest() {
+		OkHttpClient okHttpClient = Mockito.mock(OkHttpClient.class);
+		okHttpClient.setConnectionSpecs(Collections.singletonList(ConnectionSpec.MODERN_TLS));
+		WebConnection webConnection = new WebConnection(okHttpClient);
+		Call call = Mockito.mock(Call.class);
+		Mockito.when(okHttpClient.newCall(Mockito.any(Request.class))).thenReturn(call);
+		webConnection.setUrl(BASE_URL_TEST_HTTPS);
+
+		boolean isRequestSuccessfullySend = webConnection.sendWebRequest();
```
84n4n4: please also assert that this calls enque on the call.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
stefanreichenauer: I changed the code, so that all listener operations will be done in the Callback-methods.
But after checking the code of the Call-class, I don't know if I can guarantee that only one of Callback.onResponse and Callback.onFailure are called, before the WebConnection will be destroyed by the garbage collector.

Call.canceled is volatile, but if the request was first successful, it could call the listener, while on another thread, the scene is paused and thus Callback.onFailure is called, before the action is finished.

To ensure that the listener is only called once, I wrote a pop()-method, that deletes the listener afterwards.
I don't know if the scenario I described is realistic or even possible, and there might be a better way to do, what I've tried to do.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
stefanreichenauer: I now handle the timeout via the okHttpClient, which works fine.
But limiting the amount of connection doesn't work with this user story.
Every connection, which is added after reaching the limit, is put to halt until a running connaction has finished.
The user story explicitly says to abort connections over the limit, and I failed to find a way to configure this behavior.

Please note, that Catrobat uses okHttp 2.3.0, and most google searches find documentaion on version 3.0.0 and later.

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
stefanreichenauer: Calling call.cancel() directly may result in an:
_uncaughtException: 
    android.os.NetworkOnMainThreadException_
Network operations are not allowed to run in the main thread, thus the Runnable in cancelCall().

catroid/src/main/java/org/catrobat/catroid/web/WebConnection.java
84n4n4: since version 2.3.0 is now more than 4(!) years old, an update maybe overdue? and theres the possibility that this PR could be simplified using a newer version. so please investigate if updating is an option or if this causes a lot of problems with other server calls.
also research if okhttp call provides any information if a call is enqueued, either checking through dispatcher or if call itself can provide this information. if so, one could easily cancel enqueued calls. (those would be the ones when above max calls.)

wslany:
@84n4n4 wrote:
> * i strongly recommend NOT using a foreign service as default (the first joke i got started with "i bought some shoes from a drug dealer...")

Ok that was unfortunate, but it is no problem for our users. Teenagers learn about drugs in schools, see drug dealers on tv, etc. No issue. And there should be an easily understandable interesting example. Not possible with our own. 

However, this default url should not be translatable. Is strings.xml the right place for it? 
 
> * its possible to load binary files, this is a potential crash for the user variable list fragment and of course text actors

I tried reading in a png image with the brick into a variable, and it worked fine. The content could even be shown in the data section of the formula editor. Show variable crashed it, but that is a bug of the show variable brick, and it did not destroy the project, so this is not a severe issue. And it is anyway not the fault of the web brick implementation. We can write bug tickets for it later, but they are unrelated and have low priority.

wslany:
@stefanreichenauer @84n4n4 
I wrote yesterday:
> However, this default url should not be translatable. Is strings.xml the right place for it?

I thought once more about it and changed my mind: let's keep it like it is, i.e., translatable. This allows to have localized examples, i.e., examples that work in other languages. I edited my previous comment to delete that part.

stefanreichenauer:
I resolved all review annotations and merge conflicts.
Please rereview this ticket.

catroid/src/test/java/org/catrobat/catroid/test/web/WebConnectionHolderTest.java
84n4n4: with the above comment i actually meant using Collections.nCopies, or Arrays.fill, or anything but 10 adds.

stefanreichenauer: Fixed.

wslany:
Thanks @stefanreichenauer and @84n4n4 !

PR #3190, opened 2019-04-10
CATROID-137 Update broadcast messages and scenes after adding new ones from brick

ndrnour:
update Robolectric to version 4.2.1, because version 4.2 cause a problem on windows when running the test.
new tests:
* BroadcastBricksTest.java
* PointToBrickAdditionalTest.java
* SceneBricksTest.java
* BroadcastBrickMessageUpdateTest.java

done with help of @84n4n4 , [link](https://github.com/Catrobat/Catroid/compare/fb2a6d9f2dcbd67d6971d24c43b124c09c0e8938..81b558d76d89e496518dfb502e911f346608b932) to his changes, which I copied into this PR

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
@@ -410,6 +410,7 @@ public class SpriteActivity extends BaseActivity {
 						}
 						if (onNewSpriteListener != null) {
 							onNewSpriteListener.addItem(sprite);
+							refreshScriptFragment();
 						}
 					}
 				});
@@ -824,5 +825,12 @@ public class SpriteActivity extends BaseActivity {
 			getSupportFragmentManager().popBackStack();
 		}
 		StageActivity.handlePlayButton(ProjectManager.getInstance(), this);
+	}
+
+	void refreshScriptFragment() {
+		Fragment currentFragment = getCurrentFragment();
+		if (currentFragment instanceof ScriptFragment) {
+			((ScriptFragment) currentFragment).notifyDataSetChanged();
+		}
 	}
 }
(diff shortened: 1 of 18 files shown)


ndrnour:
@84n4n4 I lost your changes to this PR #3127 after force pushing the requested changes by @thmq .
so I closed that PR and copied your changes into this PR. (if you are not okay with coping your code) let me know.

catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
```diff
@@ -459,6 +460,7 @@ private void addBackgroundFromUri(Uri uri) {
 			currentSprite.getLookList().add(look);
 			if (onNewLookListener != null) {
 				onNewLookListener.addItem(look);
+				refreshScriptFragment();
```
thmq: why do you need this? it works perfectly fine for adding looks without it (see `addLookFromUri(Uri uri)`) why is this different?

sorry for being so annoying, but I just want to be sure that we do not start any new cargo cults by things like these.

ndrnour: **you are right! it works fine. so I deleted this line.**
but I'm sure that there was a problem with updating the spinner of bricks `When background changes to` and `Set background`, cause I remember I discussed it with you °_°. 
otherwise I wouldn't touch this method.

ndrnour:
### **please don't move the ticket to PO review before I got a replay from** @84n4n4 .

catroid/src/main/java/org/catrobat/catroid/ui/SpriteActivity.java
thmq: I know that we talked about it. I don't seem to remember the exact problem with the `WhenBackgroundChangesBrick` but I know that we need it for any bricks that have a spinner containing sprites (because adding a sprite is a 2 part process, where the sprite is only added after we return to the `SpriteActivity`.

wslany:
Thanks!

PR #3192, opened 2019-04-16
CATROID-166 Refactor ProjectUploadService to foreground service

LPeteR90:
- upload service is now started as foreground service
- refactored upload service flow and upload server call
-- upload service and upload server calls now use success and error callbacks
to communicate
-- extracted unit testable projectUpload transfer from upload service
- refactored parts of notification manager and notification data class


84n4n4:
please rebase

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
```diff
@@ -0,0 +1,205 @@
...
+            sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this),
+            serverCalls = ServerCalls.getInstance()
+        ).start(
+            successCallback = { projectId ->
+                Log.v(TAG, "Upload successful")
+                stopForeground(true)
+                val notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as? NotificationManager
+                notificationManager?.notify(UPLOAD_NOTIFICATION_ID, createUploadFinishedNotification(projectName))
+
+                ToastUtil.showSuccess(this, R.string.notification_upload_finished)
```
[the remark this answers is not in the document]
LPeteR90: ToastUtil returns in case no activity is present, this does no harm when we are in the background

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
```diff
@@ -0,0 +1,206 @@
...
+
+        return NotificationCompat.Builder(applicationContext, StatusBarNotificationManager.CHANNEL_ID)
+            .setContentIntent(pendingIntent)
+            .setSmallIcon(R.drawable.ic_stat)
+            .setContentTitle(
+                applicationContext.getString(R.string.notification_upload_title_pending) + " " + programName
+            )
+            .setContentText(applicationContext.getString(R.string.notification_upload_pending))
+            .setOngoing(true)
+            .setProgress(MAX_PERCENT, 0, true)
```
dpedratscher: could you set the MAX_PERCENT also in the StatusBarNotificationManager.showOrUpdateNotification, because here also 100 is used and a beautiful constant could be useful for all uses of setProgress()

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
```diff
@@ -0,0 +1,206 @@
...
+            else -> sharedPreferences.getString(EMAIL, NO_EMAIL)
+        }
+
+        val result = if (email.equals(NO_EMAIL)) {
+            DeviceSettingsProvider.getUserEmail(this)
+        } else {
+            email
+        }
+
+        return result ?: ""
```
dpedratscher: There is no check email == null for value result. Am I right, that this is an implicit second check of the email? Why is NO_EMAIL handled different than a null email here?

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -41,27 +41,31 @@
...
+import org.catrobat.catroid.transfers.project.ProjectUploadService;
 import org.catrobat.catroid.ui.MainMenuActivity;
 
 import static org.catrobat.catroid.common.Constants.EXTERNAL_STORAGE_ROOT_EXPORT_DIRECTORY;
+import static org.catrobat.catroid.common.Constants.EXTRA_PROJECT_NAME;
 
 public final class StatusBarNotificationManager {
-	public static final String EXTRA_PROJECT_NAME = "projectName";
-	public static final int MAXIMUM_PERCENT = 100;
+	private static final int MAXIMUM_PERCENT = 100;
```
dpedratscher: why not use the MAX_PERCENT from the Constants here?

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -150,85 +135,65 @@ public int createDownloadNotification(Context context, String programName) {
 		PendingIntent pendingIntent = PendingIntent.getActivity(context, notificationId, downloadIntent,
 				PendingIntent.FLAG_CANCEL_CURRENT);
 
-		NotificationData data = new NotificationData(context, pendingIntent, R.drawable.ic_stat, programName,
-				R.string.notification_download_title_pending, R.string.notification_title_open,
-				R.string.notification_download_pending, R.string.notification_download_finished);
+		NotificationData data = new NotificationData(pendingIntent, R.drawable.ic_stat, programName,
+				context.getString(R.string.notification_download_title_pending), context.getString(R.string.notification_title_open),
+				context.getString(R.string.notification_download_pending), context.getString(R.string.notification_download_finished),
+				0, 100, false, false);
```
dpedratscher: why not use MAX_PERCENT here?

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -130,7 +114,8 @@ public int createSaveProjectToExternalMemoryNotification(Context context, String
 				context.getString(R.string.notification_save_project_to_external_storage_title_pending),
 				context.getString(R.string.notification_save_project_to_external_storage_title_open),
 				context.getString(R.string.notification_save_project_to_external_storage_pending),
-				context.getString(R.string.notification_save_project_to_external_storage_open, EXTERNAL_STORAGE_ROOT_EXPORT_DIRECTORY));
+				context.getString(R.string.notification_save_project_to_external_storage_open, EXTERNAL_STORAGE_ROOT_EXPORT_DIRECTORY),
+				0, 100, false, false);
```
dpedratscher: why not use MAX_PERCENT here?

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -150,85 +135,65 @@ public int createDownloadNotification(Context context, String programName) {
...
 
 		return notificationId++;
 	}
 
-	public Notification showOrUpdateNotification(int id, int progressInPercent) {
+	public void showOrUpdateNotification(int id, int progressInPercent) {
 		NotificationData notificationData = notificationDataMap.get(id);
 		if (notificationData == null) {
-			return null;
+			return;
```
dpedratscher: is it possible / wise to give some error information here or log something?

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -150,85 +135,65 @@ public int createDownloadNotification(Context context, String programName) {
...
-			notificationBuilder.setContentTitle(notificationData.getNotificationTitleDone())
-				.setContentText(notificationData.getNotificationTextDone())
-				.setProgress(0, 0, false)
-				.setAutoCancel(true)
-				.setContentIntent(notificationData.getPendingIntent())
-				.setOngoing(false);
-			notificationManager.notify(id, notificationBuilder.build());
+
+		notificationData.setProgressInPercent(progressInPercent);
+		if (progressInPercent < 100) {
```
dpedratscher: MAX_PERCENT

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
```diff
@@ -244,58 +209,44 @@ public void showUploadRejectedNotification(int id, int statusCode, String server
...
-	}
-
-	public void cancelNotification(int id) {
-		notificationDataMap.remove(id);
-		notificationManager.cancel(id);
-	}
-
-	public int getProgressPercent() {
-		return progressPercent;
+		return builder.build();
```
dpedratscher: if the builder.build() function really does return the progress in percent then I ask myself why the builder function does this. It's not what I would expect from a builder() function to do.

catroid/src/main/java/org/catrobat/catroid/web/requests/HttpRequests.kt
```diff
@@ -0,0 +1,83 @@
...
+private const val USER_EMAIL = "userEmail"
+private const val DEVICE_LANGUAGE = "deviceLanguage"
+private val MEDIA_TYPE_ZIPFILE = MediaType.parse("application/zip")
+private const val FILE_UPLOAD_URL = FlavoredConstants.BASE_URL_HTTPS + "api/upload/upload.json"
+private const val TEST_FILE_UPLOAD_URL_HTTP = BASE_URL_TEST_HTTPS + "api/upload/upload.json"
+
+private val serverUrl: String
+    get() {
+        return if (ServerCalls.useTestUrl) {
+            TEST_FILE_UPLOAD_URL_HTTP
```
dpedratscher: hmm, I know it's always difficult with testing, but could there be an other (quick) way to inject the testing URL? the production code should not know about test stuff.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUpload.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private fun scaleSceneScreenshots(projectName: String, sceneNames: Array<String>?) {
+        sceneNames?.mapNotNull { screenshotLoader.getScreenshotFile(projectName, it, false) }
+            ?.filter { it.exists() && it.length() > 0 }
+            ?.forEach {
+                try {
+                    ImageEditing.scaleImageFile(it, UPLOAD_IMAGE_SCALE_WIDTH, UPLOAD_IMAGE_SCALE_HEIGHT)
+                } catch (ex: FileNotFoundException) {
+                    Log.e(TAG, Log.getStackTraceString(ex))
```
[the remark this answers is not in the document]
84n4n4: share has its default screenshot thingy going on if no screenshot is contained in the uploaded file.
this can happen, eg. if one manually deletes the screenshot file, OR uploads a project before running the stage at least once.

if we want to incorporate a default or error screenshot, we should add it to every project on creation, and replace it once the stage is run and a real one is created.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUpload.kt
```diff
@@ -0,0 +1,133 @@
...
+            Log.e(TAG, Log.getStackTraceString(ioException))
+            archiveDirectory.delete()
+            null
+        }
+    }
+
+    companion object {
+        private val TAG = ProjectUpload::class.java.simpleName
+        const val UPLOAD_ZIP_ERROR = 32_202
+        const val UPLOAD_ZIP_ERROR_MESSAGE = "Failed to zip directory for upload"
```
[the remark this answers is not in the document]
LPeteR90: The reason behind putting this locally to a companion object (this translates to `public static final` in java) is that these constants are only used in this class. I suppose however, it could make sense to move them to the constants file to avoid any possible collisions with other constants.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUpload.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private fun zipProjectToArchive(projectDirectory: File, archiveDirectory: File): File? {
+        return try {
+            zipArchiver.zip(archiveDirectory, projectDirectory.listFiles())
+            archiveDirectory
+        } catch (ioException: IOException) {
+            Log.e(TAG, Log.getStackTraceString(ioException))
+            archiveDirectory.delete()
+            null
```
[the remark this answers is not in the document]
LPeteR90: Since `zipProjectToArchive` is a private function, it is only used locally in this file. In this particular case, I would prefer to not rethrow another exception since `ProjectUpload` is designed to communicate with surrounding classes via the passed success and error callbacks. What benefit compared to the current implementation would a custom exception provide? We currently do pass an error message in the error callback that tells the user exactly what went wrong.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
```diff
@@ -0,0 +1,206 @@
...
+import java.util.Locale
+
+val UPLOAD_FILE_NAME = "upload$CATROBAT_EXTENSION"
+
+class ProjectUploadService : IntentService("ProjectUploadService") {
+
+    override fun onHandleIntent(projectUploadIntent: Intent?) {
+        val intent = projectUploadIntent ?: return
+
+        val projectPath = intent.getStringExtra(EXTRA_PROJECT_PATH) ?: return
```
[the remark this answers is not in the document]
LPeteR90: Hm, you've got a point here. I'd not throw an exception here, since it could cause the whole app to crash. Rather just log an error (which can easily be done by just returning a unit function which logs the error that some of the prerequisities have not been fulfilled)

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: the diff is misleading here. the `getProgressPercent()` function has been removed completely

catroid/src/main/java/org/catrobat/catroid/web/requests/HttpRequests.kt
LPeteR90: this is basically just refactored from the java source code... After having a quick look at where the useTestUrl flag is in the project, I'd rather not fix this within this ticket, but write a separate refactoring story for this.

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: will do

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: logging would be ok, don't know about error information to the user since in this case clearly we messed up and we can't provide any useful information for a user. 

Since the NotificationManager should be refactored at some point, we'll probably restrain from holding all notification data in a map (which is just plain ugly) at some point in the hopefully near future

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUpload.kt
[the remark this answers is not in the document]
LPeteR90: Additionally, I don't think this is within the scope of the ticket, so if we really do need a default screenshot, I'd prefer having a separate ticket for this.

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: 👍

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
LPeteR90: marking this resolved since it refers to other parts of the code

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: Logging warning now

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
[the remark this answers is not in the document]
LPeteR90: logging warning now in case of early return

catroid/src/main/java/org/catrobat/catroid/utils/notifications/StatusBarNotificationManager.java
LPeteR90: done

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUploadService.kt
LPeteR90: Well, the only way we can get a null email in the result value here is that somebody put a null value for any of the keys into the shared preferences or we cannot find a google account on the device in case no email has been provided by the user. 

The `?:` operator is called elvis operator btw (see https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator) and is used here to return `result` or, if it is `null`, an empty string.

wslany:
Very nice, thanks a LOT!!!!

PR #3200, opened 2019-04-22
CATROID-240 - concurrency / task blocking issue with UploadStatusPoll…

84n4n4:
…ingTask


wslany:
Let's have the code review phase for the pull request  instead during the pull request of it into the master, since it needs to be done there anyway. I checked the functionality, and it resolves the issue.

And thanks!

PR #3202, opened 2019-04-23
CATROID-224 ListWithoutDuplicates Implementation for RV

DinosaurierRex:
Custom implementation of an ArrayList which acts like a HashSet. 

- RecyclerView-Adapter needs a list, so some functions which HashSet doesnt provide can be used. (Index-based array operations)

- Theres also an internal HashSet running in background so that checks for duplicates can be executed in O(1). List.contain(...) needs O(n).

- The tests added should check the behaviour of the custom list, to act like kind of a HashSet. Operation comparison between custom List and java HashSet. Tests are parameterized so tests for every new object that uses this list implementation can be added within 4 lines of code.

- HashSet does automatically sort the data after operations. Since we need logic of a programming language, this should be users task and we'll keep insertion order like a list.

- Since this implementation relies on the correct implementation of the equal() - function (which needs to be implemented for every "custom" object in java), i also added tests for equal() of UserVariable and UserList

ListWithoutDuplicates is currently in use for UserVariable and UserList - Lists for Project and Sprite. (see ticket description [CATROID-224](https://jira.catrob.at/browse/CATROID-224))

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
@@ -1,0 +1,138 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.content;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+public class ListWithoutDuplicates<T> extends ArrayList<T> {
+
+	private final Set<T> comparingSet = new HashSet<T>();
+
+	private Collection<T> getUniqueValues(Collection<? extends T> collection) {
+		Collection<T> uniqueValues = new ArrayList<T>();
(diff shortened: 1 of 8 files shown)


catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
```diff
@@ -0,0 +1,138 @@
...
+		Collection<T> uniqueValuesToAdd = getUniqueValues(c);
+		if (uniqueValuesToAdd.isEmpty()) {
+			return false;
+		}
+
+		try {
+			return super.addAll(index, uniqueValuesToAdd);
+		} catch (IndexOutOfBoundsException ioobException) {
+			comparingSet.removeAll(uniqueValuesToAdd);
+			throw new IndexOutOfBoundsException();
```
thmq: same as above: please pass on the original exception so that we retain the stacktrace and message.

catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	@Override
+	public void add(int index, T element) {
+		if (comparingSet.add(element)) {
+			try {
+				super.add(index, element);
+			} catch (IndexOutOfBoundsException ioobException) {
+				comparingSet.remove(element);
+				throw new IndexOutOfBoundsException();
```
thmq: nicely done! just one consideration: the `ioobException` contains the information on what went wrong (i.e. what the actual index was and the size of the list) and by catching it and then throwing a new one you lose this information.

so please change it such that the stacktrace and message of the original exception are passed on:

```
			} catch (IndexOutOfBoundsException ioobException) {
				comparingSet.remove(element);
				throw ioobException;
```

catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
```diff
@@ -0,0 +1,138 @@
...
+		super.clear();
+	}
+
+	@Override
+	public T remove(int index) {
+		try {
+			T removedObject = super.remove(index);
+			comparingSet.remove(removedObject);
+			return removedObject;
+		} catch (IndexOutOfBoundsException ioobException) {
```
thmq: same again. please throw original exception.

catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
```diff
@@ -0,0 +1,138 @@
...
+
+			T oldObject = null;
+			if (comparingSet.add(element)) {
+				oldObject = super.set(index, element);
+				comparingSet.remove(oldObject);
+			}
+
+			return oldObject;
+		} catch (IndexOutOfBoundsException ioobException) {
+			throw new IndexOutOfBoundsException();
```
thmq: same again: do not create a new empty exception but reuse the old one.

wslany:
Thanks!

PR #3207, opened 2019-04-25
CATROID-76 Share DST File

shellp:
- insert FAB and show FAB if at least two “stitch” bricks used   
- create and share embroidery file when button pressed
- simple tests


catroid/src/main/java/org/catrobat/catroid/embroidery/ShareEmbroideryFile.java
```diff
@@ -0,0 +1,83 @@
...
+
+public class ShareEmbroideryFile {
+	private static final String TAG = ShareEmbroideryFile.class.getSimpleName();
+
+	public ShareEmbroideryFile() {
+	}
+
+	public void shareEmbroideryPoints(Context context, ArrayList<PointF> stitchPoints) {
+		DSTFileGenerator dstFileGenerator = new DSTFileGenerator();
+		String filename = ProjectManager.getInstance().getCurrentProject().getName() + ".dst";
```
thmq: This could be a problem because `Project.getName()` can give you a name that is invalid for a file, e.g. `project/1`. That is why we added `Project.getDirectory()` which encodes all problemeatid special chars when we want to create the project directory. You could use 
```
FileMetaDataExtractor.encodeSpecialCharsForFileSystem(ProjectManager.getInstance().getCurrentProject().getName()))
```
or **even better**:
change the method so that you can put the filename as a parameter. this way it is way more fexible and not coupled to the `ProjectManager`. Think about it, when you do it like you do you always have to be sure that a `project` is set in the ProjectManager just for the filename. You can see it in the test that this is not a good idea because of all the setup you need for the current project to be set. This is really bad practice because it firstly requres a `ProjectManager` and a current project and it forces the programmer to implicitly do things in the correct sequence (i.e. set current project, then write file).

If on the other hand you have to get the project name, or in this case file name, first it is clearer for anyone working on this that they have to provide the file name. They can still fetch it from the current project but it is no longer the `ShareEmbroideryFile`'s problem. 

By the way, please come up with a more descriptive name for this class. Maybe something like `EmbroideryFileExporter`. And the way I see it this does two things: it writes the file and when it's done it shows an intent chooser. So we might want to split that up. However, please implement the changes I suggested above first, in the meantime I'll think about this a litte more.

catroid/src/main/java/org/catrobat/catroid/embroidery/ShareEmbroideryFile.java
```diff
@@ -0,0 +1,83 @@
...
+import org.catrobat.catroid.common.Constants;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class ShareEmbroideryFile {
+	private static final String TAG = ShareEmbroideryFile.class.getSimpleName();
+
+	public ShareEmbroideryFile() {
```
thmq: if there is only the default constructor, you don't need to put it explicitly. from wikipedia:
>"In both Java and C#, a "default constructor" refers to a nullary constructor that is automatically generated by the compiler if no constructors have been defined for the class."

catroid/src/main/res/layout/dialog_stage.xml
```diff
@@ -115,4 +115,33 @@
         android:contentDescription="@string/stage_dialog_maximize"
         android:src="@drawable/stage_dialog_button_stretch_selector" />
 
+    <LinearLayout xmlns:app="http://schemas.android.com/apk/res-auto"
+        android:theme="@style/Theme.AppCompat"
```
thmq: why do you have to set the theme here explicitly? this is a little sketchy, because theming and styling in Android is something you don't want to mess with, trust me, you can only loose. One of the reasons is that it behaves differently on every API level and that it is a bit like CSS where some properties are inherited and some are passed down, all while you can never be sure which settings have precedence over others.

shellp: the constructor of the floating action button checks for the AppCompat theme, otherwise it throws an error which says "You need to use a Theme.AppCompat theme (or descendant) with the design library."

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/EmbroideryFileExporterTest.java
```diff
@@ -0,0 +1,100 @@
...
+		ArgumentCaptor<Intent> captor = ArgumentCaptor.forClass(Intent.class);
+		Mockito.verify(context, Mockito.times(1)).startActivity(captor.capture());
+
+		Intent chooserIntent = captor.getValue();
+		Intent shareIntent = chooserIntent.getParcelableExtra(Intent.EXTRA_INTENT);
+
+		File dstFile = new File(Constants.CACHE_DIR, filename);
+		Uri uriForFile = FileProvider.getUriForFile(context, context.getPackageName() + ".fileProvider", dstFile);
+
+		Intent compareShareIntent = new Intent(Intent.ACTION_SEND, uriForFile);
```
84n4n4: can you please rename those intents to expectedWhateverIntent and actualWhateverIntent.
its really hard to read the way it is now, and really hard to comprehend which part is test setup, which is produced by production code.

also please move the assertion to the respective blocks where you create your expectedIntents.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/EmbroideryFileExporterTest.java
```diff
@@ -0,0 +1,100 @@
...
+	@Category({Level.Smoke.class})
+	public void testShareNonExistentFile() {
+		File file = new File(Constants.CACHE_DIR, filename);
+		file.delete();
+		embroideryFileExporter.shareEmbroideryFile(context, file);
+		Mockito.verify(context, Mockito.never()).startActivity(Mockito.any(Intent.class));
+	}
+
+	@Test
+	@Category({Level.Smoke.class})
```
84n4n4: we dont use categories in instrumented unit tests.

catroid/src/main/java/org/catrobat/catroid/ui/ExportLaunchers.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.io.File
+
+interface ExportLauncher {
+
+    fun startActivityForResult(requestCode: Int)
+}
+
+class ExportEmbroideryFileLauncher(
+    private val activity: StageActivity,
+    private val filename: String,
```
thmq: you don't need the name. you can just use `file.getName()` instead.

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -52,14 +52,13 @@
 @RunWith(AndroidJUnit4.class)
 public class DSTFileGeneratorTest {
 	final String projectName = DSTFileGeneratorTest.class.getSimpleName();
-	DSTFileGenerator fileGenerator;
+	DSTFileGenerator fileGenerator = new DSTFileGenerator(null);
```
84n4n4: this can be made a local variable

catroid/src/main/java/org/catrobat/catroid/stage/StageActivity.java
```diff
@@ -402,6 +403,10 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
 			IntentListener asker = intentListeners.get(requestCode);
 			asker.onIntentResult(resultCode, data);
 			intentListeners.remove(requestCode);
+		} else if (requestCode == REQUEST_EXPORT_EMBROIDERY_FILE) {
+			if (resultCode == RESULT_OK) {
+				Log.d(TAG, "Exported embroidery file");
```
84n4n4: please remove this

catroid/src/main/java/org/catrobat/catroid/ui/ExportLaunchers.kt
```diff
@@ -0,0 +1,54 @@
...
+
+class ExportEmbroideryFileLauncher(
+    private val activity: StageActivity,
+    private val file: File
+) : ExportLauncher {
+
+    override fun startActivityForResult(requestCode: Int) {
+        val fileUri = FileProvider.getUriForFile(activity, activity.packageName + ".fileProvider", file)
+        val shareIntent = Intent(Intent.ACTION_SEND)
+        shareIntent.type = "text/plain"
```
84n4n4: this mimetype "text/plain" leads to all apps being listed that can deal with plain text.
for me that is:
all messenger apps,
copy to clipboard (!!)
calenmdar,
sms,
bluetooth,
etc...

check if there is a more fitting mimetype, eg. making it csv will then only list messenger apps and apps that can deal with files to be uploaded/send somewhere.

84n4n4: i just looked it up:
application/octet-stream

would be the correct mimetype for this kind of stuff

84n4n4: facebook messenger also wouldnt allow me to send the file if it was text/plain and just doesnt do anything, and for this mimetype now only apps that can really deal with it are offered.

catroid/src/main/java/org/catrobat/catroid/ui/dialogs/StageDialog.java
```diff
@@ -117,14 +138,36 @@ public void onBackPressed() {
 		new FinishThreadAndDisposeTexturesTask().execute(null, null, null);
 	}
 
+	private void shareEmbroideryFile() {
+		if (stageListener.embroideryList.size() > 1) {
+			String filename =
+					FileMetaDataExtractor.encodeSpecialCharsForFileSystem(ProjectManager.getInstance().getCurrentProject().getName());
+			DSTFileGenerator dstFileGenerator = new DSTFileGenerator(stageListener.embroideryList);
+			File dstFile = new File(Constants.CACHE_DIR, filename + ".dst");
```
thmq: please put the ".dst" into the `Constants`.

bernadettespieler:
Thank you!!

PR #3209, opened 2019-04-27
CATROID-194 Downloaded APKs cannot be installed via notification area

ndrnour:
tested on Devices with SDK 21, 23, 26 , 
Emulators with SDK 24, 25, 26, 27, 28
not testable on emulators with SDKs below 24 because of : the web-view doesn't support the new design of share website.
if you got a msg like "could n't install the Apk || or App not installed" please disable the play protect feature from play store.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
@@ -25,10 +25,8 @@ package org.catrobat.catroid.ui;
 import android.annotation.SuppressLint;
 import android.app.DownloadManager;
 import android.app.ProgressDialog;
-import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
-import android.content.IntentFilter;
 import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.net.Uri;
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -317,6 +352,7 @@ private boolean isWhatsappInstalled() {
 	protected void onDestroy() {
 		webView.setDownloadListener(null);
 		webView.destroy();
+		//unregisterReceiver(onDownloadComplete);
```
ndrnour: as far as i know every registered receiver must be unregistered (register in onCreate(), unregister in onDestroy() || register onStart(), unregister in OnStop()) but why this case cause 2 falling test cases, see the last 2 run test.
please if you have a clear answer, share it with me.

thmq: You can just delete the receiver altogether, it works without it.

ndrnour: you are right. the main point was the missing permission. tested and everything works fine.
thank you.

thmq: sure thing! happy to help :wink: . I'll look at it again tomorrow.

wslany:
Thanks!!

PR #3212, opened 2019-05-02
CATROID-25 Add WriteVariableOnDeviceBrick

gPathpp:


catroid/src/main/java/org/catrobat/catroid/content/bricks/WriteVariableOnDeviceBrick.java
```diff
@@ -0,0 +1,53 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.content.bricks;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.content.actions.ScriptSequenceAction;
+
+public class WriteVariableOnDeviceBrick extends UserVariableBrick {
```
84n4n4: by extending upon userVariableBrick, this is also a FormulaBrick but not having a formula. this can lead to weird behaviour or crashes. (eg. longclick on the brick and select edit formula -> crash)

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
```diff
@@ -0,0 +1,190 @@
...
+				deviceVariableMap = new HashMap<>();
+			}
+
+			deviceVariableMap.put(userVariable.getDeviceValueKey(), userVariable.getValue());
+			writeMapToJson(deviceVariableMap);
+		}
+	}
+
+	private UUID generateNewKey() {
+		UUID key = UUID.randomUUID();
```
84n4n4: so, this checks only if the key is duplicate in the file, while the UUID is also only put there when the variable is written.

when copying a sprite all variables in the sprite have a UUID generated, but not written to the file. so collissions can still happen.

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		synchronized (LOCK) {
+			HashMap deviceVariableMap = readMapFromJson();
+
+			if (deviceVariableMap == null) {
+				deviceVariableMap = new HashMap<>();
+			}
+
+			deviceVariableMap.put(userVariable.getDeviceValueKey(), userVariable.getValue());
```
84n4n4: this blows up when .getValue() returns NaN

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
```diff
@@ -0,0 +1,190 @@
...
+				writeMapToJson(deviceVariableMap);
+			} catch (Exception e) {
+				Log.e(TAG, e.getMessage());
+			}
+		}
+	}
+
+	private HashMap<UUID, Object> readMapFromJson() throws FileNotFoundException {
+		Type mapType = new TypeToken<HashMap<UUID, Object>>() {}.getType();
+		return new Gson().fromJson(new FileReader(deviceVariablesFile), mapType);
```
84n4n4: this blows crashes when this is not a valid json file.

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
```diff
@@ -0,0 +1,190 @@
...
+
+	public void addKeyToUserVariable(UserVariable variable) {
+		try {
+			variable.setDeviceValueKey(generateNewKey());
+		} catch (Exception e) {
+			Log.e(TAG, e.getMessage());
+		}
+	}
+
+	public static boolean checkDeviceVariableKeyEmpty(UserVariable variable) {
```
84n4n4: please rename this to deviceVariableKeyIsEmpty, or isDeviceVariableKeyEmpty. i dont like "check" because its ambiguous

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
```diff
@@ -106,10 +110,18 @@ public Sprite copyForCloneBrick(Sprite spriteToCopy) {
 
 		sprite.getSoundList().addAll(spriteToCopy.getSoundList());
 		sprite.getNfcTagList().addAll(spriteToCopy.getNfcTagList());
+		DeviceVariableAccessor accessor = new DeviceVariableAccessor(currentProject.getDirectory());
 
-		for (UserVariable userVariable : spriteToCopy.getUserVariables()) {
-			sprite.getUserVariables().add(new UserVariable(userVariable));
+		for (UserVariable originalVariable : spriteToCopy.getUserVariables()) {
+			if (checkDeviceVariableKeyEmpty(originalVariable)) {
```
84n4n4: instead of setting those UUIDs every here and there on copy, clone, etc. why not generate them everytime? so that every user variable has one at all times?

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
```diff
@@ -106,10 +110,18 @@ public Sprite copyForCloneBrick(Sprite spriteToCopy) {
...
+				accessor.addKeyToUserVariable(originalVariable);
+			}
+
+			UserVariable copyVariable = new UserVariable(originalVariable);
+			copyVariable.setDeviceValueKey(originalVariable.getDeviceValueKey());
```
84n4n4: is there any time where we want to create a new user variable from an existing one without giving it the same UUID?
if not then this should be in the constructor.

catroid/src/main/java/org/catrobat/catroid/transfers/ProjectUploadService.java
```diff
@@ -122,7 +125,14 @@ protected void onHandleIntent(Intent intent) {
 			}
 
 			File archive = new File(CACHE_DIR, UPLOAD_FILE_NAME);
-			new ZipArchiver().zip(archive, projectDir.listFiles());
+			ArrayList<File> files = new ArrayList<>(Arrays.asList(projectDir.listFiles()));
+			for (File f: files) {
```
84n4n4: you could just create a File deviceVarFile = new File(DEVICE_VARIABL....)
and use files.remove(deviceVarFile).

file equals works on abstract path:
     * Tests this abstract pathname for equality with the given object.
     * Returns <code>true</code> if and only if the argument is not
     * <code>null</code> and is an abstract pathname that denotes the same file
     * or directory as this abstract pathname.  Whether or not two abstract
     * pathnames are equal depends upon the underlying system.  On UNIX
     * systems, alphabetic case is significant in comparing pathnames; on Microsoft Windows
     * systems it is not.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SceneController.java
```diff
@@ -135,6 +136,15 @@ public Scene copy(Scene sceneToCopy, Project dstProject) throws IOException {
 
 	public void delete(Scene sceneToDelete) throws IOException {
 		StorageOperations.deleteDir(sceneToDelete.getDirectory());
+		File projectDirectory = ProjectManager.getInstance().getCurrentProject().getDirectory();
```
84n4n4: instead of adding this removal procedure to every delete here and there id rather have a cleanup once eg. on project load.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
```diff
@@ -151,6 +163,13 @@ public void delete(Sprite spriteToDelete) {
 				Log.e(TAG, Log.getStackTraceString(e));
 			}
 		}
+		File projectDirectory = ProjectManager.getInstance().getCurrentProject().getDirectory();
```
84n4n4: this just doesnt break anything right now because backpack is not handling user variables correctly. (so this is working because of another bug)

since this delete method is also used by the backpack when one deletes a sprite in backpack, it tries to delete the same thing in the currently loaded project aswell.

catroid/src/main/java/org/catrobat/catroid/io/XStreamUserVariableConverter.java
```diff
@@ -22,25 +22,46 @@
...
 
 	@Override
-	public Object fromString(String s) {
-		return new UserVariable(s);
+	public Object doUnmarshal(Object result, HierarchicalStreamReader reader, UnmarshallingContext context) {
+		String type = reader.getAttribute(TYPE);
+		if (type != null) {
+			return super.doUnmarshal(result, reader, context);
+		}
+		Log.i(TAG, "Variable written was written with old SingleValueConverter.");
```
84n4n4: is this log output necessary?

catroid/src/main/java/org/catrobat/catroid/transfers/ProjectUploadService.java
```diff
@@ -122,7 +125,14 @@ protected void onHandleIntent(Intent intent) {
...
+				if (f.getName().equals(DEVICE_VARIABLE_JSON_FILENAME)) {
```
84n4n4: also, no single char variable names please

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
```diff
@@ -0,0 +1,190 @@
...
+			try {
+				deviceVariableMap = readMapFromJson();
+			} catch (FileNotFoundException e) {
+				return key;
+			}
+
+			if (deviceVariableMap == null) {
+				return key;
+			}
+			for (int tries = 0; deviceVariableMap.containsKey(key) && tries < MAX_GENERATE_KEY_TRIES; tries++) {
```
84n4n4: why the retries?
https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions
"For example, the number of random version-4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion"

catroid/src/main/java/org/catrobat/catroid/io/BackwardCompatibleCatrobatLanguageXStream.java
```diff
@@ -620,6 +621,9 @@ private void initializeBrickInfoMap() {
 		brickInfo = new BrickInfo(HideTextBrick.class.getSimpleName());
 		brickInfoMap.put("hideTextBrick", brickInfo);
 
+		brickInfo = new BrickInfo(WriteVariableOnDeviceBrick.class.getSimpleName());
```
84n4n4: this is unnecessary.
youre adding a new brick, theres no old version of it with a different name.

(see cameraBrick below, this is the *BackwrardCompatible"-bla class. in case you ever rename your brick after we had it released, you have to add it here to let Xstream know that the "videoBrick" in the xml is a "CameraBrick" in the codebase.)

84n4n4: by adding it with the very same name, but with a small char at the beginning, although initially unnecessary, it would prevent us from ever again removing it from this list.

please do not add new bricks here.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+	@After
+	public void tearDown() throws IOException {
+		for (String name: cleanupList) {
+			TestUtils.deleteProjects(name);
+		}
+	}
+
+	private Project createSimpleProject(String name) throws IOException {
+		DefaultProjectCreator creator = new DefaultProjectCreator();
```
84n4n4: why use the default project creator? this is time intesive and builds a project with pictures, sounds, etc and all other kinds of stuff you dont need in this test.

catroid/src/main/java/org/catrobat/catroid/content/bricks/WriteVariableOnDeviceBrick.java
gPathpp: Thank you, I copied it from HideVariableBrick, because it has also no Formula.
Just checked it and the same error occurs there as well.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+	UserVariable userVariable;
+	UserVariable localUserVariable;
+	Sprite sprite;
+	Project project;
+	Context context;
+	ArrayList<String> cleanupList = new ArrayList<>();
+
+	@Before
+	public void setUp() throws IOException {
+		context = InstrumentationRegistry.getTargetContext();
```
84n4n4: please dont save context as a member. this is only used in 2 out of the 5 tests, and just calling InstrumentationRegistry whenever you need it is fine.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.ArrayList;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(AndroidJUnit4.class)
+public class DeviceVariableAccessorTest {
+
+	UserVariable userVariable;
```
84n4n4: make members private

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+		ProjectManager.getInstance().setCurrentProject(project);
+		sprite.addUserVariable(localUserVariable);
+		project.addUserVariable(userVariable);
+		XstreamSerializer.getInstance().saveProject(project);
+		cleanupList.add(project.getName());
+	}
+
+	@Test
+	public void saveUserVariableTest() throws IOException {
```
84n4n4: this tests save *and* load

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+		DeviceVariableAccessor accessor = new DeviceVariableAccessor(project.getDirectory());
+
+		Object savedVar = userVariable.getValue();
+		Object localSavedVar = localUserVariable.getValue();
+		accessor.writeVariable(userVariable);
+		accessor.writeVariable(localUserVariable);
+		userVariable.setValue("false");
+		localUserVariable.setValue("false");
+		assertTrue(accessor.readUserVariableValue(userVariable));
```
84n4n4: please test load and save separately.
just save a variable and assert that the resulting file contains the right values

and another test where you take a file from test assets or create one and load variables from there and assert that the values are correct.

also split up local and global variables into separate tests

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+		assertTrue(accessor.readUserVariableValue(localUserVariable));
+		assertEquals(savedVar, userVariable.getValue());
+		assertEquals(localSavedVar, localUserVariable.getValue());
+	}
+
+	@Test
+	public void saveProjectTest() throws IOException, LoadingProjectException {
+		UserVariable userVariable1 = project.getUserVariable(userVariable.getName());
```
84n4n4: please test that UUID is saved with the user variable in code.xml in separation.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+		UserVariable result = project1.getUserVariable(userVariable.getName());
+		accessor = new DeviceVariableAccessor(project1.getDirectory());
+		accessor.readUserVariableValue(result);
+
+		assertEquals(result.getDeviceValueKey(), userVariable1.getDeviceValueKey());
+		assertEquals(result.getValue(), userVariable1.getValue());
+	}
+
+	@Test
+	public void copyProjectTest() throws IOException, LoadingProjectException {
```
84n4n4: is this necessary in this form?
there should be one test that asserts that the DeviceVariable file is copied with the rest of the project, and that test should not be in this file but somewhere over in the IO package where all the other tests regarding IO operations are.
no loading or saving using the accessor required.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+		accessor.readUserVariableValue(userVariableProject2);
+
+		cleanupList.add(project2.getName());
+		assertEquals(userVariable.getDeviceValueKey(), userVariableProject2.getDeviceValueKey());
+		assertEquals(userVariable.getValue(), userVariableProject2.getValue());
+	}
+
+	@Test
+	public void deleteLocaleVariableTest() throws IOException {
+		DeviceVariableAccessor accessor = new DeviceVariableAccessor(project.getDirectory());
```
84n4n4: no project required for this test

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+		UserVariable clonedVar = clone.getUserVariable(localUserVariable.getName());
+		clonedVar.setValue("false");
+		assertTrue(accessor.readUserVariableValue(clonedVar));
+		assertEquals(localUserVariable.getValue(), clonedVar.getValue());
+	}
+
+	@Test
+	public void deleteUserVariableTest() throws IOException {
+		DeviceVariableAccessor accessor = new DeviceVariableAccessor(project.getDirectory());
```
84n4n4: with all those tests its kind of hard to tell what portion is setup, and what you actually want to assert.

also, you test accessor with accessor.
please unit test.
-> file doesnt contain UUID -> read returns false
-> file contains UUID -> read returns true
-> file contains UUID -> write updates value
-> file doesnt contain UUID -> write adds new entry with value
-> file doesnt contain UUID -> delete this variable doesnt break anything
-> file contains UUID -> delete really removes it from the file
-> ... etc.

-> if something breaks, you have a unit test pointing out to you what is broken.
currently if either write or read breaks, every test is red, because nothing is tested in separation ("unit").

84n4n4:
please increase language version

84n4n4:
and we should probably discuss with wolfgang if we should add some limits for value size to be stored. the only thing that kept me from writing a program that fills up the phones storage was that the brick causes pocketcode to crash when filesize reaches 40mb because it seems like somewhere this is still not 100% thread save.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
gPathpp: What do you mean with "generate them everytime"? 
First I thought it is not necessary for every UserVariable to have a valid key (!= null).
But if the clone call comes before WriteOnDevice, the Values of the Cloned and the Original Variable would be stored with different keys. So the key must be generated before. Another way would be to gernerate the key in the very beginning of the UserVariable, but I think there are way too many places where a Constructor is called (Clone, Backpack, Copy Sprite, almost every Brick).

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
gPathpp: Yes, for example if a Sprite is copied, the names of their local vars should be the same, but not their keys.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/SpriteController.java
gPathpp: Ok. A cleanup routine every project load instead would also solve this problem, wouldn't it?

catroid/src/main/java/org/catrobat/catroid/io/XStreamUserVariableConverter.java
gPathpp: No, I will remove it.

catroid/src/main/java/org/catrobat/catroid/io/DeviceVariableAccessor.java
gPathpp: There is a chance ^^
If you say so, I will remove this function.

catroid/src/main/java/org/catrobat/catroid/io/BackwardCompatibleCatrobatLanguageXStream.java
gPathpp: I will keep that in mind.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
gPathpp: It didn't work with an "empty" project. And this seemed to me a easy way to create a smal project.
But ok, i will remove this creator.

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
```diff
@@ -0,0 +1,180 @@
...
+		ProjectManager.getInstance().setCurrentProject(project);
+		sprite.addUserVariable(localUserVariable);
+		project.addUserVariable(userVariable);
+		XstreamSerializer.getInstance().saveProject(project);
+		cleanupList.add(project.getName());
+	}
+
+	@Test
+	public void saveUserVariableTest() throws IOException {
+		DeviceVariableAccessor accessor = new DeviceVariableAccessor(project.getDirectory());
```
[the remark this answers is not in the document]
gPathpp: Great idea. Thanks!

catroid/src/androidTest/java/org/catrobat/catroid/test/common/DeviceVariableAccessorTest.java
gPathpp: How do I do that?

PR #3213, opened 2019-05-02
fixed CATROID-229 clone doesn't clone current look

coderquill:


wslany:
Pls get rid of config files.

PR #3219, opened 2019-05-10
CATROID-234 Checked for global conflicts in merging projects

HanLinAung88:


catroid/src/test/java/org/catrobat/catroid/test/content/project/ProjectMergeTest.java
```diff
@@ -0,0 +1,101 @@
...
+		project2 = new Project(mockContext, "testProject2");
+	}
+
+	@Test
+	public void testProjectGlobalVariableConflicts() {
+		project1.addUserVariable(new UserVariable("testVariable1", 1));
+		List<UserVariable> globalConflicts = ProjectManager.getInstance().getGlobalVariableConflicts(project1, project2);
+		assertEquals(globalConflicts.size(), 0);
+
+		project2.addUserVariable(new UserVariable("testVariable2", 1));
```
ThomasSchwengler: Please split up your tests. Continuing to test after assertions makes a test hard to read.

For reference:
https://xp123.com/articles/3a-arrange-act-assert/

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -304,6 +306,36 @@ private static void loadLegoEV3SettingsFromProject(Project project, Context cont
...
+	}
+
+	public List<UserList> getGlobalListConflicts(Project project1, Project project2) {
+		List<UserList> project1GlobalLists = project1.getUserLists();
+		List<UserList> project2GlobalLists = project2.getUserLists();
+		List<UserList> conflicts = new ArrayList<>();
+		for (UserList project1GlobalList : project1GlobalLists) {
+			for (UserList project2GlobalList : project2GlobalLists) {
+				if (project1GlobalList.getName().equals(project2GlobalList.getName()) && !project1GlobalList.getList()
+						.equals(project2GlobalList.getList())) {
```
dpedratscher: Be careful: UserList.equals does only check the type and name of the user list itself. It does not check every item with the value.
I think the equals function should be fixed first, but then there will emerge side effects. 
maybe write an extra equals which checks every item in the list or better fix the equals?

HanLinAung88: Hi, so I was using project1GlobalList.getList() which will give me an ArrayList to check if each element is equal in the code. I think that should work?

dpedratscher: ok, right, the ArrayList.equals should do it

wslany:
Thanks!!

PR #3221, opened 2019-05-21
CATROID-243 Fix up ListWithoutDuplicates

DinosaurierRex:
- Fix for deleting renamed UserVariables/-Lists
- Added Tests for set at Index changes
- Added Tests for Adapter Logic when renaming/deleting UserData

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
@@ -120,7 +120,10 @@ public class ListWithoutDuplicates<T> extends ArrayList<T> {
 	@Override
 	public T set(int index, T element) {
 		try {
+
 			if (element.equals(this.get(index))) {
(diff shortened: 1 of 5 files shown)


catroid/src/main/java/org/catrobat/catroid/content/ListWithoutDuplicates.java
```diff
@@ -121,6 +121,8 @@ public boolean retainAll(Collection<?> c) {
 	public T set(int index, T element) {
 		try {
 			if (element.equals(this.get(index))) {
+				comparingSet.clear();
```
LPeteR90: unrelated to the two lines that were added here: The catch block of the enclosing try block is just rethrowing and should be omitted. This is also not the only occurrence of a catch block that only rethrows an exception, please also remove the others.

84n4n4:
depends on following
https://github.com/Catrobat/Catroid/pull/3229

if 3229 is merged this one has to be closed!

PR #3222, opened 2019-05-22
CATROID-250 - missing xstream alias for number of bricks

84n4n4:


thmq:
This destroys backward compatibility. All bricks, that did not have the correct alias set in old projects break deserialization. This means we have to provide some conversion for the incorrect serialization.

wslany:
Thanks! :-)

PR #3228, opened 2019-05-27
Implemented simple merge function

singlaayush:
Merge 2 projects given they have no variable or user conflicts
https://jira.catrob.at/browse/CATROID-256

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
@@ -28,9 +28,11 @@ import android.support.test.runner.AndroidJUnit4;
 import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.common.Constants;
 import org.catrobat.catroid.content.Project;
+import org.catrobat.catroid.content.Scene;
 import org.catrobat.catroid.exceptions.CompatibilityProjectException;
 import org.catrobat.catroid.exceptions.ProjectException;
 import org.catrobat.catroid.io.StorageOperations;
+import org.catrobat.catroid.io.XstreamSerializer;
 import org.catrobat.catroid.io.ZipArchiver;
 import org.catrobat.catroid.io.asynctask.ProjectSaveTask;
 import org.catrobat.catroid.test.utils.TestUtils;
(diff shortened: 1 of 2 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
```diff
@@ -172,4 +174,46 @@ public void testLoadProjectWithInvalidNestingBrickReferences() throws IOExceptio
...
 		TestUtils.deleteProjects(PROJECT_NAME_NESTING_BRICKS);
 	}
+
+	@Test
+	public void testMergeProjectWithNoConflicts() throws IOException {
+		Project project1 = TestUtils.createProjectWithLanguageVersion(CURRENT_CATROBAT_LANGUAGE_VERSION,
+				"testProject1");
+		Project project2 = TestUtils.createProjectWithLanguageVersion(CURRENT_CATROBAT_LANGUAGE_VERSION,
+				"testProject2");
+		Scene sceneOne = new Scene("test1", project1);
```
84n4n4: stick to one naming convention, either everything is One, Two,... OR 1,2,...

catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
```diff
@@ -172,4 +174,46 @@ public void testLoadProjectWithInvalidNestingBrickReferences() throws IOExceptio
 
 		TestUtils.deleteProjects(PROJECT_NAME_NESTING_BRICKS);
 	}
+
+	@Test
+	public void testMergeProjectWithNoConflicts() throws IOException {
+		Project project1 = TestUtils.createProjectWithLanguageVersion(CURRENT_CATROBAT_LANGUAGE_VERSION,
```
84n4n4: the createProjectWithLanguageVersion was intended to enable us to create projects with old language versions.

it also creates way more than what you need, so please create your own project here.

catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
```diff
@@ -172,4 +174,46 @@ public void testLoadProjectWithInvalidNestingBrickReferences() throws IOExceptio
 
 		TestUtils.deleteProjects(PROJECT_NAME_NESTING_BRICKS);
 	}
+
+	@Test
+	public void testMergeProjectWithNoConflicts() throws IOException {
```
84n4n4: please create a new class with those tests

catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
```diff
@@ -172,4 +174,46 @@ public void testLoadProjectWithInvalidNestingBrickReferences() throws IOExceptio
...
+		project2.getSceneByName("Scene 1").setName("Scene 2");
+
+		project1.addScene(sceneOne);
+		project2.addScene(sceneTwo);
+
+		XstreamSerializer.getInstance().saveProject(project1);
+		XstreamSerializer.getInstance().saveProject(project2);
+
+		Project mergedProject = ProjectManager.mergeProjects("testProject3", project1, project2);
+		assertEquals("testProject3", mergedProject.getName());
```
84n4n4: this merely asserts that there is a new project with name "testProject3". it could be empy and this test would not fail. 
one could replace the actual piece of code under test, namely ProjectManager.mergeProjects("testProject3", project1, project2); with a simple new Project("testProject3") and this test would still be green.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -405,6 +405,37 @@ public boolean initializeDefaultProject(Context context) {
 		}
 	}
 
+	public static Project mergeProjects(String mergedProjectName, Project project1, Project project2) {
+		Project mergedProject = new Project();
+		mergedProject.setXmlHeader(project1.getXmlHeader());
+		mergedProject.getSettings().addAll(project1.getSettings());
```
84n4n4: what happens with settings of project2?

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -405,6 +405,37 @@ public boolean initializeDefaultProject(Context context) {
...
+		mergedProject.getUserVariables().addAll(project1.getUserVariables());
+		mergedProject.getUserVariables().addAll(project2.getUserVariables());
+
+		mergedProject.getUserLists().addAll(project1.getUserLists());
+		mergedProject.getUserLists().addAll(project2.getUserLists());
+
+		mergedProject.setName(mergedProjectName);
+
+		for (Scene scene : project1.getSceneList()) {
+			for (Scene checkScene : project2.getSceneList()) {
```
84n4n4: please rename "scene" and "checkScene" to "sceneInProject1" and "sceneInProject2".
it was unnecessarily hard to comprehend whats going on there and what "checkScene" means.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -405,6 +405,37 @@ public boolean initializeDefaultProject(Context context) {
...
+				if (scene.getName().equals(checkScene.getName())) {
+					return null;
+				}
+			}
+			mergedProject.addScene(scene);
```
84n4n4: please separate the adding scenes and checking their names.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
84n4n4: i now found that there already is a getSceneNames method in Project.
so a 
`for (String sceneName : project1.getSceneNames()) {
    if (project2.getSceneNames().contains(sceneName) { return null }
}`

and if that succeeds 
`mergedProject.getSceneList().addAll(project1.getSceneList())
mergedProject.getSceneList().addAll(project2.getSceneList())
`

84n4n4: however, all of this is going to leave you with a broken project since none of the files (sounds, looks) are moved, and the sprites inside the scenes that contain references to those are going to be broken.
also care must be taken before doing any changes, like updating the paths, etc since the scenes, sprites, variables, etc are references that coming from the original projects.
so if youd update the filename in any sprite and then somewhere project1 is saved, it will break it.

catroid/src/androidTest/java/org/catrobat/catroid/test/ProjectManagerTest.java
Fatmangh: Resolved

wslany:
Closing this at the request of @Fatmangh

PR #3230, opened 2019-05-29
Fix Media-Library download filename

mfuerni:
Use `URLUtil.guessFileName()` to detect correct file name for media-library downloads.
It prefers content disposition header.


mfuerni:
sorry - wrong branch and error.

PR #3234, opened 2019-06-04
CATROID-270 permissions not propagated through composite bricks

84n4n4:
- fixes problems where bricks that require resources are not added to the list.


wslany:
Thanks!!

PR #3235, opened 2019-06-05
CATROID-26 Add ReadVariableFromDeviceBrick

gPathpp:


catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
```diff
@@ -0,0 +1,74 @@
...
+	public void setUserVariable(UserVariable userVariable) {
+		this.userVariable = userVariable;
+	}
+
+	private class ReadTask extends AsyncTask<UserVariable, Void, Void> {
+
+		@Override
+		protected Void doInBackground(UserVariable[] userVariables) {
+			File projectDirectory = ProjectManager.getInstance().getCurrentProject().getDirectory();
+			DeviceVariableAccessor accessor = new DeviceVariableAccessor(projectDirectory);
```
dpedratscher: is there a good way how the DeviceVariableAccessor can know about the directory without the ReadTask calling this whole lot of functions to reach the current projectDirectory? Is it necessary that the ReadTask hast to do this?

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
```diff
@@ -0,0 +1,74 @@
...
+
+			for (UserVariable variable: userVariables) {
+				accessor.readUserVariableValue(variable);
```
dpedratscher: in writeUser Variable you have at this point an exceptionhandling - would this be a good idea here too?

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
```diff
@@ -0,0 +1,74 @@
...
+			return true;
+		}
+
+		if (firstStart) {
+			firstStart = false;
+			readActionFinished = false;
+			new ReadTask().execute(userVariable);
+		}
+
+		return readActionFinished;
```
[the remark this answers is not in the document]
gPathpp: the act function is implemented like polling.
act will be called again if it returns false. else the Action of the next brick is called.
The firstStart boolean flag is reseted to true if the brick is called again.

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
```diff
@@ -0,0 +1,74 @@
...
+import org.catrobat.catroid.io.DeviceVariableAccessor;
+
+import java.io.File;
+
+public class ReadVariableFromDeviceAction extends EventAction {
+	private UserVariable userVariable;
+	private boolean readActionFinished;
+
+	@Override
+	public boolean act(float delta) {
```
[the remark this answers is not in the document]
gPathpp: delta is the delta time between two act calls, of the same Action. Our EventAction is derivated from the abstract gdx  class Action and has to be implement it with the parameter delta.

catroid/src/main/java/org/catrobat/catroid/content/actions/WriteVariableOnDeviceAction.java
```diff
@@ -34,37 +35,47 @@
...
 		if (userVariable == null) {
 			return true;
 		}
+
 		if (firstStart) {
 			firstStart = false;
-			writeThread = new WriteThread();
-			writeThread.start();
+			writeActionFinished = false;
+			new WriteVariableOnDeviceAction.WriteTask().execute(userVariable);
```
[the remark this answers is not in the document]
gPathpp: Thanks. I've missed that.

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
gPathpp: I think it would be the same thing inside the accessor. 
And it would also make the Tests for the accessor more complicated.

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
gPathpp: Maybe, but the only thing that throws an exception in the readTask is the Gson().fromJson(..) command in the readMapFromJson funktion. And there is a check if the file is corrupted and if so some handling to it.

catroid/src/main/java/org/catrobat/catroid/content/actions/ReadVariableFromDeviceAction.java
gPathpp: The other way would be in the brick, but I think the brick doesn't know the directory of the program either and it would be the same line of code to get it.

wslany:
Re:
> @wslany wslany referenced this pull request 10 hours ago
> **CATROID-270 fix permissionsBug #3245**

This was a typo in my post there. The present pull request has nothing to do with the other.

wslany:
Thanks!!!!

PR #3238, opened 2019-06-13
Hotfix 0.9.61/274 rename upload project

DinosaurierRex:
CATROID-274 Fix for renaming projects in upload dialog

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/io/asynctask/ProjectRenameTaskTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/io/asynctask/ProjectRenameTaskTest.java
@@ -1,0 +1,91 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.io.asynctask;
+
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.catrobat.catroid.common.DefaultProjectHandler;
+import org.catrobat.catroid.content.Project;
+import org.catrobat.catroid.io.asynctask.ProjectRenameTask;
+import org.catrobat.catroid.test.utils.TestUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(AndroidJUnit4.class)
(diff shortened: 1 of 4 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/activity/ProjectUploadActivity.java
```diff
@@ -258,11 +259,13 @@ private void showProgressDialogAndUploadProject() {
...
-			ProjectLoadTask
-					.task(project.getDirectory(), this);
-			project = ProjectManager.getInstance().getCurrentProject();
+			try {
+				File renamedDirectory = ProjectRenameTask.task(project.getDirectory(), name);
+				ProjectLoadTask.task(renamedDirectory, this);
+				project = ProjectManager.getInstance().getCurrentProject();
+			} catch (IOException e) {
+				Log.e(TAG, "Creating renamed directory failed!", e);
+			}
```
dpedratscher: May be a new story and too much for a hotfix:
@DinosaurierRex can you add some info for the user, that the renaming failed? What does the user see, when it didn't work? 
@wslany  Is an info enough or should we give the chance to cancel the upload or retry?

thmq: I don't really think it is necessary to think of some elaborate workflow here. This is a hotfix after all. We can think of some user interaction handling some other time.

catroid/src/main/java/org/catrobat/catroid/io/asynctask/ProjectRenameTask.java
```diff
@@ -54,21 +54,31 @@ public ProjectRenameTask setListener(ProjectRenameListener listener) {
...
 		File dstDir = new File(projectDir.getParent(), FileMetaDataExtractor.encodeSpecialCharsForFileSystem(dstName));
 
 		try {
-			return projectDir.renameTo(dstDir)
-					&& XstreamSerializer.renameProject(new File(dstDir, CODE_XML_FILE_NAME), dstName);
+			if (projectDir.renameTo(dstDir)
+					&& XstreamSerializer.renameProject(new File(dstDir, CODE_XML_FILE_NAME), dstName)) {
+				return dstDir;
+			} else {
+				return null;
```
thmq: still not loving null returns and the null checks that are needed then. :crying_cat_face: 

how about throwing a different exception here? e.g. `throw new IOException("some cool message")`

catroid/src/main/java/org/catrobat/catroid/io/asynctask/ProjectRenameTask.java
```diff
@@ -54,21 +54,30 @@ public ProjectRenameTask setListener(ProjectRenameListener listener) {
...
+			if (projectDir.renameTo(dstDir)
+					&& XstreamSerializer.renameProject(new File(dstDir, CODE_XML_FILE_NAME), dstName)) {
+				return dstDir;
+			} else {
+				throw new IOException("Cannot rename project directory " + projectDir.getAbsolutePath() + " to " + dstName);
+			}
 		} catch (IOException e) {
-			Log.e(TAG, "Cannot rename project directory " + projectDir.getAbsolutePath() + " to " + dstName, e);
-			return false;
+			throw e;
```
thmq: please remove the `try { .. } catch ( .. )` block altogether now.

catroid/src/androidTest/java/org/catrobat/catroid/test/io/asynctask/ProjectRenameTaskTest.java
```diff
@@ -0,0 +1,90 @@
...
+	@Test
+	public void projectDirectoryRenameTest() throws IOException {
+		String absoluteRenamedDirectoryPath = defaultProject.getDirectory().getParent() + "/" + renamedProjectName;
+		File renamedDirectory = ProjectRenameTask.task(defaultProject.getDirectory(), renamedProjectName);
+		assertEquals(renamedDirectory.getAbsolutePath(), absoluteRenamedDirectoryPath);
+	}
+
+	@Test
+	public void projectDirectoryRenameSlashTest() throws IOException {
+		String absoluteRenamedDirectoryPath = defaultProject.getDirectory().getParent() + "/" + slashEncodedRenamedProject;
```
thmq: you could use:

`File expectedDirectory = new File(defaultProject.getDirectory().getParent(), slashEncodedRenamedProject);`

and then:

`assertEquals(expectedDirectory, renamedDirectory);`

wslany:
Thanks!!

PR #3245, opened 2019-06-17
CATROID-270 fix permissionsBug

gPathpp:
* fixes problems where bricks that require resources are not added to the list.

Closes #3234

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/BrickPermissionTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/BrickPermissionTest.java
@@ -1,0 +1,175 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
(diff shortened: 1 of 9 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/BrickPermissionTest.java
```diff
@@ -0,0 +1,175 @@
...
+		for (Brick brick : bricks) {
+			repeatBrick1.addBrick(brick);
+		}
+		repeatBrick0.addBrick(repeatBrick1);
+
+		script.addBrick(repeatBrick0);
+		assertRuntimePermissions();
+	}
+
+	private void assertRuntimePermissions() {
```
dpedratscher: please make the function name more expressive. something like checkAllRuntimePermissionsRequested?

catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/BrickPermissionTest.java
```diff
@@ -0,0 +1,175 @@
...
+		RepeatBrick repeatBrick1 = new RepeatBrick();
+		for (Brick brick : bricks) {
+			repeatBrick1.addBrick(brick);
+		}
+		repeatBrick0.addBrick(repeatBrick1);
+
+		script.addBrick(repeatBrick0);
+		assertRuntimePermissions();
+	}
+
```
dpedratscher: could you add a test, where not all bricks are added, thus not all permissions are needed. It would be cool to check, if only the needed permissions are requested, because we would't want the app to request all permissions all the time, only the needed ones.

ThomasSchwengler:
So am I reading this correctly, that this pull request superseedes #3234 ?

wslany:
Thanks!! 

@ThomasSchwengler: yes, it supersedes #3235.

PR #3247, opened 2019-06-18
CATROID-268 Increase supported API level to 21 (Android 5.0)

ndrnour:


ThomasSchwengler:
There might now be some `@SuppressLint("NewApi")` annotations that are not necessary anymore.

E.g.:
https://github.com/Catrobat/Catroid/blob/921e5822965c52f80ce02b88505fd9293e69063b/catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java#L255

LPeteR90:
There are two commits in this PR. Is there a specific reason for this? If there isn't, please squash your PR into a single commit.

wslany:
Thanks!!

PR #3248, opened 2019-06-18
CATROID-282 Description not updated/uploaded correctly

DinosaurierRex:
When uploading a project, you can change description in the upload dialog. But Changes were made only locally.
The web server doesn't fetch the projectDescription Tag of the HTTPRequest. It fetches the description from the code.xml, which was saved only "after" Upload. So the old description still remained in Upload (code.xml)


wslany:
What about a (local) regression test, where the upload is mocked?

ThomasSchwengler:
I asked @DinosaurierRex the same question, he told me that this would not be possible with the current implementation.

wslany:
Why not just check whether before the (mocked) upload the updated description is in the xml?

ThomasSchwengler:
As far as I have been told it's not possible to mock the upload at the moment. I will talk to @DinosaurierRex about this again.

wslany:
Thanks!!

PR #3256, opened 2019-06-24
CATROID-147 Stitching should only be possible on background

shellp:
- stitch brick only available to insert for background


catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.java
```diff
@@ -773,7 +776,7 @@ public String getBrickCategory(Brick brick, boolean isBackgroundSprite, Context
 				category = res.getString(R.string.category_cast);
 			}
 		}
-		categoryBricks = setupEmbroideryCategoryList();
+		categoryBricks = setupEmbroideryCategoryList(isBackgroundSprite);
 		for (Brick categoryBrick : categoryBricks) {
 			if (brick.getClass().equals(categoryBrick.getClass())) {
 				category = res.getString(R.string.category_embroidery);
```
DinosaurierRex: As boolean flag existed before, LGTM

PR #3258, opened 2019-06-26
CATROID-159/285: Merge Project name dialog and error check

Fatmangh:
This dialog enables the user to enter the name of the newly merged project and checks if the name is valid. Thus, it checks if the name has not already been used.


DinosaurierRex:
Untestable/Unreviewable

PR #3270, opened 2019-07-03
CATROID-293: Fix GET_TASKS permission warning

ThomasSchwengler:
This permission gets merged from libARDiscovery and is a system permission starting from 21.


wslany:
Thanks!

PR #3275, opened 2019-07-04
CATROID-260 refactor SensorTests

gPathpp:


84n4n4:
please try to directly test the sensors using mocks instead of the "MindstormNXTTestModel" and the "ConnectionDataLogger" and if those can be removed

PR #3277, opened 2019-07-05
CATROID-283 touches actor/object should work for clones

DinosaurierRex:
Touches object should also interact with clones of the referenced object


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -310,6 +312,26 @@ private void removeAllClonedSpritesFromStage() {
 		StageActivity.resetNumberOfClonedSprites();
 	}
 
+	public List<Sprite> getAllClonesOfSprite(Sprite sprite) {
+		List<Sprite> clonesOfSprite = new ArrayList<Sprite>();
+		Pattern pattern = createCloneRegexPattern(sprite.getName());
+
+		for (Sprite spriteOfStage : sprites) {
+			Matcher matcher = pattern.matcher(spriteOfStage.getName());
```
ThomasSchwengler: You could skip this if `spriteOfStage.isClone` is false

DinosaurierRex: Done :+1:

wslany:
Thanks!!

PR #3281, opened 2019-07-10
CATROID-301 "Single tap at x: y:" brick for Catrobat language tests

84n4n4:


wslany:
In Catroid/catroid/src/androidTest/assets/catrobatTestRunnerTests/success/testTapBrick.catrobat, if one adds a script that places the objectToBeTapped at the coordinates x=500, y=0 and changes the tap coordinates in the Background to the same coordinates x=500, y=0, the test incorrectly *fails*:

image

image

image

84n4n4:
been a problem with the coordinates. i played around with it now, and it works now.

84n4n4:
modified test to prevent this from reaccuring

wslany:
Further thoughts need to be invested in the way this brick is defined and implemented. As it is defined now, it does not reflect the way the word "Tap" is used in the "When tapped" and "When stage is tapped" bricks, and also not how a tap is defined on https://developer.android.com/studio/debug/dev-options under "Input".

wslany:
Thanks!!!!!!!!

PR #3282, opened 2019-07-11
CATROID-307 Upgrade Facebook SDK to v5.1.1

DinosaurierRex:
As far as i can test it seems like there are no problems with just updating the reference. But to be aware, we are currently restricted, so it's still the same error message as before.


wslany:
Thanks!!

PR #3283, opened 2019-07-12
CATROID-260 refactor SensorTests and refactor MindstormsException

gPathpp:
Previously MindstormsException was derived from RuntimeException which caused crashes. (For example if the connection was not successful)


catroid/src/androidTest/java/org/catrobat/catroid/test/devices/mindstorms/nxt/SensorTests.java
```diff
@@ -23,224 +23,165 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.internal.verification.VerificationModeFactory.times;
+
+@RunWith(Parameterized.class)
 public class SensorTests {
```
84n4n4: since this is already touched anyways, please rename to "SensorTest" (our naming convention is to have all test classes end with "Test")

how is this test different from NXTI2CUltraSonicSensorTest?

gPathpp: The I2CSensor does different Stuff on getValue. 
And is difficult to test, because it calls the sendAndReceive of the mocked Connection too often and expects different return Values.. Hence I wrote tests which tests the called functions of  NXTI2CUltraSonicSensor.getValue()

catroid/src/main/java/org/catrobat/catroid/devices/mindstorms/nxt/Command.java
```diff
@@ -63,4 +66,19 @@ public int getLength() {
...
+			return false;
+		}
+
+		Command command = (Command) obj;
+		return Arrays.equals(commandData.toByteArray(), command.commandData.toByteArray());
+	}
+
+	@Override
+	public int hashCode() {
+		return super.hashCode() * TAG.hashCode();
```
84n4n4: is this on purpose? "super" here is going to be plain java "object", since this only implements a interface and does not extend anything. (so im unsure what super.hashCode() really produces here, and i doubt its going to fulfill the requirements for hashCode (see java doc))

when i generate a hash method in android studio, i get:

	@Override
	public int hashCode() {
		return Objects.hash(commandData);
	}

which makes sense.

hashCode has to:
- return the same integer on two objects that are equal according to equals(Object obj2)
- not necessarily return different integers if two objects are not equal according to equals(Object obj), but its a good idea because this behaviour could lead to really unexpected behaviour when the object is used in any kind of hashmap (also any kind of Set and some other collection objects because those use hashCode() )

gPathpp: You are right, this line makes more sense.
I've copied the other line from Sprite.
Should I refactor it this there to?

wslany:
Thanks!!

PR #3286, opened 2019-07-15
CATROID-254: Investigate XStream compatibility

ThomasSchwengler:
* Remove 64-bit excludes
* Add 64 bit libraries for gdx and mididriver
* Upgrade mididriver to 1.16 (1.17 seems to collide with our android manifest)
* Clean up jar file includes
  * Remove unneeded flowtextview
  * Remove unneeded jarchivelib
  * Remove unneeded testrunner
  * Move remaining jars to ./catroid/libs
* Upgrade xstream to 1.4.11.1
* Regarding xstream compatibility:
  * Confirm that all warnings are greylist-allowed
  * Fix a security warning from xstream by setting an allowed wildcard
  * Remove armeabi includes (They were removed later anyways)


wslany:
Thanks!

PR #3297, opened 2019-07-20
CATROID-210 Only show place at dialog for numbers

swlz:
- edited showFormulaEditorToEditFormula in order to go directly to the formula editor when a parameter contains a formula
- changed test PlaceAtBrickTest.java

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PlaceAtBrickTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PlaceAtBrickTest.java
@@ -40,6 +40,7 @@ import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
 import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.Espresso.pressBack;
 import static android.support.test.espresso.action.ViewActions.click;
 import static android.support.test.espresso.assertion.ViewAssertions.matches;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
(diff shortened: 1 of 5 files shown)


thmq:
please add the ticket identifier to the commit message, such as: `CATROID-210 ...`

swlz:
Sure, sorry. Just changed it.

swlz:
test fails because of issues with org.gradle.jvmargs=-XX:MaxPermSize=1024m -Xmx4096m in gradle.properties. It works when deleting the -Xmx4096m setting. @jampo is already informed.

thmq:
what's going on here? as far as I know, the `org.gradle.jvmargs` change is required for this PR to run on Jenkins, however this is not something a core team developer should have to change. this should be done by someone from the Jenkins team, e.g. by @jampo right? so is there any progress? have not really heard a definitive answer yet. because this is the only thing blocking this PR.

if there is no definite response from the Jenkins team, then I have an alternative solution:
@swlz since you are changing some of the same functionality in #3308 there are going to be merge conflicts with this PR anyway. + you need to remove the jvmarg setting from this PR. so if you don't want to wait for the jvmargs problem to be resolved you could close this PR and either wait for the other PR to be merged first and then add the changes from this PR to a new one ontop of the other or just add another commit to the other PR with the changes here and a separate commit message. 

since the tickets depend on each other I'd be fine with either solution.

thmq:
just now realized that the second commit on this PR was not by @swlz but by @jampo. please don't do things like these. if you want change the config please open a separate PR. and I understand that this was mos likely because you wanted to try the setting. but then just remove it again, so that we can continue with handling this PR properly. we have defined a concrete git workflow and this does not conform to it. so this just causes problems of all sorts. 

in any case I sugesst we close this PR and add the solution to the 210 issue to the other one or just move it to a new PR.

I'll close this one for now.

PR #3299, opened 2019-07-22
CATROID-284 touches Actor object should work for clones of current object

DinosaurierRex:
- Adding of all objects incl. current object should be allowed for "touches Actor or Object" formula
- Collision should be detected between a clone of current object and current object
- Collision should be detected between clones of current object
- Added two catrobat tests


catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaParser.java
```diff
@@ -273,25 +273,7 @@ private FormulaElement userVariable() throws InternFormulaParserException {
 		return lookTree;
 	}
 
-	private FormulaElement collision() throws InternFormulaParserException {
-
-		String firstSpriteName = ProjectManager.getInstance().getCurrentSprite().getName();
```
ThomasSchwengler: Thank you for removing this bs

wslany:
Thanks!

PR #3302, opened 2019-07-23
CATROID-296 Added Pick Color option for PhiroRGBLightBrick

swlz:
- Used behavior of ShowFormulaEditorStrategy to implement Pick Color option
- deleted unused import catroid/src/main/java/org/catrobat/catroid/ui/fragment/ColorSeekbar.java (getCustomView Function) in PhiroRGBLightBrick.java and SetPenColorBrick.java
- edited test for PhiroRGBLightColorBrick


ThomasSchwengler:
Please rebase to develop

catroid/src/main/java/org/catrobat/catroid/content/bricks/PhiroRGBLightBrick.java
```diff
@@ -32,15 +34,20 @@
 import org.catrobat.catroid.R;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.content.actions.ScriptSequenceAction;
+import org.catrobat.catroid.content.strategy.ShowColorPickerFormulaEditorStrategy;
+import org.catrobat.catroid.content.strategy.ShowFormulaEditorStrategy;
 import org.catrobat.catroid.formulaeditor.Formula;
 import org.catrobat.catroid.formulaeditor.FormulaElement;
+import org.catrobat.catroid.formulaeditor.InterpretationException;
+import org.catrobat.catroid.ui.UiUtils;
 import org.catrobat.catroid.ui.fragment.ColorSeekbar;
```
wslany: @ThomasSchwengler is catroid/src/main/java/org/catrobat/catroid/ui/fragment/ColorSeekbar.java still needed? I guess not, right? Better eliminate it now in this PR I would say, since it was only used in the "Set pen color" and the "Set Phiro led color" bricks, and both have now been converted to the new color picker. Or am I wrong? And is there anything else that can be eliminated alongside that also was needed only for the old color selection interface?

thmq: @wslany: yes, your're right, it is unused. that's a particularly good catch, because it isn't obvious from the code. before the new colorpicker was introduced, the seekbar fragments were actually displayed through the `FormulaEditorFragment`. Bricks that provided the option to edit their formulas through something other than the regular formulaeditor UI, called `FormulaEditorFragment.showCustomFragment(..)` this call then set a boolean switch in the `FormulaEditorFragment` (which is pretty sketchy, since it is defacto a singleton) and then the `FormulaEditorFragment` used the brick's `.getCustomView(..)` method to display the UI component (in case of bricks with color selecion the three seekbars).
the new colorpicker however, works completely differently. it disregards the `FormulaEditorFragment` entirely and displays the picker in a different way, however the overrides for the `.getCustomView(..)`  were not removed and still held the code necessary to display the old version of the picker, i.e., the three seekbars. (this is btw true for all bricks that used it, not only the one that was changed here). however, the method was never called so there was no problem, except that the `ColorSeekbar` still appeared to be used.

IMO the switch in the `FormulaEditorFragment` should be removed alltogether, BUT for now it is still used by bricks with a single seekbar (phiro motor bricks). so, it can't be done in this PR, but we should consider removing in in some other PR.

TL;DR: nice catch, appeared to be used in Android Studio because of some black magic in the formula editor and because override was cargo-culted. already told @swlz about it. should be gone soon + there is a potential follow-up ticket to remove some more unusedish stuff

wslany:
👍Thanks! 

@thmq @ThomasSchwengler Looks good to me now, but can one of you just check it and confirm that it's good to go?

> there is a potential follow-up ticket to remove some more unusedish stuff

@thmq Could you formulate that ticket?

ThomasSchwengler:
I will have a look at it asap

wslany:
Thanks!!

PR #3308, opened 2019-07-25
CATROID-349/ CATROID-210 Fix negative number in brickfield is number / Only show place at for numbers

swlz:
- created new function (isElementANumber) in FormulaElement - works for positive and negative numbers and is used in PhiroRGBLightBrick.java and SetPenColorBrick.java
- updated tests SetPenColorBrickTest and PhiroColorBrickTest

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
@@ -32,7 +32,6 @@ import org.catrobat.catroid.content.actions.ScriptSequenceAction;
 import org.catrobat.catroid.content.strategy.ShowColorPickerFormulaEditorStrategy;
 import org.catrobat.catroid.content.strategy.ShowFormulaEditorStrategy;
 import org.catrobat.catroid.formulaeditor.Formula;
-import org.catrobat.catroid.formulaeditor.FormulaElement;
 import org.catrobat.catroid.formulaeditor.InterpretationException;
 import org.catrobat.catroid.ui.UiUtils;
 
@@ -54,7 +53,7 @@ public class SetPenColorBrick extends FormulaBrick {
 		this(new Formula(red), new Formula(green), new Formula(blue));
 	}
 
-	private SetPenColorBrick(Formula red, Formula green, Formula blue) {
+	public SetPenColorBrick(Formula red, Formula green, Formula blue) {
 		this();
 		setFormulaWithBrickField(BrickField.PEN_COLOR_RED, red);
 		setFormulaWithBrickField(BrickField.PEN_COLOR_GREEN, green);
@@ -92,7 +91,7 @@ public class SetPenColorBrick extends FormulaBrick {
 	}
 
 	private boolean isBrickFieldANumber(BrickField brickField) {
-		return getFormulaWithBrickField(brickField).getRoot().getElementType() == FormulaElement.ElementType.NUMBER;
+		return getFormulaWithBrickField(brickField).isNumber();
 	}
 
 	@Override
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
@@ -90,10 +90,6 @@ public class SetPenColorBrick extends FormulaBrick {
 				&& isBrickFieldANumber(BrickField.PEN_COLOR_BLUE);
 	}
 
-	private boolean isBrickFieldANumber(BrickField brickField) {
-		return getFormulaWithBrickField(brickField).isNumber();
-	}
-
 	@Override
 	public void addActionToSequence(Sprite sprite, ScriptSequenceAction sequence) {
 		sequence.addAction(sprite.getActionFactory().createSetPenColorAction(sprite,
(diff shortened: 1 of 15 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/PhiroColorBrickTest.java
```diff
@@ -82,7 +82,7 @@ public void testPhiroLightRGBValuesWithFormulaEditor() {
 		onView(withText(R.string.brick_context_dialog_formula_edit_brick))
 				.perform(click());
 		onFormulaEditor()
-				.performEnterString("1+2 ")
+				.performEnterFormula("1+2")
```
thmq: :+1:  for fixing.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/SetPenColorBrickTest.java
```diff
@@ -38,19 +38,23 @@
...
 import static org.catrobat.catroid.uiespresso.content.brick.utils.BrickDataInteractionWrapper.onBrickAtPosition;
 import static org.catrobat.catroid.uiespresso.content.brick.utils.BrickTestUtils.createProjectAndGetStartScript;
 import static org.catrobat.catroid.uiespresso.content.brick.utils.ColorPickerInteractionWrapper.onColorPickerPresetButton;
+import static org.catrobat.catroid.uiespresso.formulaeditor.utils.FormulaEditorWrapper.onFormulaEditor;
 import static org.hamcrest.Matchers.containsString;
 
 @RunWith(AndroidJUnit4.class)
 public class SetPenColorBrickTest {
 	private int brickPosition;
+	private static final Integer INIT_COLOR = 0;
```
thmq: why `Integer`? couldn't it be `int`? or just `0` everywhere you need it?

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -1083,6 +1083,14 @@ public boolean isSingleNumberFormula() {
 		return false;
 	}
 
+	public boolean isElementANumber() {
```
thmq: please just call it `isNumber()` or something like this. you don't need to remind the element that it is an element.

think of it this way: the call is most likely going to be something like `someElement.isElementANumber()` so IMO `someElement.isNumber()` is cleaner.

thmq: btw: why do you need the function? there is literally the same function right above it? couln't you have reused (and renamed) that one?

catroid/src/main/java/org/catrobat/catroid/content/bricks/SetPenColorBrick.java
```diff
@@ -92,7 +91,7 @@ private boolean areAllBrickFieldsNumbers() {
 	}
 
 	private boolean isBrickFieldANumber(BrickField brickField) {
-		return getFormulaWithBrickField(brickField).getRoot().getElementType() == FormulaElement.ElementType.NUMBER;
+		return getFormulaWithBrickField(brickField).getRoot().isElementANumber();
```
thmq: I don't really understand the `.getRoot()` call here. I'd prefer if classes like this brick here did not have to know about the internal structure of the `Formula` meaning that it'd be better to provide a `isNumber()` function in the `Formula` class. 

An adavantage of that version would be that any changes to the `Formula` would then not require anything to change here.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
thmq: one more thing: there should be a test for this. there might already be one, since the method already exists. maybe have a look at it and see if it checks all cases that should be covered.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/SetPenColorBrickTest.java
```diff
@@ -64,6 +67,57 @@ public void setUp() {
...
+	@Test
+	public void testPenColoraluesWithFormulaEditor() {
+		onBrickAtPosition(brickPosition).checkShowsText(R.string.brick_pen_color);
+		onBrickAtPosition(brickPosition).onFormulaTextField(R.id.brick_set_pen_color_action_red_edit_text)
+				.checkShowsNumber(0)
+				.perform(click());
+		onView(withText(R.string.brick_context_dialog_formula_edit_brick))
+				.perform(click());
+		onFormulaEditor()
+				.performEnterFormula("1+2")
```
84n4n4: split this test into multiple smaller ones.
set up a brick field with the formula "1+2" before the activity is started, then do the operation to check whether formula editor is correctly opened or not.

same with the negative number, etc.

this would also be a candidate to do via robolectric, please have a look at BrickSingleFormulaFieldTest to get started on how to perform something like this.
we currently try to move as many formula editor tests over to the JVM as robolectric tests, since especially the typing operations and fragment loading is excessively slow on espresso.

swlz: ok, I'm going to change it immediately. thanks for pointing that out.

wslany:
@thmq wrote:
> please have a look at @84n4n4's suggestions, but production changes are OK IMO.

Is there still something to do? The ticket is not yet in PO review...

swlz:
Yes still need to change something.

thmq:
please fix static warning: [here](https://jenkins.catrob.at/job/Catroid/job/PR-3308/15/analysis/)

wslany:
Thanks!!

PR #3309, opened 2019-07-26
HOTFIX-0.9.63: Raise versionCode and versionName

ThomasSchwengler:


catroid/build.gradle
```diff
@@ -185,9 +185,9 @@ android {
         targetSdkVersion 26
         applicationId appId
         testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
-        versionCode 63
+        versionCode 64
```
wslany:
```suggestion
        versionCode 65
```

Necessary because we have uploaded 64 already once.

ThomasSchwengler: Done

PR #3314, opened 2019-07-26
BACKPORT/CATROID-284 touches Actor object should work for clones of current object

ThomasSchwengler:
backport of #3299


wslany:
Failed Catrobat language tests are due to their higher Catrobat language version compared to the hotfix branch based on the master, and can be ignored for the hotfix.

PR #3316, opened 2019-07-29
CATROID-209 add Arctan2 Function

SchmidtChristian:
- add String for arctan2 and parameters
- add case in InterFormulaKeyboardAdapter for arctan2
- add Entry in mapping in InternToExternGenerator
- add constant in Functions.java
- add case in FormulaElement to handle calculation
- add Tests for Arctan2 Function in ParserTestFunctions
- add 2 catrobat language tests
- change sinus, cosinus and tangens to sine, cosine and tangent


thmq:
:+1: for changing the function names to the actual english terms :laughing: 

one organizational thing though: you should only  change the `strings.xml` file in the the `values` directory, i.e., the source language strings. why? because all translations into target languages are handled by our crowdin system. so any changes to the `strings.xml` files in other directories are going to be overwritten by the system. consequently, changing them only obfuscates the PR, because then there are a lot of changed files and lines of code, while there is actually very little to check.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -454,6 +454,12 @@ private Object interpretFunction(Functions function, Sprite sprite) {
 				return doubleValueOfLeftChild == null ? defaultReturnValue : java.lang.Math.toDegrees(Math.acos(doubleValueOfLeftChild));
 			case ARCTAN:
 				return doubleValueOfLeftChild == null ? defaultReturnValue : java.lang.Math.toDegrees(Math.atan(doubleValueOfLeftChild));
+			case ARCTAN2:
+				return (doubleValueOfLeftChild == null || doubleValueOfRightChild == null) ? defaultReturnValue
```
thmq: I know that all of the other formula evaluations are written just as compact as this one, but this is really really hard to understand. the thing is: if someone has to change this they're going to have a bad time (it might even be difficult for you to know what's going on here if you have to look at it again in the future).

so what I'm trying to say is: please convert this to clean code, i.e., something that can be understood more clearly from just looking at it. especially the nested ternary operators are a nightmare.

how about something like:
```
if (doubleValueOfLeftChild == null || doubleValueOfRightChild == null) {
    return defaultReturnValue;
}

if (doubleValueOfLeftChild == 0) && (doubleValueOfRightChild == 0) {
    return -180 + java.lang.Math.random() * 360;
} else {
    return java.lang.Math.toDegrees(java.lang.Math.atan2(doubleValueOfLeftChild, doubleValueOfRightChild)));
}
```

the whole formula parsing is complicated enough as it is (I can't even count the number of people who unsuccessfully try to find someone who knows what's going on here), so I'd really appreciate some development towards a more sefl-explaining imlementation.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaKeyboardAdapter.java
```diff
@@ -456,4 +462,24 @@
 		returnList.add(new InternToken(InternTokenType.STRING, myString));
 		return returnList;
 	}
+
+	private List<InternToken> buildDoubleParameterFunction(Functions function,
+			InternTokenType firstOperator, String firstOperatorNumberValue, InternTokenType firstParameter,
```
thmq: maybe call the `InternTokenType firstOperator` agrument `InternTokenType firstOperatorType` otherwise it's a bit harder to understand if you call it from somewhere else (same for `InternTokenType secondOperator` obiviously).

and why is `firstOperatorNumberValue` a string? you pass `Operators.MINUS.name()` for this argument, so shouldn't it be something like `operatorName`? or am I missing something? what does the `numberValue` part mean?

does it have to be `firstParameterNumberValue`? I mean, is it really true that it can only be a number? I guess not, what would you need the type for then? so if I'm not mistaken i could also be a boolean or something else, right? or a string? theoretically at least. I'd say change the name, or remove the type and hardcode the number option into the method, because then it'd be clear. either way is fine with me, just this combination is a bit misleading.

catroid/src/test/java/org/catrobat/catroid/test/formulaeditor/FormulaEditorTestUtil.java
```diff
@@ -135,6 +136,18 @@ public static void testDoubleParameterFunction(Functions function, InternTokenTy
...
+	public static void testDoubleParameterFunction(Functions function, InternTokenType firstInternTokenType,
+			String firstParameter, InternTokenType secondInternTokenType, String secondParameter, Double upperBound,
+			Double lowerBound, Sprite testSprite) {
+
+		List<InternToken> internTokenList = buildDoubleParameterFunction(function, firstInternTokenType,
+				firstParameter, secondInternTokenType, secondParameter);
+		FormulaElement parseTree = new InternFormulaParser(internTokenList).parseFormula();
+
+		assertNotNull(parseTree);
+		assertTrue(((double) parseTree.interpretRecursive(testSprite) <= upperBound) && ((double) parseTree.interpretRecursive(testSprite) >= lowerBound));
```
thmq: you can use the hamcrest [isCloseTo](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/number/IsCloseTo.html) matcher here to assert that the number is somewhere close within a certain bound.

summoning @84n4n4 on this one though.

catroid/src/test/java/org/catrobat/catroid/test/formulaeditor/ParserTestFunctions.java
```diff
@@ -399,6 +399,49 @@ public void testArctan() {
...
+		secondParameter.add(new InternToken(InternTokenType.NUMBER, "1"));
+
+		FormulaEditorTestUtil.testDoubleParameterFunction(Functions.ARCTAN2, firstParameter,
+				secondParameter, -135.0, testSprite);
+
+		List<InternToken> firstParameterList = FormulaEditorTestUtil.buildBinaryOperator(InternTokenType.NUMBER, "5", Operators.PLUS,
+				InternTokenType.STRING, "datString");
+		List<InternToken> secondParameterList = FormulaEditorTestUtil.buildBinaryOperator(InternTokenType.NUMBER, "5", Operators.MULT,
+				InternTokenType.STRING, "anotherString");
+		FormulaEditorTestUtil.testDoubleParameterFunction(Functions.ARCTAN2, firstParameterList, secondParameterList,
```
thmq: that is a lot of asserts with different formulas for one test. @84n4n4 has been trying to get people to write smaller, more concise tests. his rationale is quite reasonable IMO, because with more, smaller tests you, as a developer immediately see what's wrong and in the test overview it more realistically represents what's broken and what works. e.g. if the arctan formula suddenly breaks if you put strings into it, then the whole test fails, but you don't know it was because of the strings and you have to find out if it was because of numbers, strings or something else. besides, there is one red test that says arctan not working, but if you split it you can easily see what's wrong and you see that e.g. arc tan with strings is broken, but works fine with numbers. - I guess you get the idea.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaKeyboardAdapter.java
SchmidtChristian: I need to pass a minus operator and the number so that the default parameters when using it in the formulaeditor work, so my thought was to just add a possible operator to the standard `buildDoubleParameterFunction` for both numbers.
Both `buildSingleParameterFunction`and `buildDoubleParameterFunction` we pass the type and a number value so that's why I also used them but I admit the naming could have been much clearer.
If you prefer I can delete the function entirely and replace it with something like:
`public static List<InternToken> buildDoubleParameterFunction(Functions function,
			List<InternToken> firstInternTokenList, List<InternToken> secondInternTokenList) {
		List<InternToken> tokenList = new LinkedList<InternToken>();
		tokenList.add(new InternToken(InternTokenType.FUNCTION_NAME, function.name()));
		tokenList.add(new InternToken(InternTokenType.FUNCTION_PARAMETERS_BRACKET_OPEN));
		tokenList.addAll(firstInternTokenList);
		tokenList.add(new InternToken(InternTokenType.FUNCTION_PARAMETER_DELIMITER));
		tokenList.addAll(secondInternTokenList);
		tokenList.add(new InternToken(InternTokenType.FUNCTION_PARAMETERS_BRACKET_CLOSE));
		return tokenList;
	}`

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
SchmidtChristian: Changed it, just thought I didn't want to produce the only piece of code in that file that does not look as compact as the rest.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaKeyboardAdapter.java
SchmidtChristian: yeah you are right that does look confusing, I removed the operator type and operator string and added a boolean instead which adds a minus operator if true.

catroid/src/test/java/org/catrobat/catroid/test/formulaeditor/FormulaEditorTestUtil.java
```diff
@@ -135,6 +138,19 @@ public static void testDoubleParameterFunction(Functions function, InternTokenTy
...
+			String firstParameter, InternTokenType secondInternTokenType, String secondParameter, Double median,
+			Double error, Sprite testSprite) {
+
+		List<InternToken> internTokenList = buildDoubleParameterFunction(function, firstInternTokenType,
+				firstParameter, secondInternTokenType, secondParameter);
+		FormulaElement parseTree = new InternFormulaParser(internTokenList).parseFormula();
+
+		assertNotNull(parseTree);
+		Double result = (double) parseTree.interpretRecursive(testSprite);
+		assertThat(result, closeTo(median, error));
```
84n4n4: this is fine, just so you know, assertEquals also supports delta for float and doubles.
so
assertEquals(doubleExpected, doubleActual, someSmallEpsilon)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/utils/FormulaEditorWrapper.java
```diff
@@ -110,7 +110,7 @@ public FormulaEditorWrapper performEnterString(int stringResourceId) {
 
 	public FormulaEditorWrapper checkShows(String expected) {
 		onView(FORMULA_EDITOR_TEXT_FIELD_MATCHER)
-				.check(matches(withText(expected)));
```
84n4n4: string.trim() would also cover this by removing trainling and leading whitespaces.
eg. .check(matches(withText(expected.trim())));

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
thmq: sure, I suspected as much and it's a perfectly justified approach. it's just that in a project like this, code sometimes ages badly and many design decisions (such as using very compact code here) are just no longer reasonable. so I think if you change anything you don't necessarily have to follow all conventions all the time. don't get me wrong, conventions and coding standards are a good thing, but sometimes conventions are (no longer) reasonable.

**TL;DR:** I understand why you did it like this, thanks for changing it anyway.

catroid/src/test/java/org/catrobat/catroid/test/formulaeditor/FormulaEditorTestUtil.java
thmq: yes, right. forgot about that.

thmq:
@SchmidtChristian : thanks for addressing all the comments so quickly :+1: IMO it's mergable now.

wslany:
Thanks!!

wslany:
@thmq @84n4n4 Ups, sorry --- just checked: Seems the Catrobat Language Version was not increased if I am not mistaken. Revert the merge and ask for that too?

thmq:
no, I don't think it is too much of a problem as long as it was not released. the only situation where this is a problem is when: you create a new project with this commit and a arctan2 in one of the formulas and someone else who does not have this commit or a newer one tries to open it. -> no data loss though, it just cannot be opened.

84n4n4:
this is only a problem if youd write a program using this function, and then try to open it with an older *develop* build.

the language version of the develop is already higher than the master, so we could even release without any issues.

wslany:
Language version increased with #3327

PR #3317, opened 2019-07-29
CATROID-262 refactor JobHandlerTest

gPathpp:


catroid/src/test/java/org/catrobat/catroid/test/scratchconverter/JobHandlerJobStateNotInProgressThrowsExceptionTest.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	@Test
+	public void testReceivedJobMessageWithWrongJobID() {
+		exception.expect(IllegalStateException.class);
+		JobMessage message = mock(JobMessage.class);
+		when(message.getJobID()).thenReturn(JOB_ID);
+
+		try {
+			jobHandler.onJobMessage(message);
```
84n4n4: if this is supposed to throw an exception (which i guess from the looks of this), then add a
fail statement below stating with what exception this should have thrown.

catroid/src/test/java/org/catrobat/catroid/test/scratchconverter/JobHandlerJobStateNotInProgressThrowsExceptionTest.java
```diff
@@ -0,0 +1,97 @@
...
+public class JobHandlerJobStateNotInProgressThrowsExceptionTest {
+
+	private static final long JOB_ID = 1;
+
+	@Parameterized.Parameters(name = "{0}")
+	public static Iterable<Object[]> data() {
+		ArrayList<Object[]> parameters = new ArrayList<>();
+		for (Job.State state : Job.State.values()) {
+			if (!state.isInProgress()) {
+				parameters.add(new Object[] {state.name(), state});
```
84n4n4: since this has only 5 states, of which only 4 are used here, please list them directly to make this easier for the next reader. (the amount of states isnt likely to be changed anyways, and if you want to make sure that something fails when someone does that assert on the size of the enum)

catroid/src/test/java/org/catrobat/catroid/test/scratchconverter/JobHandlerJobStateTest.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	@Test
+	public void testWrongJobIDThrowsException() {
+		exception.expect(IllegalArgumentException.class);
+		JobMessage messageWithWrongID = mock(JobMessage.class);
+		when(messageWithWrongID.getJobID()).thenReturn(WRONG_JOB_ID);
+
+		try {
+			jobHandler.onJobMessage(messageWithWrongID);
```
84n4n4: see above comments on exceptions

catroid/src/test/java/org/catrobat/catroid/test/scratchconverter/JobHandlerEventTest.java
```diff
@@ -0,0 +1,100 @@
...
+	@Before
+	public void setUp() throws Exception {
+		jobMock = mock(Job.class);
+		convertCallbackMock = mock(Client.ConvertCallback.class);
+		jobHandler = new JobHandler(jobMock, convertCallbackMock);
+	}
+
+	@After
+	public void tearDown() {
+		verify(jobMock, times(1)).getJobID();
```
84n4n4: tear down should be for cleanup stuff, this is more or less still part of each test

84n4n4: id rather have the code duplication here than having test code in teardown

catroid/src/androidTest/java/org/catrobat/catroid/test/scratchconverter/protocol/JobHandlerSuccessfulOnJobMessageTest.java
```diff
@@ -0,0 +1,195 @@
...
+	@Before
+	public void setUp() throws Exception {
+		jobSpy = spy(new Job(JOB_ID_OF_JOB_HANDLER, "My program", null));
+		convertCallbackMock = mock(Client.ConvertCallback.class);
+		jobHandler = new JobHandler(jobSpy, convertCallbackMock);
+	}
+
+	@After
+	public void tearDown() {
+		verifyNoMoreInteractions(convertCallbackMock);
```
84n4n4: this should be part of the test, teardown is for cleanup work, this seems to be integral part of the tests.

catroid/src/test/java/org/catrobat/catroid/test/scratchconverter/JobHandlerJobStateNotInProgressThrowsExceptionTest.java
84n4n4: thank you for the explanation, thats therefore resolved

wslany:
Thanks!

PR #3318, opened 2019-07-29
Restructure instrumented tests and optimize for local execution

84n4n4:


84n4n4:
- added AllEmulatorTestsSuite in androidTest package, that runs all locally executable tests at once.
- moved testsuites to parent package and renamed some for clarification
- restructured ActivityTestRules to flatten hierarchy and make FlakyTestRule available in headless instrumentation tests also

ThomasSchwengler:
@84n4n4 seems that this needs to be rebased. Some tests don't build anymore

wslany:
Thanks!!

PR #3320, opened 2019-07-30
CATROID-350 refactor GravityActionTest

gPathpp:


ThomasSchwengler:
As discussed with @84n4n4 it makes more sense to remove this test alltogether, as we are only testing libgdx and not our own code.

wslany:
@ThomasSchwengler wrote:
> As discussed with @84n4n4 it makes more sense to remove this test alltogether, as we are only testing libgdx and not our own code.

The Catrobat tests are also meant as executable specification and semantics, thus independent of the implementation details such as libgdx. Makes a lot of sense, and we need much more of them.

wslany:
THanks!!

PR #3321, opened 2019-07-31
CATROID-312 use last know position in visual placement

SchmidtChristian:
- evaluating and add parameters for intent
- transform imageview for positioning
- set coord at start to avoid different case handling for no changes
- add espressotest for VisualPlacement

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaceAtBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaceAtBrick.java
@@ -29,10 +29,12 @@ import android.support.v4.app.Fragment;
 import android.support.v7.app.AppCompatActivity;
 import android.view.View;
 
+import org.catrobat.catroid.ProjectManager;
 import org.catrobat.catroid.R;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.content.actions.ScriptSequenceAction;
 import org.catrobat.catroid.formulaeditor.Formula;
+import org.catrobat.catroid.formulaeditor.InterpretationException;
 import org.catrobat.catroid.ui.SpriteActivity;
 import org.catrobat.catroid.ui.UiUtils;
 import org.catrobat.catroid.ui.recyclerview.fragment.ScriptFragment;
@@ -41,6 +43,8 @@ import org.catrobat.catroid.visualplacement.VisualPlacementActivity;
 import static org.catrobat.catroid.content.bricks.Brick.BrickField.X_POSITION;
 import static org.catrobat.catroid.content.bricks.Brick.BrickField.Y_POSITION;
 import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_BRICK_HASH;
+import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_X_TRANSFORM;
+import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_Y_TRANSFORM;
 import static org.catrobat.catroid.ui.SpriteActivity.REQUEST_CODE_VISUAL_PLACEMENT;
 
 public class PlaceAtBrick extends FormulaBrick {
@@ -103,8 +107,22 @@ public class PlaceAtBrick extends FormulaBrick {
 			return;
 		}
 
+		Sprite currentSprite = ProjectManager.getInstance().getCurrentSprite();
+		Formula formulax = getFormulaWithBrickField(BrickField.X_POSITION);
+		Formula formulay = getFormulaWithBrickField(BrickField.Y_POSITION);
 		Intent intent = new Intent(view.getContext(), VisualPlacementActivity.class);
 		intent.putExtra(EXTRA_BRICK_HASH, hashCode());
+		int xValue;
+		int yValue;
+		try {
+			xValue = formulax.interpretInteger(currentSprite);
+			yValue = formulay.interpretInteger(currentSprite);
+		} catch (InterpretationException interpretationException) {
+			xValue = 0;
+			yValue = 0;
+		}
+		intent.putExtra(EXTRA_X_TRANSFORM, xValue);
+		intent.putExtra(EXTRA_Y_TRANSFORM, yValue);
 		activity.startActivityForResult(intent, REQUEST_CODE_VISUAL_PLACEMENT);
 	}
 
--- a/catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementActivity.java
@@ -71,6 +71,8 @@ import static org.catrobat.catroid.content.Look.ROTATION_STYLE_NONE;
 import static org.catrobat.catroid.stage.StageListener.SCREENSHOT_AUTOMATIC_FILE_NAME;
 import static org.catrobat.catroid.stage.StageListener.SCREENSHOT_MANUAL_FILE_NAME;
 import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_BRICK_HASH;
+import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_X_TRANSFORM;
+import static org.catrobat.catroid.ui.SpriteActivity.EXTRA_Y_TRANSFORM;
 
 public class VisualPlacementActivity extends BaseCastActivity implements View.OnTouchListener,
 		DialogInterface.OnClickListener, CoordinateInterface {
@@ -94,6 +96,8 @@ public class VisualPlacementActivity extends BaseCastActivity implements View.On
 	private float scaleY;
 	private float rotation;
 	private int rotationMode;
+	private float translateX;
+	private float translateY;
 
 	private float reversedRatioHeight;
 	private float reversedRatioWidth;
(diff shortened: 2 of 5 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VIsualPlacementActivityTest.java
```diff
@@ -0,0 +1,132 @@
...
+				.check(matches(withText(equalToIgnoringWhiteSpace(Integer.toString(xPos).replace("-", "- ")))));
+		onView(withId(R.id.brick_place_at_edit_text_y))
+				.check(matches(withText(equalToIgnoringWhiteSpace(Integer.toString(yPos).replace("-", "- ")))));
+	}
+
+	@Test
+	public void testNoChangeBackSave() {
+		onBrickAtPosition(brickPosition).performClick();
+		onView(withText(R.string.brick_place_at_option_place_visually))
+				.check(matches(isDisplayed()))
```
dpedratscher: do we need to check here again if it's displayed, since it's already checked in the smoke test?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VIsualPlacementActivityTest.java
```diff
@@ -0,0 +1,132 @@
...
+		Espresso.pressBack();
+
+		onView(withText(R.string.no))
+				.perform(click());
+
+		onView(withId(R.id.brick_place_at_edit_text_x))
+				.check(matches(withText(equalToIgnoringWhiteSpace(Integer.toString(xPos).replace("-", "- ")))));
+		onView(withId(R.id.brick_place_at_edit_text_y))
+				.check(matches(withText(equalToIgnoringWhiteSpace(Integer.toString(yPos).replace("-", "- ")))));
+	}
```
dpedratscher: would we need another test which actually makes some changes?

SchmidtChristian: From what I gathered, changes would be handled by the VisualPlacementTouchListener and we do have a test which checks if that aspect of the code works properly. 
catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementTouchListenerTest.java
But if you want to I can add a test which does make changes.

dpedratscher: na, if you say it's tested then it will be fine

catroid/src/main/java/org/catrobat/catroid/content/bricks/PlaceAtBrick.java
```diff
@@ -99,8 +101,18 @@ public void placeVisually() {
 			return;
 		}
 
+		Sprite currentSprite = ProjectManager.getInstance().getCurrentSprite();
+		Formula formulax = getFormulaWithBrickField(BrickField.X_POSITION);
+		Formula formulay = getFormulaWithBrickField(BrickField.Y_POSITION);
 		Intent intent = new Intent(view.getContext(), VisualPlacementActivity.class);
 		intent.putExtra(EXTRA_BRICK_HASH, hashCode());
+		try {
+			intent.putExtra("X", formulax.interpretInteger(currentSprite));
```
84n4n4: id prefer to have the interpretInteger portions in a try catch block, saving it into temporary variables and then putting the intent extra outside of the try / catch.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementActivityTest.java
```diff
@@ -0,0 +1,129 @@
...
+		baseActivityTestRule.launchActivity();
+	}
+	@Category({Cat.AppUi.class, Level.Smoke.class})
+	@Test
+	public void testNoChangeConfirm() {
+
+		onBrickAtPosition(brickPosition).performClick();
+		onView(withText(R.string.brick_place_at_option_place_visually))
+				.check(matches(isDisplayed()))
+				.perform(click());
```
84n4n4: leaving the activity under test (as in the test rule) is problematic in espresso and can lead to very weird failures of other tests, timeouts, activities getting stuck, etc. this in turn will lead to *other* tests randomly failing, which can make those failures very hard to locate and fix.

the nicer way of doing that would be testing each activity in isolation.
so you could test the VisualPlacementActivity in standalone by just giving it the intent extra on launch and see that it works as expected, aswell as that the activity result is correct.

using espressos Intended and Intending you can test the script activity for the brick by capturing the intent and asserting that the contents are correct.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementActivity.java
```diff
@@ -124,6 +126,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
 	protected void onCreate(@Nullable Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setContentView(R.layout.visual_placement_layout);
+		Bundle extras = getIntent().getExtras();
+		translateX = extras.getInt("X");
```
84n4n4: "X" and "Y" should be constants in this class, and also used outside of here when creating the intent.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementActivityTest.java
```diff
@@ -0,0 +1,118 @@
...
+	private static final int YPOS = 500;
+
+	@Rule
+	public final ActivityTestRule<VisualPlacementActivity> baseActivityTestRule =
+			new ActivityTestRule<>(VisualPlacementActivity.class, false, false);
+
+	@Before
+	public void setUp() {
+		createProjectAndGetStartScript(VisualPlacementActivity.class.getSimpleName());
+		Intent i = new Intent();
```
84n4n4: please call this variable "intent", and try to avoid single character variables.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementActivityTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+	@Test
+	public void testResultWithClickOnCenter() {
+		onView(withId(R.id.frame_container)).perform(click());
+		onView(withId(R.id.confirm)).perform(click());
+
+		Assert.assertTrue(baseActivityTestRule.getActivity().isFinishing());
+		Intent resultIntent = baseActivityTestRule.getActivityResult().getResultData();
+		Assert.assertEquals(0, resultIntent.getExtras().get(X_COORDINATE_BUNDLE_ARGUMENT));
+		Assert.assertEquals(1, resultIntent.getExtras().get(Y_COORDINATE_BUNDLE_ARGUMENT));
```
84n4n4: very nice!!!

if you could just add another test for the spriteActivity that checks that the outgoing intent also contains the right value we would have both sides tested!

84n4n4:
the iml file has to be removed from the commit!

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementActivityInputTest.java
```diff
@@ -0,0 +1,100 @@
...
+		Intents.release();
+		TestUtils.deleteProjects(PlaceAtBrickTest.class.getSimpleName());
+	}
+
+	@Test
+	public void testCorrectInput() {
+		bricktoAdd.setCoordinates(XPOS, YPOS);
+		baseActivityTestRule.launchActivity();
+		onView(withId(R.id.brick_place_at_edit_text_x)).perform(click());
+		onView(withText(R.string.brick_place_at_option_place_visually)).perform(click());
```
84n4n4: this will still launch the VisualPlacementActivity though.
the test is over at this point, but you leave the environment in a different activity than what youre actually testing.

theres two ways around this:
intending(intentMatcher).respondWith(result);

or 

		Instrumentation.ActivityMonitor activityMonitor = new Instrumentation.ActivityMonitor
				(VisualPlacementActivity.class.getName(), null, true);
		InstrumentationRegistry.getInstrumentation().addMonitor(activityMonitor);
		assertTrue(InstrumentationRegistry.getInstrumentation().checkMonitorHit(activityMonitor, 1));


both prevent the VisualPlacementActivity from being launched, while asserting that it would be and instead return a activityResult right away.

that also points you towards the untested parts of your PR.

you got:
- VisualPlacementActivity returns a valid activityResult if it is started with this very coordinates
- ScriptActivity starts VisualPlacementActivity with a valid intent containing these coordinates

whats missing: (untested)
- ScriptActivity puts the returned value in the brick and uses them
- VisualPlacement activity returning something different than what it got in its start intent.

this can be easily tested in the scope of those two tests. please move them to the intents package, and name them accordingly to the activity they test.

VisualPlacementActivity:
check that incoming Intent sets the x and y coords in VisualPlacementActivity
and check that outcoing Intent sends the modified x and y coords back (theres already setter methods in this activity)

SpriteActivity
check that it launches an intent for the VisualPlacementActivity with the given formula field values  (already implemented)
and check that SpriteActivity handles the activityResult and does something with those values (missing)

wslany:
Thanks!!

PR #3326, opened 2019-08-06
CATROID-231 Embroidery - Switch between layers

shellp:
- all objects are able to use the stitch brick
- colorchanges indicate change of layer which will be stitched
- simple tests


catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTFileGeneratorTest.java
```diff
@@ -185,26 +186,4 @@ public void testEndBytes() {
 		byte[] endBytes = fileGenerator.getFileEndBytes();
 		assertArrayEquals(expectedEndBytes, endBytes);
 	}
-
-	@Test
-	public void testStitchPointsEquals() {
```
84n4n4: why has this been removed?

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -185,15 +196,15 @@ private String getHeaderString(int size, int colorChanges, int[] boundingBox, Po
...
+		final int colorChanges = getColorChanges(stitchPoints);
 		PointF firstStitch = stitchPoints.get(0).point;
 		PointF lastStitch = stitchPoints.get(stitchPoints.size() - 1).point;
 		int[] boundingBox = getBoundingBox(stitchPoints);
 
 		return getHeaderString(stitchPoints.size(), colorChanges, boundingBox, firstStitch, lastStitch).getBytes();
 	}
 
-	private byte[] getBytesForStitchPoint(int x, int y, boolean jump) {
+	private byte[] getBytesForStitchPoint(int x, int y, boolean jump, boolean colorChange) {
```
84n4n4: please refer from using boolean flags.
use intdefs instead.

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryList.java
```diff
@@ -32,16 +32,25 @@
 	@Override
 	public synchronized boolean add(Object o) {
 		boolean status = false;
-		PointF point = (PointF) o;
```
84n4n4: if this would be:
public class EmbroideryList extends ArrayList<PointF>

the add method yould be add(PointF point) and no cast required. so that the compiler and static analysis would already complain when youd try to perform eg. add(3). while with the Object solution this would blow up at runtime.

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryList.java
```diff
@@ -32,16 +32,25 @@
...
-
-		if (super.isEmpty() || isNewStitchPoint(point)) {
+		if (o instanceof PointF) {
+			PointF point = (PointF) o;
+
+			if (super.isEmpty() || isNewStitchPoint(point)) {
+				status = super.add(new StitchPoint(point));
+			}
+		} else if (o instanceof StitchPoint) {
```
84n4n4: stitchpoint should be a PointF and not have a PointF. if you want to have an ArrayList working this way, or youd have to remove the possibility to add PointF directly, and rather do the new StitchPoint(point) outside when youre trying to add a PointF
this whole EmbroideryList is not how one would normally deal with polymorphism in java.

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryList.java
```diff
@@ -32,16 +32,25 @@
...
+			StitchPoint point = (StitchPoint) o;
 			status = super.add(point);
 		}
 
 		return status;
 	}
 
 	private Boolean isNewStitchPoint(PointF point) {
```
84n4n4: since "StitchPoint" is now a class, isNewStitchPoint is kind of misleading to take PointF as parameter.

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryListManager.java
```diff
@@ -0,0 +1,85 @@
...
+
+import android.graphics.PointF;
+
+import org.catrobat.catroid.content.Sprite;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class EmbroideryListManager {
```
84n4n4: wouldnt it be easier to just have the layer set in the stitchpoint, instead of all this map-ception?

you could write a comparator to sort, or filter for stitchpoints based on their layer, and this way round create the dst stitchblocks.

also, without having a deeper look into it, i think using the name of the sprite as identification will probably break when you use clones.

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryListManager.java
```diff
@@ -0,0 +1,85 @@
...
+
+				embroideryList.addAll(list);
+				StitchPoint lastStitchOfSprite = (StitchPoint) list.get(list.size() - 1);
+				embroideryList.add(new StitchPoint(lastStitchOfSprite.point, false, true));
+				embroideryList.add(new StitchPoint(lastStitchOfSprite.point, true, false));
+			}
+		}
+
+		if (!embroideryList.isEmpty()) {
+			embroideryList.remove(embroideryList.size() - 1);
```
84n4n4: this is going to break as soon as you have 1 element in it due to out of bounds exception

catroid/src/main/java/org/catrobat/catroid/embroidery/EmbroideryList.java
```diff
@@ -23,25 +23,94 @@
...
 
-import android.graphics.PointF;
+import org.catrobat.catroid.content.Sprite;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 
-public class EmbroideryList extends ArrayList{
+public class EmbroideryList extends ArrayList<StitchPoint> {
```
ThomasSchwengler: Please do not override from ArrayList. In about 99% of all cases this is a violation of the SOLID principles
In this case it violates:
* *Single responsibility principle*:
  * The list creates elements
  * The list has way more logic in, than a simple list should have
* *Liskov subsitution principle*:
  * `add` has side effects that are not to be expected from an arraylist
  * `addAll` does not work
  * `set` does not work
  * `add(index, element)` does not work
  * Most other `ArrayList` functions are broken as well
* *Interface segregation principle*:
  * You only ever use a very small subset of what an `ArrayList` or even a `List` offers, most of it is even unusable

Apart from that, you should prefer composition over inheritance.

I'd like to support you in finding a better solution.
From my point of view you should start at finding a better name for this. It's probably either some kind of Controller or Adapter. Best by starting to create an interface for this class.

catroid/src/main/java/org/catrobat/catroid/embroidery/JumpStitchPoint.java
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.embroidery;
+
+import android.graphics.PointF;
+
+import org.catrobat.catroid.content.Sprite;
+
+public class JumpStitchPoint implements StitchPoint {
```
ThomasSchwengler: The intent was, that Stitchpoints actually had a purpose apart from being containers. If you let the points act on some kind of workspace this should simplify things.
Regarding color and layer changes: Can you explain to me how this algorithm should actually work?
It seems to me that having a stack/array of layers and letting the stich points act on those layers might simplify the logic here even more. (Maybe I am wrong here :/ )

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchCommandTest.java
```diff
@@ -0,0 +1,171 @@
...
+		Mockito.verify(stream, Mockito.times(2)).addStitchPoint(Mockito.eq(0.0f), Mockito.eq(0.0f));
+		Mockito.verify(stream, Mockito.times(1)).addStitchPoint(Mockito.eq(xCoord),
+				Mockito.eq(yCoord));
+	}
+
+	@Test
+	public void testLayerSwitchTriggeredColorChange() {
+		EmbroideryWorkSpace workSpace = Mockito.mock(DSTWorkSpace.class);
+		ArrayList<StitchPoint> pointList = Mockito.spy(new ArrayList<>());
+		Mockito.when(pointList.isEmpty()).thenReturn(false);
```
ThomasSchwengler: Please add an element instead of using a spy

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchCommandTest.java
```diff
@@ -0,0 +1,171 @@
...
+		Mockito.when(stream.getPointList()).thenReturn(pointList);
+
+		StitchCommand previousCommand = new DSTStitchCommand(1, 1, layer - 1, Mockito.mock(Sprite.class));
+		StitchCommand command = new DSTStitchCommand(xCoord, yCoord, layer, Mockito.mock(Sprite.class));
+		command.act(workSpace, stream, previousCommand);
+
+		Mockito.verify(stream, Mockito.times(1)).addColorChange();
+		Mockito.verify(stream, Mockito.times(1)).addJump();
+		Mockito.verify(stream, Mockito.times(3)).addStitchPoint(Mockito.eq(1.0f), Mockito.eq(1.0f));
+		Mockito.verify(stream, Mockito.times(1)).addStitchPoint(Mockito.eq(xCoord), Mockito.eq(yCoord));
```
ThomasSchwengler: Please use inOrder (https://static.javadoc.io/org.mockito/mockito-core/2.6.9/org/mockito/InOrder.html)

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchCommandTest.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	@Test
+	public void testAddPreviousCommandOfSpriteOfOtherLayer() {
+		EmbroideryWorkSpace workSpace = Mockito.mock(DSTWorkSpace.class);
+		ArrayList<StitchPoint> pointList = Mockito.spy(new ArrayList<>());
+		Mockito.when(pointList.isEmpty()).thenReturn(true);
```
ThomasSchwengler: Same as above

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchCommandTest.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	@Test
+	public void testStitchCommandEquals() {
+		Sprite sprite = new Sprite("firstSprite");
+		StitchCommand command1 = new DSTStitchCommand(0, 0, 0, sprite);
+		StitchCommand command2 = new DSTStitchCommand(0, 1, 0, sprite);
+		StitchCommand command3 = new DSTStitchCommand(0, 0, 1, sprite);
+		StitchCommand command4 = new DSTStitchCommand(0, 0, 0, new Sprite("secondSprite"));
+		StitchCommand command5 = command1;
```
ThomasSchwengler: should probably be `new DSTStitchCommand(0, 0, 0, sprite);`

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchCommandTest.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	@Test
+	public void testStitchCommandHashCode() {
+		Sprite sprite = new Sprite("firstSprite");
+		StitchCommand command1 = new DSTStitchCommand(0, 0, 0, sprite);
+		StitchCommand command2 = new DSTStitchCommand(0, 1, 0, sprite);
+		StitchCommand command3 = new DSTStitchCommand(0, 0, 1, sprite);
+		StitchCommand command4 = new DSTStitchCommand(0, 0, 0, new Sprite("secondSprite"));
+		StitchCommand command5 = command1;
```
ThomasSchwengler: Should probably be `new DSTStitchCommand(0, 0, 0, sprite);`

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchPointTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(AndroidJUnit4.class)
+public class DSTStitchPointTest {
```
ThomasSchwengler: Please also test regular point (`stitchPoint.isConnectingPoint` is true)

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchPointTest.java
```diff
@@ -0,0 +1,118 @@
...
+	public void testColorChangeStitchPoint() {
+		StitchPoint stitchPoint = new DSTStitchPoint(0, 0);
+		stitchPoint.setColorChange(true);
+
+		assertTrue(stitchPoint.isColorChangePoint());
+		assertFalse(stitchPoint.isConnectingPoint());
+	}
+
+	@Test
+	public void testStitchBytes() {
```
ThomasSchwengler: Please separate in smaller tests

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStitchPointTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+	@Test
+	public void testStitchPointsHashCode() {
+		StitchPoint stitchPoint1 = new DSTStitchPoint(1, 2);
+		StitchPoint stitchPoint2 = new DSTStitchPoint(0, 0);
+		stitchPoint2.setJump(true);
+		StitchPoint stitchPoint3 = new DSTStitchPoint(0, 0);
+		stitchPoint1.setColorChange(true);
+		StitchPoint stitchPoint4 = new DSTStitchPoint(2, 1);
+		StitchPoint stitchPoint5 = stitchPoint1;
```
ThomasSchwengler: Please assign a new point

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStreamTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(AndroidJUnit4.class)
+public class DSTStreamTest {
+
+	@Test
+	public void testAddJumpStitchPoint() {
+		EmbroideryStream stream = new DSTStream();
```
ThomasSchwengler: Please move to a `@before` function

catroid/src/androidTest/java/org/catrobat/catroid/test/embroidery/DSTStreamTest.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Test
+	public void testInterpolatedStitchPoints() {
+		EmbroideryStream stream = new DSTStream();
+		stream.addStitchPoint(0, 0);
+		stream.addStitchPoint(80, 80);
+
+		assertEquals(5, stream.getPointList().size());
+		assertEquals(new PointF(0, 0), stream.getPointList().get(0).getPoint());
```
ThomasSchwengler: Please create a helper function to fight repetition

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStitchPoint.java
```diff
@@ -0,0 +1,128 @@
...
+	private int relativeY;
+	private byte[] dstPointBytes = new byte[] {0, 0, (char) 0x03};
+
+	public DSTStitchPoint(float x, float y) {
+		this.xCoord = x;
+		this.yCoord = y;
+	}
+
+	@Override
+	public PointF getPoint() {
```
ThomasSchwengler: please refactor to `getX()` and `getY()`

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTFileGenerator.java
```diff
@@ -23,277 +23,39 @@
...
-		PointF lastStitch = stitchPoints.get(stitchPoints.size() - 1).point;
-		int[] boundingBox = getBoundingBox(stitchPoints);
-
-		return getHeaderString(stitchPoints.size(), colorChanges, boundingBox, firstStitch, lastStitch).getBytes();
+		FileOutputStream fileStream = new FileOutputStream(dstFile);
+		writeHeader(fileStream);
+		writeStitchPoints(fileStream);
+		writeEnd(fileStream);
+		fileStream.flush();
+		fileStream.close();
```
ThomasSchwengler:
```
	public void writeToDSTFile(File dstFile) throws IOException {
		try (FileOutputStream fileStream = new FileOutputStream(dstFile)) {
			writeHeader(fileStream);
			writeStitchPoints(fileStream);
			writeEnd(fileStream);
		}
	}
```

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTHeader.java
```diff
@@ -0,0 +1,106 @@
...
+	private float firstX;
+	private float firstY;
+	private float lastX;
+	private float lastY;
+	private int colorChangeCount;
+	private int pointAmount;
+
+	@Override
+	public void initialize(float currentX, float currentY) {
+		minX = Math.round(currentX * DSTFileConstants.STITCH_POINT_UNIT_FACTOR);
```
ThomasSchwengler: Replace repetition with helper function. E.g.:
```
	private int toEmbroideryUnit(float value) {
		return Math.round(value * DSTFileConstants.STITCH_POINT_UNIT_FACTOR);
	}
```

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStitchPoint.java
```diff
@@ -0,0 +1,128 @@
...
+		return new PointF(xCoord, yCoord);
+	}
+
+	@Override
+	public byte[] getBytes() {
+		return Arrays.copyOf(dstPointBytes, dstPointBytes.length);
```
ThomasSchwengler: Please do this at runtime, don't save different representations of your data

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStitchPoint.java
```diff
@@ -0,0 +1,128 @@
...
+	public void setRelativeCoordinatesToPreviousPoint(float previousX, float previousY) {
+		relativeX =
+				Math.round(this.xCoord * DSTFileConstants.STITCH_POINT_UNIT_FACTOR) - Math.round(previousX * DSTFileConstants.STITCH_POINT_UNIT_FACTOR);
+		relativeY =
+				Math.round(this.yCoord * DSTFileConstants.STITCH_POINT_UNIT_FACTOR) - Math.round(previousY * DSTFileConstants.STITCH_POINT_UNIT_FACTOR);
+		setDSTPointBytes();
+	}
+
+	private void setDSTPointBytes() {
+		int xValue = (relativeX < 0 ? DSTFileConstants.CONVERSION_TABLE[(relativeX * (-1)) + 121]
```
ThomasSchwengler: please create a helper function for this

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStitchPoint.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Override
+	public boolean isColorChangePoint() {
+		return (((0x3 << 6) & dstPointBytes[2]) >>> 6) == 0x3;
+	}
+
+	@Override
+	public void setRelativeCoordinatesToPreviousPoint(float previousX, float previousY) {
+		relativeX =
+				Math.round(this.xCoord * DSTFileConstants.STITCH_POINT_UNIT_FACTOR) - Math.round(previousX * DSTFileConstants.STITCH_POINT_UNIT_FACTOR);
```
ThomasSchwengler: Please create a global helper function for this

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStream.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	public void addStitchPoint(float x, float y) {
+		DSTStitchPoint point = new DSTStitchPoint(x, y);
+		point.setColorChange(colorChangeBitSet);
+		point.setJump(jumpBitSet);
+		colorChangeBitSet = false;
+		jumpBitSet = false;
+		if (firstPoint) {
```
ThomasSchwengler: please check if list is empty instead

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStream.java
```diff
@@ -0,0 +1,130 @@
...
+
+package org.catrobat.catroid.embroidery;
+
+import android.graphics.PointF;
+
+import java.util.ArrayList;
+
+public class DSTStream implements EmbroideryStream {
+	private boolean firstPoint = true;
+	private EmbroideryHeader header = new DSTHeader();
```
ThomasSchwengler: Please move to assign in constructor

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTStream.java
```diff
@@ -0,0 +1,130 @@
...
+
+	@Override
+	public EmbroideryHeader getHeader() {
+		return header;
+	}
+
+	@Override
+	public void reset() {
+		colorChangeBitSet = false;
+		firstPoint = false;
```
ThomasSchwengler: Please check if `points.clear()` should be called

catroid/src/main/java/org/catrobat/catroid/embroidery/DSTWorkSpace.java
```diff
@@ -0,0 +1,55 @@
...
+import org.catrobat.catroid.content.Sprite;
+
+public class DSTWorkSpace implements EmbroideryWorkSpace {
+
+	private float currentX;
+	private float currentY;
+	private Sprite lastSprite;
+
+	@Override
+	public void update(float currentX, float currentY, Sprite currentSprite) {
```
ThomasSchwengler: Please rename to `set`

wslany:
Thanks! 

Note: Build warnings are unrelated to this PR.

PR #3327, opened 2019-08-07
MAINTENENCE Update language version after adding new function

thmq:
- added arctan2 function in previous commit, this
   increases the language version.


wslany:
Thanks!

PR #3328, opened 2019-08-07
CATROID-347 Refactor languageVersion check

SchmidtChristian:
- add function to check if Version is valid
- remove all "if" which only update language version to the next
- keep those which do other things
- set to current language version in the end

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -99,108 +99,33 @@ public final class ProjectManager {
 			restorePreviousProject(previousProject);
 			throw new OutdatedVersionProjectException(context.getString(R.string.error_outdated_version));
 		}
-
-		if (project.getCatrobatLanguageVersion() == 0.8f) {
-			project.setCatrobatLanguageVersion(0.9f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.9f) {
-			project.setCatrobatLanguageVersion(0.91f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.91f) {
-			project.setCatrobatLanguageVersion(0.92f);
-			project.setScreenMode(ScreenModes.STRETCH);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.92f) {
-			project.setCatrobatLanguageVersion(0.93f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.93f) {
-			project.setCatrobatLanguageVersion(0.94f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.94f) {
-			project.setCatrobatLanguageVersion(0.95f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.95f) {
-			project.setCatrobatLanguageVersion(0.96f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.96f) {
-			project.setCatrobatLanguageVersion(0.97f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.97f) {
-			project.setCatrobatLanguageVersion(0.98f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.98f) {
-			project.setCatrobatLanguageVersion(0.99f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.99f) {
-			project.setCatrobatLanguageVersion(0.991f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.991f) {
-			project.setCatrobatLanguageVersion(0.992f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.992f) {
-			updateCollisionFormulasTo993(project);
-			project.setCatrobatLanguageVersion(0.993f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.993f) {
-			updateSetPenColorFormulasTo994(project);
-			project.setCatrobatLanguageVersion(0.994f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.994f) {
-			updateArduinoValuesTo995(project);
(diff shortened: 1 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -441,6 +366,17 @@ private static void updateScriptsToTreeStructure(Project project) {
 		}
 	}
 
+	private static boolean checkValidLanguageVersion(Float languageVersion) {
+		if (languageVersion <= CURRENT_CATROBAT_LANGUAGE_VERSION) {
+			if (languageVersion == 0.8f) {
+				return true;
+			}
```
PlantinumX: pls research if it is possible to include the if statement in the regex

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -441,6 +366,17 @@ private static void updateScriptsToTreeStructure(Project project) {
 		}
 	}
 
+	private static boolean checkValidLanguageVersion(Float languageVersion) {
```
84n4n4: i personally dislike "check..." methods because its never really clear what they do without looking it up.

when something returns a boolean, i prefer the method to read as a question, eg. "isValidLanguageVersion" not as a command "checkValidLanguageVersion"

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
84n4n4: the regex seems ok, since there proably wasnt a published 0.81 i guess, but all in all wouldnt this read a little better as:

`return languageVersion == 0.8f || 
            (languageVersion <= CURRENT_CATROBAT_LANGUAGE_VERSION &&
                       languageVersion >= 0.9f)`

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -99,105 +99,30 @@ public void loadProject(File projectDir, Context context) throws ProjectExceptio
...
-			project.setCatrobatLanguageVersion(0.9998f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.9998f) {
-			project.setCatrobatLanguageVersion(0.9999f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.9999f) {
-			project.setCatrobatLanguageVersion(0.99991f);
-		}
-		if (project.getCatrobatLanguageVersion() == 0.99991f) {
+		if (checkValidLanguageVersion(project.getCatrobatLanguageVersion())) {
```
84n4n4: if youd do the check here and return right away if invalid you save one level of intentation and have the parts belonging together in one place:

`if (!checkValidLanguageVersion(project.getCatrobatLanguageVersion())) {
			restorePreviousProject(previousProject);
			throw new CompatibilityProjectException(context.getString(R.string.error_project_compatibility));
}`

84n4n4:
tests would be nice.
use powermock to mock the XstreamSerializer.getInstance().loadProject(projectDir, context);

so it returns a project mock where you can test all exceptions this is supposed to through or even test that the update methods are called.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
thmq: you could even invert the method to already say `isInvalidLanguageVersion()` then this approach would look even cleaner (look at the snippet below). besides you could even leave out the `else` because the exception stops method execution anyway. so that's two levels of indentation and felt cyclomatic complexity. 

if you go this way, just don't forget to invert the return value in the method FKA `checkValidLanguageVersion`.

```
if (isInvalidLanguageVersion(project.getCatrobatLanguageVersion()) {
    restorePreviousProject(previousProject);
    throw new CompatibilityProjectException(context.getString(R.string.error_project_compatibility));
}

if (project.getCatrobatLanguageVersion() <= 0.91f) {
   project.setScreenMode(ScreenModes.STRETCH);
}
if (project.getCatrobatLanguageVersion() <= 0.992f) {
    updateCollisionFormulasTo993(project);
}
if (project.getCatrobatLanguageVersion() <= 0.993f) {
    updateSetPenColorFormulasTo994(project);
}
if (project.getCatrobatLanguageVersion() <= 0.994f) {
    updateArduinoValuesTo995(project);
}
if (project.getCatrobatLanguageVersion() <= 0.995f) {
    updateCollisionScriptsTo996(project);
}
if (project.getCatrobatLanguageVersion() <= 0.999f) {
    makeShallowCopiesDeepAgain(project);
}
if (project.getCatrobatLanguageVersion() <= 0.9993f) {
    updateScriptsToTreeStructure(project);
}			
project.setCatrobatLanguageVersion(CURRENT_CATROBAT_LANGUAGE_VERSION);

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
thmq: yes, that looks better. I'd just turn the second part around (just to be extra nitpicky) so that it says: either 0.8f or greater than 0.9f and below current language version.

`return languageVersion == 0.8f || (languageVersion >= 0.9f && languageVersion <= CURRENT_CATROBAT_LANGUAGE_VERSION)`

SchmidtChristian: I removed the regex entirely, my thought was that, it works for now but, once the language version reaches 1.0 the regex in it's current form would allow invalid versions. So the regex would have to be dynamic, something like `mystring.matches("[" + x + " ][.][" + y + "]+..."` where x would be the first char in the 0.9 string and y the third one. Which would make it harder to read in the first place and we would have to remember all the highest versions per subversion so 0.8, 0.99992 if we would switch from 0.99992 to 1.0 etc. So I added a map which has all known versions, to check we only have to access 1 entry and see if it is null. (assuming the java map is as efficient as c++ maps)

wslany:
Thanks!

PR #3329, opened 2019-08-07
CATROID-308 remove permissions.txt

SchmidtChristian:
- remove Function in standalone gradle which removes values from permission file
- change Test in XstreamSerializerTest which uses the file
- add function in ProjectManager to remove Permissionfile if upgraded to higher language version
- add new if in language Version check
- increase language Version
- remove creation of permissionFile in XstreamSerializer

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -64,6 +64,7 @@ import java.util.Iterator;
 import java.util.List;
 
 import static org.catrobat.catroid.common.Constants.CURRENT_CATROBAT_LANGUAGE_VERSION;
+import static org.catrobat.catroid.common.Constants.PERMISSIONS_FILE_NAME;
 
 public final class ProjectManager {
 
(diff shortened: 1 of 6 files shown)


84n4n4:
ah, this is what this was good for...

maybe we should discuss with @wslany if we should fix this instead of removing.

in short:
this removes unused permissions from apk generator created apks. and was meant to remove eg. bluetooth or camera permissions if they are not required for the project.

thmq:
> ah, this is what this was good for...
> 
> maybe we should discuss with @wslany if we should fix this instead of removing.
> 
> in short:
> this removes unused permissions from apk generator created apks. and was meant to remove eg. bluetooth or camera permissions if they are not required for the project.

I knew there was going to be some black magic that relied on this.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -441,6 +445,13 @@ private static void updateScriptsToTreeStructure(Project project) {
 		}
 	}
 
+	private static void removePermissionsFile(Project project) {
+		File permissionsFile = new File(project.getDirectory(), "permissions.txt");
```
84n4n4: please put "permissions.txt" somewhere in constants

84n4n4:
i clarified with @wslany, were going to remove this!
so this PR is fine, if you could just move the "permissions.txt" string somewhere in constants and use it from there.

thank you!

wslany:
Thanks!!

PR #3334, opened 2019-08-19
CATROID-355: Check that new project name is valid

Fatmangh:
Check that new project name is valid

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
@@ -1,0 +1,167 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.uiespresso.ui.fragment;
+
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+import android.widget.EditText;
+
+import org.catrobat.catroid.R;
+import org.catrobat.catroid.common.BrickValues;
+import org.catrobat.catroid.content.Project;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.SingleSprite;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.content.StartScript;
+import org.catrobat.catroid.content.bricks.SetXBrick;
+import org.catrobat.catroid.formulaeditor.Formula;
+import org.catrobat.catroid.io.XstreamSerializer;
+import org.catrobat.catroid.testsuites.annotations.Cat;
+import org.catrobat.catroid.testsuites.annotations.Level;
+import org.catrobat.catroid.ui.ProjectListActivity;
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityTestRule;
+import org.hamcrest.Matchers;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import static android.support.test.InstrumentationRegistry.getInstrumentation;
+import static android.support.test.espresso.Espresso.closeSoftKeyboard;
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu;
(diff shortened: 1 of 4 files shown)


catroid/src/main/java/org/catrobat/catroid/merge/MergeItemTextWatcher.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+public class MergeItemTextWatcher<T extends Nameable> extends TextInputDialog.TextWatcher {
+	private List<T> scope;
+
+	public MergeItemTextWatcher(List<T> scope) {
+		this.scope = scope;
+	}
+
+	private boolean isNameUnique(String name) {
```
84n4n4: if this textwatcher is for merge of only projects this is not necessary, since Utils.checkIfProjectExistsOrIsDownloadingIgnoreCase checks exactly that.

if this is intended for use also for merging scenes in the future, it will not work correctly, because i wouldnt be able to create a scene with a name another project has.

catroid/src/main/java/org/catrobat/catroid/merge/MergeItemTextWatcher.java
```diff
@@ -0,0 +1,70 @@
...
+
+		if (input.isEmpty()) {
+			return context.getString(R.string.name_empty);
+		}
+
+		input = input.trim();
+
+		if (input.isEmpty()) {
+			error = context.getString(R.string.name_consists_of_spaces_only);
+		} else if (!isNameUnique(input) && Utils.checkIfProjectExistsOrIsDownloadingIgnoreCase(input)) {
```
84n4n4: see above comment, once !isNameUnique(input) && is removed from the statement, this is *exaclty* the same text watcher as in NewProjectDialogFragment.
and then this should be refactored accordingly to also use this, instead of having the duplicate code. (and of course have its name changed accordingly, eg. NewProjectNameTextWatcher)

84n4n4: if you want to generalize this, then youd have to *only* check for isNameUnique, and add the currently downloading projects to the list outside, wherever you call this watcher.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+				.performCheckItem();
+
+		onRecyclerView().atPosition(1)
+				.performCheckItem();
+
+		onView(withId(R.id.confirm)).perform(click());
+
+		onView(allOf(withText(""), isDisplayed(), instanceOf(EditText.class)))
+				.perform(replaceText(validName));
+		closeSoftKeyboard();
```
84n4n4: this "closeSoftKeyboard()" is not doing anything, because you are using the wrong import.
you are using android.support.test.espresso.action.ViewActions.closeSoftKeyboard, which is a ViewAction, so it has to be done in a perform. if you just state it outside, you create a new ViewAction object and not doing anything with it.

the correct one for this would be: android.support.test.espresso.Espresso.closeSoftKeyboard, which is a standalone action like "pressBack()".

so either:

import static android.support.test.espresso.action.ViewActions.closeSoftKeyboard;

... .perform(replaceText(validName), closeSoftKeyboard());

or:

import static android.support.test.espresso.Espresso.closeSoftKeyboard;
...
closeSoftKeyboard();

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+public class MergeProjectNameTest {
+
+	@Rule
+	public BaseActivityTestRule<ProjectListActivity> baseActivityTestRule = new
+			BaseActivityTestRule<>(ProjectListActivity.class, true, false);
+
+	private String firstProjectName = "firstProjectName";
+	private String secondProjectName = "secondProjectName";
+	private String validName = "ValidName";
+	private String inValidName = "firstProjectName";
```
84n4n4: this should be:
private String inValidName = firstProjectName;

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+import static org.hamcrest.Matchers.not;
+
+@RunWith(AndroidJUnit4.class)
+public class MergeProjectNameTest {
+
+	@Rule
+	public BaseActivityTestRule<ProjectListActivity> baseActivityTestRule = new
+			BaseActivityTestRule<>(ProjectListActivity.class, true, false);
+
+	private String firstProjectName = "firstProjectName";
```
84n4n4: all of these strings should be constants, therefore static final and in caps.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+		onRecyclerView().atPosition(1)
+				.performCheckItem();
+
+		onView(withId(R.id.confirm)).perform(click());
+
+		onView(allOf(withText(""), isDisplayed(), instanceOf(EditText.class)))
+				.perform(replaceText(inValidName));
+		closeSoftKeyboard();
+
+		onView(Matchers.allOf(withId(android.R.id.button1), withText(R.string.ok)))
```
84n4n4: please use the static imported method (allOf) as in the onView above.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MergeProjectNameTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+		onRecyclerView().atPosition(0)
+				.performCheckItem();
+
+		onRecyclerView().atPosition(1)
+				.performCheckItem();
+
+		onView(withId(R.id.confirm)).perform(click());
+
+		onView(allOf(withText(""), isDisplayed(), instanceOf(EditText.class)))
```
84n4n4: this will break if someone gives this a default text.
why not access the edit text with
onView(withId(R.id.input_edit_text), isDisplayed())

PR #3336, opened 2019-08-20
Fix layout margins of delete key in formula editor

ThomasSchwengler:
* Fixes the margins of the delete key in formula editor

| Before | After |
|----------|--------|
|  |  |


wslany:
Thanks!

PR #3340, opened 2019-08-23
CATROID-345 whenBackGroundChanges crash fix

SchmidtChristian:
- change WhenBackgroundChangesBrick from BasicBrickType to ScriptBrickBaseType
- add Testcase for the error

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/bricks/WhenBackgroundChangesBrick.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/bricks/WhenBackgroundChangesBrick.java
@@ -146,4 +146,14 @@ public class WhenBackgroundChangesBrick extends BrickBaseType implements ScriptB
 		spinner.add(item);
 		spinner.setSelection(item);
 	}
+
+	@Override
+	public List<Brick> getDragAndDropTargetList() {
+		return getScript().getBrickList();
+	}
+
+	@Override
+	public int getPositionInDragAndDropTargetList() {
+		return -1;
+	}
 }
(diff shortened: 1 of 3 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/rtl/WhenBackgroundChangesHoverTest.java
```diff
@@ -0,0 +1,98 @@
...
+import static android.support.test.espresso.action.ViewActions.click;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+
+import static org.catrobat.catroid.uiespresso.content.brick.utils.BrickDataInteractionWrapper.onBrickAtPosition;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.core.Is.is;
+
+@RunWith(AndroidJUnit4.class)
+public class WhenBackgroundChangesHoverTest {
```
84n4n4: instead of a UI test that only checks this very script brick to not regress, it would be better to add unit tests that verify the correct behavior of getDragAndDropTargetList() for all script bricks.

84n4n4:
please unit test instead of ui test

84n4n4:
instead of changing the bricks base type to ScriptBrickBaseType, you could also just override the methods in the brick itself. (like its done in WhenConditionBrick)

84n4n4:
testwise you could try writing a test that gathers all non-abstract implementations of the interface ScriptBrick, and then test if they properly return their script list.

have a look at ScriptsXmlSerializerTest for example, that gathers all classes implementing Script and then tests that they behave all the same.

84n4n4:
ive written a test as described above:

https://gist.github.com/84n4n4/012d47072f8cc3d727205d6922c61b04

this also shows that theWhenBounceOffBrick has the same issue.

without changing the base class of those bricks you could just add this block to both of them:

	@Override
	public List<Brick> getDragAndDropTargetList() {
		return getScript().getBrickList();
	}

	@Override
	public int getPositionInDragAndDropTargetList() {
		return -1;
	}

84n4n4:
this of course does not require a UI test anymore, and it also makes sure that new script bricks will show up as failing tests if someone does this again. (because the test dynamically tests ALL scriptBricks in the codebase)

wslany:
Thanks!

PR #3341, opened 2019-09-02
CATROID-329 Set Arduino and RaspberryPi Bricks to default in Arduino flavor

PlantinumX:
* Add new Flavor Arduino with extensions enabled
* Add new icon with several resolutions
* add fastlane setup
* add gradle setup


catroid/src/arduino/res/xml/preferences.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="utf-8"?>
```
84n4n4: this poses the problem that if we add another feature, someone has to remember that preferences have now be modified here and in the main package.

maybe moving the default values of preferences somewhere else like the flavoredConstants or creating another array xml to store those values.

wslany:
Thanks!

PR #3342, opened 2019-09-02
CATROID-239 Formula Editor does not crash after wrapping anything in Brackets

DollyVolley:
* Bracket Elements get initialized with empty string instead of null. String compare operations now don't crash anymore with brackets.

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorTest.java
@@ -71,9 +71,7 @@ public class FormulaEditorTest {
 
 		//typeText not working for formula editor, so use CustomActions.typeInValue
 		onFormulaEditor()
-				.performEnterNumber(12345.678);
-
-		onFormulaEditor()
+				.performEnterNumber(12345.678)
 				.performCloseAndSave();
 	}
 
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/InternFormulaParser.java
```diff
@@ -217,7 +217,7 @@ private FormulaElement term() throws InternFormulaParserException {
 
 			case BRACKET_OPEN:
 				getNextToken();
-				currentElement.replaceElement(new FormulaElement(FormulaElement.ElementType.BRACKET, null, null, null,
+				currentElement.replaceElement(new FormulaElement(FormulaElement.ElementType.BRACKET, "", null, null,
```
84n4n4: this would be nicer if youd fix this by checking for null in stringWithoutTrailingZero, to make call to "getValue" save for all kinds of tokens.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorTest.java
```diff
@@ -71,10 +73,23 @@ public void numericValuesTest() {
...
 		//typeText not working for formula editor, so use CustomActions.typeInValue
 		onFormulaEditor()
-				.performEnterNumber(12345.678);
+				.performEnterNumber(12345.678)
+				.performCloseAndSave();
+	}
+
+	@Category({Cat.AppUi.class, Level.Smoke.class})
+	@Test
+	public void wrappedInBracketsTest() {
```
84n4n4: write an instrumented unit test for this. (there are already a handful testing the getResultForComputeDialog method)
and remove the UI test.

catroid/src/test/java/org/catrobat/catroid/test/formulaeditor/FormulaTest.java
```diff
@@ -197,5 +197,13 @@ public void testComputeDialogResult() {
 		Formula regexFormula = new Formula(regexFunctionFormulaElement);
 		computeDialogResult = regexFormula.getResultForComputeDialog(null);
 		assertEquals("elephant", computeDialogResult);
+
+		FormulaElement bracketOpenFormulaElement = new FormulaElement(ElementType.BRACKET, null, null);
```
PlantinumX: pls split into smaller tests for readability

wslany:
Thanks!

PR #3343, opened 2019-09-03
CATROID-287 Create ReadListFromDevice and WriteListOnDevice Brick

swlz:
- Created two new bricks
- Changed UserVariable, UserList and UserData in order to use several functions for the DeviceAccessor
- Created new file DeviceList
- Created new tests for DeviceList


swlz:
test also fails with exit code 137 which @84n4n4 mentioned in jenkins-general

catroid/src/androidTest/java/org/catrobat/catroid/test/io/devicelistaccessor/DeviceUserListAccessorParameterizedValueTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+	@Parameterized.Parameters(name = "{0}")
+	public static Iterable<Object[]> data() {
+		return Arrays.asList(new Object[][] {
+				{"Double", Arrays.asList(12.123)},
+				{"Negative Double", Arrays.asList(-2312.123)},
+				{"String", Arrays.asList("initial String Value")},
+				{"Empty String", Arrays.asList("")},
+				{"String special char", Arrays.asList("{}\\%&($%(/(/§")},
+				{"Boolean", Arrays.asList(true)},
```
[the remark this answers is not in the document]
swlz: ok changed that

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+public abstract class DeviceUserDataAccessor {
+
+	public static final String TAG = DeviceUserDataAccessor.class.getSimpleName();
+	private static final Object LOCK = new Object();
```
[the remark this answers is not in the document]
swlz: ok thx. applied these changes.

catroid/src/androidTest/java/org/catrobat/catroid/test/io/devicelistaccessor/DeviceUserListAccessorParameterizedValueTest.java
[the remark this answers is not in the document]
thmq: thanks.

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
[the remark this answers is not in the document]
thmq: perfect, thank you.

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,200 @@
...
+		if (!deviceMap.containsKey(userData.getDeviceKey())) {
+			userData.reset();
+			return false;
+		}
+		userData.setValue(deviceMap.get(userData.getDeviceKey()));
+		return true;
+	}
+
+	public void writeUserData(UserData userData) throws IOException {
+		getLock();
```
thmq: you missed the `synchronized` keyword here. just calling the method does not by lock anything ->

the `synchronized` keyword in Java allows only one thread to execute the statements it synchronizes. if you put the keyword in front of a method such as:
```
synchronized void someMethod() 
```
then the method can be invoked only once per object simultaneously. meaning that for each object of the class only one thread can call it.

alternatively you can use the keyword around a code block, such as:
```
sychronized(someObject) {
   doStuffHere
}
```
note that in this version you can specify on which object to synchronize -> e.g. the object itself by using `this`, or any other object. using a static object here means that the block is synchronized between all objects of the class (this is what you want to do here, I guess).

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,200 @@
...
+		if (deviceVariableMap == null) {
+			deviceVariableMap = new HashMap<>();
+		}
+
+		deviceVariableMap.put(userData.getDeviceKey(), userData.getValue());
+		writeMapToJson(deviceVariableMap);
+	}
+
+	public void removeDeviceValue(UserData userData) {
+		getLock();
```
thmq: same as above.

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
swlz: oh, sorry. just changed it.

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,203 @@
...
+				deviceVariableMap.remove(userData.getDeviceKey());
+				writeMapToJson(deviceVariableMap);
+			} catch (Exception e) {
+				Log.e(TAG, e.getMessage());
+			}
+		}
+	}
+
+	@VisibleForTesting
+	public HashMap<UUID, Object> readMapFromJson() {
```
thmq: from SonarLint:


The purpose of the Java Collections API is to provide a well defined hierarchy of interfaces in order to hide implementation details.
Implementing classes must be used to instantiate new collections, but the result of an instantiation should ideally be stored in a variable whose type is a Java Collection interface.
This rule raises an issue when an implementation class:
is returned from a public method.
is accepted as an argument to a public method.
is exposed as a public member.
Noncompliant Code Example
  public class Employees {    private HashSet<Employee> employees = new HashSet<Employee>(); // Noncompliant - "employees" should have type "Set" rather than "HashSet"      public HashSet<Employee> getEmployees() { // Noncompliant      return employees;    }  }  
Compliant Solution
  public class Employees {    private Set<Employee> employees = new HashSet<Employee>(); // Compliant      public Set<Employee> getEmployees() { // Compliant      return employees;    }  }

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,203 @@
...
+			userData.reset();
+			return false;
+		}
+		userData.setValue(deviceMap.get(userData.getDeviceKey()));
+		return true;
+	}
+
+	public void writeUserData(UserData userData) throws IOException {
+		synchronized (getLock()) {
+			if (!deviceFile.exists()) {
```
thmq: you disregard the return value of `createNewFile()`, here. so if it fails it can be very difficult to find out why it did not work because it then crashes somewhere during the json serialization. so I suggest you do something like this:
```
if (!deviceFile.exists() && !deviceFile.createNewFile()) {
	throw new IOException("Cannot create " + deviceFile.getAbsolutePath());
}
```

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	protected abstract String getDeviceFileName();
+
+	protected abstract List<? extends UserData> getUserData(Sprite sprite);
+
+	protected abstract List<? extends UserData> getUserData(Project project);
+
+	public List<UUID> getKeyList(List<? extends UserData> userDataList) {
```
thmq: can be private

catroid/src/main/java/org/catrobat/catroid/io/DeviceUserDataAccessor.java
```diff
@@ -0,0 +1,203 @@
...
+			if (deviceFile.exists()) {
+				Log.e(TAG, "Device Variable File corrupted!");
+				deviceFile.delete();
+			}
+			return null;
+		}
+	}
+
+	@VisibleForTesting
+	public void writeMapToJson(HashMap map) throws IOException {
```
thmq: It is preferable to use try-with resource handling, such as this:

```
	@VisibleForTesting
	public void writeMapToJson(Map map) {
		try (BufferedOutputStream bos =
				new BufferedOutputStream(new FileOutputStream(deviceFile))) {
			Type mapType = new TypeToken<HashMap<UUID, Object>>() {
			}.getType();
			String jsonString = new Gson().toJson(map, mapType);
			bos.write(jsonString.getBytes());
			bos.flush();
		} catch (IOException e) {
			Log.e(TAG, e.getMessage());
		}
	}
```

swlz:
instrumented unit test issues like develop

swlz:
Ok, already looking at it. 👍

swlz:
> if a new brick is added to the Catrobat language there is unfortunately no automatic way to add the correct alias for the .xml file, so all of them have to be added manually, see [here](https://github.com/Catrobat/Catroid/blob/07b20dda7c29f8b173a12396aaf472b3b2ce44b5/catroid/src/main/java/org/catrobat/catroid/io/XstreamSerializer.java#L444). please add the two new bricks.

already did that.

thmq:
> > if a new brick is added to the Catrobat language there is unfortunately no automatic way to add the correct alias for the .xml file, so all of them have to be added manually, see [here](https://github.com/Catrobat/Catroid/blob/07b20dda7c29f8b173a12396aaf472b3b2ce44b5/catroid/src/main/java/org/catrobat/catroid/io/XstreamSerializer.java#L444). please add the two new bricks.
> 
> already did that.

sorry, completely missed it. thanks for checking though :wink:

swlz:
Don't worry. 😊

PR #3344, opened 2019-09-04
CATROID-365 fix crash of WhenBackGroundChanges without selection

SchmidtChristian:
- add null check in setBackGroundEventId
- add testcase for hashCode Function


wslany:
Thanks!!

PR #3348, opened 2019-09-12
CATROID-212 + CATROID-391 Add list view for User Bricks

DollyVolley:
* Add fragment_brick_add_userbrick.xml
* Add query to distinguish userbricks and others for loading the specific xmls
* Add UI Test
* Rename "My bricks" to "Your bricks"

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/AddUserbrickTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/AddUserbrickTest.java
@@ -1,0 +1,86 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.uiespresso.ui.fragment;
+
+import android.support.test.InstrumentationRegistry;
+
+import org.catrobat.catroid.ProjectManager;
+import org.catrobat.catroid.R;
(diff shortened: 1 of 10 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/fragment/AddBrickFragment.java
```diff
@@ -79,7 +78,12 @@ public static AddBrickFragment newInstance(String selectedCategory, ScriptFragme
 
 	@Override
 	public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
-		View view = inflater.inflate(R.layout.fragment_brick_add, container, false);
+		View view;
```
84n4n4: dont do this.

this makes this a "tagged class", doing two completely different things with different layouts in a single class. the next guy will again just add behaviour here with help of some boolean flag.

https://books.google.at/books?id=ka2VUBqHiWkC&pg=PA101&lpg=PA101&dq=java++%22tagged+class%22&source=bl&ots=y_IoKgr_M4&sig=ACfU3U2qgbGTUfeD2HkpOgmn_FV7C6OCEA&hl=de&sa=X&ved=2ahUKEwj6h9mst-bkAhWEKVAKHeZSAgUQ6AEwB3oECAkQAQ#v=onepage&q=java%20%20%22tagged%20class%22&f=false

http://thefinestartist.com/effective-java/20

https://medium.com/@appmattus/effective-kotlin-item-23-prefer-class-hierarchies-to-tagged-classes-de99d37f815a

this should be a separate fragment.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/AddUserbrickTest.java
```diff
@@ -0,0 +1,107 @@
...
+		project.getDefaultScene().addSprite(firstSprite);
+		ProjectManager.getInstance().setCurrentProject(project);
+		ProjectManager.getInstance().setCurrentSprite(firstSprite);
+	}
+
+	@Test
+	public void checkPlusButtonExists() {
+		onView(withId(R.id.button_add))
+				.perform(click());
+		onData(anything())
```
84n4n4: theres already tests accessing the categories, there should be no additional matchers required.

		onData(allOf(is(instanceOf(String.class)), is(UiTestUtils.getResourcesString(R.string.category_motion))))
				.inAdapterView(BrickCategoryListMatchers.isBrickCategoryView())
				.perform(click());

when using on data, use the onData(HERE) to specify the type of data you want to have, and use inAdapterView to only specify the adapter you want to get.

DollyVolley: Ok, I already checked other tests but I must have overlooked it.
Big thanks for the specific feedback!

catroid/src/main/java/org/catrobat/catroid/ui/fragment/AddBrickFragment.java
DollyVolley: Just to make sure: You mean both fragments should inherit from ListFragment?
This would be a strong case of code duplication IMO.

Or did you mean to inherit from AddBrickFragment and just overwrite the onCreateView method?
Sorry for the inconvenience

wslany:
Thanks!

PR #3350, opened 2019-09-12
CATROID-338 fix touches actor name change does not affect formulas

SchmidtChristian:
- add check in RecyclerViewFragment to call rename instead of setter
- change renameSpriteInCollisionFormulas iteration over bricks
- add Testcases similar to CATROID-258


wslany:
Thanks!

PR #3352, opened 2019-09-15
JENKINS increased jvm heap to overcome memory leakage in JVM tests

PlantinumX:
* fixed response code in testLoginWithNotExistingUser 803 instead of 764

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,7 +97,7 @@ pipeline {
 
                                 // Checks that the creation of standalone APKs (APK for a Pocketcode app) works, reducing the risk of breaking gradle changes.
                                 // The resulting APK is not verified itself.
-                                sh """./gradlew assembleStandaloneDebug ${useWebTestParameter()} -Papk_generator_enabled=true -Psuffix=generated817.catrobat \
+                                sh """./gradlew copyAndroidNatives assembleStandaloneDebug ${useWebTestParameter()} -Papk_generator_enabled=true -Psuffix=generated817.catrobat \
                                             -Pdownload='https://share.catrob.at/pocketcode/download/817.catrobat'"""
 
                                 // Build the flavors so that they can be installed next independently of older versions.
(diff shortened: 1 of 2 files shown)


84n4n4:
this does not concern "instrumented" tests, commit message misleading-

as far as i found this mostly occurs on robolectric tests, therefore JVM tests.

Jenkinsfile
```diff
@@ -73,6 +73,7 @@ pipeline {
 
     stages {
         stage('All') {
+
```
84n4n4: are all of the changes in this file just linebreaks and formatting, or are there any value changes aswell?

catroid/src/androidTest/java/org/catrobat/catroid/test/web/AuthenticationCallsTest.java
```diff
@@ -68,7 +68,7 @@
 	private static final int STATUS_CODE_USER_ADD_EMAIL_EXISTS = 757;
 	private static final int STATUS_CODE_USER_EMAIL_INVALID = 765;
 	private static final int STATUS_CODE_AUTHENTICATION_FAILED = 601;
-	private static final int STATUS_CODE_USER_NOT_EXISTING = 764;
+	private static final int STATUS_CODE_USER_NOT_EXISTING = 803;
```
84n4n4: what is this doing in this PR?

Jenkinsfile
PlantinumX: just one change with new gradle update the gradle task copyAndroidNatives is not working anymore

84n4n4: ok, now i see it

catroid/src/androidTest/java/org/catrobat/catroid/test/web/AuthenticationCallsTest.java
PlantinumX: the share guys changed the status code of user non existing to 803 and because these the loginWithNonExisitingUser failed I wanted to fix it as well within this PR. I know it should be a separated ticket but it was a overkill in my eyes

Jenkinsfile
PlantinumX: task copyAndrioidNatives is not being executed. so both paralles stages are executing the task.

catroid/src/androidTest/java/org/catrobat/catroid/test/web/AuthenticationCallsTest.java
PlantinumX: the share guys changed their status code from 764 to 803. Therefore the testcase was failing all the time. To see a whole pass on jenkins changed it to 803. And to reduce opening a new PR i changed it in these PR.

84n4n4:
changing the title in the PR does not change the commit message.

wslany:
Thanks!

PR #3353, opened 2019-09-19
Add dependency injection with Dagger 2

SimonStefan:
Hi,
I've seen that you still don't have dependency injection in your project. I've implemented a base version that currently only injects the `ProjectManager` into the `MainActivity`. With regard to the refactoring design pattern it does not introduce any breaking API changes.
Open for discussion

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -45,6 +45,7 @@ import org.catrobat.catroid.content.bricks.Brick;
 import org.catrobat.catroid.content.bricks.Brick.BrickField;
 import org.catrobat.catroid.content.bricks.FormulaBrick;
 import org.catrobat.catroid.content.bricks.SetPenColorBrick;
+import org.catrobat.catroid.dagger.EagerSingleton;
 import org.catrobat.catroid.exceptions.CompatibilityProjectException;
 import org.catrobat.catroid.exceptions.LoadingProjectException;
 import org.catrobat.catroid.exceptions.OutdatedVersionProjectException;
@@ -66,9 +67,9 @@ import java.util.List;
 import static org.catrobat.catroid.common.Constants.CURRENT_CATROBAT_LANGUAGE_VERSION;
 import static org.catrobat.catroid.common.Constants.PERMISSIONS_FILE_NAME;
 
-public final class ProjectManager {
+public final class ProjectManager implements EagerSingleton {
 
-	private static final ProjectManager INSTANCE = new ProjectManager();
+	private static ProjectManager instance;
 	private static final String TAG = ProjectManager.class.getSimpleName();
 
 	private Project project;
@@ -77,13 +78,34 @@ public final class ProjectManager {
 	private Scene startScene;
 	private Sprite currentSprite;
 
-	private ProjectManager() {
+	private Context applicationContext;
+
+	public ProjectManager(Context applicationContext) {
+		if (instance != null) {
+			throw new RuntimeException("ProjectManager should be instantiated only once");
+		}
+		this.applicationContext = applicationContext;
+		instance = this;
 	}
 
+	/**
+	 * Replaced with dependency injection
+	 *
+	 * @deprecated use dependency injection with Dagger instead.
+	 */
+	@Deprecated
 	public static ProjectManager getInstance() {
-		return INSTANCE;
+		return instance;
 	}
 
+	public void loadProject(File projectDir) throws ProjectException {
+		loadProject(projectDir, applicationContext);
+	}
+
+	/**
+	 * @deprecated use {@link #loadProject(File projectDir)} without Context instead.
+	 */
+	@Deprecated
 	public void loadProject(File projectDir, Context context) throws ProjectException {
 
 		Project previousProject = project;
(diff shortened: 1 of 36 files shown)


thmq:
Interesting idea. Looks promising, but I'd have to give it more thought. I'm just answering now so that you know that someone's thinking about it and so that you don't assume that no one cares :smirk_cat:. Besides, we should definitely discuss how to approach integrating it into all Activities. Would make sense to include: @ThomasSchwengler and @84n4n4 into the discussion.

SimonStefan:
@thmq @ThomasSchwengler @84n4n4 thank's for the fast reply. It took some time to make the necessary adjustments to the unit tests and instrumented tests without having to make changes in every file. Regarding your request to extend dependency injection to all activities I would prefer to implement this in another pull request to keep the reviews short. If you want we can even discuss this at university

ThomasSchwengler:
You are of course not expected to expand this into every activity in this pull request. We as a team should however think about defining _a_ standard way of writing activities, meaning we should in the future add dagger2 to all activities (and possibly move to one of the common MVC patterns, be it MVVM or MVP).
Adding a DI framework is a good thing, I also like the idea of deprecating the current singletons.
We have to enforce the usage of DI more, with or without dagger2.

For me it's an 'aye'.

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -77,13 +78,34 @@
...
 	}
 
+	/**
+	 * Replaced with dependency injection
+	 *
+	 * @deprecated use dependency injection with Dagger instead.
+	 */
+	@Deprecated
 	public static ProjectManager getInstance() {
 		return INSTANCE;
```
ThomasSchwengler:
```
[2019-10-01T07:16:01.665Z] /home/catroid/jenkins_slave/workspace/Catroid_PR-3353/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java:98: error: cannot find symbol
[2019-10-01T07:16:01.665Z] 		return INSTANCE;
[2019-10-01T07:16:01.665Z] 		       ^
[2019-10-01T07:16:01.665Z]   symbol:   variable INSTANCE
[2019-10-01T07:16:01.665Z]   location: class ProjectManager
```

build.gradle
```diff
@@ -24,7 +24,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.10'
+    ext.kotlin_version = '1.3.30'
```
ThomasSchwengler: Is there any reason why the version is raised in this PR?

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
SimonStefan: You should never rename a variable manually 😉
Resolved

build.gradle
SimonStefan: No. I've set it to a version were I used Dagger successfully to avoid possible incompatibilities during development. Everything works fine with `1.3.10`. Should I revert it?

ThomasSchwengler: No, I guess it's fine.

SimonStefan:
@ThomasSchwengler It seems that the current test suite is broken. The standalone project cannot be found (https://share.catrob.at/pocketcode/download/817.catrobat)

I've rebased the branch onto the latest develop. Is there anything I can do to help you with the decision? At our company we are using ViewModel from the Android Jetpack with automated constructor injection of all dependencies with Dagger2. That's nice for writing independent tests that can be executed really quick on the local machine.

SimonStefan:
ok, project seems to be online again. 
@catrobatjenkins test this please

SimonStefan:
@thmq @ThomasSchwengler @84n4n4 
how went your discussion regarding this pull request? In #3365 and #3374 I extend the usage.
After that, I would like to contribute more by refactoring existing code. Would you prefer me to 

1. increase the usage of dagger and get rid of all static singletons
2. show improved testability with the usage of ViewModels or
3. use POJOs for persistency

The latter is should include some planning beforehand.

LPeteR90:
Seems to me nobody has any objections currently... Code LGTM, I think can help us in the long run & would like to approve this PR - provided contributors who will have to shoulder the load of introducing people to the concept of DI and working with the changes themselves are ok with it.  
 
The proposed order of refactoring (get rid of static singletons, introduce VMs) appears sensible to me. An architecture pattern like MVVM could really benefit our codebase - if used consistently throughout the application. However, this would require Bricks not to be views anymore (point nr. 3 on your list), which would be a massive change & require planning, as you pointed out correctly. 

Ping @ThomasSchwengler @84n4n4

ThomasSchwengler:
@SimonStefan please squash some commits (all fixup commits). Doesn't have to be one commit, but please remove those fixups.
After that I will approve it :+1:

LPeteR90:
Just retriggered a Jenkins run that failed due to two very strange test failures...

Currently, we are in a rather cumbersome release phase (Backend Changes) so this PR should probably be merged only after the release branch has been merged back to master and develop... Will approve after that. 

Sorry for the ongoing delay :/

SimonStefan:
@ThomasSchwengler everything should be fine now. please merge

wslany:
Could you resolve the conflict? Thanks!

wslany:
Thanks!! (note: the failing tests on jenkins are due to network authentication issues that are unrelated to this PR)

PR #3355, opened 2019-09-24
CATROID-360 fix broadcast messages lost in compound bricks

SchmidtChristian:
- change iteration through bricks
- add testcase for nested and nested with secondary list


wslany:
Thanks!

PR #3357, opened 2019-09-25
CATROID-378 fix Formulaeditor crashes after variable/list rename and discard

SchmidtChristian:
- add function to update uservariable and userlist in InternFormulaState
- call function in FormulaEditorEditText
- add testcase for function


wslany:
Thanks!

PR #3358, opened 2019-09-25
MAINTENANCE: Solve lint issues and reenable lint rules

ThomasSchwengler:
Fixed Lint warnings
---

* Introduce a baseline file that will set our lint violation count to zero
  but will fail when there are new violations. This means that, while many
  issues are not fixed in this commit, we hopefully will see fewer violations
  in the future
* Fix InvalidPackage
  * Add a list of offenders to the lint ignore file
* Fix RtlHardcoded
* Fix UnusedAttribute
* Fix CommitPrefEdits
  * Add some changes to ensure that a `commit`/`apply` happens on every `SharedPreferencesEditor`
* Fix WifiManagerLeak
  * This was never an issue
* Fix ObsoleteSdkInt
* Fix IconColors
* Fix RelativeOverlap
* Fix StaticFieldLeak
  * These have to be solved on a case-by-case basis
* Fix WrongRegion
* Fix RtlSymmetry
* Fix ContentDescription
* Fix ClickableViewAccessibility
* Fix IconMissingDensityFolder
* Fix HandlerLeak
* Fix GradleCompatible
  * Update the support library to `28.0.0` as recommended by this rule
* Fix ApplySharedPref
  * According to the android documentation any `commit` can be replaced with `apply`
    The difference is while `apply` also sets the values to `SharedPreferences` instantly,
    it writes to persistent storage in the background.
* Fix InflateParams
  * Seems that this has been addressed already
* Fix IconDensities
* Fix UnusedResource (partially)

Additionally
---

* Gradle adds a lint task for each `flavor * build type = 6 * 3 = 18` lint tasks when just running `lint`
* Enabling those lint rules slows down each lint task by about *10 seconds*, adding up to *3 minutes* additional runtime overall
* This seems unsustainable and very unnecessary, as we have pretty much no flavor specific code.
* Suggested resolution: Only lint the catroid flavor
* Expected outcome: linting only the catroid flavor reduces linting time from `8` minutes to `~1` minute

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
@@ -196,7 +196,8 @@ public class StageResourceHolder implements GatherCollisionInformationTask.OnPol
 										.findViewById(R.id.dialog_terms_of_use_check_box_agree_permanently);
 								PreferenceManager.getDefaultSharedPreferences(stageActivity)
 										.edit()
-										.putBoolean(SETTINGS_PARROT_AR_DRONE_CATROBAT_TERMS_OF_SERVICE_ACCEPTED_PERMANENTLY, checkBox.isChecked());
+										.putBoolean(SETTINGS_PARROT_AR_DRONE_CATROBAT_TERMS_OF_SERVICE_ACCEPTED_PERMANENTLY, checkBox.isChecked())
+										.apply();
 								onDroneTermsOfUseAgreed();
 							}
 						})
@@ -244,7 +245,8 @@ public class StageResourceHolder implements GatherCollisionInformationTask.OnPol
 								PreferenceManager.getDefaultSharedPreferences(stageActivity)
 										.edit()
 										.putBoolean(SETTINGS_PARROT_JUMPING_SUMO_CATROBAT_TERMS_OF_SERVICE_ACCEPTED_PERMANENTLY,
-												checkBox.isChecked());
+												checkBox.isChecked())
+										.apply();
 								onJSDroneTermsOfUseAgreed();
 							}
 						})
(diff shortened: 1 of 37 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
```diff
@@ -196,7 +196,8 @@ public void onClick(DialogInterface dialog, int which) {
 										.findViewById(R.id.dialog_terms_of_use_check_box_agree_permanently);
 								PreferenceManager.getDefaultSharedPreferences(stageActivity)
 										.edit()
-										.putBoolean(SETTINGS_PARROT_AR_DRONE_CATROBAT_TERMS_OF_SERVICE_ACCEPTED_PERMANENTLY, checkBox.isChecked());
```
thmq: no `.commit()` :thinking: how did that even work?

ThomasSchwengler: I have honestly no idea. In my opinion this should never have worked 🤷

ThomasSchwengler:
Interestingly lint still has one warning - that we have a baseline file 😅 - I will see what I can do about that tomorrow

ThomasSchwengler:
image

wslany:
What's with these warnings in the builds, can we ignore them and merge this?

thmq:
> What's with these warnings in the builds, can we ignore them and merge this?

seems strange. @ThomasSchwengler what do you think?

ThomasSchwengler:
@thmq I agree, this should only get merged after the release to avoid conflicts between the release branch and develop

thmq:
@ThomasSchwengler I think it's time to rebase :crossed_fingers:

ThomasSchwengler:
> @ThomasSchwengler I think it's time to rebase

Rebased :+1:

ThomasSchwengler:
Rebase 2, the rebase strikes back

wslany:
Thanks!

PR #3361, opened 2019-09-30
MAINTENANCE: Fix Authentication test

ThomasSchwengler:
* New web test server returns StatusCode 803 (Username valid but not found)
  See https://github.com/Catrobat/Catroweb-Symfony/blame/1d420c9b075aca14ff203226792bc988082bdd21/src/Catrobat/StatusCode.php#L77
* Fix test to reflect this change
* Neither the old status code nor the new one are used in production, so production is left unchanged


ThomasSchwengler:
@HCrane @dmetzner has there been any announcement that there will be a new version code? As far as I can see this changed about 2 months ago.

dmetzner:
@ThomasSchwengler The version code gets incremented with every release. The exact number can be found in our changelog. https://github.com/Catrobat/Catroweb-Symfony/blob/release/v2.3.0/CHANGELOG.md

ThomasSchwengler:
@dmetzner whoops, I meant status code

dmetzner:
@ThomasSchwengler 
The change was not announced by us. Only pointed out by @PlantinumX in our Slack channel.
API changes will be announced in the future and old API functions will only be deprecated and not directly changed. There will also be a swagger documentation for the API.
So something like this should not happen again. Sorry for any inconvenience.

wslany:
Thanks!!

PR #3362, opened 2019-10-02
CATROID-379 media download dialog update progress crashes

84n4n4:


catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java
```diff
@@ -219,14 +219,18 @@ public void createProgressDialog(String mediaName) {
...
+			progressDialog.show();
+		}
 	}
 
 	public void updateProgressDialog(long progress) {
 		if (progress == 100) {
-			progressDialog.setProgress(progressDialog.getMax());
-			setResult(RESULT_OK, resultIntent);
-			progressDialog.dismiss();
+			if (progressDialog.isShowing() && !isFinishing()) {
```
PlantinumX: what if we completely delete the progressDialog. As a user you mostly never see the dialog anyway

84n4n4: i thought about that.
in the long run, it would be probably better, BUT....
the problem that multiple downloads could be started by fast tapping two different files still persists, and the way the webview activity handles downloads it is not possible to quickfix this.

so the webview would have to be refactored or better written from scratch to deal with all of this properly. (for example, the cache file and result intent are created on download start and saved as members (!),and then used again when download finished, which is triggered not by the download service/task callback directly, but the progress update of the dialog to "100") :-P

for now i think it would be best to just try to prevent the crashes. (which is not guaranteed with this fix anyways, since i still cannot reproduce them)

wslany:
Thanks. There must still be other issues with this whole media download fails, but probably unrelated since no crashes are occurring: I just tried to select some media gallery looks with the apk from this pr, and the app was just returned to the overview of looks of the object, but without any new look being actually downloaded from the media gallery. Repeatedly. Somehow selecting a look from the media gallery works fine, but then it never gets downloaded. It is an intermittent issue, however, as sometimes later (after unsuccessfully trying it 5 times) it worked again, with the same look from the media lib...

PR #3365, opened 2019-10-03
Extend dependency injection proposed in #3353 

SimonStefan:
requires #3353 

Adds dependency injection for DefaultProjectHandler, XstreamSerializer, StatusBarNotificationManager and DownloadUtil


SimonStefan:
The code of this PR is also available in #3374

PR #3367, opened 2019-10-04
CATROID-346 Remove logging from FormulaElement

84n4n4:
Logging in FormulaElement causes crashes and massiv slowdown with
nested functions. (eg. join) due to the huge amount of log output
that is generated.


84n4n4:
from what i found online there is NO possible feature in java reflections that can be used to find out what imports a .class has. therefore it is not possible to write a test that prevents the next guy from just adding the logging back. there is checkstyle that can do this, but this is not intended to be used on a single files.

wslany:
Thanks!

PR #3369, opened 2019-10-09
CATROID-367 Find BLE devices

PlantinumX:
- add request for location permission in onCreate of Activity
- add ScanCallbackMethod which adds dive to adapter
- add BLE Scan to DoDiscoveryMethod
- call classic discovery after BLE Scan and when there are no permissions

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/common/bluetooth/ConnectionDataLogger.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/common/bluetooth/ConnectionDataLogger.java
@@ -27,8 +27,6 @@ import android.support.annotation.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 
 import org.catrobat.catroid.bluetooth.base.BluetoothConnection;
-import org.catrobat.catroid.bluetooth.base.BluetoothDevice;
-import org.catrobat.catroid.common.bluetooth.models.DeviceModel;
 
 import java.util.ArrayList;
 import java.util.concurrent.BlockingQueue;
@@ -51,10 +49,6 @@ public final class ConnectionDataLogger {
 		return getNextSentMessage(0, 0);
 	}
 
-	public byte[] getNextSentMessage(int messageOffset) {
-		return getNextSentMessage(messageOffset, 0);
-	}
-
 	public byte[] getNextSentMessage(int messageOffset, int messageByteOffset) {
 		return getNextMessage(sentMessages, messageOffset, messageByteOffset);
 	}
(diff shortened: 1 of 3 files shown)


catroid/src/androidTest/java/org/catrobat/catroid/common/bluetooth/ConnectionDataLogger.java
```diff
@@ -181,38 +155,10 @@ private ConnectionDataLogger(boolean local) {
...
-	}
-
-	public static ConnectionDataLogger createBluetoothConnectionLoggerWithTestDevice(BluetoothDevice testDevice) {
-		BluetoothTestUtils.hookInTestDevice(testDevice);
-		return new ConnectionDataLogger(false);
-	}
-
 	public void disconnectAndDestroy() {
 		if (connectionProxy != null) {
 			connectionProxy.disconnect();
```
DinosaurierRex: LGTM since every dependence is also deleted. Also unused at current develop :+1:

catroid/src/main/java/org/catrobat/catroid/bluetooth/ConnectBluetoothDeviceActivity.java
```diff
@@ -135,7 +139,17 @@ public void onReceive(Context context, Intent intent) {
...
+					if ((device.getType() == DEVICE_TYPE_CLASSIC || device.getType() == DEVICE_TYPE_DUAL)
+							&& newDevicesArrayAdapter.getPosition(deviceInfo) < 0) {
+						newDevicesArrayAdapter.add(deviceInfo);
+					}
+					if (device.getType() == DEVICE_TYPE_LE || device.getType() == DEVICE_TYPE_DUAL) {
+						String deviceInfoBLE = "BLE - " + deviceInfo;
+						if (newDevicesArrayAdapter.getPosition(deviceInfoBLE) < 0) {
+							newDevicesArrayAdapter.add(deviceInfoBLE);
+						}
+					}
```
DinosaurierRex: Since only the string representation is splitting the device type, someone could exploit these name by calling himself "BLE - " which leads to false behaviour ... You could try writing a custom adapter to show only strings but save type in an object behind. https://stackoverflow.com/questions/17818969/displaying-custom-objects-in-arrayadapter-the-easy-way

catroid/src/main/java/org/catrobat/catroid/bluetooth/ConnectBluetoothDeviceActivity.java
```diff
@@ -117,12 +124,14 @@ private String getSelectedBluetoothAddress(View view) {
...
 			String address = getSelectedBluetoothAddress(view);
-			if (address == null) {
+			Pair pair = newDevicesArrayAdapter.getItem(position);
+			if (address == null || pair == null) {
 				return;
 			}
-			connectDevice(address);
+			if (pair.second.equals(DEVICE_TYPE_CLASSIC)) {
+				connectDevice(address);
+			}
```
DinosaurierRex: Since the logic for connecting is done in another ticket, this seems to be fine :+1:

catroid/src/main/java/org/catrobat/catroid/bluetooth/ConnectBluetoothDeviceActivity.java
```diff
@@ -221,7 +245,15 @@ public void onClick(View view) {
...
-		newDevicesArrayAdapter = new ArrayAdapter<String>(this, R.layout.device_name);
+		newDevicesArrayAdapter = new ArrayAdapter<Pair>(this, R.layout.device_name,
+				new ArrayList<Pair>()) {
+			@Override
+			public View getView(int position, View convertView, ViewGroup parent) {
+				TextView view = (TextView) super.getView(position, convertView, parent);
+				view.setText((String) getItem(position).first);
+				return view;
+			}
+		};
```
DinosaurierRex: :+1:

wslany:
Also tested successfully with Arduino and Bluetooth 2.x - thanks!

PR #3379, opened 2019-10-23
Share Update Fix - Http Headers contentDisposition has changed

84n4n4:
under discussion...

either web team fixes this to format the contentDisposition in http header the same as they did before, or we can fix this via this. (whats going on with catty regarding this issue?)


84n4n4:
fixed on share side

PR #3382, opened 2019-10-29
Catroid-167 Refactor ProjectDownloadService to foreground service

gPathpp:
delete DownloadUtil
remove context from StatusBarNotificationManager
add ProjectDownloader and MediaDownloader


catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectDownloadService.kt
```diff
@@ -0,0 +1,165 @@
...
+                DownloadSuccessCallback(this, projectName, destinationFile, renameAfterDownload, resultReceiver, id),
+                DownloadErrorCallback(this, resultReceiver, id),
+                DownloadProgressCallback(this, resultReceiver, id))
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            stopForeground(STOP_FOREGROUND_DETACH)
+        }
+    }
+
+    private class DownloadSuccessCallback(
```
LPeteR90: Refactor private classes to functions

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
```diff
@@ -0,0 +1,183 @@
...
+) : Serializable {
+    private val callbackWeakReference = WeakReference<ProjectDownloadCallback>(callback)
+
+    companion object {
+        private const val serialVersionUID = 42L
+        private const val FILENAME_TAG = "fname="
+        private val TAG = ProjectDownloader::class.java.simpleName
+
+        fun getProjectNameFromUrl(url: String): String? {
+            val projectNameIndex = url.lastIndexOf(FILENAME_TAG) + FILENAME_TAG.length
```
84n4n4: please check out URLUtil from webkit package: guessFileName()

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
```diff
@@ -0,0 +1,183 @@
...
+                    callbackWeakReference.get()?.onDownloadFinished(projectName, url)
+                    queue.finished(projectName)
+                }
+                ERROR_CODE -> queue.finished(projectName)
+            }
+        }
+    }
+
+    interface ProjectDownloadQueue {
+        fun projectNameNotAvailable(projectName: String): Boolean
```
84n4n4: please rename to projectNameAlreadyTakenLocally

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
```diff
@@ -0,0 +1,183 @@
...
+
+object GlobalProjectDownloadQueue {
+    val queue = ProjectDownloadQueue()
+
+    class ProjectDownloadQueue : ProjectDownloader.ProjectDownloadQueue {
+        private var projectNameSet: Set<String> = Collections.synchronizedSet(HashSet<String>())
+
+        @Synchronized
+        override fun projectNameNotAvailable(projectName: String): Boolean =
+            alreadyInQueue(projectName) || File(DEFAULT_ROOT_DIRECTORY, projectName).exists()
```
84n4n4: please remove this method from the download queue, theres already some functionality somewhere that provides you a list of local projects that you can check against. (without having this class know about the default root directory or files there)

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    fun downloadOverwriteExistingProject(
+        context: Context,
+        projectName: String,
+        renameProject: Boolean
+    ) {
+        synchronized(queue) {
+            if (queue.alreadyInQueue(projectName)) {
```
84n4n4: check the escaping of special characters!!!
pocket code escapes spec. chars as far as i know,.so will share for display in the url (eg. "?",""",etc)

catroid/src/test/java/org/catrobat/catroid/test/transfers/ProjectDownloaderTest.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ProjectDownloader.class, ReplaceExistingProjectDialogFragment.class, ToastUtil.class})
+public class ProjectDownloaderTest {
+
+	//Doesn't has to be an original link to an project just the project name has to be in the link
```
LPeteR90: Is this comment obsolete? Not sure what additional information a dev gets from it...

gPathpp: Is obsolete, thanks!

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
```diff
@@ -0,0 +1,183 @@
...
+                URLDecoder.decode(projectNameUTF8, "UTF-8")
+            } catch (e: UnsupportedEncodingException) {
+                Log.e(TAG, "Could not decode project name: $projectNameUTF8", e)
+                null
+            }
+        }
+    }
+
+    fun download(activity: AppCompatActivity) {
+        val projectName = getProjectNameFromUrl(url) ?: return
```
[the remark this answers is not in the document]
gPathpp: The log output is in the function "getProjectNameFromUrl". I don't know if it is necessary to inform the User about this, what's your opinion on this?

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectDownloadService.kt
```diff
@@ -0,0 +1,165 @@
...
+            ToastUtil.showError(this, R.string.error_project_download)
+            return
+        }
+
+        val id = downloadIntent.getIntExtra(EXTRA_NOTIFICATION_ID, 0)
+        val statusBarNotificationManager = StatusBarNotificationManager.getInstance()
+        val notification = statusBarNotificationManager.getNotification(this, id)
+
+        // startForeground does't do the same thing in previous versions of Android
+        // (it makes the notification buggy)
```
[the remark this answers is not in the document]
gPathpp: Exact same problem. StartForeground attaches the notification to the Service and StopForeground (false) should detach it, but unfortunatly it doesn't (The notification gets dismissed after the execution of the service). Luckily stopForeground(STOP_FOREGROUND_DETACH) works fine, but is only available after version O. Since we don't need the StartForeground method before version O...  I'll try to write better comments ^^ Or should I refactor it to your version?

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
gPathpp: The project name in the the URL is utf8 encoded and is decoded before putting it into the queue.
To eliminate the problem that (maybe, in some way) the project name in the XML and the name in the URL are not equal, I'll remove the boolean flag (renameProjectAfter download), so that the project name will be always renamed after download.

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
gPathpp: Unfortunately this function doesn't work for our URL..

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
gPathpp: This function makes the whole thing way too slow (validate string while renaming etc..) => wrote a new function. See in ReplaceExistingProjectDialogFragment

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectDownloadService.kt
```diff
@@ -0,0 +1,182 @@
...
+                    override fun onError(code: Int, message: String) {
+                        downloadErrorCallback(context, resultReceiver, id)
+                    }
+                },
+                object : DownloadProgressCallback {
+                    override fun onProgress(progress: Long) {
+                        downloadProgressCallback(context, resultReceiver, id, progress)
+                    }
+                }
+            )
```
LPeteR90: you don't need to store context in a separate var. Use `this@ProjectDownloadService` instead.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectDownloadService.kt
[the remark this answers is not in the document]
LPeteR90: I'd prefer to have as few switches regarding the current platform version as possible. So if `ProjectUploadService` is working across platforms, I'd rather have a similar solution to that (removing the id with `stopForeground(true)`, then issuing a new notification with a different id in success/error callbacks).

catroid/src/main/java/org/catrobat/catroid/web/ProjectDownloader.kt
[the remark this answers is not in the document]
LPeteR90: As a user I would like to be notified about why my download was not started. So something like a "Your download could not be started, seems there is a problem with our server" toast would be nice. I would not expect my download to just not start. Further, based on the message users would receive here, we might also receive more sensible bug reports.

catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectDownloadService.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    private fun downloadSuccessCallback(
+        context: Context,
+        projectName: String,
+        destinationFile: File,
+        resultReceiver: ResultReceiver
+    ) {
+        val statusBarNotificationManager = StatusBarNotificationManager.getInstance()
+        val notificationId = statusBarNotificationManager.createProjectDownloadNotification(this,
+                                                                                            projectName)
```
LPeteR90: this linebreak seems kind of funny :)

gPathpp: Ups, must have been the auto-correct of android studio ^^

PR #3385, opened 2019-10-31
CATROID-324 Phiro Flavor wrong Icon and missing extension

SchmidtChristian:
- add icons (copys of launcher icon) for every resolution 
- add string in gradle file like arduino
- add the new string as default value for phiro in preferences.xml


wslany:
Merged it into the release branch, as we then can release Phiro Code with this fix already for the forthcoming release (note: there's no need to increase the version code for this PR, because I haven't uploaded Phiro Code yet to Google Play).

wslany:
Thanks!!

PR #3391, opened 2019-11-06
CATROID-399: Fix project infix for share

ThomasSchwengler:
* Move switch between uris into code
* Revert change to `/program/` infix until web has released their new version


wslany:
Thanks!

PR #3393, opened 2019-11-07
CATROID-311 Visual Placement position correction

SchmidtChristian:
- add variables for the last 2 event timestamps
- add variable for the last 2 positions
- check the last 2 movement events ins ACTION_UP
- add testcase

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
@@ -23,16 +23,33 @@
 
 package org.catrobat.catroid.visualplacement;
 
-import android.os.SystemClock;
 import android.view.MotionEvent;
 import android.widget.ImageView;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class VisualPlacementTouchListener {
 
 	private Mode mode;
-	private float previousY;
+	private static final long TAP_DELAY_THRESHOLD = 100;
+	private static final long JITTER_DELAY_THRESHOLD = 30;
+	private static final float JITTER_DISTANCE_THRESHOLD = 10.f * 10.f;
 	private float previousX;
-	private long lastTimeMove = 0;
+	private float previousY;
+	private List<TouchEventData> recentTouchEventsData = new ArrayList<>();
+
+	private final class TouchEventData {
+		private final long timeStamp;
+		private final float xCoordinate;
+		private final float yCoordinate;
+
+		private TouchEventData(long time, float x, float y) {
+			timeStamp = time;
+			xCoordinate = x;
+			yCoordinate = y;
+		}
+	}
 
 	private void setMode(Mode mode) {
 		this.mode = mode;
@@ -42,31 +59,50 @@ public class VisualPlacementTouchListener {
 		if (event.getPointerId(0) == 0) {
 			float currentX = event.getRawX();
 			float currentY = event.getRawY();
+			long motionEventTime = event.getEventTime();
 
 			switch (event.getAction()) {
 				case MotionEvent.ACTION_DOWN:
 					setMode(Mode.TAP);
 					previousX = currentX;
 					previousY = currentY;
-					lastTimeMove = SystemClock.elapsedRealtime();
+					recentTouchEventsData.add(new TouchEventData(motionEventTime, currentX, currentY));
 					break;
 				case MotionEvent.ACTION_CANCEL:
 				case MotionEvent.ACTION_UP:
 					if (mode == Mode.TAP) {
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						removeObsoleteTouchEventsData(motionEventTime);
+						float dX = currentX - previousX;
+						float dY = currentY - previousY;
+						if (!recentTouchEventsData.isEmpty() && recentTouchEventsData.size() > 1) {
+							TouchEventData oldestEntry = recentTouchEventsData.get(0);
+							float distanceCorrectionX = currentX - oldestEntry.xCoordinate;
+							float distanceCorrectionY = currentY - oldestEntry.yCoordinate;
+							double distance = distanceCorrectionX * distanceCorrectionX + distanceCorrectionY * distanceCorrectionY;
+							if (distance < JITTER_DISTANCE_THRESHOLD && distance != 0) {
+								dX -= distanceCorrectionX;
+								dY -= distanceCorrectionY;
+							}
+						}
+						imageView.setX(imageView.getX() + dX);
+						imageView.setY(imageView.getY() + dY);
 					}
 					break;
 				case MotionEvent.ACTION_MOVE:
-					long delayTime = SystemClock.elapsedRealtime() - lastTimeMove;
-					if (delayTime > 100) {
+					long delayTime = motionEventTime - event.getDownTime();
+					if (delayTime > TAP_DELAY_THRESHOLD) {
 						setMode(Mode.MOVE);
 						float dX = currentX - previousX;
 						float dY = currentY - previousY;
 						imageView.setX(imageView.getX() + dX);
 						imageView.setY(imageView.getY() + dY);
+						recentTouchEventsData.add(new TouchEventData(motionEventTime, currentX, currentY));
 						previousX = currentX;
 						previousY = currentY;
+						removeObsoleteTouchEventsData(motionEventTime);
 					}
 					break;
 			}
@@ -76,6 +112,20 @@ public class VisualPlacementTouchListener {
 		} else {
 			return false;
 		}
+	}
+
+	private void removeObsoleteTouchEventsData(long timeStamp) {
+		List<TouchEventData> obsoleteTouchEventsData = new ArrayList<>();
+		for (TouchEventData touchEventData : recentTouchEventsData) {
+			if (touchEventData != null) {
+				if ((timeStamp - touchEventData.timeStamp) > JITTER_DELAY_THRESHOLD) {
+					obsoleteTouchEventsData.add(touchEventData);
+				} else {
+					break;
+				}
+			}
+		}
+		recentTouchEventsData.removeAll(obsoleteTouchEventsData);
 	}
 
 	public enum Mode {MOVE, TAP}
--- a/catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
+++ b/catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
@@ -1,0 +1,173 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2018 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.visualplacement;
+
+import android.view.MotionEvent;
+import android.widget.ImageView;
+
+import org.catrobat.catroid.visualplacement.CoordinateInterface;
+import org.catrobat.catroid.visualplacement.VisualPlacementTouchListener;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.mockito.ArgumentMatchers.anyFloat;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class VisualPlacementPositionCorrectionTest {
+
+	@Mock
+	ImageView imageView;
+
+	@Mock
+	MotionEvent motionEvent;
+
+	@Mock
+	CoordinateInterface coordinateInterface;
+
+	@InjectMocks
+	VisualPlacementTouchListener listener;
+
+	private final Float initalPositionX = 1.f;
+	private final Float initalPositionY = 1.f;
+	private final Float firstMovementPositionX = 5.f;
+	private final Float firstMovementPositionY = 5.f;
+	private final Float secondMovementPositionX = 6.f;
+	private final Float secondMovementPositionY = 6.f;
+	private final Float actionUpMovementPositionX = 7.f;
+	private final Float actionUpMovementPositionY = 7.f;
+	private final long firstMovementTimestamp = 150;
+	private final long secondMovementTimestamp = 170;
+	private final long actionUpTimestamp = 175;
+
+	@Before
+	public void setUp() {
+		when(motionEvent.getDownTime()).thenReturn((long) 0);
+	}
+
+	public void triggerTouchDownEvent(float startPositionX, float startPositionY) {
+		when(motionEvent.getAction()).thenReturn(MotionEvent.ACTION_DOWN);
+		when(imageView.getX()).thenReturn(startPositionX);
+		when(imageView.getY()).thenReturn(startPositionY);
+		when(motionEvent.getRawX()).thenReturn(startPositionX);
+		when(motionEvent.getRawY()).thenReturn(startPositionY);
+		listener.onTouch(imageView, motionEvent, coordinateInterface);
+	}
+
+	public void triggerMovementEvent(long timestamp, float positionX, float positionY) {
+		when(motionEvent.getAction()).thenReturn(MotionEvent.ACTION_MOVE);
+		when(motionEvent.getRawX()).thenReturn(positionX);
+		when(motionEvent.getRawY()).thenReturn(positionY);
+		when(motionEvent.getEventTime()).thenReturn(timestamp);
+
+		listener.onTouch(imageView, motionEvent, coordinateInterface);
+		when(imageView.getX()).thenReturn(positionX);
+		when(imageView.getY()).thenReturn(positionY);
+	}
+
+	public void triggerTouchUpEvent(long timestamp) {
+		when(motionEvent.getEventTime()).thenReturn(timestamp);
+		when(motionEvent.getAction()).thenReturn(MotionEvent.ACTION_UP);
+		when(motionEvent.getRawX()).thenReturn(actionUpMovementPositionX);
+		when(motionEvent.getRawY()).thenReturn(actionUpMovementPositionY);
+		listener.onTouch(imageView, motionEvent, coordinateInterface);
+	}
+
+	@Test
+	public void testTouchCorrection() {
+		triggerTouchDownEvent(initalPositionX, initalPositionY);
+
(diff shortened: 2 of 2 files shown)


catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -30,9 +30,13 @@
...
 	private Mode mode;
-	private float previousY;
 	private float previousX;
-	private long lastTimeMove = 0;
+	private float previousY;
+	private float oldX;
+	private float oldY;
+	private long touchDownTimestamp = 0;
+	private long lastTimestamp = 0;
+	private long oldTimestamp = 0;
```
DinosaurierRex: oldTimeStamp is the one before lastTimeStamp, but we couldn't find a better name. Open for suggestions

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -48,25 +52,42 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
...
 						imageView.setY(imageView.getY() + dY);
+						oldX = previousX;
+						oldY = previousY;
 						previousX = currentX;
 						previousY = currentY;
+						oldTimestamp = lastTimestamp;
+						lastTimestamp = SystemClock.elapsedRealtime();
 					}
 					break;
 			}
```
DinosaurierRex: LGTM :+1:

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,136 @@
...
+		float invalidPositionY = 20.0f;
+
+		triggerTouchDownEvent(initalPositionX, initalPositionY);
+		triggerMovementEvent(firstMovementTimestamp, firstMovementPositionX, firstMovementPositionY);
+		triggerMovementEvent(secondMovementTimestamp, invalidPositionX, invalidPositionY);
+		triggerTouchUpEvent();
+
+		verify(imageView, times(2)).setX(anyFloat());
+	}
+}
```
DinosaurierRex: Maybe checking the coordinates would be a better verification. But currently the ones holding the coordinates are mocked. Maybe @84n4n4 has some better ideas about that ?

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -48,25 +52,42 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
...
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						long delayTime = lastTimestamp - oldTimestamp;
+						Double distance =
+								Math.sqrt((previousX - oldX) * (previousX - oldX)
+										+ (previousY - oldY) * (previousY - oldY));
+						if (delayTime < 30 && distance < 10) {
+							float dX = previousX - oldX;
+							float dY = previousY - oldY;
```
ThomasSchwengler: please reuse in calculation of distance

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -48,25 +52,42 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
...
 					break;
 				case MotionEvent.ACTION_CANCEL:
 				case MotionEvent.ACTION_UP:
 					if (mode == Mode.TAP) {
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						long delayTime = lastTimestamp - oldTimestamp;
+						Double distance =
+								Math.sqrt((previousX - oldX) * (previousX - oldX)
```
ThomasSchwengler: if you compare `distance < 10*10` you can ditch the sqrt

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -48,25 +52,42 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
...
+					oldTimestamp = touchDownTimestamp;
 					break;
 				case MotionEvent.ACTION_CANCEL:
 				case MotionEvent.ACTION_UP:
 					if (mode == Mode.TAP) {
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						long delayTime = lastTimestamp - oldTimestamp;
+						Double distance =
```
ThomasSchwengler: please use `double`

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -48,25 +56,40 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
...
 					if (mode == Mode.TAP) {
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						double delayTime = currentTimestamp - previousTimestamp;
+						float dX = previousX - oldX;
+						float dY = previousY - oldY;
+						double distance = dX * dX + dY * dY;
+						if (delayTime < 30 && distance < 10 * 10) {
+							imageView.setX(imageView.getX() - dX);
```
wslany: As discussed via Slack, only the last position should be used.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+		triggerTouchUpEvent(actionUpTimestamp);
+
+		verify(imageView).setX(firstMovementPositionX);
+		verify(imageView).setY(firstMovementPositionY);
+		verify(imageView, times(3)).setX(anyFloat());
+	}
+
+	@Test
+	public void testTouchCorrectionWithInvalidDelayBetweenMovements() {
+		long invalidActionUpTimestamp = secondMovementTimestamp + 15;
```
wslany:
```suggestion
		long tooLateTimestamp = secondMovementTimestamp + 15;
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+		triggerTouchDownEvent(initalPositionX, initalPositionY);
+		triggerMovementEvent(firstMovementTimestamp, firstMovementPositionX, firstMovementPositionY);
+		triggerMovementEvent(secondMovementTimestamp, secondMovementPositionX, firstMovementPositionY);
+		triggerTouchUpEvent(invalidActionUpTimestamp);
```
wslany:
```suggestion
		triggerTouchUpEvent(tooLateTimestamp);
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+		verify(imageView, times(2)).setX(anyFloat());
```
wslany: A better test would here check that the X and Y values at the end are the ones from the secondMovement. This makes it clearer for others who want to understand the test, and also is the functional end result. The number of invocations is OK too for this specific implementation and test case.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+	@Test
+	public void testTouchCorrection() {
+		triggerTouchDownEvent(initalPositionX, initalPositionY);
+		triggerMovementEvent(firstMovementTimestamp, firstMovementPositionX, firstMovementPositionY);
+		triggerMovementEvent(secondMovementTimestamp, secondMovementPositionX, secondMovementPositionY);
+		triggerTouchUpEvent(actionUpTimestamp);
+
+		verify(imageView).setX(firstMovementPositionX);
+		verify(imageView).setY(firstMovementPositionY);
+		verify(imageView, times(3)).setX(anyFloat());
```
wslany: I don't think that this is doing what you want, though I am not sure. From my understanding, this just checks that getX was invoked with these values at some time, but not that they were necessarily the final invocations. Please change that so that it is made sure that the the final values are the ones from the first movement. This may then be clearer for others who want to understand the test. 

The number of invocations test in the next line  is OK too for this specific implementation and test case.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+	private final Float firstMovementPositionY = 5.f;
+	private final Float secondMovementPositionX = 6.f;
+	private final Float secondMovementPositionY = 6.f;
+	private final long firstMovementTimestamp = 150;
+	private final long secondMovementTimestamp = 170;
+	private final long actionUpTimestamp = 175;
+
+	@Before
+	public void setUp() {
+		when(timeProvider.elapsedRealtime()).thenReturn((long) 0);
```
wslany:
```suggestion
		when(motionEvent.getDownTime()).thenReturn((long) 0);
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+	@Mock
+	MotionEvent motionEvent;
+
+	@Mock
+	CoordinateInterface coordinateInterface;
+
+	@Mock
+	TimeProvider timeProvider;
+
```
wslany: Delete.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+		when(imageView.getY()).thenReturn(startPositionY);
+		listener.onTouch(imageView, motionEvent, coordinateInterface);
+	}
+
+	public void triggerMovementEvent(long timestamp, float positionX, float positionY) {
+		when(motionEvent.getAction()).thenReturn(MotionEvent.ACTION_MOVE);
+		when(motionEvent.getRawX()).thenReturn(positionX);
+		when(motionEvent.getRawY()).thenReturn(positionY);
+
+		when(timeProvider.elapsedRealtime()).thenReturn(timestamp);
```
wslany:
```suggestion
		when(motionEvent.getDownTime()).thenReturn(timestamp);
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+		when(imageView.getX()).thenReturn(positionX);
+		when(imageView.getY()).thenReturn(positionY);
+		listener.onTouch(imageView, motionEvent, coordinateInterface);
+	}
+
+	public void triggerTouchUpEvent(long timestamp) {
+		when(timeProvider.elapsedRealtime()).thenReturn(timestamp);
```
wslany:
```suggestion
		when(motionEvent.getDownTime()).thenReturn(timestamp);
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementTouchListenerTest.java
```diff
@@ -56,9 +58,17 @@
...
 
+	@Mock
+	TimeProvider timeProvider;
+
 	@InjectMocks
 	VisualPlacementTouchListener listener;
 
+	@Before
+	public void setUp() {
+		when(timeProvider.elapsedRealtime()).thenReturn((long) 0);
```
wslany:
```suggestion
		when(firstEvent.getDownTime()).thenReturn((long) 0);
```
I am not sure whether that's the right MotionEvent here. Please correct if it is incorrect for setUp().

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementTouchListenerTest.java
```diff
@@ -27,7 +27,9 @@
 import android.widget.ImageView;
 
 import org.catrobat.catroid.visualplacement.CoordinateInterface;
+import org.catrobat.catroid.visualplacement.TimeProvider;
```
wslany:
```suggestion
```

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementTouchListenerTest.java
```diff
@@ -56,9 +58,17 @@
 	@Mock
 	CoordinateInterface coordinateInterface;
 
+	@Mock
+	TimeProvider timeProvider;
+
```
wslany: Delete.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.test.visualplacement;
+
+import android.view.MotionEvent;
+import android.widget.ImageView;
+
+import org.catrobat.catroid.visualplacement.CoordinateInterface;
+import org.catrobat.catroid.visualplacement.TimeProvider;
```
wslany:
```suggestion
```

wslany:
I don't know why, but https://github.com/Catrobat/Catroid/pull/3393/files shows more change suggestions. Some seem to be missing above. However, they are also visible when one is not logged in on the https://github.com/Catrobat/Catroid/pull/3393/files --- maybe a bug in Github? Please check there too... sorry.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -23,50 +23,88 @@
...
 				case MotionEvent.ACTION_CANCEL:
 				case MotionEvent.ACTION_UP:
 					if (mode == Mode.TAP) {
 						imageView.setX(event.getX() - (float) imageView.getWidth() / 2);
 						imageView.setY(event.getY() - (float) imageView.getHeight() / 2);
+					} else {
+						removeOldEvents(currentTimestamp);
+						if (!eventData.isEmpty()) {
+							TouchEventData oldestEntry = eventData.get(0);
+							float dX = previousX - oldestEntry.xCoordinate;
```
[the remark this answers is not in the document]
SchmidtChristian: Thats the point I was talking with you about on friday, between the last movement and the up event is about ~5 ms so previousX is in 99% of the cases the same as currentX.
From the tests I made I can see that  in rare cases I get a deviation on one parameter of about 0.5 at most. Only if I remove the finger really fast with a swipe movement after holding it I get noticable deviations and as we discussed this isn't really a behaviour we want to correct with this anyways.

I think because of the small delay between the last movement and the up it wouldn't really make a noticable difference if we use the previousX or the currentX.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -23,50 +23,88 @@
...
+							float dY = previousY - oldestEntry.yCoordinate;
+							double distance = dX * dX + dY * dY;
+							if (distance < DISTANCE_THRESHOLD && distance != 0) {
+								imageView.setX(imageView.getX() - dX);
+								imageView.setY(imageView.getY() - dY);
+							}
+						}
```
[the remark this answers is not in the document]
SchmidtChristian: this basically depends on (as described in my comment on Line 86) if you want the current or the previous, if you want the current then this doesn't even need an else because then we have to do this always before (possible) position correction.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -23,50 +23,88 @@
...
 
+	public VisualPlacementTouchListener(TimeProvider timeProvider) {
+		this.timeProvider = timeProvider;
+	}
+
 	public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterface coordinateInterface) {
 		if (event.getPointerId(0) == 0) {
 			float currentX = event.getRawX();
 			float currentY = event.getRawY();
+			long currentTimestamp = timeProvider.elapsedRealtime();
```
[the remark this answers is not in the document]
SchmidtChristian: Oh neat, with this we don't need the interface to avoid powermocking.
I changed it in my repo but getDownTime() only gives us the timestamp of the touchdown event, but getEventTime() fits here pretty well.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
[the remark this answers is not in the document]
wslany: I would still use currentX/Y, because it is more likely that these values then will be congruent with the values given by getX/Y in lines 84-85 below. In that way, there can never be any drift. It's on the safe side to use currentX/Y, even if in 99% of the cases it should be the same as previousX/Y. But with currentX/Y, it is 100%, so why not use it?

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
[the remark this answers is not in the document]
wslany: Let's go with the currentX/Y please.

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
```diff
@@ -42,31 +59,46 @@ public boolean onTouch(ImageView imageView, MotionEvent event, CoordinateInterfa
 		if (event.getPointerId(0) == 0) {
 			float currentX = event.getRawX();
 			float currentY = event.getRawY();
+			long timeSinceTouchDown = event.getEventTime();
```
wslany: Oh, thanks for correcting this. I misunderstood the function. OK, in that case, however, the name I suggested is not correct anymore, since it now again is only a timestamp. I thus propose to rename it here (and at all occurrences) to:
```suggestion
			long motionEventTime = event.getEventTime();
```

catroid/src/main/java/org/catrobat/catroid/visualplacement/VisualPlacementTouchListener.java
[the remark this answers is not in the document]
SchmidtChristian: if we use previous it is always congruent with the values given by getX/Y in lines 84-85
We would have to set them before we get to this point in the programm on the current values, I changed it to current but I am not sure if a deviating value for the current Value here wouldn't be the same as an unwanted jitter right when lifting the finger.

catroid/src/test/java/org/catrobat/catroid/test/visualplacement/VisualPlacementPositionCorrectionTest.java
SchmidtChristian: the values get set onto the mock of CoordinateInterface using this code
`coordinateInterface.setXCoordinate(imageView.getX());` 
We already mock the getX() Method of the imageView mock before calling the method, inside it would be set to the right value with setX() but if I try to verify in the end it would still use the old value so I can't check it like that.

wslany: Ah, OK, thanks!

wslany:
Thanks!!

PR #3394, opened 2019-11-07
Stitching types

shellp:
CATROID-395 Stop current stitch
CATROID-232 Start zigzag stitch with length 2 and width 10
CATROID-396 Start triple stitch with length 10


PlantinumX:
handled by @adnaribo from now on

PR #3395, opened 2019-11-08
Revert "CATROID-287 Create ReadListFromDevice and WriteListOnDevice Brick"

wslany:
Reverts Catrobat/Catroid#3343

PR #3396, opened 2019-11-08
Merge CATROID-287 into release-0.9.65

thmq:
CATROID-287 was merged into develop. However, as per @wslany 's request it should be included in the release. There were 5 other PRs that were originally merged into develop while the release branch was open. These were merged into release through #3388. Hence, it is IMO the best course of action to just merge the develop into release now in order to get #3343 into the release as well (all other options are very likely to end in merge-hell).

**TL;DR:** this PR puts CATROID-287 into the release. it replaces #3395 to avoid merge-hell.


wslany:
Thanks!

PR #3397, opened 2019-11-09
CATROID-401:  Privacy policy update 

PlantinumX:
When User updates PocketCode the AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY is already set to true so need additional check with the versionCode which is changed in each release

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/common/SharedPreferenceKeys.java
+++ b/catroid/src/main/java/org/catrobat/catroid/common/SharedPreferenceKeys.java
@@ -29,7 +29,8 @@ public final class SharedPreferenceKeys {
 	}
 
 	public static final String ACCESSIBILITY_PROFILE_PREFERENCE_KEY = "AccessibilityProfile";
-	public static final String AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY = "AgreedToPrivacyPolicy";
+	public static final String AGREED_TO_PRIVACY_POLICY_VERSION =
+			"AgreedToCurrentPrivacyPolicy";
 
 	public static final String DEVICE_LANGUAGE = "deviceLanguage";
 	public static final String LANGUAGE_TAG_KEY = "applicationLanguage";
--- a/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
@@ -63,7 +63,7 @@ import java.io.IOException;
 import java.io.InputStream;
 
 import static org.catrobat.catroid.common.FlavoredConstants.DEFAULT_ROOT_DIRECTORY;
-import static org.catrobat.catroid.common.SharedPreferenceKeys.AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY;
+import static org.catrobat.catroid.common.SharedPreferenceKeys.AGREED_TO_PRIVACY_POLICY_VERSION;
 
 public class MainMenuActivity extends BaseCastActivity implements
 		ProjectLoadTask.ProjectLoadListener {
@@ -79,10 +79,11 @@ public class MainMenuActivity extends BaseCastActivity implements
 		PreferenceManager.setDefaultValues(this, R.xml.preferences, true);
 		ScreenValueHandler.updateScreenWidthAndHeight(this);
 
-		boolean hasUserAgreedToPrivacyPolicy = PreferenceManager.getDefaultSharedPreferences(this)
-				.getBoolean(AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY, false);
-
-		if (hasUserAgreedToPrivacyPolicy) {
+		int oldPrivacyPolicyHash = PreferenceManager.getDefaultSharedPreferences(this)
+						.getInt(AGREED_TO_PRIVACY_POLICY_VERSION, 0);
+		int currentPrivacyPolicyHash = getResources().getString(R.string.dialog_privacy_policy_text)
+						.hashCode();
+		if (oldPrivacyPolicyHash == currentPrivacyPolicyHash) {
 			loadContent();
 		} else {
 			setContentView(R.layout.privacy_policy_view);
@@ -92,7 +93,9 @@ public class MainMenuActivity extends BaseCastActivity implements
 	public void handleAgreedToPrivacyPolicyButton(View view) {
 		PreferenceManager.getDefaultSharedPreferences(this)
 				.edit()
-				.putBoolean(AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY, true)
+				.putInt(AGREED_TO_PRIVACY_POLICY_VERSION, getResources()
+						.getString(R.string.dialog_privacy_policy_text)
+						.hashCode())
 				.commit();
 		loadContent();
 	}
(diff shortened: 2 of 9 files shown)


catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -82,7 +82,9 @@ protected void onCreate(Bundle savedInstanceState) {
 		boolean hasUserAgreedToPrivacyPolicy = PreferenceManager.getDefaultSharedPreferences(this)
 				.getBoolean(AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY, false);
 
-		if (hasUserAgreedToPrivacyPolicy) {
+		boolean hasUserUpdatedPocketCode = hasUserUpdatedPocketCode();
```
84n4n4: unnecessary boolean field. just call the method inline below, the variable doesnt add any value for the reader since it literally has the same name as the method.

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -97,6 +99,26 @@ public void handleAgreedToPrivacyPolicyButton(View view) {
...
+		try {
+			long versionCode =
+					getPackageManager().getPackageInfo(getPackageName(), 0).getLongVersionCode();
+
+			if (PreferenceManager.getDefaultSharedPreferences(this).getInt("lastUpdateVersion", 0)
+					!= versionCode) {
+
+				PreferenceManager.getDefaultSharedPreferences(this)
+						.edit()
+						.putLong("lastUpdateVersion", versionCode)
```
84n4n4: this should be defined as a constant in SharedPreferenceKeys.java

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -97,6 +99,26 @@ public void handleAgreedToPrivacyPolicyButton(View view) {
 		loadContent();
 	}
 
+	private boolean hasUserUpdatedPocketCode() {
+		try {
+			long versionCode =
+					getPackageManager().getPackageInfo(getPackageName(), 0).getLongVersionCode();
+
+			if (PreferenceManager.getDefaultSharedPreferences(this).getInt("lastUpdateVersion", 0)
```
84n4n4: see comment below

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -97,6 +99,26 @@ public void handleAgreedToPrivacyPolicyButton(View view) {
 		loadContent();
 	}
 
+	private boolean hasUserUpdatedPocketCode() {
+		try {
+			long versionCode =
```
84n4n4: name this currentVersionCode

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -97,6 +99,26 @@ public void handleAgreedToPrivacyPolicyButton(View view) {
...
+			if (PreferenceManager.getDefaultSharedPreferences(this).getInt("lastUpdateVersion", 0)
+					!= versionCode) {
+
+				PreferenceManager.getDefaultSharedPreferences(this)
+						.edit()
+						.putLong("lastUpdateVersion", versionCode)
+						.commit();
+				return true;
+			}
+		} catch (Exception exception) {
```
84n4n4: please narrow down exceptions, this throws PackageManager.NameNotFoundException

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -82,7 +82,9 @@ protected void onCreate(Bundle savedInstanceState) {
 		boolean hasUserAgreedToPrivacyPolicy = PreferenceManager.getDefaultSharedPreferences(this)
```
84n4n4: i personally would go another way and rename this to "hasUserAgreedToCurrentPrivacyPolicy", and AGREED_TO_CURRENT_PRIVACY_POLICY_PREFERENCE_KEY.

while defining it as follows in SharedPreferenceKeys.java:
	public static final String AGREED_TO_CURRENT_PRIVACY_POLICY_PREFERENCE_KEY = "AgreedToPrivacyPolicy_0_9_65";

whenever Privacy policy changes the value of the preference field above should be updated.
this way would not require any changes or additional methods in MainMenuActivity.

however, both versions have tradeoffs:
your version:
+ doesnt need any manual work when theres a change in privacy policy
- privacy policy pops up on every update
- additional code

my version:
+ no additional code
+ privacy policy pops up only when changed
- needs manual update of value of "AgreedToPrivacyPolicy_0_9_65"

pick your poison...

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
```diff
@@ -97,6 +99,26 @@ public void handleAgreedToPrivacyPolicyButton(View view) {
 		loadContent();
 	}
 
+	private boolean hasUserUpdatedPocketCode() {
+		try {
+			long versionCode =
+					getPackageManager().getPackageInfo(getPackageName(), 0).getLongVersionCode();
```
84n4n4: just had a look at jenkins and the documentation, you cant do this, getLongVersionCode was just recently added in API28

catroid/src/main/java/org/catrobat/catroid/ui/MainMenuActivity.java
84n4n4: ok, i thought about this a bit more, one could load the privacy policy string and hash it, and use this as the value that is written in the sharedPreferences.
this way, whenever the privacy policy string is changed, it will automatically show the privacy policy, not requiring any additional manual work.
this however would end up showing the privacy policy dialog whenever user changes language, or whenever its string changed eg. due to it beeing translated to this language. (which depending on how you look at it could be a good thing aswell, showing users the policy again when its available in their language, and looking at how often users will actually switch languages through the settings menu....(rarely i guess)
)

wslany: I like the hash value approach. It has the benefit of working automatically. Note that it is ok to check it for the currently set language (of Catroid), since this would show it again once it changes in one's language at a later time, e.g., when it has finally been translated, which is desirable. Sounds like the best way to me.

catroid/src/main/java/org/catrobat/catroid/common/SharedPreferenceKeys.java
```diff
@@ -29,7 +29,8 @@ private SharedPreferenceKeys() {
 	}
 
 	public static final String ACCESSIBILITY_PROFILE_PREFERENCE_KEY = "AccessibilityProfile";
-	public static final String AGREED_TO_PRIVACY_POLICY_PREFERENCE_KEY = "AgreedToPrivacyPolicy";
+	public static final String AGREED_TO_CURRENT_PRIVACY_POLICY_PREFERENCE_KEY =
```
84n4n4: why not just use "int"?
that would spare us all the conversions everytime we use hashCode.
then naming would be AGREED_TO_PRIVACY_POLICY_WITH_THIS_HASH or AGREED_TO_PRIVACY_POLICY_VERSION

wslany:
Thanks!

PR #3403, opened 2019-11-16
CATROID-418 Handle user deleting sound file used by PlaySoundAndWaitBrick

thmq:
fixed issue where deleting sound used by PlaySoundAndWaitBrick caused an IllegalArgumentException.


wslany:
Thanks!

PR #3404, opened 2019-11-16
CATROID-420 Catch runtime exception in ContentResolver.

thmq:
- ContentResolver.query(..) seems to throw a runtime exception
   because of some security issue -> as seen in Play Console.
   This catches said exception.


wslany:
Thanks!

PR #3405, opened 2019-11-17
CATROID-427: Fix fabric logged crashes in logs on debug builds

ThomasSchwengler:
Fixed like proposed on [Stack Overflow](https://stackoverflow.com/questions/16932248/crashlytics-could-not-be-initialized-api-key-missing-from-androidmanifest-xml).


wslany:
Thanks!

PR #3406, opened 2019-11-19
CATROID-344 fix no reaction to bounce off clones of sprite

SchmidtChristian:
- add reference to sprite I am cloned from
- set reference during cloningprocess
- check when collision if collision with clone and then trigger event with original


wslany:
Thanks!

PR #3407, opened 2019-11-19
MAINTENANCE: Update Copyright regex to allow dates other than 2018

ThomasSchwengler:


ThomasSchwengler:
I reversed this. instead of updating all copyright notices to 2019, any date after 2010 is now allowed. This should make more sense than updating all our copyrights year after year.

I think the proper thing to do would be to only update the date if you make changes to a file.

wslany:
Thanks!

PR #3408, opened 2019-11-19
CATROID-173 Correct Position of Say and Think Bubble

mpirrer:
* Correct position of say/think bubble
* Add function to check if a pixel is transparent
* Add tests

Co-Authored-By: geschob <geschob@users.noreply.github.com>

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
+++ b/catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
@@ -27,6 +27,8 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Matrix;
 import android.graphics.PointF;
+import android.support.annotation.VisibleForTesting;
+import android.support.v4.util.Pair;
 import android.util.Log;
 
 import com.badlogic.gdx.math.Polygon;
@@ -45,14 +47,26 @@ import ar.com.hjg.pngj.PngjInputException;
 
 public class CollisionInformation {
 	private static final String TAG = CollisionInformation.class.getSimpleName();
+	private static final String TAG_COLLISION_POLYGON = "CollisionPolygon";
 
 	public Polygon[] collisionPolygons;
 	public Thread collisionPolygonCalculationThread;
 	private boolean isCalculationThreadCancelled = true;
 	private LookData lookData;
 
+	private Pair<Integer, Integer> leftBubblePos;
+	private Pair<Integer, Integer> rightBubblePos;
+
 	public CollisionInformation(LookData lookData) {
 		this.lookData = lookData;
+	}
+
+	public Pair<Integer, Integer> getLeftBubblePos() {
+		return leftBubblePos;
+	}
+
+	public Pair<Integer, Integer> getRightBubblePos() {
+		return rightBubblePos;
 	}
 
 	public void calculate() {
@@ -60,15 +74,6 @@ public class CollisionInformation {
 		CollisionPolygonCreationTask task = new CollisionPolygonCreationTask(lookData);
 		collisionPolygonCalculationThread = new Thread(task);
 		collisionPolygonCalculationThread.start();
-	}
-
-	public boolean isCalculationCancelled() {
-		return isCalculationThreadCancelled;
-	}
-
-	public void cancelCalculation() {
-		isCalculationThreadCancelled = true;
-		Log.i(TAG, "Collision Polygon Calculation of " + lookData.getName() + " cancelled!");
 	}
 
 	public int getNumberOfVertices() {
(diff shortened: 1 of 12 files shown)


catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static junit.framework.Assert.assertEquals;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ShowBubbleActor.class, Image.class, Pixmap.class, Bitmap.class, Texture.class})
+public class ShowBubbleActorTest {
+	private Image image;
+	private static final String NAME = "Test";
```
SchmidtChristian: maybe something like ACTORNAME for this would be better

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,180 @@
...
+	@Test
+	public void testLeftBubblePosition() {
+		Float expectedXLeft = -495.0f;
+		Float expectedYLeft = 347.0f;
+
+		Sprite sprite = new Sprite(NAME);
+		sprite.look.setX(-540.0f);
+		sprite.look.setY(-897.0f);
+		sprite.look.setHeight(1794.0f);
+		sprite.look.setWidth(1080.0f);
```
SchmidtChristian: there are multiple uses of these values in different tests would be nice if you could move them to constants with a descriptive name. especially height and width because they are always the same

catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/CollisionInformationTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static junit.framework.Assert.assertEquals;
+
+import static org.catrobat.catroid.common.Constants.IMAGE_DIRECTORY_NAME;
+
+@RunWith(AndroidJUnit4.class)
+public class CollisionInformationTest {
+	private LookData lookData;
+	private File imageFolder;
+	private static final int EXPECTED_Y = 285;
```
ThomasSchwengler: Please move the value into the tests (inconsistently right now)

catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/CollisionInformationTest.java
```diff
@@ -0,0 +1,97 @@
+/*
```
ThomasSchwengler: We already have a test with exactly this name in a different folder - I would prefer separate names

catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/CollisionInformationTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.File;
+
+import static junit.framework.Assert.assertEquals;
+
+import static org.catrobat.catroid.common.Constants.IMAGE_DIRECTORY_NAME;
+
+@RunWith(AndroidJUnit4.class)
+public class CollisionInformationTest {
+	private LookData lookData;
```
ThomasSchwengler: The naming of this test suggests you should be testing the `CollisionInformation` class here, not the `LookData` class.

Is it possible to mock the `LookData` class and instantiate the `CollisionInformation` class with it?

catroid/src/main/java/org/catrobat/catroid/stage/ShowBubbleActor.java
```diff
@@ -74,18 +77,52 @@ private void init() {
 	}
 
 	private Image getImageForDraw() {
-		if (drawRight) {
-			image.setX(sprite.look.getXInUserInterfaceDimensionUnit() + (sprite.look
-					.getWidthInUserInterfaceDimensionUnit() / 2));
+
+		if (sprite.look.getLookData() != null) {
```
ThomasSchwengler: I would prefer the use of local variables to resolve calls to `something.somethingelse.somefunction()`

catroid/src/main/java/org/catrobat/catroid/utils/ImageEditing.java
```diff
@@ -245,4 +246,8 @@ public static synchronized void writeMetaDataStringToPNG(String absolutePath, St
 			Log.e(TAG, "writeMetaDataStringToPNG: Failed to rename new file");
 		}
 	}
+
+	public static boolean isPixelTransparent(Bitmap source, int[] pixels, int x, int y) {
```
ThomasSchwengler: I would prefer to pass `width` as argument instead of passing the whole bitmap.

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
```diff
@@ -79,73 +83,102 @@ public int getNumberOfVertices() {
...
+						rightBubblePos = new Pair<>(xDiff, yDiff);
+					}
+
+					leftBubblePos = new Pair<>(xDiff, yDiff);
+				}
+			}
+		}
+	}
+
 	public void loadOrCreateCollisionPolygon() {
```
ThomasSchwengler: I would prefer this function to be split up into smaller functions

ThomasSchwengler: Naming also implies this - There should never be an OR or an AND in a function name.

mpirrer: I didn't write this function, just modified it a little. But I can try to split it up.

mpirrer: Changing this function would result in modifyng four files and four tests.
Should I do it nonetheless?

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+				Mockito.anyInt()).thenReturn(pixmap);
+
+		PowerMockito.whenNew(Texture.class).withParameterTypes(Pixmap.class)
+				.withArguments(Mockito.any()).thenReturn(texture);
+	}
+
+	@Test
+	public void testLeftBubblePosition() {
+		Float expectedXLeft = EXPECTED_X_STD_LEFT;
+		Float expectedYLeft = EXPECTED_Y_STD;
```
DinosaurierRex: This is unnecessary since you just need those variables to compare with your result. Just use your defined constants here.

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+		ShowBubbleActor bubbleActor = new ShowBubbleActor(OBJECT_NAME, sprite, Constants.SAY_BRICK);
+
+		assertEquals(expectedXLeft, bubbleActor.calculateLeftImageX(CALC_X_LEFT));
+		assertEquals(expectedYLeft, bubbleActor.calculateImageY(CALC_Y_LEFT));
+	}
+
+	@Test
+	public void testRightBubblePosition() {
+		Float expectedXRight = EXPECTED_X_STD_RIGHT;
+		Float expectedYRight = EXPECTED_Y_STD;
```
DinosaurierRex: Same as above.

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+		ShowBubbleActor bubbleActor = new ShowBubbleActor(OBJECT_NAME, sprite, Constants.SAY_BRICK);
+
+		assertEquals(expectedXRight, bubbleActor.calculateRightImageX(CALC_X_RIGHT));
+		assertEquals(expectedYRight, bubbleActor.calculateImageY(CALC_Y_RIGHT));
+	}
+
+	@Test
+	public void testLeftBubblePositionAfterMoveX() {
+		Float expectedXLeft = 5.0f;
```
DinosaurierRex: Is there something special about this value? Else could you also declare this variable as a constant.

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+		Float expectedYLeft = EXPECTED_Y_STD;
```
DinosaurierRex: Same as above

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+		ShowBubbleActor bubbleActor = new ShowBubbleActor(OBJECT_NAME, sprite, Constants.SAY_BRICK);
+
+		assertEquals(expectedXLeft, bubbleActor.calculateLeftImageX(CALC_X_LEFT));
+		assertEquals(expectedYLeft, bubbleActor.calculateImageY(CALC_Y_LEFT));
+	}
+
+	@Test
+	public void testRightBubblePositionAfterMoveX() {
+		Float expectedXRight = 847.0f;
+		Float expectedYRight = EXPECTED_Y_STD;
```
DinosaurierRex: Same as above.

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+		ShowBubbleActor bubbleActor = new ShowBubbleActor(OBJECT_NAME, sprite, Constants.SAY_BRICK);
+
+		assertEquals(expectedXRight, bubbleActor.calculateRightImageX(CALC_X_RIGHT));
+		assertEquals(expectedYRight, bubbleActor.calculateImageY(CALC_Y_RIGHT));
+	}
+
+	@Test
+	public void testLeftBubblePositionAfterMoveY() {
+		Float expectedXLeft = EXPECTED_X_STD_LEFT;
+		Float expectedYLeft = EXPECTED_Y_MOVE_Y;
```
DinosaurierRex: Same as above

catroid/src/test/java/org/catrobat/catroid/test/stage/ShowBubbleActorTest.java
```diff
@@ -0,0 +1,170 @@
...
+		ShowBubbleActor bubbleActor = new ShowBubbleActor(OBJECT_NAME, sprite, Constants.SAY_BRICK);
+
+		assertEquals(expectedXLeft, bubbleActor.calculateLeftImageX(CALC_X_LEFT));
+		assertEquals(expectedYLeft, bubbleActor.calculateImageY(CALC_Y_LEFT));
+	}
+
+	@Test
+	public void testRightBubblePositionAfterMoveY() {
+		Float expectedXRight = EXPECTED_X_STD_RIGHT;
+		Float expectedYRight = EXPECTED_Y_MOVE_Y;
```
DinosaurierRex: Same as above.

wslany:
Thanks!!

PR #3414, opened 2019-11-27
CATROID-363 no Internet error messages and refactoring

SchmidtChristian:
- added Utils function to also print error message
- call this method instead of the old one
- add function call to catrobat community and media library


wslany:
Thanks!

PR #3420, opened 2019-12-05
JENKINS-331: Dynamic URL option when building an apk

jtrattner:


catroid/build.gradle
```diff
@@ -140,6 +140,11 @@ android {
         versionName "0.9.66"
         println "VersionCode is $versionCode"
         println "VersionName is $versionName"
+        buildConfigField "String", "GIT_COMMIT_INFO", "\"${getGitCommitInfo()}\""
```
jampo: what do you need the GIT_COMMIT_INFO for?

Jenkinsfile
```diff
@@ -39,8 +39,8 @@ def postEmulator(String coverageNameAndLogcatPrefix) {
     archiveArtifacts "${coverageNameAndLogcatPrefix}_logcat.txt"
 }
 
-def useWebTestParameter() {
-    return env.USE_WEB_TEST?.toBoolean() ? '-PuseWebTest' : ''
+def webTestUrlParameter() {
+    return env.WEB_TEST_URL?.isEmpty() ? '' : "-Pweb_test_url='${params.WEB_TEST_URL}'"
```
jampo: for -P I think camel case is normally used for naming and not underline so it would be -PwebTestUrl like before the -PuseWebTest

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -98,9 +98,10 @@
 	public static final String TEXT_TO_SPEECH_TMP_PATH = TMP_PATH + "/textToSpeech";
 
 	// Web:
-	private static final String MAIN_URL_PRODUCTION = "https://share.catrob.at";
-	private static final String MAIN_URL_WEBTEST = "https://web-test.catrob.at";
-	public static final String MAIN_URL_HTTPS = BuildConfig.USE_WEB_TEST ? MAIN_URL_WEBTEST : MAIN_URL_PRODUCTION;
+	//private static final String MAIN_URL_PRODUCTION = "https://share.catrob.at";
```
jampo: delete code if it isn't needed any more only leave comments if really necessary / helping to explain something

catroid/build.gradle
```diff
@@ -140,6 +140,11 @@ android {
...
+        buildConfigField "String", "MAIN_URL", (project.hasProperty('web_test_url') ?
```
jampo: would then also need to change "web_test_url" to "webTestUrl" here

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -98,9 +98,10 @@
...
+	//private static final String MAIN_URL_WEBTEST = "https://web-test.catrob.at";
+	//public static final String MAIN_URL_HTTPS = BuildConfig.USE_WEB_TEST ? MAIN_URL_WEBTEST : MAIN_URL_PRODUCTION;
+	public static final String MAIN_URL_HTTPS = BuildConfig.MAIN_URL;
```
jampo: you moved the logic from here to the buildconfig.MAIN_URL string any particular reason for that?
I think its better readable here

catroid/build.gradle
```diff
@@ -140,6 +140,11 @@ android {
...
+                "\"" + project.getProperties().get('web_test_url') + "\"" :
+                '"https://share.catrob.at"')
```
jampo: The "https://share.catrob.at" is a magic String here -> would be better to have it in own constant with name as before "MAIN_URL_PRODUCTION"

jtrattner:
test this please

catroid/build.gradle
jtrattner: seems to be there before, weird diff

jtrattner:
This PR will be closed and a new one will be opened, due to a Jenkins error (wrong PR is tested).

PR #3422, opened 2019-12-10
CATROID-393 fix rename Scene Bug

gPathpp:


catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/SceneListFragment.java
```diff
@@ -259,6 +260,20 @@ public void onItemClick(Scene item) {
 	public void onLoadFinished(boolean success) {
 		if (!success) {
 			ToastUtil.showError(getActivity(), R.string.error_load_project);
+			return;
 		}
+		refreshFragment();
```
thmq: what you should do here instead is to update the adapter's item list.

```
adapter.setItems(ProjectManager.getInstance().getCurrentProject().getSceneList());
```
or 
```
initializeAdapter();
```
although the initializeAdapter does more than that (would still work though).

why? because when you rename the scene, the whole project is reloaded but the adapter's list not updated accordingly (i.e. the `ProjectManager` get's a new project but the `SceneAdapter` still holds the old scene list). this is really straightforward and there is absolutely no need to detach and reattach the fragment.

gPathpp: initializeAdapter does not work right (it leads to an Activity crash), but the other line works fine. Thanks

wslany:
Thanks!!

PR #3423, opened 2019-12-26
CATROID-415 - ProjectActivity crash onBack when current project is null

84n4n4:


84n4n4:
we got rid of a lot of cascading crashes with the last release, this one here still seems to be the last of those.
if the current project == null, the project activity should be finished (which is what the users seem to try here anyways) since nothing can be performed there anyways in this state.

84n4n4:
btw. this is currently the highest ranking crash with over 100 occurences. (second highest with ~40 in CATROID-439)

wslany:
Just saying...: https://kotlinlang.org/docs/reference/null-safety.html

wslany:
Thanks!!

PR #3424, opened 2019-12-26
CATROID-439 - StageActivity onDestroy crashes when no current project

84n4n4:
currently second highest crash in play console, again cascading crash after an actual crash for some other reason.


wslany:
Thanks!

PR #3427, opened 2019-12-27
CATROBAT-TESTS - refactor catrobatTest runner

84n4n4:
refactor the internal work / data flow when running catrobat tests.
results now delivered back to runner / user via activity result, instead of using uservariables.

this made it possible to:

* remove all espresso dependencies from catrobat runner.
* test catrobat tests on the cell itself for debugging (test results are displayed via toast and copied to clipboard) (currently only when starting from projectActivity)

additionally:

* this now requires a finish stage brick at the end of the test
* it would be possible to have more than one assertion brick in the test (currently deactivated in ui)
* this opens up the testing framework for further assertion bricks with special behaviour to be added

and finally:
* this reduces the size and complexity of the test runner aswell as the assertion brick, aswell as limiting the connections to project internals (variables, etc) to an absolute minimum, providing a better reference implementation for the iOS guys.


84n4n4:
all unused variables removed from catrobat tests.

wslany:
Thanks!! (note: failing tests on jenkins are unrelated to this PR)

PR #3435, opened 2020-01-16
CATROID-430 Refactor import of projects activity

DinosaurierRex:
Moved Import logic from ProjectListFragment to new ProjectImportController
Calling Activity gets now notified when Controller finished


catroid/src/main/java/org/catrobat/catroid/io/ProjectImportController.java
```diff
@@ -0,0 +1,105 @@
...
+		if (!uriToImport.getScheme().equals("file")) {
+			throw new IllegalArgumentException("importProject has to be called with a file uri. (not a content uri");
+		}
+
+		File cacheFile = new File(CACHE_DIR, CACHED_PROJECT_ZIP_FILE_NAME);
+		if (cacheFile.exists()) {
+			cacheFile.delete();
+		}
+
+		File src = new File(uriToImport.getPath());
```
PlantinumX: pls move src before creating the cache file

DinosaurierRex: While reviewing we triggered a situation which seems like an endless Loop in the ImportTask. PR now set to WIP

PlantinumX:
Status on this PR,please?

DinosaurierRex:
> Status on this PR,please? -> I see the androidX integration. Should we do a seperate PR for the androidX migration and on top of that this PR

Yes that would probably be a good idea :+1:

DinosaurierRex:
Android X Migration was done already in [PR-3471](https://github.com/Catrobat/Catroid/pull/3471)

PR #3437, opened 2020-01-17
JENKINS-366 quickfix disable fastlane installation to fix Dockerfile.…

jampo:
Unfortunately the Dockerfile is not working as we have incompatible versions of gem and fastlane at the moment - this is a quick fix to fix this.

However we need to fix the situation to  be able to have fastlane for releases again. See Jenkins Ticket https://jira.catrob.at/browse/JENKINS-330 for that (is already in progress)

### DIFF
--- a/docker/Dockerfile.jenkins
+++ b/docker/Dockerfile.jenkins
@@ -54,19 +54,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 # Install Fastlane
-RUN apt-get update && apt-get install -y --no-install-recommends \
-        rubygems \
-        ruby-dev \
-        g++ \
-        make \
-        less \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.jenkins
```diff
@@ -54,19 +54,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+#        g++ \
+#        make \
+#        less \
+#    && rm -rf /var/lib/apt/lists/* \
+#    && gem update --system \
+#    && gem install fastlane -NV \
+#    && apt-get purge -y --auto-remove \
+#        ruby-dev \
+#        g++ \
+#        make
```
ThomasSchwengler: Please remove these lines instead of commenting them out. They can be restored from git if needed.

wslany:
Thanks!

PR #3440, opened 2020-01-22
CATROID-394 Fix Google OAuth2 Login

PlantinumX:


catroid/src/main/java/org/catrobat/catroid/transfers/GoogleLoginHandler.java
```diff
@@ -47,56 +41,44 @@
...
 		GoogleExchangeCodeTask.OnGoogleExchangeCodeCompleteListener {
 
 	private AppCompatActivity activity;
-	public static final int GPLUS_REQUEST_CODE_SIGN_IN = 0;
 	public static final int REQUEST_CODE_GOOGLE_PLUS_SIGNIN = 100;
-	public static final int STATUS_CODE_SIGN_IN_CURRENTLY_IN_PROGRESS = 12502;
-	private GoogleApiClient googleApiClient;
-	private static final String GOOGLE_PLUS_CATROWEB_SERVER_CLIENT_ID = "427226922034-r016ige5kb30q9vflqbt1h0i3arng8u1.apps.googleusercontent.com";
+	private GoogleSignInClient googleSignInClient;
+	private static final String GOOGLE_LOGIN_CATROWEB_SERVER_CLIENT_ID = "427226922034-r016ige5kb30q9vflqbt1h0i3arng8u1.apps.googleusercontent.com";
```
DinosaurierRex: Pleas put this into Constants class.

catroid/src/main/java/org/catrobat/catroid/transfers/GoogleLoginHandler.java
```diff
@@ -47,56 +41,44 @@
...
 
-	public GooglePlusLoginHandler(AppCompatActivity activity) {
+	@SuppressWarnings("RestrictedApi")
```
DinosaurierRex: Is there a reason for a SuppressWarnings

catroid/src/main/java/org/catrobat/catroid/transfers/GoogleLoginHandler.java
```diff
@@ -47,56 +41,44 @@
...
+import static com.google.android.gms.auth.api.signin.GoogleSignIn.getSignedInAccountFromIntent;
+
+public class GoogleLoginHandler implements CheckOAuthTokenTask.OnCheckOAuthTokenCompleteListener,
 		GoogleLogInTask.OnGoogleServerLogInCompleteListener,
 		CheckEmailAvailableTask.OnCheckEmailAvailableCompleteListener,
 		GoogleExchangeCodeTask.OnGoogleExchangeCodeCompleteListener {
 
 	private AppCompatActivity activity;
-	public static final int GPLUS_REQUEST_CODE_SIGN_IN = 0;
 	public static final int REQUEST_CODE_GOOGLE_PLUS_SIGNIN = 100;
```
DinosaurierRex: When renaming from GOOGLE PLUS to only GOOGLE, please refactor also this variables.

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -408,9 +410,9 @@ public static void invalidateLoginTokenIfUserRestricted(Context context) {
 	public static void logoutUser(Context context) {
 		SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
 		String userName = sharedPreferences.getString(Constants.USERNAME, Constants.NO_USERNAME);
+		new GoogleLoginHandler((AppCompatActivity) context).getGoogleSignInClient().signOut();
```
DinosaurierRex: May split this up and make function calls on object

catroid/src/main/java/org/catrobat/catroid/transfers/GoogleLoginHandler.java
PlantinumX: As seen in this link https://stackoverflow.com/questions/49464246/when-using-firebase-googlesignin-is-marked-as-internal-and-should-not-be-access it is a known issue in version 12.0.0

wslany:
Thanks!!

PR #3441, opened 2020-01-22
CATROID-445 Adapt Help URLs for bricks to xwiki and test

DinosaurierRex:
We have switched from our old, now obsolete mediawiki (wiki.catrob.at) to our new xwiki platform (wiki.catrobat.org).
The "Help" entries in the context menus of bricks point to these wiki pages, using a combination of url paths and anchors.
The url scheme has changed and thus had to be adapted.
The user's language is now automatically set through the user's browser setting, and can be changed only there.

FYI: The high number of files changed is caused by the refactoring of moving the physics bricks to the same package as the other bricks. Main Changes to be reviewed in BricksHelpUrlTest.


wslany:
Thanks!!

PR #3442, opened 2020-01-22
JENKINS-366 quickfix disable fastlane installation to fix Dockerfile.…

AndreasFinsterbusch:
Unfortunately the Dockerfile is not working as we have incompatible versions of gem and fastlane at the moment - this is a quick fix to fix this.

However we need to fix the situation to be able to have fastlane for releases again. See Jenkins Ticket https://jira.catrob.at/browse/JENKINS-330 for that (is already in progress)


wslany:
Thanks!

PR #3444, opened 2020-01-23
CATROID-291: Refactoring

ThomasSchwengler:
This refactoring is necessary for CATROID-291

* Create a `EnumUtils` helper class to avoid code duplications among enums.
* Simplify readability of `NumberFormatsTest` by fixating the expected output
* `FormulaElement`
  * Move access to static variables from within functions into parameters
  * Move common functions into a `common` package to reduce code duplication
    and enhance readability
  * Add a `function` package and move all function interpretation into classes following the open-closed principle
  * Split up many functions to improve readability
  * Simplify many functions
* `Formula`
  * Simplify many functions by reducing the number of concerns per function
* `CollisionDetection`
  * Improve memory consumption by reducing the number of memory allocations in loops
* Hide direct references to android context in wrappers
* Rename functions to be more precise
* Replace some class instances with mocks in tests where possible

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/TouchesEdgeTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/test/sensing/TouchesEdgeTest.java
@@ -23,6 +23,8 @@
 
 package org.catrobat.catroid.test.sensing;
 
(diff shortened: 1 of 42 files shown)


PlantinumX:
talked with Thomas it is ok to close PR until it is ready again :)

PR #3452, opened 2020-01-29

adnaribo:


PlantinumX:
please fix failing testcases :)

PR #3455, opened 2020-02-05
Empty formulas are now correctly displayed, if the ShowVariable brick is used

trummerp:
With this update the ShowVariable brick are now correctly displaying empty variables/formulas as such instead of the predefined text "No value set". 

Examples: 
The formula "5" still results in "5". 
The formula "" now results in "", instead of "No value set".


wslany:
Thanks!!

PR #3456, opened 2020-02-05
CATROID-182 Add visual placement to the GlideToBrick

mstoeg:
In my opinion the variable name of the string resource "Place Visually" should be renamed from brick_place_at_option_place_visually to brick_option_place_visually because it would be misleading otherwise. As this string is now both used for PlaceAtBrick and GlideToBrick. What can be done there without changing all the string.xml files?

New Commit Message:
CATROID-182 Add visual placement to the GlideToBrick

- Derive a new class VisualPlacementBrick from FormulaBrick and move the logic for visual placement from the PlaceAtBrick class to this new base class to prevent code duplication.

- Add an entry "Place Visually" also in the ScriptFragment of the GlideToBrick.

- Refactor the testcase PlaceAtBrickTest and make a new Test for VisualPlacementBricks to test that visual placement both works for GlideToBricks and PlaceAtBricks.


catroid/src/main/java/org/catrobat/catroid/content/bricks/FormulaBrick.java
```diff
@@ -149,6 +170,81 @@ public void showFormulaEditorToEditFormula(View view) {
 		}
 	}
 
+	public void showFormulaEditorVisualPlacementDecision(final View view) {
+		Context context = view.getContext();
+		Fragment currentFragment = UiUtils.getActivityFromView(view).getSupportFragmentManager()
+				.findFragmentById(R.id.fragment_container);
+
+		boolean showVisualPlacementDialog;
+		if (this instanceof PlaceAtBrick) {
```
gPathpp: Please make a new interface or a class (VisualPlacementBrick or smth) to minimize code duplication.

gPathpp:
Please add a Test.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementBrickTest.java
```diff
@@ -51,62 +56,129 @@
...
+	public void testIsVisualPlacementShown() {
+		onBrickAtPosition(1).checkShowsText(brickString);
+		onView(withId(editTextXId))
 				.perform(click());
 		onView(withText(R.string.brick_place_at_option_place_visually))
 				.check(matches(isDisplayed()));
 		onView(withText(R.string.brick_context_dialog_formula_edit_brick))
-				.check(matches(isDisplayed()))
+				.check(matches(isDisplayed()));
+		pressBack();
```
gPathpp: You could separate this test too.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementBrickTest.java
```diff
@@ -51,62 +56,129 @@
...
 		onView(withText(R.string.brick_place_at_option_place_visually))
 				.check(matches(isDisplayed()));
 		onView(withText(R.string.brick_context_dialog_formula_edit_brick))
 				.check(matches(isDisplayed()));
 		pressBack();
-
-		onBrickAtPosition(brickPosition).performClick();
-		onView(withText(R.string.brick_place_at_option_place_visually))
-				.check(matches(isDisplayed()));
 	}
```
gPathpp: Please add a test case that verifies that the visual placement activity is shown after the click (or a similar test ;)). Have a look at Mockito :D.

wslany:
Thanks!!

PR #3461, opened 2020-02-07

adnaribo:


catroid/src/main/java/org/catrobat/catroid/content/actions/RunningStitchAction.java
```diff
@@ -0,0 +1,58 @@
...
+public class RunningStitchAction extends TemporalAction {
+
+	private Sprite sprite;
+	private Formula length;
+
+	@Override
+	protected void update(float delta) {
+		int lengthInterpretation;
+		try {
+			lengthInterpretation = length == null ? 0 : length.interpretInteger(sprite);
```
DinosaurierRex: Please apply same changes as requested in first review here.

wslany:
Thanks!! :-)

PR #3474, opened 2020-02-11
CATROID-447 Added catrobatTests for BroadcastsAndWait

Koell:
they include tests for Broadcast and ReceiveMessage


Koell:
updated tests & squashed them into a single commit

wslany:
Thanks!!

PR #3476, opened 2020-02-13
JENKINS-365 add jenkinsfile to build and push docker image

jampo:
JENKINS-366 actually push images

JENKINS-365 use image for build

JENKINS-365 set default label to stable and add parameter to overwrite for debug reasons

JENKINS-365 always pull docker images

JENKINS-365 add build trigger for midnight develop


wslany:
Thanks!

PR #3478, opened 2020-02-15
CATROID-464 Refactor string resource 'Place Visually'

mstoeg:
Refactor the variable name of the string resource as it is now possible that the GlideToBrick can be placed visually as well. In order to prevent misleading.

Change exception handling in VisualPlacementBrickTest class, as there were sometimes problems with removing the project.


wslany:
Thanks!

PR #3481, opened 2020-02-17
CATROID-396 Start triple stitch with length 10

adnaribo:


catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -38,7 +38,7 @@
 
 public final class Constants {
 
-	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = 0.99996f;
+	public static final float CURRENT_CATROBAT_LANGUAGE_VERSION = 0.99997f;
```
DinosaurierRex: @wslany since we updated version to ....96 for the last three stitch bricks, is this correct to increase by one here again?

wslany: Yes, increasing it here again is fine. It is better to increase it without harm done than to forget. By increasing each time the xml changes, we don't have to think and check anymore whether it happened already, i.e., we follow the isolation principle. Could also be important if some things need to be retracted later.

wslany:
Thanks!

PR #3482, opened 2020-02-19
CATROID-374 add WebViewActivityLoginCookieTest

gPathpp:
Added a JUnit4 test and made a small refactoring.


gPathpp:
> PLEASE FIX:
> 1 Login does not work (at the same time, it works fine with other apks). It fails silently. Tried multiple times.
> 2 The community site cannot be accessed (at the same time, it works fine with other apks). The webview only shows a blue screen. Tried multiple times.
> 
> QUESTION: @DinosaurierRex @gPathpp Have you tried it once? Did the above work for you?

Just tested it with my Android-10 device, works fine. Which android version do you use? Does the other PR to CATROID-374 work with your device? (We've tried both versions with my phone)

wslany:
Thanks!

PR #3487, opened 2020-02-23
CATROID- 472: Completed StringFinderTest

pallav12:
Changes: Added more unit test, and made StringFinder null safe.


pallav12:
@wslany please review what's the issue, it says cannot find the symbol R at 71 places in android test, but the app installs successfully. It happens even without my commit.

DinosaurierRex:
This is the wrong ticket number and also the purpose of this PR is not quite clear. When looking at the function there would be a Nullpointer Exception to handle (from `pattern.matcher(null) --> will raise an NPE`). Please dont throw unhandled Exceptions when theres no actual need. Please reopen under correct title and/or description. Thanks in advance :pray:

PR #3491, opened 2020-02-25
CATROID-469 Stitching brick behaves incorrectly over 60 steps

adnaribo:


catroid/src/test/java/org/catrobat/catroid/stage/EmbroideryActorCircleTest.java
```diff
@@ -0,0 +1,108 @@
...
+		ShapeRenderer renderer = mock(ShapeRenderer.class);
+		embroideryActorSpy = spy(new EmbroideryActor(1.0f, embroideryPatternManager, renderer));
+	}
+
+	@Test
+	public void testDraw() {
+		embroideryActorSpy.draw(mock(Batch.class), 1.0F);
+
+		VerificationMode mode = firstPointDrawn ? times(1) : never();
+		verify(embroideryActorSpy, mode).drawCircle(stitchPoints.get(0));
```
gPathpp: Please verify (maybe in an extra test method) that the second stitch-point is always drawn.

wslany:
Thanks!

PR #3492, opened 2020-02-25
CATROID-468 Make visual placement option available in formula editor

mstoeg:
- Adapt FormulaEditorDiscardChangesTest so that it works correctly again with the new functionalty of the formula editor
- Add further testcases in VisualPlacementBrickTest
- Separate a GlideToBrickTest from VisualPlacementBrickTest because this one cannot be parameterized


catroid/src/main/java/org/catrobat/catroid/content/bricks/VisualPlacementBrick.java
```diff
@@ -107,11 +114,8 @@ public boolean areAllBrickFieldsNumbers() {
 				&& isBrickFieldANumber(getYBrickField());
 	}
 
-	public boolean showVisualPlacementDialog(View view) {
-		Fragment currentFragment = UiUtils.getActivityFromView(view).getSupportFragmentManager()
-				.findFragmentById(R.id.fragment_container);
-
-		return currentFragment instanceof ScriptFragment
+	public boolean showVisualPlacementDialog(View view, Fragment currentFragment) {
```
gPathpp: Please rename this function.

wslany:
Thanks!

PR #3496, opened 2020-02-26
CATROID-446  Fix Google sign-in button

malisale:
Fixed Google Sign-in Button


PlantinumX:
there is a second pr for that I will evaluate which PR fullfills the requirement more

PR #3497, opened 2020-02-27
CATROID-466 Fixed Rate Us Feature

rheo-chiti:
CATROID-466 Fixed Rate Us Feature
Fix bug CATROID-466 

**BEFORE**

**AFTER**


rheo-chiti:
I tried logging the package name and I got ```org.catrobat.catroid.arduino``` instead of ```org.catrobat.catroid```. This issue is likely because of bad merging of these two packages. I tried changing the package name from ```org.catrobat.catroid.arduino```  to ```org.catrobat.catroid``` at all files of AndroidManifest.xml but didn't got any success.

PlantinumX:
this ticket is solved when arduino is released

PR #3500, opened 2020-02-28
CATROID-169 Sound continues on scene continue transition

maxf24:
- added test for correct behavior of sound on scene transition

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
@@ -1,0 +1,167 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2020 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.uiespresso.content.brick.stage;
+
+import android.media.MediaPlayer;
+
+import org.catrobat.catroid.ProjectManager;
+import org.catrobat.catroid.common.SoundInfo;
+import org.catrobat.catroid.content.Project;
+import org.catrobat.catroid.content.Scene;
+import org.catrobat.catroid.content.Script;
+import org.catrobat.catroid.content.Sprite;
+import org.catrobat.catroid.content.StartScript;
+import org.catrobat.catroid.content.bricks.PlaySoundBrick;
+import org.catrobat.catroid.content.bricks.SceneTransitionBrick;
+import org.catrobat.catroid.content.bricks.WaitBrick;
+import org.catrobat.catroid.io.ResourceImporter;
+import org.catrobat.catroid.io.SoundManager;
+import org.catrobat.catroid.io.XstreamSerializer;
+import org.catrobat.catroid.stage.StageActivity;
+import org.catrobat.catroid.test.utils.TestUtils;
+import org.catrobat.catroid.uiespresso.stage.utils.ScriptEvaluationGateBrick;
+import org.catrobat.catroid.uiespresso.util.rules.BaseActivityTestRule;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.File;
+import java.io.IOException;
+
+import androidx.test.core.app.ApplicationProvider;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.platform.app.InstrumentationRegistry;
+
+import static junit.framework.TestCase.assertFalse;
+import static junit.framework.TestCase.assertTrue;
+
+import static org.catrobat.catroid.common.Constants.SOUND_DIRECTORY_NAME;
+
+@RunWith(AndroidJUnit4.class)
+public class SceneTransitionWithSoundBrickStageTest {
+
+	private ScriptEvaluationGateBrick lastBrickFirstScript;
+	private ScriptEvaluationGateBrick lastBrickSecondScript;
+
+	private String firstSceneName;
+	private Script secondScript;
+
+	private static ProjectManager projectManager;
+	private File soundFile;
+
+	@Rule
+	public BaseActivityTestRule<StageActivity> baseActivityTestRule = new
+			BaseActivityTestRule<>(StageActivity.class, false,
+			false);
+
+	@BeforeClass
+	public static void setUpProjectManager() {
+		projectManager = ProjectManager.getInstance();
+	}
+
+	@Before
+	public void setUp() throws Exception {
+		createProject(getClass().getSimpleName());
+		baseActivityTestRule.launchActivity(null);
+	}
+
+	@After
+	public void tearDown() throws IOException {
+		TestUtils.deleteProjects(getClass().getSimpleName());
+		if (soundFile != null && soundFile.exists()) {
+			soundFile.delete();
+		}
+	}
+
+	@Test
+	public void testStopSoundOnSceneTransition() {
+		runProject(lastBrickSecondScript);
+		assertFalse(getMediaplayer().isPlaying());
+	}
+
+	@Test
+	public void testContinueSoundAfterSceneTransition() {
+		secondScript.addBrick(new SceneTransitionBrick(firstSceneName));
+		runProject(lastBrickFirstScript);
+		assertTrue(getMediaplayer().isPlaying());
+	}
+
+	@Test
+	public void testContinueSoundDoesNotStartFromBeginning() {
+		secondScript.addBrick(new SceneTransitionBrick(firstSceneName));
+		runProject(lastBrickFirstScript);
+		MediaPlayer mediaPlayer = getMediaplayer();
+		assertTrue(mediaPlayer.isPlaying());
+		assertTrue(mediaPlayer.getCurrentPosition() > 50);
+	}
+
+	private void runProject(ScriptEvaluationGateBrick scriptBrick) {
+		scriptBrick.waitUntilEvaluated(3000);
+	}
+
+	private MediaPlayer getMediaplayer() {
+		return SoundManager.getInstance().getMediaPlayers().get(0);
+	}
+
+	private void createProject(String projectName) throws IOException {
+		Project project = new Project(ApplicationProvider.getApplicationContext(), projectName);
+		Script script = new StartScript();
+		Sprite sprite = project.getDefaultScene().getBackgroundSprite();
+		XstreamSerializer.getInstance().saveProject(project);
+
+		firstSceneName = project.getDefaultScene().getName();
+		Scene secondScene = new Scene("Scene 2", project);
+
+		sprite.addScript(script);
+		project.getDefaultScene().addSprite(sprite);
+		projectManager.setCurrentProject(project);
+		projectManager.setCurrentSprite(sprite);
+		project.addScene(secondScene);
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -148,6 +162,20 @@ public synchronized void stopAllSounds() {
 		}
 	}
 
+	public List<String> getSoundFiles() {
+		return soundFiles;
+	}
+
+	public List<Integer> getCurrentPositionOfSounds() {
```
gPathpp: Please rename this function.
Maybe it is useful to map sounds to positions.

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -148,6 +162,20 @@ public synchronized void stopAllSounds() {
...
+		List<Integer> positionList = new ArrayList<>();
+		for (MediaPlayer mediaPlayer : mediaPlayers) {
+			int position = mediaPlayer.isPlaying() ? mediaPlayer.getCurrentPosition()
```
gPathpp: You may not want to add sounds that are not playing.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -875,10 +876,24 @@ private StageBackup saveToBackup() {
...
+		saveSoundsToBackup(backup);
 		return backup;
 	}
 
+	private void saveSoundsToBackup(StageBackup backup) {
+		List<String> soundFiles = SoundManager.getInstance().getSoundFiles();
+		List<Integer> positionList =
+				SoundManager.getInstance().getCurrentPositionOfSounds();
+		backup.soundsDurationMap = new HashMap<>();
+		if (soundFiles.size() == positionList.size()) {
```
gPathpp: Please refactor this as discussed.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
```diff
@@ -0,0 +1,169 @@
...
+	@Test
+	public void testContinueSoundDoesNotStartFromBeginning() {
+		secondScript.addBrick(new SceneTransitionBrick(firstSceneName));
+		MediaPlayer mediaPlayer = startAndGetMediaplayer(lastBrickFirstScript);
+		assertTrue(mediaPlayer.isPlaying());
+		assertTrue(mediaPlayer.getCurrentPosition() > 0);
+	}
+
+	private MediaPlayer startAndGetMediaplayer(ScriptEvaluationGateBrick scriptBrick) {
+		onView(withId(R.id.button_play))
```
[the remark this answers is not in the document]
gPathpp: As discussed with Thomas, please refactor it to StageActivity.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
```diff
@@ -0,0 +1,169 @@
...
+		Scene secondScene = new Scene("Scene 2", project);
+
+		sprite.addScript(script);
+		project.getDefaultScene().addSprite(sprite);
+		ProjectManager.getInstance().setCurrentProject(project);
+		ProjectManager.getInstance().setCurrentSprite(sprite);
+		project.addScene(secondScene);
+
+		script.addBrick(new PlaySoundBrick());
+		script.addBrick(new WaitBrick(500));
```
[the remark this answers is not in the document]
maxf24: it is necessary to let a sound play for a short time so that it is possible to check that it did not start from the beginning when continuing a scene.

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -42,6 +44,7 @@
 
 	private final List<MediaPlayer> mediaPlayers = new ArrayList<MediaPlayer>(MAX_MEDIA_PLAYERS);
 	private float volume = 70.0f;
+	private List<String> soundFiles = new ArrayList<>(MAX_MEDIA_PLAYERS);
 
 	public SoundManager() {
```
gPathpp: This line seems not right. Could you please check why this is public? (I know you didn't write this line ^^)
(if it's only used in tests => visible for testing and hopefully, there is no occurrence in the code)

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -148,6 +164,27 @@ public synchronized void stopAllSounds() {
 		}
 	}
 
+	public Map<String, Integer> getPlayingSoundDurationMap() {
+		List<Integer> positionList =
+				SoundManager.getInstance().getCurrentPositionOfSounds();
+		Map<String, Integer> soundsDurationMap = new HashMap<>();
+		for (String sound : soundFiles) {
+			int position = positionList.get(soundFiles.indexOf(sound));
```
gPathpp: This seems a bit too complicated for this task. Instead of two separate lists of sound files and media Players. You could map one sound file to one MediaPlayer from the beginning (in play sound).
For further processing see https://mkyong.com/java8/java-8-filter-a-map-examples/
 map<String, MediaPlayer> => filter => map => collect = map<string, Integer>

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
```diff
@@ -51,11 +54,23 @@ public static SoundManager getInstance() {
 	}
 
 	public synchronized void playSoundFile(String pathToSoundfile) {
+		playSound(pathToSoundfile, 0);
+	}
+
+	public synchronized void playSoundFileWithStartTime(String pathToSoundfile,
```
gPathpp: The name is ok, but it is the only difference to playSound. Please remove it and/or rename Play Sound)

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+		secondScript = new StartScript();
+		Sprite secondSprite = new Sprite("Sprite2");
+		secondSprite.addScript(secondScript);
+		secondScene.addSprite(secondSprite);
+
+		lastBrickFirstScript = ScriptEvaluationGateBrick.appendToScript(script);
+		lastBrickSecondScript = ScriptEvaluationGateBrick.appendToScript(secondScript);
+	}
+}
```
gPathpp: Nice!

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -918,6 +919,9 @@ private void restoreFromBackup(StageBackup backup) {
 		if (CameraManager.getInstance() != null && backup.cameraRunning) {
 			CameraManager.getInstance().resumeForScene();
 		}
+		for (Map.Entry<String, Integer> entry : backup.soundsDurationMap.entrySet()) {
+			SoundManager.getInstance().playSoundFileWithStartTime(entry.getKey(), entry.getValue());
+		}
 		initStageInputListener();
 	}
```
gPathpp: Nice fix!

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
gPathpp: Or if it's in the tests => mock. Thank you!

catroid/src/main/java/org/catrobat/catroid/io/SoundManager.java
maxf24: Good Points, thanks!
Unfortunatly using stream is only possible starting with API 24 and as per my IDE is the current min 21.

gPathpp: Ok, then its probably fine as it is. You can decide if you want to refactor this or leave it like that.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
```diff
@@ -0,0 +1,167 @@
...
+	public void tearDown() throws IOException {
+		TestUtils.deleteProjects(getClass().getSimpleName());
+		if (soundFile != null && soundFile.exists()) {
+			soundFile.delete();
+		}
+	}
+
+	@Test
+	public void testStopSoundOnSceneTransition() {
+		runProject(lastBrickSecondScript);
```
wslany:
```suggestion
		runProject(lastBrickFirstScript);
```

PLEASE CHECK: Shouldn't this also start with the first scene here? Otherwise the sound never starts, right?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
```diff
@@ -0,0 +1,167 @@
...
+		assertTrue(getMediaplayer().isPlaying());
+	}
+
+	@Test
+	public void testContinueSoundDoesNotStartFromBeginning() {
+		secondScript.addBrick(new SceneTransitionBrick(firstSceneName));
+		runProject(lastBrickFirstScript);
+		MediaPlayer mediaPlayer = getMediaplayer();
+		assertTrue(mediaPlayer.isPlaying());
+		assertTrue(mediaPlayer.getCurrentPosition() > 0);
```
wslany: PLEASE EXPLAIN: Have you checked this test in test-first style, i.e., before you implemented the fix? And what is the unit for the current position here? Can't this asset be true even if the sound was restarted, because of timing issues? Please clarify and confirm.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
maxf24: In runProject the whole Project is run, both scripts get executed - the parameter is only used to wait until this Script is fully evaluated.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
maxf24: I did start my tests without my fix and all of them failed. 
The current position is the time played in milliseconds.
Yes, I think timing issues are a potential problem here, but I did not find a better way to check this.

wslany: Thanks! How about, after starting the sound in the first scene, to wait for 0.1 second while still in the first scene using a "wait 0.1 seconds" brick, and then when returning back from the second scene immediately checking that the current position is > 100? Would that work?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
wslany: Thank you for explaining!

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/SceneTransitionWithSoundBrickStageTest.java
maxf24: Oh I forgot to mention that, I already have a "wait 0.5 seconds" brick and after some testing it seems that I have to wait at least 0.3 seconds for the sound to start.
This seems to be very inconsistent and unfortunately I don't think a more precise check is possible.

wslany: Well, shouldn't it then be >200? You could compare to > 150 instead of > 0, right?

wslany: I just want to make these tests future proof (Moore's law...)

maxf24: After running the test three times the position of the Sound when transitioning back were: 227, 82, 141 ms
It really varies a lot.

wslany: > 50?

wslany: Thanks!

wslany:
Thanks!

PR #3507, opened 2020-03-02
CATROID-500 Fix variable interpretation in WebRequestBrick

stefanreichenauer:
Moved the interpretation of the input formula of the WebRequestBrick
from the ActionFactory to the action, so that the formula is
interpreted at run-time.

[CATROID-500](https://jira.catrob.at/browse/CATROID-500)

### DIFF
--- a/catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
+++ b/catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
@@ -28,6 +28,8 @@ import org.catrobat.catroid.content.SingleSprite;
 import org.catrobat.catroid.content.Sprite;
 import org.catrobat.catroid.content.actions.WebRequestAction;
 import org.catrobat.catroid.formulaeditor.Formula;
+import org.catrobat.catroid.formulaeditor.FormulaElement;
+import org.catrobat.catroid.formulaeditor.InterpretationException;
 import org.catrobat.catroid.formulaeditor.UserVariable;
 import org.catrobat.catroid.stage.StageActivity;
 import org.catrobat.catroid.stage.StageListener;
@@ -38,6 +40,7 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.Mockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
@@ -62,6 +65,8 @@ public class WebRequestActionTest {
 	private static final Double ERROR_BAD_REQUEST = 400d;
 	private static final String TEST_URL = "https://catroid-test.catrob.at/pocketcode/";
 	private static final String TEST_USERVARIABLE = "testUservariable";
+	private static final String TEST_INPUTVARIABLE = "testInputvariable";
+	private static final String RESPONSE_STRING = "Response";
 
 	private Sprite testSprite;
 	private UserVariable userVariable;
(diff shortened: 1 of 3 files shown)


catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -185,4 +190,50 @@ public void testCancelledCallAndResendRequest() {
...
+	public void testSuccessfulResponseWithInputVariable() {
+		UserVariable inputVariable = new UserVariable(TEST_INPUTVARIABLE, TEST_URL);
+		FormulaElement formulaElement = new FormulaElement(
+				FormulaElement.ElementType.USER_VARIABLE, inputVariable.getName(), null);
+		final String responseString = "Response";
+
+		Formula formula = Mockito.spy(new Formula(formulaElement));
+
+		testSprite.addUserVariable(userVariable);
+		testSprite.addUserVariable(inputVariable);
```
DinosaurierRex: Could you split up the test in a way that its sepparated from setup and the code lines which really do test something. Also please think about implementing the teardown function because you are using static members like the `StageActivity.stageListener. ...`. This mocks will remain for further tests and may bring them to fail if not reset

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -185,4 +190,50 @@ public void testCancelledCallAndResendRequest() {
...
+
+		when(StageActivity.stageListener.webConnectionHolder.addConnection(any())).thenReturn(true);
+
+		try {
+			doAnswer(invocation -> {
+				Sprite sprite = invocation.getArgument(0);
+				return String.valueOf(sprite.getUserVariable(TEST_INPUTVARIABLE).getValue());
+			}).when(formula).interpretString(any(Sprite.class));
+		} catch (InterpretationException e) {
+			fail("Formula could not be interpreted");
```
DinosaurierRex: An Exception will automatically cause the test to fail. So please dont catch Excpetions just to make the test "fail". Also think about another test where this failing could be triggered in a way that you want the test to pass if the Exception is thrown. Hint: Annotations `@Test(expected = NullPointerException.class)` like this will make the test pass if the mentioned exception is thrown.

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
```diff
@@ -185,4 +190,50 @@ public void testCancelledCallAndResendRequest() {
...
+			return webConnection;
+		}).when(webConnectionFactory).createWebConnection(anyString(), any());
+
+		doAnswer(invocation -> {
+			action.onRequestFinished(responseString);
+			return null;
+		}).when(webConnection).sendWebRequest();
+
+		assertTrue(action.act(0f));
+		assertEquals(responseString, userVariable.getValue());
```
DinosaurierRex: Split these asserts up in several TCs as far as possible. If a setup function is implemented, for each Testcase your setup will be loaded and then its easier to split up

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
stefanreichenauer: Setup is now in a seperate function.

Can you please specify, how the existing teardown-function is not sufficient?
I have only seen other tests setting the **Stagelister.stageListener** to null.

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
stefanreichenauer: I removed the try-catch-block, and added the exception to the method signature.

I choose to not add a test for testing the failing exception, because the exception is triggered by the function **Formula.interpretString**, which is not to test in this test-class, and the function is mocked anyway.
Please let me know, if I should add a test for it.

catroid/src/test/java/org/catrobat/catroid/test/content/actions/WebRequestActionTest.java
stefanreichenauer: Done.

wslany:
Thanks!!

PR #3509, opened 2020-03-05
CATROID-388 Implement <item's index('abc', *list name*) - function

trummerp:

### DIFF
--- a/catroid/src/main/res/values/strings.xml
+++ b/catroid/src/main/res/values/strings.xml
@@ -1248,6 +1248,7 @@
     <string name="brick_broadcast_default_value">"new message"</string>
     <string name="brick_add_item_to_userlist_add">Add</string>
     <string name="brick_add_item_to_userlist">to list</string>
+    <string name="brick_clear_userlist">Delete all items from list</string>
     <string name="brick_delete_item_from_userlist_delete">Delete item from list</string>
     <string name="brick_delete_item_from_userlist">at position</string>
     <string name="brick_insert_item_into_userlist_insert_into">Insert</string>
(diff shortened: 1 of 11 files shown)


catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
```diff
@@ -574,6 +576,18 @@ private Object interpretFunctionContains(Object right, Sprite sprite) {
 		return 0d;
 	}
 
+	private Object interpretFunctionIndexOfItem(Object left, Sprite sprite) {
+		if (rightChild.getElementType() == ElementType.USER_LIST) {
+			Project currentProject = ProjectManager.getInstance().getCurrentProject();
```
PlantinumX: pls ask @DinosaurierRex possible to avoid ProjectManager.getInstance().getCurrentProject()

catroid/src/main/res/values/strings.xml
```diff
@@ -1801,6 +1801,8 @@ needs read and write access to it. You can always change permissions through you
     <string name="formula_editor_function_list_item_parameter">(1,*list name*)</string>
     <string name="formula_editor_function_contains">contains</string>
     <string name="formula_editor_function_contains_parameter">(*list name*,1)</string>
+    <string name="formula_editor_function_index_of_item">index of item</string>
+    <string name="formula_editor_function_index_of_item_parameter">(1,*list name*)</string>
```
PlantinumX: pls check again if it is possible with apostroph

trummerp: Renamed it, so that it is equal as in the ticket description

catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java
trummerp: I will leave that as it is for now, because there is no dependency injection for the project manager implemented yet

trummerp:
I think there went something wrong. I wanted to replace the testcase from @wslany with the one i commited, but reseted the commit accidentally. I made a new PR with the original commit and the wanted changes.

PR #3512, opened 2020-03-09
CATROID-503 Add language test for decimal logarithm function

christophsiegl:
created a simple catrobat test for decimal logarithm function


christophsiegl:
Thanks for the sample-project, I appreciate it!
Did not know about the possibility to use 'NaN'.
And I am not sure what the correct result of log('abc') would be? I would assume "ERROR" instead of "FALSE"? But currently it is "FALSE"?!

wslany:
> Thanks for the sample-project, I appreciate it!
You are welcome.

> Did not know about the possibility to use 'NaN'.
Yeah, it's little known, but important. And I think this may be the first Catrobat language test where it is documented.

> And I am not sure what the correct result of log('abc') would be? I would assume "ERROR" instead of "FALSE"? But currently it is "FALSE"?!
Well, ERROR is not very useful since at the moment it cannot be checked during runtime in a project. It could be something for the future, but as of now, false may be easier to check. Anyway, this is a borderline case that is not important. Most likely, no project will ever rely on this behavior, so changing it later once better exception handling becomes available will most likely not result in any broken projects. We could also leave it away, but equally could change the test if the behavior ever changes.

wslany:
Thanks!

PR #3513, opened 2020-03-10
CATROID-505 Add language test for tangent function

christophsiegl:
created a catrobat test for the tangent function


BrandyMan1:
PLEASE FIX: 
-  Please use consistent variable capitalization and do not use abbreviations.
-  Also, please make use of the possibility to use normal spaces as in English inside variable names.
-  For instance, please use "tangent zero" instead of "tanZero".
-  Note that you can delete and rename variables, as follows: in the formula editor, tap on the Data button. Then long-press on a variable name to delete or rename it.
- Proof if tangent(9x45) = 1 instead of proofing for '<= 1' and '<0.999'
- I think it would be enough to test things like "tangent(360 + 45) = tangent(45)" and "tangent(315) = tangent(-45)" once. Then you can focus on angels between 0 and 360°. This would lead to better readable code.

wslany:
Thanks!

PR #3519, opened 2020-03-12
CATROID-504 Fix Lost changes due to reloading of projects

Pratik1999:
Ticket :- https://jira.catrob.at/projects/CATROID/issues/CATROID-504

The Bug was Caused due to Reloading of Current Project in ProjectUploadActivity, When the project is reloaded the project variable is assigned a new reference, whereas the other variables like currentlyEditedScene, etc still refer to the old project reference Scene. So when the ProjectUploadActivity finishes and the ProjectActivity and corresponding fragments resume the variable don't point to correct reference.

The ProjectUploadActivity has been modified to not load the projects when it is already loaded(which is the case when the ProjectUploadActivity is launched from ProjectActivity)


wslany:
Please use https://github.com/Catrobat/Catroid/blob/develop/.github/PULL_REQUEST_TEMPLATE.md in the topmost commit comment.

PR #3522, opened 2020-03-14
CATROID-511 Refactor distance calculations to use standard Java math …

Abdelaty:
Instead of using our own implementation based on the Pythagorean theorem to calculate 
distances, use the standard math functions from Java.

Signed-off-by: abdelaty <abdelaty.mohammedmagdi@gmail.com>

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -55,13 +55,9 @@ import androidx.annotation.VisibleForTesting;
 
 public class Look extends Image {
 
-	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
-		int touchIndex = TouchUtil.getLastTouchIndex();
-
-		return (float) Math.hypot((TouchUtil.getX(touchIndex) - getXInUserInterfaceDimensionUnit()),
-				(TouchUtil.getY(touchIndex) - getYInUserInterfaceDimensionUnit()));
-	}
-
+	@Retention(RetentionPolicy.SOURCE)
+	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
+	public @interface RotationStyle {}
 	public static final int ROTATION_STYLE_LEFT_RIGHT_ONLY = 0;
 	public static final int ROTATION_STYLE_ALL_AROUND = 1;
 	public static final int ROTATION_STYLE_NONE = 2;
@@ -308,9 +304,11 @@ public class Look extends Image {
 		setY(y - getHeight() / 2f);
 	}
 
-	@Retention(RetentionPolicy.SOURCE)
-	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
-	public @interface RotationStyle {
+	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
+		int touchIndex = TouchUtil.getLastTouchIndex();
+
+		return (float) Math.hypot((TouchUtil.getX(touchIndex) - getXInUserInterfaceDimensionUnit()),
+				(TouchUtil.getY(touchIndex) - getYInUserInterfaceDimensionUnit()));
 	}
 
 	public float getAngularVelocityInUserInterfaceDimensionUnit() {
--- a/catroid/src/main/java/org/catrobat/catroid/content/Look.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Look.java
@@ -55,9 +55,13 @@ import androidx.annotation.VisibleForTesting;
 
 public class Look extends Image {
 
-	@Retention(RetentionPolicy.SOURCE)
-	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
-	public @interface RotationStyle {}
+	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
+		int touchIndex = TouchUtil.getLastTouchIndex();
+
+		return (float) Math.hypot((TouchUtil.getX(touchIndex) - getXInUserInterfaceDimensionUnit()),
+				(TouchUtil.getY(touchIndex) - getYInUserInterfaceDimensionUnit()));
+	}
+
 	public static final int ROTATION_STYLE_LEFT_RIGHT_ONLY = 0;
 	public static final int ROTATION_STYLE_ALL_AROUND = 1;
 	public static final int ROTATION_STYLE_NONE = 2;
@@ -304,12 +308,9 @@ public class Look extends Image {
 		setY(y - getHeight() / 2f);
 	}
 
-	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
-		int touchIndex = TouchUtil.getLastTouchIndex();
-
-		return (float)
-				Math.sqrt(Math.pow((TouchUtil.getX(touchIndex) - getXInUserInterfaceDimensionUnit()), 2)
-						+ Math.pow((TouchUtil.getY(touchIndex) - getYInUserInterfaceDimensionUnit()), 2));
+	@Retention(RetentionPolicy.SOURCE)
+	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
+	public @interface RotationStyle {
 	}
 
 	public float getAngularVelocityInUserInterfaceDimensionUnit() {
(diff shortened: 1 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
```diff
@@ -370,7 +370,7 @@ private static float pointToLineDistance(PointF lineStart, PointF lineEnd, Point
 	}
 
 	private static float pointToPointDistance(PointF p1, PointF p2) {
-		return (float) Math.sqrt(((p1.x - p2.x) * (p1.x - p2.x) + (p1.y - p2.y) * (p1.y - p2.y)));
+		return (float) Math.hypot(p1.x - p2.x, p1.y - p2.y);
```
wslany: SUGGESTION: Consider using https://refactoring.guru/inline-method here instead. The code is so simple that it seems unnecessary to have a separate function here, which is called only once. Also, in the other case a unit test for the function would be needed --- an overkill here IMHO.

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -55,9 +55,13 @@
 
 public class Look extends Image {
 
-	@Retention(RetentionPolicy.SOURCE)
-	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
-	public @interface RotationStyle {}
+	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
```
wslany: PLEASE CHECK: For what is this actually *really* needed?

catroid/src/main/java/org/catrobat/catroid/content/Look.java
```diff
@@ -304,12 +308,9 @@ public void setYInUserInterfaceDimensionUnit(float y) {
...
 
-	public float getDistanceToTouchPositionInUserInterfaceDimensions() {
-		int touchIndex = TouchUtil.getLastTouchIndex();
-
-		return (float)
-				Math.sqrt(Math.pow((TouchUtil.getX(touchIndex) - getXInUserInterfaceDimensionUnit()), 2)
-						+ Math.pow((TouchUtil.getY(touchIndex) - getYInUserInterfaceDimensionUnit()), 2));
+	@Retention(RetentionPolicy.SOURCE)
+	@IntDef({ROTATION_STYLE_LEFT_RIGHT_ONLY, ROTATION_STYLE_ALL_AROUND, ROTATION_STYLE_NONE})
+	public @interface RotationStyle {
```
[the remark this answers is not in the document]
Abdelaty: Sorry @wslany I edited now.

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
Abdelaty: I used the inline-method and removed the "pointToPointDistance" function, because it's not necessary now.
Thanks for this suggestion. and for the refrence ^_^

catroid/src/main/java/org/catrobat/catroid/content/Look.java
Abdelaty: When I used Android Studio Refactoring tool, it just split the sorting of functions!
Sorry for that, I have edit it now by another commit.

Abdelaty:
@wslany Thank you sir for the review. And sorry because it's more than 1 commit. but I had a problem in merging it.

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
```diff
@@ -164,10 +164,9 @@ private void createCollisionPolygon(String path) {
 				ArrayList<PointF> points = getPointsFromPolygonVertices(boundingPolygon.get(i));
 				ArrayList<PointF> simplified = simplifyPolygon(points, 0, points.size() - 1, epsilon);
 
-				if (pointToPointDistance(simplified.get(0), simplified.get(simplified.size() - 1)) < epsilon) {
+				if (Math.hypot(simplified.get(0).x - simplified.get(simplified.size() - 1).x, simplified.get(0).y - simplified.get(simplified.size() - 1).y) < epsilon) {
```
wslany: PRAISE: Actually, after seeing this solution, I am thinking your previous code for this was more readable and easily understandable. Sorry, my bad for suggesting the additional refactoring, I guess I didn't think enough about the complexity in the parameters. 

Could you please change it  back to your previous version, i.e., with the Math.hypot in pointToPointDistance? I guess also no explicit unit test is needed for the pointToPointDistance, though other reviewers might disagree. I apologize.

Thanks!

catroid/src/main/java/org/catrobat/catroid/content/Look.java
wslany: Ahh, what I meant was: For what is this method used, what functionality does it serve, how does it impact Catroid? I may be a kind-of trick question, but actually, I am not sure myself ;-)

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
Abdelaty: Yes, I think this also, after this change it will take more time to be understood. I reverted now and add the pointToPointDistance method again. Thank you for reviewing ^_^

catroid/src/main/java/org/catrobat/catroid/content/Look.java
Abdelaty: After investigation and reading about annotations and @IntDef, its alternative to use enumeration, since enumeration is bad for performance. 
this method is being used at 
[ L54 SetRotationStyleAction.java](https://github.com/Catrobat/Catroid/blob/fb5dbaa40516c24e16b31d639174bd74ea54f2fe/catroid/src/main/java/org/catrobat/catroid/content/actions/SetRotationStyleAction.java#L54)
[ L283 ActionFactory.java](https://github.com/Catrobat/Catroid/blob/fb5dbaa40516c24e16b31d639174bd74ea54f2fe/catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java#L283) 
[ L102 SetRotationStyleAction.java](https://github.com/Catrobat/Catroid/blob/fb5dbaa40516c24e16b31d639174bd74ea54f2fe/catroid/src/main/java/org/catrobat/catroid/content/bricks/SetRotationStyleBrick.java#L102)

wslany: Nono, I meant: For what do we need getDistanceToTouchPositionInUserInterfaceDimensions?

Abdelaty: we are using it here [L921 FormulaElement](https://github.com/Abdelaty/Catroid/blob/fb5dbaa40516c24e16b31d639174bd74ea54f2fe/catroid/src/main/java/org/catrobat/catroid/formulaeditor/FormulaElement.java#L921)
it's kinda confusing for me, but I think it's return the distance from base point to touch position. 
I think also the FormulaElement class need to get refactoring.

wslany: Sure, but please compare it to the other formula elements --- this one does not seem to be used anywhere in the app, I mean it is not part of anything the user can use in a formula in the Catroid app. Sure, there are tests, there's the code, but it cannot be used by the end-user, unless I am mistaken. I wonder what happened here. Can you check via "blame" who was responsible for this? Maybe we can clear that up, and either add the feature, or remove the code. What do you think?

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
```diff
@@ -167,7 +167,7 @@ private void createCollisionPolygon(String path) {
 				if (pointToPointDistance(simplified.get(0), simplified.get(simplified.size() - 1)) < epsilon) {
 					simplified.remove(simplified.size() - 1);
 				}
-
+
```
wslany: Why space?

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
```diff
@@ -370,9 +370,9 @@ private static float pointToLineDistance(PointF lineStart, PointF lineEnd, Point
 	}
 
 	private static float pointToPointDistance(PointF p1, PointF p2) {
-		return (float) Math.sqrt(((p1.x - p2.x) * (p1.x - p2.x) + (p1.y - p2.y) * (p1.y - p2.y)));
-	}
-
+		return (float) Math.hypot(p1.x - p2.x, p1.y - p2.y);
+	}	
+
```
wslany: Why spaces?

catroid/src/main/java/org/catrobat/catroid/content/Look.java
Abdelaty: Sure, I will check "blame", And will investigate about the point from those code lines.

Abdelaty: I checked blame, and found that @robertriedl created it. But I don't think that he remembers it's 4 years already 😄 
image
I traced it and now I'm completing a tracing report to post it here before we can do any action.
And I investigated and it doesn't seem to be used in any part of the app, but I can't figure out what is this written for. So I may need some time to discover that.

catroid/src/main/java/org/catrobat/catroid/sensing/CollisionInformation.java
Abdelaty: After I moved the functions it added space, will fix it now.

Abdelaty:
I'm unable to squash commits so I will create another PR.

PR #3551, opened 2020-03-27
Fix typos in strings.xml

NBAMIR:
**Description:**
Change occurrences of "signin" to "sign_in".


wslany:
Thanks, but we only do such changes in catroid/src/main/res/values/strings.xml --- all the other languages are translated via https://translate.catrobat.org/ and the files are then automatically imported from there before releases. However, in this case the change in the id seems not worth retriggering the translators, so I guess we better skip it --- sorry for your work! Hope that's OK. But thank you for your help!!!!

PR #3565, opened 2020-03-31
Fix typo in attr.xml, styles.xml and styles_accessibility.xml

NBAMIR:
**Description:**
Renamed category_icon_visiblity to category_icon_visibility.


wslany:
Thanks!!

PR #3567, opened 2020-03-31
Replaced manual min max calculation with Math.min and Math.max.

NBAMIR:
**Description:**
Replaced manual min max calculation with Math.min and Math.max.


wslany:
Thanks!

PR #3583, opened 2020-04-18
JENKINS-377 change number of builds to keep

jampo:
Change the number of builds to keep to use less disk space on Jenkins see https://jira.catrob.at/browse/JENKINS-377


wslany:
Thanks!

PR #3591, opened 2020-04-28
CATROID-534

mireiallaurado:
NFCTagList used to save the same NFC tag several times when scanned.

Now it saves the NFC tags that aren't on the list but doesn't save the ones that have already been scanned.

https://jira.catrob.at/browse/CATROID-534


DinosaurierRex:
There are two tickets within one PR, and for the second ticket there are only code style changes. Please reopen one PR for each ticket. Thanks in advance :pray:

PR #3601, opened 2020-05-08
CATROID-477 Switch option from backpack

juliajulie95:
https://jira.catrob.at/browse/CATROID-477
Switching the options in the context menu, which influences not only the backpack, but also delete and copy action. Due to how the action are coded, adroid decides if the item is shown as text or overflow menu by the length of the text. In englisch it shows the text, in german the overflow menu is shown.

Due to android choosing to create a overflow menu automatically if the text is too long, I only switched the options and did not force an overflow menu. Usability wise this this is better, because forcing an overflow menu for one item will have the user do one more action to get to their target although it is not necessary and this weighs more than having german and englisch version identical. The english version now shows "select_all" as text, the german version shows an overflow menu. If the overflow menu would be forced in the englisch version, a hack would be needed to test this, because the action is shown over the project view and both overflow menus are exactly the same, which leads to the matcher not being able to match the overflow menu correctly. This could potentially make the "deleteAllScenes" test flaky.

- [ ] Confirm that the changes follow the project’s coding guidelines
- [ ] Perform a self-review of the changes


wslany:
Thanks!!

PR #3623, opened 2020-06-04
Catroid-279 Fix unnecessary vibrations when pausing a project containing a vibration brick

trummerp:
This PR should solve the problem with the unnecessary Vibration, by starting the Vibration synchronous and letting the vibration thread just checking when to stop. This PR consists of 2 commits.

Commit 1: Solves the base problem of the unnecessary vibration. This commit also intended a mocked regression test, which should have proved that the behavior is correct now. But although Powermockito provides a wide various range to test this methods, i faced a lot of other complications like time dependencies (code reads time from device <-> local tests always failing/read always zero if not executed on a device), so i decided not to go with a test for now. Help is welcome.

Commit 2: Refactoring of unnecessary calls and some other things:
- Resume call right after start
- Pause call if pressed the "Back"-button while already paused.
- Removing all VibrationUtil interactions, if the script contains NO vibration brick
- Refactored deprecated code (conversions, ...) 

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -435,20 +435,23 @@ public class StageListener implements ApplicationListener {
 			return;
 		}
 		this.stageDialog = stageDialog;
-		if (!ProjectManager.getInstance().getStartScene().getName().equals(scene.getName())) {
-			transitionToScene(ProjectManager.getInstance().getStartScene().getName());
+		ProjectManager projectManager = ProjectManager.getInstance();
+		Project currentProject = projectManager.getCurrentProject();
+
+		if (!projectManager.getStartScene().getName().equals(scene.getName())) {
+			transitionToScene(projectManager.getStartScene().getName());
 		}
 		stageBackupMap.clear();
 		embroideryPatternManager.clear();
 
-		FlashUtil.reset();
 		VibrationUtil.reset();
+		FlashUtil.reset();
 		TouchUtil.reset();
 		removeAllClonedSpritesFromStage();
 
-		UserDataWrapper.resetAllUserData(ProjectManager.getInstance().getCurrentProject());
+		UserDataWrapper.resetAllUserData(currentProject);
 
-		for (Scene scene : ProjectManager.getInstance().getCurrentProject().getSceneList()) {
+		for (Scene scene : currentProject.getSceneList()) {
 			scene.firstStart = true;
 		}
 		reloadProject = true;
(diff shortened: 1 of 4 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -435,20 +436,26 @@ public void reloadProject(StageDialog stageDialog) {
...
+		Project currentProject = projectManager.getCurrentProject();
+
+		if (!projectManager.getStartScene().getName().equals(scene.getName())) {
+			transitionToScene(projectManager.getStartScene().getName());
 		}
 		stageBackupMap.clear();
 		embroideryPatternManager.clear();
 
+		if (currentProject.getRequiredResources().contains(Brick.VIBRATION)) {
+			VibrationUtil.reset();
```
mandelbrot2357, line 449: Is this check necessary? Getting all required resources is probably quite slow, so if resetting the time to 0 when vibration is not used doesn't affect anything, I would just leave out the conditional statement.

trummerp: Update:
- Killed "context" as a member in VibrationUtil
- Removed some time consuming if conditions


ad:
> Also, the StageResourceHolder already starts the thread. I would either eliminate that line so you don't create an unnecessary thread at the beginning or keep it and simply leave one vibration thread running and use a mutex instead of the binary semaphore to update the vibration time to tell the thread to start vibrating again. The latter option would probably cut down complexity quite a bit.

I will leave that as it is. Im not sure if there is an explicit mutex in java. There is some similar stuff like reentrantlocks, binary semaphores or synchronized objects, which behaves like mutual expressions, but i think the solution with the binary semaphore is just fine. Also the thread needs to get started at the beginning with the existing solution.

trummerp:
Before consider merging this please check [CATROID-608](https://jira.catrob.at/browse/CATROID-608), since there is a old bug related to this ticket, which changes it's behavior.

catroid/src/main/java/org/catrobat/catroid/utils/VibrationUtil.java
```diff
@@ -124,52 +112,41 @@ public static void reset() {
 		setTimeToVibrate(0.0);
 	}
 
-	public static void setContext(Context stageContext) {
-		context = stageContext;
+	public static void setVibrator(Vibrator vibrator) {
+		vibration = vibrator;
```
mandelbrot2357, line 116: The member function should be called vibrator as well, in accordance with the setter name,

catroid/src/main/java/org/catrobat/catroid/stage/StageResourceHolder.java
```diff
@@ -318,7 +318,7 @@ public void onShow(DialogInterface dialog) {
 		if (requiredResourcesSet.contains(Brick.VIBRATION)) {
 			Vibrator vibration = (Vibrator) stageActivity.getSystemService(VIBRATOR_SERVICE);
 			if (vibration != null) {
-				VibrationUtil.setContext(stageActivity);
+				VibrationUtil.setVibrator(vibration);
```
mandelbrot2357, line 321: The naming doesn't really matter as much here, but while you're at it...

PR #3634, opened 2020-06-15
CATROID-176 IfOnEdgeBounce and Steps with Physics Objects

maxf24:
Physics Objects did only bounce due to physical collisions, not due to steps taken explicitly.

- Physics Objects touching the edge of the screen by steps taken now also bounce off and travel in the opposite direction.
- Physics Objects using the Left-Right rotation-style are now correctly flipped when set to look left on start.

As discussed with @wslany only "testIfOnEdgeBouncePhysicsMoveStepsRegression.catrobat" is needed for the Acceptance Criteria found on Jira. 

JIRA: https://jira.catrob.at/browse/CATROID-176


trummerp:
Hi, good work! 
There are some things i came across while testing:

- If bouncing on the **left or right edge**, the object's flip seems to be depending on the starting angle. 
E.g.: I added a starting angle of 10° to [https://share.catrob.at/pocketcode/program/79684](url) and what happened was that
the object always reflecting with a degree of 0°+-(10°) or -180°+-(10°).
- Also the gravity seems a little strange. Depending on the angle (I tested it with 45°) at which the object hits the edge, it seems to accelerate or slow down and never really stops flying through the space(screen). I guess it is expected, that sooner or later the object should be held on the ground by gravity (At least that is what i expect about gravity:D)

maxf24:
Hi!

_If bouncing on the left or right edge, the object's flip seems to be depending on the starting angle._
This is intentional as it's the same behaviour as a non physic object that touches the wall with steps and an specific angle.

 _it seems to accelerate or slow down and never really stops flying through the space(screen)_
If an object is flying up with steps and normal gravity is pulling it down, it starts to fall when the velocity > steps/min. When touching the ground the velocity is much greater then we think it is because the steps slowed it down before, therefore bouncing much harder as one would think. 
The problem here is that steps and the physics environment work very differently and do not mix well.

Koell (approved):
👍

wslany:
Thanks!

PR #3865, opened 2020-11-09
CATROID-740 Change dialog text for deletion of variables

ChriMoser:
Change dialog text for deletion of variables to "Delete this variable?" "Deleting a variable that is still in use could cause errors. You can't undo this!"

[](https://jira.catrob.at/browse/CATROID-740)

### DIFF
--- a/catroid/src/main/res/values/strings.xml
+++ b/catroid/src/main/res/values/strings.xml
@@ -2051,8 +2051,9 @@ needs read and write access to it. You can always change permissions through you
     <string name="cast_screen_pause_text">Paused</string>
     <!-- -->
 
-    <string name="deletion_alert_title">Warning!</string>
-    <string name="deletion_alert_text">Deleting a variable that is still in use is dangerous. Are you sure?</string>
+    <string name="deletion_alert_title">Delete?</string>
+    <string name="deletion_alert_text">Deleting a variable or list that is still in use can cause
+        errors. You can\'t undo this!</string>
     <string name="deletion_alert_yes">Yes, delete anyway!</string>
 
     <!-- Upload progress dialog -->


juliajulie95 (approved):

catroid/src/main/res/values/strings.xml
```diff
@@ -2051,8 +2051,9 @@ needs read and write access to it. You can always change permissions through you
     <string name="cast_screen_pause_text">Paused</string>
     <!-- -->
 
-    <string name="deletion_alert_title">Warning!</string>
-    <string name="deletion_alert_text">Deleting a variable that is still in use is dangerous. Are you sure?</string>
+    <string name="deletion_alert_title">Delete this variable?</string>
+    <string name="deletion_alert_text">Deleting a variable that is still in use could cause errors.
```
wslany, line 2055:
```suggestion
    <string name="deletion_alert_title">Delete?</string>
    <string name="deletion_alert_text">Deleting a variable or list that is still in use can cause errors.
```
Reason: This text is shown also when more than one variables and/or lists were selected, so plurals would have to be considered, as well as lists mentioned, see below. Also, the proposed shorter title line is completely understandable from the context, i.e., from the option that the user selected in the overflow menu.
image

Note: There would be a way to differentiate between singulars and plurals in texts shown in Android, but that involves more code, and I also think that it is not necessary in this case as the phrase is understood equally well. In this case it would also be quite complicated, e.g., if one variable and two lists were selected etc. Also, in other languages it is not necessary, for instance in Japanese there is no difference between singular and plural. So, IMHO it is not be worth the effort to differentiate here between singular and plural.

wslany (changes requested):

ndrnour (approved):

ndrnour:
not sure why you create a ticket for such a fix? this could be fixed on Crowdin and then by updating the string.xml files

wslany:
> not sure why you create a ticket for a text readability improvement? this could be fixed on Crowdin and then by updating the string.xml files

This was a TRAINING ticket meant for first time contributors.

wslany (approved):

wslany:
Thanks!

PR #3884, opened 2020-11-25
CATROID-770 "index of current touch(__)" multi-touch sensor

hofcle78:
Adapted code to return the index of the current touch.

https://jira.catrob.at/browse/CATROID-770


dandrawes (approved):
Looks good!

wslany (changes requested):
The index seems to be one too small, i.e., if the screen is being touched the third time, and only the finger of the third touch remains on the screen (i.e., the third touch is still going on), _index of current touch(1)_ incorrectly is equal to 2 instead of 3. Please note that the first touch has number one, second number two, etc. If, like in the current implementation of _index of current touch()_ the screen has not yet been touched, then _index of current touch(1) = 0_, but once it is being touched for the first time (ongoing), _index of current touch(1)_ incorrectly still remains 0, instead of correctly being 1. 

I'm not sure why the following is happening, which is rather strange behavior: If I exchange the "Wait until screen is touched is true" brick by a "Wait 0.5 seconds"brick, and change the durations in the "Touch at" bricks in the broadcast scripts to a longer time, e.g., 10 seconds, the test succeeds when I execute it the first time, but fails the second and third etc time when it is executed without leaving the project. It only works again if the project is exited completely and entered again. The value of _index of current touch(2)_ is 3 only the first time. In later executions, the value of _index of current touch(2)_ remains 0. Here's the relevant part with changes marked (without the second Wait until brick and following fourth Assert brick, but they change nothing w.r.t. the bug):

image

Also, testCurrentIndexOfTouch.catrobat is prone to flakiness: 

1. In the third Assert brick, _index of current touch(2)_ is asserted to be equal to 3. Replacing the 2 by 1, the result of _index of current touch(1)_ must be asserted to be 2 to make the test successful on my phone (because of above bug --- correctly, the numbers reported by the sensor need to be increased by 1, so it should correctly be 3 instead of 2). However, this assignment depends on the order of execution of the two broadcast scripts, which can be arbitrary, so it could also be the other way around, and in that case, _index of current touch(2)_ would evaluate to 2 (in fact should still be 3, but only after the index offset correction) in the third Assert brick, and the test would thus fail. 

1. Additionally, since it is not guaranteed that the two broadcast scripts start exactly at the same time, the "Wait until stage is touched is true" brick could be triggered and the following Assert brick could be executed before the second broadcast script executes the Touch at brick, and in that case, _index of current touch(2)_ would evaluate to 0 in the third Assert brick, and the test would thus fail for another reason. 

In summary, the test's success depends on timing conditions coming from unknown implementation details. This could lead to different results on different hardware or with different implementations (e.g., Catty instead of Catroid), and especially also on our slow Android emulators on jenkins, and thus the test could fail from time to time or depending on the circumstances. Therefore, please refactor the test to make it safe against such implementation details but so that it still tests the correct implementation of the sensor reporter.

There's a small typo in the "lets touch" broadcast message: grammatically correct it would be "let's touch". It's not really important, but since the test needs to be changed anyway, I would be grateful if you could also fix this (thanks!).

hofcle78:
> The index seems to be one too small

done.

> If I exchange the "Wait until screen is touched is true" brick by a "Wait 0.5 seconds"brick, and change the durations in the "Touch at" bricks in the broadcast scripts to a longer time, e.g., 10 seconds, the test succeeds when I execute it the first time, but fails the second and third etc time when it is executed without leaving the project. It only works again if the project is exited completely and entered again.

done.

> Therefore, please refactor the test to make it safe against such implementation details but so that it still tests the correct implementation of the sensor reporter.

done.

> There's a small typo in the "lets touch" broadcast message: grammatically correct it would be "let's touch".

done.

petbuer (approved):

wslany (approved):

wslany:
Thanks!!

PR #3902, opened 2020-12-21
CATROID-384 Enable edit of broadcast messages

mstoeg:
https://jira.catrob.at/browse/CATROID-384

- Refactor old testcases and add new ones for editing broadcast messages (including the test case for a message with the same text in a second scene, which should not be changed by the renaming in the first scene) - as it was asked for in the ticket specification


dandrawes (approved):
Well done!

wslany (changes requested):
As implied by _Tapping "edit..." shall open the selected message (in the same text edit field as when defining a new message) and allow to edit it._ in https://jira.catrob.at/browse/CATROID-384 the existing name should be prefilled in the edit field. However, it currently is empty in the PR. Please change this so the old message is initially inserted when edited it. Reason: often, one only wants to correct a typo, or see the full text of the message, which may be longer than the one line field shown in the script view.

Also, when putting a script containing broadcast scripts into the backpack and unpacking them in another scene, leads to all messages being replaced visually by "edit...", see screenshot below. However, when selecting "edit..." and changing the message to, e.g., abc, and then pressing OK, the message remains "edit..." unchanged, and the new message abc is nowhere to be found in the spinner of messages. This also happened with only one scene, so it is not necessary to have two scenes, but this issue appears possible in a more repeatable way by creating a second scene first, then copying a script from the first scene to the second scene using the backpack. 
image

mstoeg:
Thank you @wslany for pointing out the first issue - I'm sorry for not noticing that.

As far as the second issue is concerned, I think there is a bug in the Backpack rather than in this implementation. The problem is that broadcast messages within the backback are not copied into the new scene when they are unpacked. This already leads to vanishing messages and weird message selections in the current develop branch version (see screencast). I'm a bit confused now about how the correct behavior should actually be.

https://user-images.githubusercontent.com/60399652/103475351-c59c4a80-4dac-11eb-9f78-52ea5c7ee3e8.mp4

dandrawes (approved):
Looks good!

wslany:
> As far as the second issue is concerned I fixed the issue with the freezed "edit..." messages being displayed. However, I think there is already a bug in the Backpack rather than in this implementation. The problem is that broadcast messages within the backback are not copied into the new scene when they are unpacked. This already leads to vanishing messages and weird message selections in the current develop branch version (see screencast). If this behavior is not intendet, I would suggest to add separate a bug ticket for this.

Thank you => https://jira.catrob.at/browse/CATROID-886

wslany (approved):

wslany:
Thanks!

PR #3908, opened 2020-12-26
CATROID-760 Fast Loops

blindguardian50:
Execute loop delays only if specific bricks are contained within
https://jira.catrob.at/browse/CATROID-760

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java
@@ -915,30 +915,34 @@ public class ActionFactory extends Actions {
 		return action;
 	}
 
-	public Action createRepeatAction(Sprite sprite, Formula count, Action repeatedAction) {
+	public Action createRepeatAction(Sprite sprite, Formula count, Action repeatedAction,
+			boolean isLoopDelay) {
 		RepeatAction action = Actions.action(RepeatAction.class);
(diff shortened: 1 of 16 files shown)


adnaribo (approved):

wslany (changes requested):
The loops correctly remain slowed down if, for instance, a "Next look" brick is used _directly_ in a loop. However, as soon as the "Next look" brick is _inside_ another nested brick or loop, e.g., in an "If then endif" brick, where the condition is true, i.e., where the "Next look" brick is executed during _each_ iteration, the loop delay nevertheless is **_incorrectly_** ignored.

E.g., the values of the variables x and y reported after running the stage a few seconds and then looking up the values in the formula editor for the following two loops, _incorrectly_ are the same. Correct would be that the number reported for the loop with the "Next look" brick should be much _lower_ (in the same amount of time).
image
image

image

Here the reported numbers if the "Next look" is outside of the nested "If then endif" brick, but still in the loop:
image

image

adnaribo (approved):

wslany (changes requested):
Please reduce the amount of duplicate code: 6 times the same block of 40+ lines that checks the loop delay. This could lead to divergent, difficult to maintain code if the parts are edited non-uniformly. See https://medium.com/grensesnittet/the-key-to-eliminating-duplicated-code-patterns-ddd23b958ded

PR #3914, opened 2020-12-30
Jenkins script for Pocket Code beta release and upload

juliajulie95:
Allows to upload Pocket Code beta to Google Play alpha of org.catrobat.catroid.test flavor
Like signed release, but debug flag for share.


wslany:
Thanks!

PR #3932, opened 2021-01-08
CATROID-864 Show keyboard for TextInputDialogs

mstoeg:
https://jira.catrob.at/browse/CATROID-864

- Fixed bug of Scratch2Catrobat search view, where the results of the old search persisted even if no more character was within the textfield.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/ViewUtils.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/ViewUtils.kt
@@ -1,0 +1,50 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2021 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


ndrnour (approved):
well done. this will fix it for all alert dialogs. but not for custom dialogs that contain Editable textView like `RegistrationDialogFragment` and `LoginDialogFragment`, which should be fixed once we have a ticket to refactor them to kotlin.

wslany:
Very nice!

Two questions, just to make sure: 
1. Is this a behavior that can be tested? It would be great, so we don't forget it should work.
2. Also, I found that the dialog for the Scratch converter interface does not automatically display the keyboard for the search text field. It would be awesome to have the keyboard shown there too, since one would normally always need it and the screen anyway is half empty, since there are just four sample entries. However, if this should better be kept for another ticket, that would be OK from my side. So, only if it is straightforward with this ticket, if at all.

wslany:
If you change anything, please also update the license header date range to -2021.

wslany (changes requested):
As discussed on slack. Thank you!

mstoeg:
I implemented the requested changes regarding custom dialogs (like `RegistrationDialogFragment` and `LoginDialogFragment`) and for the `ScratchConverter` Activity. In the course of that I also changed the header date. 

As far as a testcase is concerned I could not really come up with a sensible way to test this. With a Espresso Test testing all the occurences would be extremely inefficient. Moreover, the tests would most likely be flaky because of timing varieties until the keyboard is displayed. Also a Unit Test makes no sense in this case as it would require to mock the whole building process of the dialogs which is not really feasible.

ndrnour:
it looks good, but some optimization can be done. let me know when you have time for a call in order to do it.

ndrnour (approved):

ndrnour:
regarding the testing, until now there is no way to check whether the keyboard is shown or hidden on Android.
Therefore no testing in this PR.

wslany (approved):

wslany:
Thanks!!

PR #3945, opened 2021-01-22
CATROID-897 Added CLT for ForVariableFromTo brick

s-swapnil:
https://jira.catrob.at/projects/CATROID/issues/CATROID-897

Added CLT for "For Variable From To" brick with different values and test cases. Notes also present with the test cases.


wslany (changes requested):
Thanks a lot!!

Please make the following changes: 
- Add a loop with a negative number as start, a positive as end
- Add a loop with two negative numbers, the second smaller (more negative) than the first
- Show that, after the loop is finally exited, the value of the variable is the final value from the For brick, even if there was a (futile) attempt to set it to a different value in the loop. You can do this with the existing tests where you assert that data is 55, i.e., add an assert that x3 should be 10
- Show what happens when the numbers are not an integers, e.g., from -2.7 to 3.14 (the loop shall not be executed even once).
- Rename the folder control.ForVariableFromTo to control (I know that there's also control.ForItemInUserList, and that should be corrected too --- I would be grateful if you could move the test in it also into the same control folder and remove the control.ForItemInUserList folder).
- Add two loops, one inside the other, to show what happens when there are nested loops (with different variables)
- Add two loops, one inside the other, to show what happens if there are nested loops that use the _same_ variable

s-swapnil:
Thanks @wslany for suggesting these interesting test cases,I have added them to the CLT.I've also rearranged the above mentioned CLT's into the 'control' folder as per your directive.
Just one thing- I believe that the test case of looping from -2.7 to 3.14 works a bit differently(as evident from the CLT, please see). The loop does execute, but it does so by dropping the decimal parts from these non-integral limits.(So, it runs from -2 to 3 in this particular instance)

s-swapnil:
Hi @wslany , are there any more changes to be made? If yes,please say.

wslany (approved):

qlin93 (approved):
Looks good!

wslany:
Thanks!

PR #4000, opened 2021-03-03
CATROID-935 "Recently used" Brick category

devsebastian:
Ticket: https://jira.catrob.at/browse/CATROID-935


catroid/src/main/res/drawable/recently_used_pos.xml
```diff
@@ -0,0 +1,32 @@
...
+  -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="24dp"
+    android:width="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path android:fillColor="#FFFFFF"
+        android:pathData="M13.5,8H12V13L16.28,15.54L17,14.33L13.5,12.25V8M13,3A9,9 0 0,0 4,12H1L4.96,16.03L9,12H6A7,7 0 0,1 13,5A7,7 0 0,1 20,12A7,7 0 0,1 13,19C11.07,19 9.32,18.21 8.06,16.94L6.64,18.36C8.27,20 10.5,21 13,21A9,9 0 0,0 22,12A9,9 0 0,0 13,3" />
+</vector>
```
mstoeg, line 32: Add newline at EOF here.

catroid/src/main/res/values/styles.xml
```diff
@@ -583,6 +584,10 @@
         <item name="android:textSize">?attr/x_large</item>
     </style>
 
+    <style name="BrickCategory.RecentlyUsed">
+        <item name="android:background">@drawable/brick_selection_background_recently_used</item>
```
mstoeg, line 588: You have to add the .9.png file for different solutions as well (not just for xhdpi as you did). This is probably why you got lint-warnings in the first place.

catroid/config/lint-baseline.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
   ~ Catroid: An on-device visual programming system for Android devices
-  ~ Copyright (C) 2010-2020 The Catrobat Team
+  ~ Copyright (C) 2010-2021 The Catrobat Team
```
mstoeg, line 4: Please check if changing the `lint-baseline.xml` file is really necessary after adding .png files for different resolutions.

catroid/src/main/java/org/catrobat/catroid/io/BackpackSerializer.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Catroid: An on-device visual programming system for Android devices
- * Copyright (C) 2010-2018 The Catrobat Team
+ * Copyright (C) 2010-2021 The Catrobat Team
```
mstoeg, line 3: Please undo the changes on this file as there is just a change of formatting.

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -81,6 +81,10 @@
 	public static final File TEMP_LOOK_FILE =
 			new File(DEFAULT_ROOT_DIRECTORY, "temporary_look_file" + DEFAULT_IMAGE_EXTENSION);
 
+	// Recent Bricks Directory
+	public static final String RECENT_BRICKS_DIRECTORY = "recent_bricks";
```
mstoeg, line 85: This field is never used in your source code. This ticket does not really have to do something with the Backpack - so it would be nice if you actually used this directory instead of interweaving things with the Backpack, which is very confusing.

catroid/src/main/java/org/catrobat/catroid/ui/controller/RecentBrickListManager.java
```diff
@@ -0,0 +1,83 @@
...
+	private void createBackpackDirectories() {
+		DEFAULT_ROOT_DIRECTORY.mkdir();
+	}
+
+	public List<Brick> getRecentBricks() {
+		return recentBricksHolder.getRecentBricks();
+	}
+
+	public void addBrick(Brick brick) {
+		if (!recentBricksHolder.contains(brick)) {
```
mstoeg, line 67: Please rethink the logic of this method. When a brick is already contained in the `recentBricksHolder` then remove it and add it again, such that it is displayed on top again - as the most recently used brick should always be on top.

catroid/src/main/java/org/catrobat/catroid/common/RecentBricksHolder.java
```diff
@@ -0,0 +1,67 @@
...
+	public boolean contains(Brick brick) {
+		for (Brick b : recentBricks) {
+			if (b.getClass().getName().equals(brick.getClass().getName())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public void setRecentBricks(List<Brick> recentBricks) {
```
mstoeg, line 52: This method is actually not used so please remove it if it is not necessary.

mstoeg (changes requested):

catroid/src/main/res/values/styles.xml
devsebastian: but if I'm not mistaken, all the other selection brick backgrounds like control, look etc. have just a single solution file right? (in drawable-xhdpi) wouldn't it be redundant to have more for just the recent bricks?

catroid/src/main/java/org/catrobat/catroid/io/BackpackSerializer.java
devsebastian: Actually to reduce code redundancy I had modified BackPackSerializer so that I could use the same class to serialize both the backpack as well as the Recent Bricks, to achieve this I modified the BackpackSerializer constructor to accept a "File" object which will be used to save the serialized file for either use case. In this way, the same serializer could be used to save RecentBricksFile for Recent Bricks and Backpack File for Backpack. I had created the RECENT_BRICKS_DIRECTORY name for the same purpose but forgot to use it in code, I will do the necessary and push the code again? although I could make a different serializer for the Recent Bricks but I don't think that would make much sense, considering it will contain the same code?

catroid/src/main/res/values/styles.xml
mstoeg: I guess you are right. As there are no actual instances of the recently used type (I mean actual bricks with that grayish background) a single solution file should suffice. I am sorry.

catroid/src/main/java/org/catrobat/catroid/io/BackpackSerializer.java
mstoeg: Ok that is fine then! No need to produce duplicate code.

mstoeg (changes requested):
Looks good now! However, there is still a problem if I select a brick from the "recently used" category. Because it then appears multiple times in the list. (see screenshot)
recently used

catroid/src/main/java/org/catrobat/catroid/ui/controller/RecentBrickListManager.java
```diff
@@ -0,0 +1,83 @@
...
+		recentBrickListDirectory.mkdir();
+	}
+
+	public List<Brick> getRecentBricks() {
+		return recentBricksHolder.getRecentBricks();
+	}
+
+	public void addBrick(Brick brick) {
+		int index = recentBricksHolder.find(brick);
+		if (index >= 0) {
```
devsebastian, line 68: > Looks good now! However, there is still a problem if I select a brick from the "recently used" category. Because it then appears multiple times in the list. (see screenshot)
> recently used

Sorry, had forgot to add an '=' sign here, after '>' so it was adding a duplicate of the brick if the matched brick index in the recent bricks list was '0', stupid mistake 🙈 !

catroid/src/main/java/org/catrobat/catroid/common/RecentBricksHolder.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	public List<Brick> getRecentBricks() {
+		return recentBricks;
+	}
+
+	public int find(Brick brick) {
+		for (int i = 0; i < recentBricks.size(); i++) {
+			Brick b = recentBricks.get(i);
+			if (b.getClass().getName().equals(brick.getClass().getName())) {
```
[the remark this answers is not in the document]
devsebastian: Noted! thanks @lyndseyBeil :)

mstoeg (changes requested):
Thank you, the bug with the double occurences is fixed. However, there is a problem with UserDefinedBricks. If I define a UserDefinedBrick and add it to the script view the app crashes, which is due to this PR - I guess there are some problems with storing the UserDefinedBricks in the "RecentlyUsed" category. Handling UserDefinedBricks is rather problematic in this case, because those bricks could also be deleted and thus should be deleted from the recently used category as well. If this is too hard to implement I would suggest to not add UserDefinedBricks to the "recently used" category in the first place.

catroid/src/main/java/org/catrobat/catroid/common/RecentBricksHolder.java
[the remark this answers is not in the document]
lyndseyBeil: :)

mstoeg (commented):
In a first approach I would suggest to exclude the UserDefinedBricks from the "Recently used" category, as they are also different in different Sprites and Projects.

devsebastian:
I was thinking of the same approach! maybe in a future ticket, we could add the user-defined Brick as well and consider all its intricacies! nevertheless, I've ignored user-defined bricks in this ticket!

mstoeg (changes requested):
UserDefinedBricks are not considered any more - thank you! I am sorry to complain again, but there is also an issue with bricks that should only be available if the current sprite is not the BackgroundSprite and vice versa. Have a look at the `CategoryBricksFactory.java` class to see to which bricks this applies. The problem with the recently used brick category would then be that it would become possible to add bricks to the background sprite although they are not intended to work for the background sprite. Maybe in a first approach again exclude these specific bricks. Furthermore, it would be really nice if you could add a testcase to check the bricks that were excluded, which could be really helpful when implementing those special cases in another ticket.

devsebastian:
I think excluding the bricks all together might be poor for UX? I made the changes in the PR such that the bricks for Non Background Sprites won't be visible for background sprites in the recent Bricks Category! Let me know if you have a differing view, will make changes accordingly! I have added the changes in a new commit for easier review and rebase, will squash it once the commits are confirmed and reviewed! I've also added a test to check the brick visibility in the recent bricks category for background sprite and actors. And thank you so much for your patient assessment and review:)

catroid/src/main/java/org/catrobat/catroid/ui/controller/RecentBrickListManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+public final class RecentBrickListManager {
+
+	final Class[] nonBackgroundSpriteClasses = {WhenBounceOffBrick.class, IfOnEdgeBounceBrick.class,
+			GoNStepsBackBrick.class, ComeToFrontBrick.class, WhenBounceOffBrick.class,
+			SetLookBrick.class, SetLookByIndexBrick.class, SayBubbleBrick.class,
+			SayForBubbleBrick.class, ThinkBubbleBrick.class, ThinkForBubbleBrick.class,
+			LookRequestBrick.class, PenDownBrick.class, PenUpBrick.class,
+			SetPenSizeBrick.class, SetPenColorBrick.class, StampBrick.class,
+			LookRequestBrick.class, LookRequestBrick.class};
```
mstoeg, line 61: I think you have added the `LookRequestBrick.class` three times.

mstoeg (commented):
Thank you for fixing the problem with the background bricks immediately - works like a charm. Just change the minor thing I mentionend and squash the commits. Then I will approve this PR.

catroid/src/androidTest/java/org/catrobat/catroid/test/content/bricks/RecentBricksTest.java
```diff
@@ -0,0 +1,135 @@
...
+		onView(withText(R.string.category_recently_used)).perform(click());
+		onBrickAtPosition(1).checkShowsText(R.string.brick_forever);
+	}
+
+	@Test
+	public void testCheckNonBackgroundBricksAreHiddenForBackgroundSprites() {
+		onView(withText(spriteName)).perform(click());
+
+		onView(withId(R.id.button_add)).perform(click());
+		onView(withText(R.string.category_pen)).perform(scrollTo(), click());
```
mstoeg, line 91: Unfortunately, I have seen that this test does not pass on Jenkins. The reason for this is that the emulators are quit slow. You could try to invoke `onView(withText(R.string.category_pen)).perform(waitFor(isDisplayed(), waitThreshold));` with a certain threshold, e.g. 5000ms to be sure that the view is actually displayed before clicking it.

devsebastian: Thank you so much! was gawking over this error since morning 😂 will try this and see if it works on jenkins!

devsebastian: Done, seems like the problem has been solved, the error is no longer visible:)

mstoeg (approved):

catroid/src/main/res/values/strings.xml
```diff
@@ -2079,5 +2079,6 @@ needs read and write access to it. You can always change permissions through you
     <string name="serif_font" translatable="false">serif</string>
     <string name="edit">Edit</string>
     <string name="data_value">Value</string>
+    <string name="category_recently_used">Recently Used</string>
```
wslany, line 2082:
```suggestion
    <string name="category_recently_used">Recently used</string>
```

wslany (approved):

wslany:
Thansk!

PR #4011, opened 2021-03-11
CATROID-895 Embedded survey campaign

dandrawes:
https://jira.catrob.at/browse/CATROID-895
- agreed with Stefan to only save the last two URLs independently of the language

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/common/Survey.java
+++ b/catroid/src/main/java/org/catrobat/catroid/common/Survey.java
@@ -40,24 +40,24 @@ import static org.catrobat.catroid.common.SharedPreferenceKeys.LAST_USED_DATE_KE
 import static org.catrobat.catroid.common.SharedPreferenceKeys.SHOW_SURVEY_KEY;
 import static org.catrobat.catroid.common.SharedPreferenceKeys.SURVEY_URL1_HASH_KEY;
 import static org.catrobat.catroid.common.SharedPreferenceKeys.SURVEY_URL2_HASH_KEY;
-import static org.catrobat.catroid.common.SharedPreferenceKeys.TIME_SPENT_IN_APP_KEY;
+import static org.catrobat.catroid.common.SharedPreferenceKeys.TIME_SPENT_IN_APP_IN_SECONDS_KEY;
 
 public class Survey implements GetSurveyTask.SurveyResponseListener {
 	@VisibleForTesting
-	public static final int MINIMUM_TIME_SPENT_IN_APP = 3600;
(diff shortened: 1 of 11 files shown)


mstoeg (approved):
Well done!

wslany (changes requested):
Please add tests.

mstoeg (approved):

wslany (approved):

catroid/src/main/java/org/catrobat/catroid/common/Survey.java
```diff
@@ -0,0 +1,197 @@
...
+
+	public void showSurvey(Context context) {
+		if (fulfilledSurveyRequirements) {
+			long oldDate = PreferenceManager.getDefaultSharedPreferences(context)
+					.getLong(LAST_USED_DATE_KEY, new Date(System.currentTimeMillis()).getTime());
+
+			boolean showSurvey = PreferenceManager.getDefaultSharedPreferences(context)
+					.getBoolean(SHOW_SURVEY_KEY, false);
+
+			if (!DateUtils.isToday(oldDate) || (showSurvey && getUploadFlag())) {
```
wslany, line 131: I am not sure about this condition. To show the survey, how come `DateUtils.isToday(oldDate)` is allowed to be true, since the conditions are joined by a `||`?

wslany (changes requested):
Thank you!

Besides the fixed typo (spend => spent) in catroid/src/test/java/org/catrobat/catroid/test/web/EmbeddedSurveyTest.kt, which I directly committed, there are a few inlined comments. 

Sorry for the nitpicking, but it is confusing that some temporal values are in milliseconds and others in seconds without saying so in the name. For consistency, please add `InSeconds` or `InMilliseconds` or `_IN_SECONDS` or `_IN_MILLISECONDS` to differentiate clearly the units used for each. For instance, in Survey.java `totalTimeSpentInApp` is in seconds, but `totalTimeSpentIntage` is in milliseconds. 

There are some tests that would be good: 
* What about not showing a survey twice? 
* What about not showing the same first survey again after showing another survey in between?
* What about if there's is no survey for a language? 
* Is the correct language survey shown for the user's language set inside the app? 
* What if the API does not reply correctly (maybe it is turned off for maintenance and thus is not available, or it does not run anymore at all)?
* Are the timeSpentInApp values from catroid/src/main/java/org/catrobat/catroid/common/Survey.java correctly computed in all cases? Tests are needed to document the intended behavior, since it is quite complex and actually it is not time spent in the app, but time spent in the IDE.
* Can you test whether the right values are set in the shared preferences?

wslany (approved):

wslany:
Thanks!!

PR #4013, opened 2021-03-12
CATROID-962 Catrobat Language Test for "wait" brick

frit96:
https://jira.catrob.at/browse/CATROID-962


petbuer:
Thanks. 👍 
However, the problem is that the wait brick is really hard to test. First, if you wanted to test the correct timing, you would need a separate time source to validate the result. Second, synchronization is not ensured, since there's no guarantee in which order the "When scene starts" scripts are executed, and which bricks are executed first in an interlaced, multithreaded order. In general, waits should be avoided in language tests, since they can cause flakyness on slow emulators (e.g. Jenkins).

petbuer (approved):

wslany (changes requested):
Thanks, but unfortunately Catrobat is not clever enough to understand the transitivity of numerical comparisons. So `0.8 < timer < 1.2` actually would be also true if timer would be 100, since the `true` resulting from `0.8 < timer` is then cast to the number 1 because there's a numerical comparison of `true < 1.2`, and 1 is smaller than 1.2, so altogether it is `true`. Similarly, `0.8 < 10000 < 1.2` is also evaluated to true in Catrobat even though this clearly is mathematically incorrect (but this is also the incorrect behavior in other programming languages).

Thus, please replace the formula for this test by  `0.8 < timer and timer < 1.2` to test as you intended it.

However, I have seen similar formulas written by our young users, so I think we need to somehow handle that in the future, and also it would be cool. It makes much more sense to handle it like it is normally handled in math, i.e., like you wrote it, than how Catrobat interpreted it. What to do? We could add a kind of wizard that detects "transitively chained" logical formulas and offers to rewrite them with "and"s in between. What do you think?

petbuer (approved):

wslany (approved):

PR #4017, opened 2021-03-13
CATROID-958 Textblock detection x and y sensors return the wrong value

ritiksp2411:
Changed the value of the sensor variable X to Y and Y to -X.

https://jira.catrob.at/browse/CATROID-958

I tested the work on the Honor 9i android device.
 
Signed-off-by: ritiksp2411 <ritik.rjt@gmail.com>

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/TextBlockUtil.java
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/TextBlockUtil.java
@@ -1,6 +1,6 @@
 /*
  * Catroid: An on-device visual programming system for Android devices
- * Copyright (C) 2010-2020 The Catrobat Team
+ * Copyright (C) 2010-2021 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
@@ -102,13 +102,13 @@ public final class TextBlockUtil {
 				float aspectRatio = (float) imageWidth / imageHeight;
 
 				if (ProjectManager.getInstance().isCurrentProjectLandscapeMode()) {
-					float relativeY = textBlockBounds.exactCenterX() / imageWidth;
-					relativeY = invertAxis ? relativeY : 1 - relativeY;
+					float relativeX = textBlockBounds.exactCenterX() / imageWidth;
+					relativeX = invertAxis ? 1 - relativeX : relativeX;
 					return coordinatesFromRelativePosition(
+							relativeX,
+							SCREEN_WIDTH,
 							1 - textBlockBounds.exactCenterY() / imageHeight,
-							SCREEN_WIDTH / aspectRatio,
-							relativeY,
-							(float) SCREEN_WIDTH
+							(float) SCREEN_WIDTH / aspectRatio
 					);
 				} else {
 					float relativeX = textBlockBounds.exactCenterX() / imageHeight;


qlin93:
Thanks for making the changes. 
I tested your changes with this test:
testTextblockDetection
and the returned coordinates are not on the screen. The X should be exactly where the text is that is being detected. 
Have a look at FaceAndTextDetector.kt, where the coordinate translation is implemented correctly. You also need to check how it works when using the front camera. Please make the necessary changes and amend your commit.

ritiksp2411:
Thanks for the review, 
I have one doubt like we have a bug in landscape mode right, but the screenshot is in Portrait mode? I didn't get that part.

ritiksp2411:
In the file FaceAndTextDetector.kt, we have Y as X if the camera is facing front, and Y as 1-X if not.
Along with that if we have oldAPI, then X is Y otherwise, X is 1-Y. So do I need to check oldAPI too right?

qlin93:
Thanks. Creating a project in landscape mode does not change the orientation of the Pocket Code app itself only of the project when it is run.

Yes please also take the oldAPI into account.  
Have you installed the app via Android Studio to test the changes you made? 
This is important because the solution for TextblockUtil might not be exactly the same as for the FaceAndTextDetector, 
as the x and y might be different. (It could have different origins)

ritiksp2411:
thanks, I check the changes in my phone but couldn't recognize that properly, now I will make the changes and update accordingly.

ritiksp2411:
When I tested with your test, I have noticed that returned coordinates of the X are sometimes on the screen and sometimes off the screen. Any specific reason for that?

ritiksp2411:
WhatsApp Image 2021-03-25 at 7 45 36 PM
It is showing the X properly now.

qlin93:
Looks good. Please push your changes on this branch by amending your commit.

ritiksp2411:
Thanks for helping, I updated the PR.

qlin93 (changes requested):
Hi sorry for the delay. Unfortunately it still does not work correctly. When a text is detected and you move your camera around the x and y coordinates or in the example testcase the X should follow the text. With the change you made the problem still persists, the movement along the x-axis is shown on the y-axis and vice versa.

ritiksp2411:
okay, yeah so the shift of X was shown in Y, but bow I Fixed it such that when we move our camera left side that means Text in the image will go right and the same will be shown by X on the screen. Similarly, if we move the camera upwards, the text in the image will go down and the same will be shown by X on the screen.

catroid/src/main/java/org/catrobat/catroid/utils/TextBlockUtil.java
```diff
@@ -102,10 +102,10 @@ public static Point getCenterCoordinates(int index) {
 				float aspectRatio = (float) imageWidth / imageHeight;
 
 				if (ProjectManager.getInstance().isCurrentProjectLandscapeMode()) {
-					float relativeY = textBlockBounds.exactCenterX() / imageWidth;
-					relativeY = invertAxis ? relativeY : 1 - relativeY;
+					float relativeY = textBlockBounds.exactCenterY() / imageWidth;
+					relativeY = invertAxis ? 1 - relativeY : relativeY;
```
qlin93, line 106:
```suggestion
					float relativeX = textBlockBounds.exactCenterX() / imageWidth;
					relativeX = invertAxis ? 1 - relativeX : relativeX;
```

catroid/src/main/java/org/catrobat/catroid/utils/TextBlockUtil.java
```diff
@@ -102,10 +102,10 @@ public static Point getCenterCoordinates(int index) {
...
 					return coordinatesFromRelativePosition(
-							1 - textBlockBounds.exactCenterY() / imageHeight,
+							1 - textBlockBounds.exactCenterX() / imageHeight,
 							SCREEN_WIDTH / aspectRatio,
 							relativeY,
 							(float) SCREEN_WIDTH
```
qlin93, line 111:
```suggestion
							relativeX,
							SCREEN_WIDTH,
							1 - textBlockBounds.exactCenterY() / imageHeight,
							(float) SCREEN_WIDTH / aspectRatio
```

qlin93 (changes requested):
Thanks, that is better but the movement on the axes are still inverted and the scaling is off. I looked into the issue more closely and it turned out to be a little more complicated to fix than stated in the ticket. 

Please apply the suggested changes and test whether they work on your device before updating your commit.

ritiksp2411:
Thanks for the suggestion, I saw that it exactly behaves the same as my previous PR, like axis are behaving the same way. I am unable to justify scaling there but ya, the axis are working in the same manner.

qlin93:
I tested it again and you are right the axes are behaving the same way, but the scaling is completely off. I cannot approve something that does not work correctly and you should always test and make sure that the behaviour is exactly as it should be.
In this case the returned coordinates should be exactly in the center of the text.  

Here are some screenshots I took while testing. 

With the existing changes: 
https://drive.google.com/file/d/1OMjYDEh03vS4Rflgldxu5ruN8qzsN7uK/view?usp=sharing
https://drive.google.com/file/d/12aaEzku-waJbM_BHLTA3g5oBf_O4pKG9/view?usp=sharing

With suggested changes: 
https://drive.google.com/file/d/1SPbV4IUhP4bC6CRjCtLW-iRux71CgdN_/view?usp=sharing
https://drive.google.com/file/d/1fu0pbYGCF2lXVn19H2psCJqQpdOFv_eh/view?usp=sharing

ritiksp2411:
Thanks for the insight and for helping me out. I will change the code as suggested and update the PR soon.

catroid/src/main/java/org/catrobat/catroid/utils/TextBlockUtil.java
```diff
@@ -102,13 +102,13 @@ public static Point getCenterCoordinates(int index) {
...
+					float relativeX = textBlockBounds.exactCenterX() / imageWidth;
+					relativeX = invertAxis ? 1 - relativeX : relativeX;
 					return coordinatesFromRelativePosition(
-							1 - textBlockBounds.exactCenterY() / imageHeight,
-							SCREEN_WIDTH / aspectRatio,
-							relativeY,
-							(float) SCREEN_WIDTH
+							relativeX,
+							SCREEN_WIDTH,
+							1 - textBlockBounds.exactCenterY() / imageWidth,
```
qlin93, line 110:
```suggestion
							1 - textBlockBounds.exactCenterY() / imageHeight,
```

qlin93 (changes requested):
Thanks just a small error, please change imageWidth to imageHeight in line 110.

qlin93 (approved):
Thanks!

ritiksp2411:
Thanks for helping,
Learned a lot from the issue

wslany:
Thanks a lot!!

PR #4022, opened 2021-03-16
CATROID-919 Delete exif from imported looks

Tamdav:
Used ExifInterface to remove Exifdata

Ticket: https://jira.catrob.at/browse/CATROID-919

*Please enter a short description of your pull request and add a reference to the Jira ticket.*


catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -137,4 +147,25 @@ public void addToRequiresPermissionTaskList(RequiresPermissionTask task) {
...
+	public static void removeExifFromPNG(File file) {
+		try {
+			FileInputStream fileInputStream = new FileInputStream(file.getAbsolutePath());
+			BufferedInputStream bufferedInputStream = new BufferedInputStream(fileInputStream);
+			byte[] bMapArray = new byte[bufferedInputStream.available()];
+			bufferedInputStream.read(bMapArray);
+			Bitmap bitmap = BitmapFactory.decodeByteArray(bMapArray, 0, bMapArray.length);
+
+			ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+			bitmap.compress(Bitmap.CompressFormat.JPEG, 100, byteArrayOutputStream);
```
mstoeg, line 160: Unfortunately, compressing all images to the `JPEG` format is not a good idea. For example if the chosen image has `png` format we loose all the transparency information. You can easily try this out with the "normal" actors from the Media library as they all have transparent background.
Moreover if I use image file formats, which are not supported by Catroid, e.g. `.svg` instead of showing an error message the app crashes. For reproduction see the attached screencast (the file I selected the second time has `.svg` format)

https://user-images.githubusercontent.com/60399652/111780212-603dfd00-88b7-11eb-8c57-0b9a7b376e91.mp4

mstoeg (changes requested):

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -137,4 +149,23 @@ public void addToRequiresPermissionTaskList(RequiresPermissionTask task) {
 	public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
 		permissionRequestActivityExtension.onRequestPermissionsResult(this, requestCode, permissions, grantResults);
 	}
+
+	public static void removeExifData() {
+		File oldFile = new File(CAMERA_CACHE_DIR, TMP_IMAGE_FILE_NAME + ".jpg");
+		File file = new File(CAMERA_CACHE_DIR, TMP_IMAGE_FILE_NAME + "___temp.jpg");
```
mstoeg, line 155: Using those hardcoded values and file types here is problematic. E.g. what if the original type of the image is not of type ".jpg" or if the file is named other than `TMP_IMAGE_FILE_NAME`, wich is for example the case if I select an image from the storage. Therefore I would suggest to directly pass the right directory and fileName as a parameter, e.g. something like:
 `	public static void removeExifData(File directory, String fileName) {
		File oldFile = new File(directory, fileName);
		File file = new File(directory, "temp_" + fileName);`
You can get those parameters if you directly call the `removeExifData()` function after the actual file is retrieved from the URI `File file = StorageOperations.copyUriToDir(getContentResolver(), uri, imageDirectory, lookFileName);` in ProjectActivity and SpriteActivity.
Moreover, I would suggest to move this method to another class as it is static anyways and has nothing to do with the `BaseActivity`. I would suggest to add it to the `Utils` class where already other file operation functions are implemented.

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -137,4 +149,23 @@ public void addToRequiresPermissionTaskList(RequiresPermissionTask task) {
...
+			new ExifRewriter().removeExifMetadata(inputStream, outputStream);
+
+			if (!oldFile.delete()) {
+				Log.e(TAG, "removeExifData: Failed to delete old file");
+			}
+			if (!file.renameTo(new File(CAMERA_CACHE_DIR, TMP_IMAGE_FILE_NAME + ".jpg"))) {
+				Log.e(TAG, "removeExifData: Failed to rename new file");
+			}
+		} catch (IOException | ImageWriteException | ImageReadException e) {
+			Log.e(TAG, "removeExifData: Failed to remove exif data");
```
mstoeg, line 168: I think you have to remove the temp file here as well, otherwise if there is an Exception you generated two files and did not remove one of them.

catroid/src/main/java/org/catrobat/catroid/ui/BaseActivity.java
```diff
@@ -137,4 +149,23 @@ public void addToRequiresPermissionTaskList(RequiresPermissionTask task) {
...
+		File file = new File(CAMERA_CACHE_DIR, TMP_IMAGE_FILE_NAME + "___temp.jpg");
+		try (FileInputStream inputStream = new FileInputStream(oldFile);
+				FileOutputStream outputStream = new FileOutputStream(file)) {
+
+			new ExifRewriter().removeExifMetadata(inputStream, outputStream);
+
+			if (!oldFile.delete()) {
+				Log.e(TAG, "removeExifData: Failed to delete old file");
+			}
+			if (!file.renameTo(new File(CAMERA_CACHE_DIR, TMP_IMAGE_FILE_NAME + ".jpg"))) {
```
mstoeg, line 164: You can simply use `file.renameTo(oldFile)` here as you already defined this file.

mstoeg (changes requested):
Please have a look at my remarks. If you feel overwhelmed just contact me on Slack - we can also finish this ticket in Pair Programming if you like.

ndrnour:
**Please let me have the last review before merging this.** 
by the way, your approach (loading the image into a bitmap and then compress it) has many downsides. 
remember is the task to remove the metadata, and **NOT** to compress the image again.

mstoeg (commented):
@ndrnour what would be your suggestion to solve this?

ndrnour:
> @ndrnour what would be your suggestion to solve this?

* for JPEG 
using `ExifInterface` [link](https://developer.android.com/reference/android/media/ExifInterface) 
short example would be:
`val exif = ExifInterface(pathToImageFile)`
`exif.setAttribute(ExifInterface.TAG_GPS_LATITUDE, "")`
`exif.setAttribute(ExifInterface.TAG_GPS_LONGITUDE, "")`
`exif.setAttribute(ExifInterface.TAG_GPS_ALTITUDE_REF, "")`
`........`
`exif.saveAttributes()`

* for PNG
it depends, reading this [discussion](https://stackoverflow.com/questions/9542359/does-png-contain-exif-data-like-jpg) could help you.

mstoeg (approved):
Looks good! Requested changes from Nur were also implemented properly.

wslany:
Thanks!!

PR #4045, opened 2021-03-25
CATROID-759 "number of looks" object property

frit96:
https://jira.catrob.at/browse/CATROID-759
"number of looks" object property
adds a property with which you can check how many looks have already been created


catroid/src/main/java/org/catrobat/catroid/formulaeditor/Sensors.java
```diff
@@ -41,7 +41,8 @@
 	DRONE_RECORDING, DRONE_NUM_FRAMES, COLLIDES_WITH_EDGE(true), COLLIDES_WITH_FINGER(true), OBJECT_X_VELOCITY(true),
 	OBJECT_Y_VELOCITY(true), OBJECT_ANGULAR_VELOCITY(true),
 	LAST_FINGER_INDEX, FINGER_X, FINGER_Y, FINGER_TOUCHED, NUMBER_CURRENT_TOUCHES, OBJECT_LOOK_NUMBER(true), OBJECT_LOOK_NAME(true),
```
petbuer, line 43:
```suggestion
	LAST_FINGER_INDEX, FINGER_X, FINGER_Y, FINGER_TOUCHED, NUMBER_CURRENT_TOUCHES, 
```

catroid/src/main/java/org/catrobat/catroid/formulaeditor/Sensors.java
```diff
@@ -41,7 +41,8 @@
...
-	OBJECT_BACKGROUND_NUMBER(true), OBJECT_BACKGROUND_NAME(true),
+	OBJECT_NUMBER_OF_LOOKS(true), OBJECT_BACKGROUND_NUMBER(true), OBJECT_BACKGROUND_NAME(true),
+	OBJECT_NUMBER_OF_BACKGROUNDS(true),
```
petbuer, line 45:
```suggestion
	OBJECT_LOOK_NUMBER(true), OBJECT_LOOK_NAME(true), OBJECT_NUMBER_OF_LOOKS(true), 
	OBJECT_BACKGROUND_NUMBER(true), OBJECT_BACKGROUND_NAME(true), OBJECT_NUMBER_OF_BACKGROUNDS(true),
```

catroid/src/main/java/org/catrobat/catroid/formulaeditor/Sensors.java
petbuer: Please put the different sensors in separate lines (see also next suggestion in line 44 and 45).

petbuer (changes requested):
Thanks!
Please update the copyright year at the beginning of all the files you edited to 2021.
The language test looks good, but you could leave out the "Delete background" brick at the end. Additionally, a second language test would be good, with an object other than the background.

petbuer (approved):

wslany (changes requested):
Thanks! 

Please increase the Catrobat language version number whenever anything that changes the xml structure is introduced. 

Also, please include a CLT test case for an object that does not have any looks. Note that you do not have to use intermediate variables, and that you can use several asserts in one test.

petbuer (approved):

wslany (changes requested):
Sorry for being late. 🙏 Please move the CLT from the "data" folder to the "formula" folder, and please use the "expected" parameter field like 
image

frit96:
Sorry Wolfgang but I don't quite get what you want me to do with the parameter filed besides expected.
Do you want me to exchange the two fields besides assert and expected so that the 1 is besides assert?

wslany (changes requested):
I meant you could compare to the expected value of "number of looks" instead of to true in each assert. In case of a failure, this gives the developer a more meaningful failure messages. Remember, these tests will be reused for Catty and possible future implementation. It is also clearer for maintenance, since future developers need one less cognitive step in each assert brick to understand the test. That is also why the location should be changed from `catroid/src/androidTest/assets/catrobatTests/data` (where CLTs regarding bricks from the Data category belong) to `catroid/src/androidTest/assets/catrobatTests/formula` (where everything that can happen in the formula editor belong).

I.e., the brick could look like this:
image
instead of 
image

Please use the attached CLT. Please change the extension back to .catrobat --- .catrobat files apparently cannot be uploaded in github comment. It contains above suggestion, is safe from a race condition on very slow emulators, and also tests the following additional cases:
- Reduction of the number of looks.
- If the value is already 0, it cannot be further decreased.
- Looks imported from the web should increase the number of looks by 1, but only during the lifetime of the look.

When I added the last test, I found that `number of looks` does not work as expected when looks are loaded from the web. The value is not increased. So, the calculation is not yet correct, and the tests fail. Please fix this. 
[testNumberOfLooks.zip](https://github.com/Catrobat/Catroid/files/6470526/testNumberOfLooks.zip)

wslany (changes requested):
I found another problem: 
Put the following script from the background into the backpack: 
image
Now unpack it in a normal object and observe that while the name correctly changed in the the "Next look" brick, the `number of backgrounds` has incorrectly not been renamed to `number of looks`:
image

 This can be solved by using a common identifier for the two, i.e., instead of 
```
                      <value>OBJECT_NUMBER_OF_BACKGROUNDS</value>
or
                      <value>OBJECT_NUMBER_OF_LOOKS</value>
```
just use
```
                      <value>OBJECT_NUMBER_OF_LOOKS</value>
```
in all cases.

catroid/src/main/java/org/catrobat/catroid/content/actions/LookRequestAction.kt
```diff
@@ -54,6 +54,7 @@ open class LookRequestAction : WebAction() {
             scope?.sprite?.look?.lookData = this
             collisionInformation?.collisionPolygonCalculationThread?.join()
             file?.delete()
+            webRequest = true
```
petbuer, line 57:
```suggestion
            isWebRequest = true
```

catroid/src/main/java/org/catrobat/catroid/common/LookData.java
```diff
@@ -90,6 +92,14 @@ public void setName(String name) {
 		this.name = name;
 	}
 
+	public boolean getWebRequest() {
+		return isWebRequest;
+	}
+
+	public void setWebRequest(Boolean webRequest) {
+		this.isWebRequest = webRequest;
+	}
```
petbuer, line 101:
```suggestion
	public boolean getIsWebRequest() {
		return isWebRequest;
	}

	public void setIsWebRequest(Boolean isWebRequest) {
		this.isWebRequest = isWebRequest;
	}
```

catroid/src/main/java/org/catrobat/catroid/formulaeditor/common/FormulaElementOperations.kt
```diff
@@ -109,6 +109,15 @@ object FormulaElementOperations {
     @JvmStatic
     fun getLookName(lookData: LookData?) = lookData?.name ?: ""
 
+    @JvmStatic
+    fun getNumberOfLooks(lookData: LookData?, lookDataList: List<LookData>): Int {
+        return if (lookData?.webRequest == true) {
```
petbuer, line 114: " == true" is redundant with the new name of the Getter function.
```suggestion
        return if (lookData?.isWebRequest) {
```

petbuer (changes requested):
Please rename the Getter, Setter and the parameter of the Setter as commented.
Also increase the Language version number.

catroid/src/main/java/org/catrobat/catroid/formulaeditor/common/FormulaElementOperations.kt
frit96: the problem here is that if i leave the == true out it wants a not null assertion and i thought we should avoid them

petbuer (approved):

wslany (approved):

wslany:
Thanks!

PR #4068, opened 2021-04-02
CATROID-850 Insert flatten(*list*) for backwards compatibility

petbuer:
https://jira.catrob.at/browse/CATROID-850

Insert flatten(*list*) around all userlists when opening old projects where "flatten" was not yet available.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
@@ -167,9 +167,13 @@ public final class ProjectManager {
 		if (project.getCatrobatLanguageVersion() <= 0.9999995) {
 			updateBackgroundIndexTo9999995(project);
 		}
+		if (project.getCatrobatLanguageVersion() < 999.9 && !BuildConfig.FEATURE_LIST_AS_BASIC_DATATYPE) {
+			ProjectManager.flattenAllLists(project);
+		}
 		if (project.getCatrobatLanguageVersion() <= 1.03) {
 			ProjectManager.updateDirectionProperty(project);
 		}
+
 		project.setCatrobatLanguageVersion(CURRENT_CATROBAT_LANGUAGE_VERSION);
 
 		localizeBackgroundSprites(project, context.getString(R.string.background));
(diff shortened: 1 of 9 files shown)


qlin93 (approved):

wslany (changes requested):
Thanks! However, I got the following error after downloading a project with the following formula (otherwise the default project):
image

image

qlin93 (approved):

catroid/src/main/java/org/catrobat/catroid/ProjectManager.java
```diff
@@ -167,9 +167,13 @@ public void loadProject(File projectDir, Context context) throws ProjectExceptio
 		if (project.getCatrobatLanguageVersion() <= 0.9999995) {
 			updateBackgroundIndexTo9999995(project);
 		}
+		if (project.getCatrobatLanguageVersion() < 1.01 && BuildConfig.DEBUG) {
```
wslany, line 170:
```suggestion
		if (project.getCatrobatLanguageVersion() < 999.9 && BuildConfig.DEBUG) {
		// Delete the DEBUG flag and replace 999.9 by the newest Catrobat language version number 
		// once https://jira.catrob.at/browse/CATROID-851 has been implemented and merged.
```
The version number in the condition here will have to be set when the transition to lists as primitive data structures is made. 

Note to myself: The current Google Play version supports CatrobatLanguageVersion 1.03, so I tested it with an earlier version, and the insertion worked fine. Test project to check for insertion was https://share.catrob.at/app/project/c4b5c4ed-c616-11eb-abc2-005056a36f47 (this project is only visible with debug builds).

petbuer: The check "< 999.9" would always be true and the way it is implemented now, flattenAllLists would be called at every LoadProject. Then Flatten would be inserted again and again.

The version check prevents this at the moment. Flatten is added once when an old project is loaded. Then the version number of the project is updated and next time Flatten is not executed again.

Perhaps it would be better to set the version number for comparison to 0 and NOT execute flattenAllLists at the moment and only include this functionality when we know the language version number of "lists as primitive data structure"?

wslany: Sorry for the late reply. Inserting the flatten function again and again is no problem, as long as this is changed to the actual version number required before the release with the full feature with the "lists as primitive data structure", since we use the debug versions for testing only. Since it is for testing only, inserting again and again is no problem. If we keep 1.01 or 0 here, we cannot test the flatten functionality with the debug version, since we cannot easily create projects with the old version number or even earlier ones. And this line will anyway have to be changed once the feature can be released.

wslany (changes requested):
There was one bug that reinserted the flatten list each time inside a flatten function when a corresponding project was reopened. I suggested one way to solve this. Tests were also missing for this and a few other cases. Note that the code for the tests is just a stub that still needs to be completed.

Please also add the FLATTEN keyword in https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/formulaeditor/Functions.java next to the other list functions LIST_ITEM, CONTAINS, INDEX_OF_ITEM, NUMBER_OF_ITEMS --- this was missed in https://jira.catrob.at/browse/CATROID-849

Please also move the flatten function strings up from lines https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/res/values/strings.xml#L2092-L2093 to place them directly below to the other list functions at https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/res/values/strings.xml#L1807-L1819

wslany (approved):

wslany:
Thanks!

PR #4123, opened 2021-04-18
CATROID-967 Convert brick pngs to xml

Tamdav:
- Create an XML file for each brick and changed the bricks in the project.
- It is not possible to create just one brick file per size and change the color of one path dynamically. So we need one file per color per size.
- Also had to make small changes to some brick files.
- As it is not possible to define a textarea in xml vector files - like it was the case with 9-Patch images - the padding on the left was added manually within the BrickLayout (and changes dynamically with the screen width).

https://jira.catrob.at/browse/CATROID-967


mstoeg (approved):
Looks good - reviewed in PP.

wslany (commented):
Thanks, looks very nice. One question though: why do some bricks need more space than others? See, e.g., 
image

Tamdav:
Each brick is assigned a style. For example, when creating the brick "Vibrate for 1 second", the style 'small' was assigned. The other two in your example have been assigned the style 'medium' and this style is by default larger than the style 'small'. But I can change these bricks to the smaller style if you like.

wslany:
> Each brick is assigned a style. For example, when creating the brick "Vibrate for 1 second", the style 'small' was assigned. The other two in your example have been assigned the style 'medium' and this style is by default larger than the style 'small'. But I can change these bricks to the smaller style if you like.

Please make them all small by default. I wonder why we did that --- probably it was not a conscious decision. 

Note: There may be some accessibility setting where the spacings and fonts are larger, but I am not sure whether that influences above size type or not. 

One more thing is that the overview of categories seems to need more space per category. Here how it was before:
image

With this PR, it now looks like this:
image

I would say that making that spread out is not useful --- actually, the less vertical space needed, the less scrolling is needed for the user. Can we make that more compact without loosing the nice design?

wslany:
Thanks!

PR #4141, opened 2021-04-22
CATROID-853 fix "Set rotation style" in physics bug

gPathpp:
https://jira.catrob.at/browse/CATROID-853

- refactor PhyicsLook
- fixed rotation for Left/Right and no Rotation at all Type. (For L/R rotation can only be changed via setRotation Brick etc..)
  for Gravitiy/Motion


wslany (changes requested):
Thanks. The CLTs seem prone to flakiness: Executing testFixedLeftRightRotationWithGravity.catrobat yields a position y value of -899.4305... on a Samsung Galaxy Note 20 Ultra, which barely is above -900. If there would be a slight delay, for whatever reason, e.g., a garbage collection, between the last "If one edge bounce" and the assert, the test would fail. Same also for the other test.

Also, the tests do not sufficiently test that the two rotation styles work as expected for physics objects. The tests for the two cases are exactly the same, so there's no way to ensure that one of the rotation styles is not implemented like the other: If only the "don't rotate" style would work correctly, we would not know that the "left-right only" doesn't work with just these tests. Having said that, I checked the behavior, and indeed the mirroring is not working correctly, _nor_ is the bouncing behavior working correctly, since an object with the left-right-only does not correctly mirror its direction according to its vertical axis when touching the walls, nor does it bounce from the walls in the same way as the all-around type does. Instead, it somehow sticks to the walls without bouncing.

Please remember that the CLTs should be able to serve as a specification by example for future implementations and for Catty (which does not yet have a physics engine).

gPathpp:
As discussed,  this PR has to wait until the separation of look-direction and motion-direction is completed.

PR #4173, opened 2021-05-04
CATROID-826 Cut/copy/paste in formula editor

devsebastian:
Ticket: https://jira.catrob.at/browse/CATROID-826

Cut/copy/paste for current session in formula editor

### DIFF
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementBrickTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/app/VisualPlacementBrickTest.java
@@ -229,6 +229,7 @@ public class VisualPlacementBrickTest {
 		onFormulaEditor()
 				.check(matches(isDisplayed()));
 		pressBack();
+		pressBack();
 		isBackInScriptFragment();
 	}
 
@@ -236,6 +237,7 @@ public class VisualPlacementBrickTest {
 	public void testNoRecursiveOpeningOfFormulaEditors() {
 		openFormulaEditorFragmentFromEditTextX();
 		openFormulaEditorFragmentFromEditTextX();
+		pressBack();
 		pressBack();
 		isBackInScriptFragment();
 	}
--- a/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/ClipboardTest.java
+++ b/catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/ClipboardTest.java
@@ -1,0 +1,167 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2021 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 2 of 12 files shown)


Walcho1125 (approved):
Good Job!
Looks good to me after the discussed changes

wslany (changes requested):
Thanks a lot, this is very nice. After testing it, I have a request.

I know this contradicts what was requested in the ticket, I think that the floating menu initially would be better placed _below_ (not above) the selected text. Since most formulas are short, the floating menu would then not cover anything, and also be better visible. Compare the current placement:
image
with what I imagine the revised placement would look like:
image
Even for long formulas, it would still only cover part of the keyboard, but that would occur with more experienced users anyway, and the top of the keyboard by then is also well known and therefore not something that the user needs to see immediately. On most phones, the buttons will also be so big that they are still visible even when the floating menu partly covers them.

wslany:
Bugs: 

In the scripts of the animal object of the sample project:
1. Tap on the x parameter of the Glide brick to enter the formula editor.
2. Tap on Copy.
3. Tap on the y parameter field in the Glide brick at the top while staying in the formula editor to switch to the other field in the edit field below the brick.
4. Long press on the whole formula (marked in blue) to make the floating menu appear.
5. Tap on Paste.
6. Observe that the copied formula _incorrectly_ has _not_ been inserted (the 3s should have been replaced by 2s but are not). Instead, the first parameter of the old random function is now selected in blue color. What should have happened is that the copied formula from the x field should have replaced the old y formula, and the focus removed.
7. Since the first parameter -200 of the random function in the y parameter is now selected in blue, long press on it to make the floating menu reappear and tap on Copy. 
8. Tap between the "," and the "2" to place the cursor in front of the 200 of the second parameter of the random function. 
9. Observe that the Undo button now incorrectly becomes bright. Why? Tapping it does nothing besides making the Redo button bright. 
9. Long press and tap on Paste.
10. Observe that the "-200" has been inserted, but that _incorrectly_ there is a small space between the -200 and the 200.
11. Note that if you tap between this space and the 2 to place the cursor there, and then tap the delete button on the virtual keyboard of the formula editor, the space is not deleted, but instead the 0 before it is. The number with the cursor inserted now looks like this: "- 20 |200". 
12. Tap on the back button to leave the formula editor. A toast that the formula is not valid is correctly shown, and the whole formula is marked in red. This is correct, but the space should never have been inserted. 

In the scripts of the animal object of the sample project:
1. Tap on the x parameter of the glide brick to enter the formula editor.
2. Tap on Cut.
3. Observe that the formula has been cut and the edit field is empty, which is as expected.
4. Observe that the Undo arrow in the top bar _incorrectly_ is still greyed out and cannot be used. 
3. Tap on the y parameter field in the brick at the top.
4. Observe that, correctly, a toast message "The formula is not valid" is shown, the focus (blue thin line in brick at top below x field) remains at the x parameter field, and the the input field is still empty. 
5. However, at the top in the brick, the "random ..." formula _incorrectly_ is shown again in the x field. 
6. Undo _incorrectly_ still cannot be tapped. 
7. Also note that the formula editor cannot be left since the formula is not valid (this is expected behavior, but since Undo cannot be tapped, the user has no clear way how to recover).

devsebastian:
So I found a couple of things that resulted in the bugs and have fixed them in the updated PR:
1. Regarding line 6: if the selection was **not** null on pasting any content, I had called the predefined `replaceSelection()` method to replace the selected tokens with the clipboard content, but on further examination, I found the `replaceSelection()` method internally calls the `replaceFunctionButKeepParameters()` method which actually replaces the functions by **not** changing the parameters if the function name (or type) is same! fixed that by not calling the internal `replaceSelection()` method.
2. line 9: The undo button actually became bright on tapping on any other token after cutting/pasting any content because the invalidateOptionsMenu() method was called only in the onTouch() method, fixed that by calling it after every copy/paste operation is complete. (in the previous PR even if the parameters didn't change the token was still replaced - by the predefined `replaceSelection()` method - hence the enabling of undo button even if it seemed nothing was pasted or changed)
3. line 11: the small space between -200 and 200 in "-200, -200 200", unfortunately, comes because the internFormula internally calls the `insertLeftToCurrentToken(tokens)` or `insertRightToCurrentToken(tokens)` depending on whether the cursorTokenPosition is `LEFT` or `RIGHT` ie. depending on whether the cursor is to the left of the blank_space or to the right:- "-200, |200" or "-200,| 200", I don't think there is a simple way to change that behaviour without the change affecting other operations in the internFormula? Please let me know if I'm wrong? I could try and check whether there is a blank space after the cursor before every paste operation and then delete it, but I think that would have unexpected effects in other cases!
4. I believe the second bug is also related to my second point, and has been resolved after the change!

Walcho1125 (approved):
Well done!
Looks good after the discussed changes!

wslany (approved):

wslany:
Thanks!

PR #4197, opened 2021-05-27
CATROID-1074 REFACTOR PlaySoundAction to Kotlin

Rawbat:
https://jira.catrob.at/browse/CATROID-1074


reichli (changes requested):
Looks good!
There are just two things that I think could be improved further IMO:
1) If we refactor Java classes to Kotlin we should refactor the associated tests as well.
2) I would change the `sound` field from `lateinit` back to `SoundInfo?`. If the user does not choose a sound right after placing the brick this will be `NULL`. If the user then starts the program the whole app won't crash but you will still be thrown back to the main menu, which is a different behaviour than before the refactoring.

catroid/src/main/java/org/catrobat/catroid/content/actions/PlaySoundAction.kt
```diff
@@ -20,36 +20,25 @@
...
-
-	public void setSprite(Sprite sprite) {
-		this.sprite = sprite;
-	}
-
-	public void setSound(SoundInfo sound) {
-		this.sound = sound;
-	}
+    override fun update(percent: Float) {
+        if (sprite.soundList.contains(sound) && sound!!.file != null) {
```
reichli, line 36: Can you please use `let` here to perform a null-check on `sound`?

Like this:
`sound?.let {sound -> ... }`

reichli (changes requested):

catroid/src/main/java/org/catrobat/catroid/content/actions/PlaySoundAction.kt
```diff
@@ -20,36 +20,27 @@
...
-	public void setSprite(Sprite sprite) {
-		this.sprite = sprite;
-	}
-
-	public void setSound(SoundInfo sound) {
-		this.sound = sound;
-	}
+    override fun update(percent: Float) {
+        if (sprite.soundList.contains(sound)) {
+            sound?.let {
```
reichli, line 37: Please access the context object inside the `let`-block (either as `it` or by renaming it) instead of the `sound`-object. Since the `let`-block is only executed if `sound` is not null, the context object is also guaranteed to be not null and we can get rid of the `!!`.

Like this:

`sound?.let {
  it.isMidiFile
} // no "it!!.isMidiFile" here`

reichli (changes requested):

reichli (approved):
Looks good, thank you :)

adnaribo:
Thank you

adnaribo (approved):

PR #4216, opened 2021-06-12
CATROID-1129 Fade particle effect in/out brick

devsebastian:
Ticket: https://jira.catrob.at/browse/CATROID-1129

### Introduce Particle Effects in Catrobat
* added the new ParticleEffectsFade Brick which has 2 options to fade in or to fade out.
* the particle effects scale will depend on the size of the look, so the bigger the look bigger the "flame"
* particle default color: red 
* A new particle file will take care of all the default properties of the particle effects so as to not overwhelm the user with many options, some of these properties could now be controlled using custom bricks that will be added later...
* Added a test to check whether the boolean property of the particle effect is switched on or off on the sprite on adding the fade In or fade out brick, please let me know if any further tests are required to be added?


Rawbat:
Please also include entries for this brick in the files BricksHelpUrlTest.java and RtlBrickTest.java

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static androidx.test.espresso.action.ViewActions.click;
+import static androidx.test.espresso.matcher.ViewMatchers.withId;
+
+public class ParticleEffectsTest {
+
+	@Rule
+	public FragmentActivityTestRule<SpriteActivity> baseActivityTestRule = new
+			FragmentActivityTestRule<>(SpriteActivity.class, SpriteActivity.EXTRA_FRAGMENT_POSITION, SpriteActivity.FRAGMENT_SCRIPTS);
+
+	Script script;
```
Rawbat, line 52: Please declare this as private

catroid/src/main/java/org/catrobat/catroid/content/actions/FadeParticleEffectAction.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.catrobat.catroid.content.actions;
+
+import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction;
+
+import org.catrobat.catroid.content.Sprite;
+
+public class FadeParticleEffectAction extends TemporalAction {
+
+	Sprite sprite;
```
Rawbat, line 32: Please declare this as private

Rawbat (changes requested):
Please change the copyright to 2021 in all edited files.

devsebastian:
Done 👍 and thank you so much @Rawbat for helping me fix the failing tests in Jenkins!!

Rawbat (approved):
Great, Thanks!

wslany (changes requested):
Wow, awesome!!

What I observed and hope you can change or fix:
- The particle effects continue to move in the pause menu (when exiting the stage, see screenshot). Instead, it should be frozen (but continue again if the resume button is tapped in the pause menu). Please add a test. 
image
- Please move the new brick in the green Looks category upwards, between "Change color by" and "Clear graphics effects"
- The "Clear graphics effects" shall immediately cancel the particle effect for the object (no fading out; also, if the object has clones, only the one clone or original executing the "Clear graphics effect" brick should cancel its particle effect). Please add a test. 
- If an object has a "Fade in particle effect" brick in any of its scripts, and that brick has been executed at least once successfully in the current session for this project, then, even if it is never executed later (e.g., if it has been commented out, or there is another reason why it is not immediately executed, e.g., because of a "Wait 100 seconds" brick), the particle effect is _immediately_ shown when the scene starts, and all waits or other bricks before it are ignored for the particle effect. Exiting the current project and reentering resets this, and the particle effect is again correctly shown only when the Fade in brick is executed. Please add tests. 
- There is no fading out, the particle effect incorrectly stops abruptly. Instead, the fading out should also take some time, like the fading in. Please add a test (also for the fading in). Note: Later we can add parameters for it that all ow to specify the amount of time, but not for this first version.
- Please rename the brick to "Fade <in/out> particle effect", i.e., the pull down should be after the first word in English. Can this be done with a parameter field in the string resource so that the  positions of the words are flexible for translations? Reason: The ordering of words in Japanese grammar for instance is often reversed compared to English, and if it is just one string resource, it is much clearer for translators in that way. I have tried to suggest a possible format in my in-lined comment.
- If there is no background look, or there is a white background look, or if there is another object (i.e., different from the background) _behind_ the object, and that second object's look covers the complete stage in white below the first object (the one with the particle effect), then the particles of a foreground object are not visible. Note that this is quite strange, because when the background's look is, e.g., completely black, then the particles have a nice red color. Why are they white if the underground is white or transparent? Instead, the particles should be visible in these cases. Please also add a test. Note that this is _not_ necessary if the look _below_ the object, either from the background or from a second object's look that covers the complete stage has the same color as the particles (i.e., red).  
- The color of the particles seem to depend on the background color. How is the color decided? Note: Later we should add other parameters that allow to set the color, size, and length of the particles. But that should be done in later, separate tickets.
- I haven't checked yet everything regarding layers, clones, and starting or resuming scenes. Is it all working fine? If so, please add corresponding test cases.
- The particle effect by default always seems to point upwards unless the object moves. Changing the gravity vector does not influence the particles' direction. Instead, the direction unless the object is moving shall point in the opposite direction of the gravity vector. Also, please make sure that it works correctly in landscape mode. Please also add tests.

devsebastian:
To your questions how is the color decided, and why the particle effect is not visible in lighter backgrounds: 

By default the particle have the "additive" property set to true, this gives a realistic look to the effect, where the core is lighter than, the outer regions of the flame which gradually become darker like shown in the image:

Screenshot (191)

This is in contrast to when we set the additive property to false, this makes the entire effect completely opaque rendering an unrealistic and uniformly colored effect as shown in the below image:

Screenshot (192)

although the additive effect looks more interesting, but it won't be visible in lighter backgrounds.

Screenshot (193)

of course we can always set the additive property to false, to make the effect visible on all backgrounds.

Screenshot (194)

Though I thought, we could add this property with the fade particle effect brick in a later ticket, so that the user can toggle the property anytime and the final control lie with the user, please let me know what you think?

wslany:
One idea: 
1. Since the empty background is the default for new projects, i.e., exactly where users might want to explore the particle effects, what we could do is the following: as a special case (and only in that case) the "additive" property shall be turned off at the beginning if and only if there is no background look. 
2. Immediately add a feature (a separate brick?) that allows to turn off and on this "additive property". This then also allows to explicitly override the default defined in the step 1 above. 

What do you think? Note that we already have several such special cases to make things easier and more intuitive for users.

devsebastian:
Changes:
1. particle gravity is directly equal to scene gravity.
2. fade out brick now has a gradual fade-out effect
3. ClearGraphicsBrick stops the particle effect without any fade effect
4. additive property has been added to the fadeParticlebrick itself. the user can now switch the additive effect on or off, and correspondingly the particle visibility and fade color.
5. the particle effects work on all sprites and clones (particle effect on clones behaves exactly the same way as on any other sprite).
6. the particles pause on their positions and resume back on pressing the pause and resume button respectively!
7. added tests for most of the use cases including stage pause, stage resume, add clone and add fade particle effect on the clone, check whether the additive property is enabled or disabled, etc.

wslany:
Thanks.
1. Please write "particle" with a small letter in the bricks and remove the ":" before the spinner parameters.
2. Please make two bricks out of it. Reason: the additive parameter makes no sense when the first spinner for in and out is set to "out". 
3. The second brick should be "Turn the particle effect's additivity on/off" (with a spinner for on and off, default is off). Note: can this be set for the object that executes the brick only, or does it have to be global for all objects? Both are OK, but the first would be preferable. 
4. Please set the additivity automatically to "off" whenever the background is empty (either no background look, or the background's look is hidden). As soon as a background look is visible, revert the additivity to its earlier value, before there was no background.

wslany:
6. Please make the bricks' texts parameterized so that the spinners can be anywhere in the text, and the translators can decide whether there is a front and/or rear part before and/or after the spinner. It is ok that the spinner is in its own line.

wslany:
7. The bug is still present that the particle effect is immediately visible after a _**re**_-start of the stage (without first leaving the project by returning to Catroid's main menu) or a resume from the stage's pause menu, even if the scripts of the object actually turn it on only a few seconds after the start of the scene, and even if the scripts have been edited by the user since executing the project earlier. It works fine at the first time after entering the project. Please, as always, _**first**_ add a test for the corrected behavior, then correct it.

wslany:
> 6. the particles pause on their positions and resume back on pressing the pause and resume button respectively!

This didn't work when I just tried it out.

wslany (approved):

catroid/src/main/java/org/catrobat/catroid/common/Constants.java
```diff
@@ -42,7 +42,7 @@
 
 public final class Constants {
 
-	public static final double CURRENT_CATROBAT_LANGUAGE_VERSION = 1.04;
+	public static final double CURRENT_CATROBAT_LANGUAGE_VERSION = 1.05;
```
wslany, line 45: Please rebase and further increase compared to the develop branch.

Rawbat (approved):
Looks good, thanks!

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        script.addBrick(FadeParticleEffectBrick(0))
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Assert.assertTrue(sprite.look.isAdditive)
+    }
+
+    @After
+    fun tearDown() {
+        baseActivityTestRule.finishActivity()
+        try {
+            StorageOperations.deleteDir(File(FlavoredConstants.DEFAULT_ROOT_DIRECTORY, projectName))
```
ndrnour, line 178: no need for that, the test rule will take care of that for you.
for more info see `BaseActivityTestRule.java`

catroid/src/main/java/org/catrobat/catroid/content/bricks/FadeParticleEffectBrick.kt
```diff
@@ -0,0 +1,103 @@
...
+import android.widget.Spinner
+import org.catrobat.catroid.R
+import org.catrobat.catroid.content.Sprite
+import org.catrobat.catroid.content.actions.ScriptSequenceAction
+
+class FadeParticleEffectBrick : BrickBaseType {
+
+    companion object {
+        private const val FADE_IN = 0
+        private const val FADE_OUT = 1
```
ndrnour, line 39: please make them public (just remove private access modifier). and use them in the test instead of 0 and 1.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    @Before
+    fun setUp() {
+        createProject()
+        baseActivityTestRule.launchActivity(Intent())
+    }
+
+    @Test
+    fun particleEffectFadeInTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 82: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Assert.assertTrue(projectManager.currentSprite.look.hasParticleEffect)
+    }
+
+    @Test
+    fun particleEffectFadeOutTest() {
+        script.addBrick(FadeParticleEffectBrick(1))
```
ndrnour, line 89: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Assert.assertFalse(projectManager.currentSprite.look.hasParticleEffect)
+    }
+
+    @Test
+    fun particleEffectPauseTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 96: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Espresso.pressBack()
+        Assert.assertTrue(projectManager.currentSprite.look.isParticleEffectPaused)
+    }
+
+    @Test
+    fun particleEffectResumeTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 104: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Espresso.pressBack()
+        Espresso.onView(ViewMatchers.withId(R.id.stage_dialog_button_continue))
+            .perform(ViewActions.click())
+        Assert.assertFalse(projectManager.currentSprite.look.isParticleEffectPaused)
+    }
+
+    @Test
+    fun particleEffectGravityTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 114: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        var particleGravity = emitter.gravity.highMax.toDouble()
+        Assert.assertEquals(particleGravity, -10.0, 0.00)
+        Espresso.onView(ViewMatchers.isRoot()).perform(CustomActions.wait(500))
+        particleGravity = emitter.gravity.highMax.toDouble()
+        Assert.assertEquals(particleGravity, 20.0, 0.0)
+    }
+
+    @Test
+    fun particleEffectOnCloneTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 131: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        }
+        if (!cloneFound) {
+            Assert.fail("No Clone Found")
+        }
+    }
+
+    @Test
+    fun particleAfterWaitTest() {
+        script.addBrick(WaitBrick(2000))
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 150: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        val projectManager = projectManager
+        val look = projectManager.currentSprite.look
+        Assert.assertFalse(look.hasParticleEffect)
+        Espresso.onView(ViewMatchers.isRoot()).perform(CustomActions.wait(2000))
+        Assert.assertTrue(look.hasParticleEffect)
+    }
+
+    @Test
+    fun particleEffectDisableAdditiveByDefaultWhenNoBackgroundLookTest() {
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 161: @here

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/content/brick/stage/ParticleEffectsTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        Espresso.onView(ViewMatchers.withId(R.id.button_play)).perform(ViewActions.click())
+        Assert.assertFalse(sprite.look.isAdditive)
+    }
+
+    @Test
+    fun particleEffectEnableAdditiveByDefaultWhenNoBackgroundLookTest() {
+        projectManager.currentlyEditedScene.backgroundSprite.lookList.add(LookData())
+        script.addBrick(FadeParticleEffectBrick(0))
```
ndrnour, line 169: @here

catroid/src/main/java/org/catrobat/catroid/content/bricks/FadeParticleEffectBrick.kt
ndrnour: see @here annotations

catroid/src/main/java/org/catrobat/catroid/content/bricks/FadeParticleEffectBrick.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private var fadeSpinnerSelectionId: Int
```
ndrnour, line 42:
```suggestion
    private var fadeSpinnerSelectionId: Int = FADE_IN
```
and please remove the constructor at line 44

ndrnour (changes requested):
please fix the warnings too.

wslany (changes requested):
Thanks a lot!! Please change tests as suggested.

Please also add two tests for the scene transition bricks (Start scene and Continue scene), as follows:
In both tests, there is one object in "Scene 1" with the following script, as well as a second scene "Scene 2" that, in the **particleEffectAfterSceneRestartsTest** test only executes a "Start Scene 1" brick when the scene starts (e.g., via a brick in the Background), and in the **particleEffectAfterSceneContinuesTest** test only executes a "Continue Scene 1" brick when the scene starts.

The script of the object in Scene 1 in both tests is the following:
```
When scene starts
Wait 0.2 seconds
Fade particle effect in
Start Scene 1
```

In case of **particleEffectAfterSceneRestartsTest**, the test part should then be:
```
        onView(ViewMatchers.withId(R.id.button_play)).perform(click())
        onView(ViewMatchers.isRoot()).perform(wait(300))
        assertFalse(projectManager.currentSprite.look.hasParticleEffect)
        onView(ViewMatchers.isRoot()).perform(wait(300))
        assertTrue(projectManager.currentSprite.look.hasParticleEffect)
```

In case of **particleEffectAfterSceneContinuesTest**, the test part should then be:
```
        onView(ViewMatchers.withId(R.id.button_play)).perform(click())
        onView(ViewMatchers.isRoot()).perform(wait(300))
        assertTrue(projectManager.currentSprite.look.hasParticleEffect)
```

wslany (approved):

ndrnour (approved):

PR #4265, opened 2021-07-18
CATROID-847 Fix floating + and play buttons in script view

vinayak0505:
CATROID-847 Fix floating + and play buttons in script view

I have implemented TDD but I have an issue where the button id is used several places, so the test is not running but I have checked it from my side it is working and also written the test code which should work when a similar ids issue is resolved.


mseiser:
A new PR for this issue was opened https://github.com/Catrobat/Catroid/pull/4486, this PR is not up to date anymore.

PR #4355, opened 2021-10-16
CATROID-1242 "Username" property reporter

seopei:
CATROID-1242 "Username" property reporter
https://jira.catrob.at/browse/CATROID-1242


wslany (changes requested):
In view of possible abuses of this feature through the disclosure of the Catrobat user's avatar name (even if it is only an account name which is supposed to be anonymous --- but even if it always were, it is still a complex online presence that can be mobbed etc) via the web access bricks, I have come to the realization that it is necessary to ask the user for permission before sharing the username with a downloaded project. 

The question for you, @seopei now is: since this will make the ticket somewhat more complex, would you be still ready to implement it additionally in the same ticket, or would you prefer to keep it separate? In the latter case, we still would need some change for this pull request, albeit a simpler one. Namely, we would need to add a compile flag that hides the new feature in release versions, and only makes it available in debug versions, until the security dialog has been added. However, if you would be willing to add the permission dialog to this ticket, then the compile flag would not be necessary.

Basically, what needs to be done is this: if a downloaded project contains the "username" reporter function AND also a brick that accesses the web (for this we have already some flag), it has to be marked as "tainted". If the function is executed, a dialog that interrupts the execution of the stage must be shown, asking the user "This projects requests access to your username. If the project accesses the internet, it could disclose your username to third parties without your knowledge. If in doubt, do not allow this.

Allow this project to access your Catrobat user name?"

With options:
```
YES              CANCEL   NOT NOW
```

In case the answer was YES, the project can be marked as untainted for the username reporter function. This should be a property for this device and project only.

wslany:
Please increase the Catrobat language version number whenever new functionality of the language is added.

reichli:
Closed due to inactivity. Can serve as reference for next developer.

PR #4389, opened 2021-11-15
Fix jenkins testsuite failing

juliajulie95:
Jenkins testsuite fails with error "Category annotations on Parameterized classes are not supported on individual methods."
Fixed through annotating the categories for the whole instead for the (one and only) test in this file.
No change in behaviour, but now complies to rules

+ additionally changed the quarantined testsuite to always be triggert

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -206,10 +206,6 @@ pipeline {
                         }
 
                         stage('Quarantined Tests') {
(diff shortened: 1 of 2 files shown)


wslany (approved):
LGTM!

wslany:
Thanks!

PR #4396, opened 2021-11-23
CATROID-1245 Duplicated definition of userdefined bricks should not be possible

Rawbat:
Implemented that when a duplicated userdefined brick definition is unpacked the newly unpacked one is assigned a new id and an extra upwards counting label is attached to it.

https://jira.catrob.at/browse/CATROID-1245


adnaribo:
Looks good!

adnaribo (approved):

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/ScriptController.kt
```diff
@@ -162,8 +164,15 @@ class ScriptController {
         destinationSprite.scriptList.add(script)
     }
 
-    @Throws(CloneNotSupportedException::class)
+    @Throws(CloneNotSupportedException::class, UnpackUserDefinedBrickException::class)
     fun unpack(scriptToUnpack: Script, destinationSprite: Sprite) {
+        if (scriptToUnpack is UserDefinedScript) {
+            for (script: Script in destinationSprite.scriptList) {
+                if (script is UserDefinedScript && script.userDefinedBrickID == scriptToUnpack.userDefinedBrickID) {
```
mstoeg, line 171: I am sorry to jump in here, but I was just curious about this ticket. Please do not compare `UUID`s with `==` as this just checks for reference equality. Use `.equals()` to check for the same **value** instead.

mstoeg (changes requested):
I would also change the behavior when I copy mutliple scripts with the backpack (containing a UserDefinedScript as well). Because then the error message is hardly readable as it is closed immediately and all the other scripts (without the UserDefinedScript) get copied. So the error message is not shown in a "blocking" manner.

wslany (changes requested):
Renaming by adding something like " (1)" as a last label or extension of the last label would be more useful than disallowing the pasting. We already have a method in other cases to generate the next numbered "name", e.g., when copying objects or projects. A similar method can easily be implemented when pasting user-defined bricks from the backpack, including in case several user defined bricks are pasted. If this is done, please avoid code duplication. 

Note that it is planned to later add a new feature (in another ticket) that will add the possibility to change the labels and inputs of already existing user-defined bricks. Then changing the "numbered" name to a more useful one will be possible.

catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/controller/ScriptController.kt
Rawbat: The == should be the correct way to handle this because it is written in kotlin. Referential equality in kotlin is ===. As far as I know the == in kotlin basically boils down to a?.equals(b) ?: (b === null).

mstoeg: You are perfectly right! It just would have been a problem when written in Java. Sorry for the inconvenience.

wslany:
Please note #4417

adnaribo (approved):

wslany (changes requested):
Names should be unique, i.e., if the brick is unpacked twice, the second one should get a (2) appended instead of two bricks with a (1) appended, etc:
Screenshot_20220417-091118

Use something like org.catrobat.catroid.ui.recyclerview.util.UniqueNameProvider

mseiser:
Closed due to inactivity, Please reopen if needed

PR #4441, opened 2022-02-03
CATROID-880 Formula editor buttons for list, text, maths

Piki2000:
https://jira.catrob.at/browse/CATROID-880


catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorListsListTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@Category(Cat.AppUi::class, Level.Smoke::class)
+@RunWith(Parameterized::class)
+class FormulaEditorListsListTest(
+    private val formulaEditorFunction: String,
+    private val formulaEditorFunctionParameter: String,
+    @Suppress("unused")
+    private val testName: String
```
juliajulie95, line 57: Please check why this ios suppressed, could we delete the variable?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorLogicListTest.kt
```diff
@@ -0,0 +1,222 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@Category(Cat.AppUi::class, Level.Smoke::class)
+@RunWith(Parameterized::class)
+class FormulaEditorLogicListTest(
+    private val formulaEditorFunction: String,
+    private val formulaEditorFunctionParameter: String,
+    @Suppress("unused")
+    private val testName: String
```
juliajulie95, line 53: Please check why this ios suppressed, could we delete the variable?

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/formulaeditor/FormulaEditorTextListTest.kt
```diff
@@ -0,0 +1,225 @@
...
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.matcher.ViewMatchers.withId
+
+@Category(AppUi::class, Smoke::class)
+@RunWith(Parameterized::class)
+class FormulaEditorTextListTest(
+    private val formulaEditorFunction: String,
+    private val formulaEditorFunctionParameter: String,
+    @Suppress("unused")
+    private val testName: String
```
juliajulie95, line 54: Please check why this is suppressed, could we delete the variable?

juliajulie95 (changes requested):

juliajulie95 (approved):
Thank you!

wslany (changes requested):
Thanks from me as well. Several things that caught my eye and I would like you to correct:

There are several reporters, sensors etc in the new tabs that are shown there, even though the corresponding AI feature has not been activated, and even though they are then correctly not shown in the other, existing categories. For instance, the OCR or voice recognition related sensors should only be shown in the text category when the OCR and/or voice recognition has been turned on in the settings. Here some corresponding screenshots:
image
image
image
image

This implies that there are also missing tests for the AI extension features in the formula editor, because the incorrect visibility of these sensors in the lists under the AI settings should otherwise have been detected.

On the other hand, the following sensor is not listed under Logic when the corresponding AI extension "Object recognition" has been activated by the user (the screenshot is from the sensor list in the formula editor, where it correctly is shown):
image

Note: Most of the above was not specified in https://jira.catrob.at/browse/CATROID-880 since at the time when the ticket was written, the settings screen for the AI extension had not yet been introduced, and thus this should be changed even though it was not specified in the ticket. Sorry about that!

For the same reasons, there need to be a few other changes, on the one hand in the order the AI sensors are shown in the lists, and on the other hand, subtitles are needed in the "text" and "logic" lists. These subtitles have already been introduced in the sensors category because of the AI extension since the ticket was written, as follows:

When an AI extension is activated, the corresponding sensors shall be listed _above_ other sensors and reporters in the lists (like it is already done in the existing sensors list), together with subtitles above the corresponding sensors. Here how it looks in the sensor list:
image

So, for instance, when the AI sensors in the new "text" list are shown, they should be preceded by corresponding subtitles (the same as in the sensor category), and these subcategories should be shown at the top instead at the bottom, see below:
image

Same order and subtitles in a similar way also for the sensors in the "logic" list. 

Since there must be some delimiter between the special AI extension sensors and the normal text reporters and functions (since the AI sensors shall be shown _above_ the others), please also introduce a new subtitle called "Other text reporters". This new subtitle shall, however, only be shown when some corresponding AI extensions has been activated that needs to be shown in the text list _above_ the normal text reporters, since otherwise there are no subcategories for text. 

Note that no new subtitle for regular reporters is not needed in the logic list, since there is already a subtitle shown anyway because of the different subcategories of regular logic reporters and constants.  

Also note that no subtitle needs to be shown at the top of the "list" functions, since there are no subcategories.

Please also add corresponding tests in a test-first (test driven) way.

wslany:
One more question: Why did you change the comparison color here?
image

PR #4618, opened 2022-08-27
CATROID-1159 implement "When you collide with __" brick

aemkic:
- Implement the "When you collide with __" brick
with a Formula as condition.
- There are three cases, either we collide with the
finger, egde or a sprite.
- We used two different formulas. For finger and
edge we use the sensor formula element and for
the sprite we use the collision formula element.

New Test
- testWhenCollideWithBrick

https://jira.catrob.at/browse/CATROID-1159

- [ ] Include the name of the Jira ticket in the PR’s title
- [ ] Include a summary of the changes plus the relevant context
- [ ] Confirm that the changes follow the project’s coding guidelines
- [ ] Perform a self-review of the changes
- [ ] Verify to commit no other files than the intentionally changed ones
- [ ] Check that the commits’ message style matches the [project’s guideline](https://github.com/Catrobat/Catroid/wiki/Commit-Message-Guidelines)
- [ ] Stick to the project’s gitflow workflow


catroid/src/main/java/org/catrobat/catroid/content/WhenCollideWithScript.java
```diff
@@ -0,0 +1,163 @@
...
+		for (Brick brick : brickList) {
+			brick.addRequiredResources(requiredResourcesSet);
+		}
+	}
+
+	public ConcurrentFormulaHashMap getFormulaMap() {
+		return formulaMap;
+	}
+
+	public Sprite getSpriteWithName(String name) {
```
reichli, line 97: Apparently this is method is not used. If so please remove it.

catroid/src/main/java/org/catrobat/catroid/content/WhenCollideWithScript.java
```diff
@@ -0,0 +1,163 @@
...
+		Scene thisScene = ProjectManager.getInstance().getCurrentlyEditedScene();
+
+		return thisScene != null ? thisScene.getSprite(name) : null;
+	}
+
+	public Sprite getBackgroundSprite() {
```
reichli, line 103: This method is only used internally within the class. Please mark it (and others like it) as private.

reichli (changes requested):
Hi!

I noticed that you need to add references to the brick/script in a few more places: BackwardCompatibleCatrobatLanguageXStream, XstreamSerializer and also in a couple of parameterized test suites. Just look where the WhenBounceOffScript and WhenConditionScript and their corresponding bricks are used. This should be a good reference. Some of those places where the new brick is missing are also reported as failing tests when you look at the latest Jenkins run.

There are two more test cases that I think would make sense in your CTL:

- own object selected in spinner: hit clone of self
- trigger only once for each collision: pass through object and then turn around to pass through it again -> counter only increased twice. Though this might not work completely as expected until https://jira.catrob.at/browse/CATROID-766 is merged. Might wanna keep this ticket on your radar as there might be merge conflicts or additional changes depending on who merges the ticket first.

Oh and could you please rewrite the classes for both the brick and the script to Kotlin? :) (since they are completely new classes anyway).

moberer (approved):

catroid/src/main/java/org/catrobat/catroid/content/bricks/WhenCollideWithBrick.kt
```diff
@@ -0,0 +1,118 @@
...
+        items.add(StringOption(context.getString(R.string.touches_finger)))
+        items.add(StringOption(context.getString(R.string.touches_edge)))
+        items.addAll(ProjectManager.getInstance().currentlyEditedScene.spriteList)
+        spinner = BrickSpinner(R.id.brick_when_you_collide_with_spinner, view, items)
+        spinner!!.setOnItemSelectedListener(this)
+        spinner!!.setSelection(script.getSpriteToCollideWithName())
+        return view
+    }
+
+    override fun getPositionInScript() = -1
```
KarlDerMaisbauer, line 78: Can be removed

PR #4766, opened 2023-04-02
IDE-152 Refactor FormatNumberUtil.java to kotlin

Aayush2111:
Jira ticket: https://jira.catrob.at/browse/IDE-152

Files that has been refactored and converted to kotlin: FormatNumberUtil

I have also covered range of test for different input scenarios and ensure that the cutTrailingZeros method works as expected.

- testCutTrailingZeros with input "123.0" expects output "123": This test case ensures that the method removes the trailing zero(s) correctly and returns the expected output.

- testCutTrailingZeros with input "123.00" expects output "123": This test case is similar to the first one, but with two trailing zeros instead of one.

- testCutTrailingZeros with input "123.45" expects output "123.45": This test case is for a number without any trailing zeros. It ensures that the method doesn't modify the input value when there are no trailing zeros to remove.

- testCutTrailingZeros with input "123.4500" expects output "123.45": This test case is similar to the third one, but with trailing zeros at the end of the decimal part. It ensures that the method removes only the trailing zeros and doesn't modify the rest of the value.

- testCutTrailingZeros with input "0" expects output "0": This test case is for a zero input. It ensures that the method returns "0" when the input is "0".

- testCutTrailingZeros with input "0.0" expects output "0": This test case is similar to the fifth one, but with a decimal point. It ensures that the method removes all the trailing zeros from the decimal part and returns "0".

- testCutTrailingZeros with input "123.0" expects output "123.0": This test case ensures that the method returns the input as is, if there are no trailing zeros to remove.

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/utils/FormatNumberUtil.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/utils/FormatNumberUtil.kt
@@ -1,0 +1,45 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2023 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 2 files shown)


Aayush2111:
@sebastiangollob @mseiser 
Please review the changes and provide any feedback that you may have.
Thank you! :)

catroid/src/main/java/org/catrobat/catroid/utils/FormatNumberUtil.kt
```diff
@@ -20,26 +20,26 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+package org.catrobat.catroid.utils
 
-package org.catrobat.catroid.utils;
+import java.math.BigDecimal
 
-import java.math.BigDecimal;
+class FormatNumberUtil private constructor() {
```
ma-zea, line 27: using object FormatNumberUtil instead of class FormatNumberUtil private constructor() would reduce code... then you could just declare method inside of it with @JvmStatic before. No need for init throw assertionerror and companion object

ma-zea (changes requested):

harshsomankar123-tech:
Done #4932

PR #4977, opened 2024-04-26
CATROID-CLT Add test for the changeSize brick

dajminks:
In the test different ways of giving an input for the changeSize brick are used and assertions made to catch if the changes were applied.

There was no CLT yet for this brick.

CLT TEST


reichli:
please recreate with corresponding Jira-Ticket attached

PR #5099, opened 2025-07-06
CATROID-1623 Refactor GlideToAction to Kotlin

doppem15:
The following files were refactored and converted to Kotlin by me and simplified as much as possible:
     - GlideToAction.kt
     - GlideToActionTest.kt
Furthermore, the file "ActionPhysicsFactory.kt" was modified to ensure compatibility with the new Kotlin files.

https://catrobat.atlassian.net/browse/CATROID-1623


dajminks:
Hi Markus,

4 suggestions; 

1. in ActionPhysicsFactory.kt maybe out of consistency reasons use a setter for duration and scope
2. for easier reading, maybe in GlideToAction.kt Line 85, change the if to check for absolute values to be bigger than 0.1f instead of -0.1f > deltaX || deltaX > 0.1f and same for deltaY 

3. If possible, check does the restart (in the update function) work / does the duration really change after the restart
4. You tested for null, Nan and positive numbers, what if you put in negative numbers? 

And squash all commits into one commit

BR Dominik

doppem15:
Hi Dominik,

i tried to follow your suggestions 2–4!

As for point 1), I understand your argument regarding consistency. However, I believe that in Kotlin it's cleaner to work this way, since set/get are generated automatically and their usage is preferred.
Maybe it would make sense to open a ticket to align this across the codebase — but at the moment, other actions are still written in Java, which is where the inconsistency comes from.

dorianpercic (approved):

catroid/src/test/java/org/catrobat/catroid/test/content/actions/GlideToActionTest.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        action.act(0f)
+        val durationBefore = action.durationValue
+
+        sprite.look.setPositionInUserInterfaceDimensionUnit(10f, 10f)
+
+        action.act(0.5f)
+        val durationAfter = action.durationValue
+
+        assert(durationAfter < durationBefore)
```
harshsomankar123-tech: This assertion currently uses assert(...), which depends on JVM assertions being enabled. I could not find -ea in our test setup, so I think this check may not actually fail in CI. Could you please replace it with a JUnit assertion such as assertTrue(durationAfter < durationBefore)?

catroid/src/test/java/org/catrobat/catroid/test/content/actions/GlideToActionTest.kt
```diff
@@ -0,0 +1,188 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2022 The Catrobat Team
```
harshsomankar123-tech: Change Copyright year to 2026

catroid/src/main/java/org/catrobat/catroid/content/actions/conditional/GlideToAction.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private var restart = false
+
+    override fun begin() {
+
+        val durationInterpretation: Float = interpretFormula(duration, scope, "durationInterpretation")
+        val endXInterpretation : Float = interpretFormula(endX,scope, "endXInterpretation")
+        val endYInterpretation : Float = interpretFormula(endY,scope, "endYInterpretation")
+
+        if (!restart) {
```
harshsomankar123-tech, line 67: Here the refactor seems to change the previous behavior: in the Java version, if endX or endY interpretation failed, durationInterpretation was also reset to 0, so the action finished immediately. Now the invalid coordinate falls back to 0f, but the duration is still applied, which can make the sprite glide to (0, 0) over time instead.

catroid/src/main/java/org/catrobat/catroid/content/actions/conditional/GlideToAction.kt
```diff
@@ -0,0 +1,117 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2022 The Catrobat Team
```
harshsomankar123-tech: Here Also

harshsomankar123-tech (changes requested):
Hi @doppem15 , thanks for the refactoring! I have a few minor concerns.

harshsomankar123-tech (approved):
@doppem15 @dorianpercic I’ve addressed my concerns, and one failing test case has now been fixed. Now LGTM.
Thanks!

wslany (approved):
PO reviewed and found working as expected.

PR #5143, opened 2026-01-26
IDE-307 Fix failure in quarantined test suite

Kale2605:
getActivity() sometimes null... not yet attached. Fix by wait and try again.

https://catrobat.atlassian.net/browse/IDE-307?atlOrigin=eyJpIjoiNjk3NDllZDk2YmUyNDk2NmFjM2NkYjkwNjRhYzFhNGYiLCJwIjoiaiJ9

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
+++ b/catroid/src/main/java/org/catrobat/catroid/ui/fragment/FormulaEditorFragment.java
@@ -1062,9 +1062,21 @@ public class FormulaEditorFragment extends Fragment implements ViewTreeObserver.
 		this.chosenUserDataItem = chosenUserDataItem;
 	}
 
+	private void retryOnHiddenChanged() {
+		new Handler(android.os.Looper.getMainLooper()).post(() -> {
+			if (isAdded()) {
+				onHiddenChanged(false);
+			}
+		});
+	}
+
 	@Override
 	public void onHiddenChanged(boolean hidden) {
 		if (!hidden) {
+			if (getActivity() == null) {
+				retryOnHiddenChanged();
+				return;
+			}
 			ActionBar actionBar = ((AppCompatActivity) getActivity()).getSupportActionBar();
 			boolean isRestoringPreviouslyDestroyedActivity = actionBar == null;
 			if (!isRestoringPreviouslyDestroyedActivity) {


Frajhamster (approved):
Looks good, except that on SonarQube your code introduced a new issue, which still passes the QualityGate check, so you can decide to take a look at it and try to fix it, or leave it like it is now, up to you.

Kale2605:
> Looks good, except that on SonarQube your code introduced a new issue, which still passes the QualityGate check, so you can decide to take a look at it and try to fix it, or leave it like it is now, up to you.

Done

harshsomankar123-tech (changes requested):
Hii @Kale2605 Thanks For the PR 
I have found some problems with this implementation:

1. Infinite retry risk
If getActivity() keeps returning null, this ends up reposting to the main thread without any real stopping condition. That can quietly pile up work on the UI thread. It would be safer to cap the number of retries or introduce a clear exit path.

2. Hardcoded false
onHiddenChanged already gives us the correct hidden value from the system. Passing false ignores that and can lead to the wrong behavior  for example, running “show” logic even when the fragment is actually being hidden. We should just forward the original hidden value.

3. Retry counter not reset
The retry counter needs to be reset after both success and failure. Otherwise, it can carry over into future hide/show cycles on the same fragment instance and cause inconsistent behavior.

U can refer This snippet:
```
// Add these fields to the class
private int retryCount = 0;
private static final int MAX_RETRIES = 5;

// Add this helper method
private void retryOnHiddenChanged(final boolean hidden) {
    if (retryCount >= MAX_RETRIES) {
        retryCount = 0; // Reset so future calls can attempt again
        Log.e(TAG, "Failed to update UI in onHiddenChanged after " + MAX_RETRIES + " retries.");
        return;
    }
    retryCount++;
    new Handler(Looper.getMainLooper()).post(() -> {
        if (isAdded()) {
            onHiddenChanged(hidden);
        }
    });
}

// Update the existing onHiddenChanged method
@Override
public void onHiddenChanged(boolean hidden) {
    if (hidden) {
        return; // UI update logic is only needed when becoming visible
    }

    if (getActivity() == null) {
        retryOnHiddenChanged(hidden);
        return;
    }

    retryCount = 0; // Reset count on success
    
    ActionBar actionBar = ((AppCompatActivity) getActivity()).getSupportActionBar();
    boolean isRestoringPreviouslyDestroyedActivity = actionBar == null;
    
    if (!isRestoringPreviouslyDestroyedActivity) {
        actionBar.setTitle(R.string.formula_editor_title);
        BottomBar.hideBottomBar(getActivity());
        updateButtonsOnKeyboardAndInvalidateOptionsMenu();
        updateBrickView();
    }
    
    if (chosenCategoryItem != null) {
        addResourceToActiveFormula(chosenCategoryItem.getNameResId());
        chosenCategoryItem = null;
    }
    
    if (chosenUserDataItem != null) {
        if (chosenUserDataItem instanceof UserVariable) {
            addUserVariableToActiveFormula(chosenUserDataItem.getName());
        } else if (chosenUserDataItem instanceof UserList) {
            addUserListToActiveFormula(chosenUserDataItem.getName());
        } else if (chosenUserDataItem instanceof UserDefinedBrickInput) {
            addUserDefinedBrickInputToActiveFormula(chosenUserDataItem.getName());
        }
        chosenUserDataItem = null;
    }
}
```

PR #5161, opened 2026-03-11
CATROID-1615 "Write embroidery data to file" Brick Not Saving .dst File on Smartphone

doppem15:
I fixed a bug where the "Save plot to" and "Write embroidery data" bricks failed on older Android devices (SDK <29). While writing the tests, I noticed a lot of duplicated code, so I created a new base class to unify the logic. This makes the storage handling cleaner and ensures it now works reliably across all Android versions.

Additionally, many new tests were added to make the code more reliable:

-Instrumented Tests were required for SVG, Laser, and Embroidery.

-Instrumented Tests were also used for the Base class.

-Unit Tests were required for Variable actions.

https://catrobat.atlassian.net/browse/CATROID-1615

- [ ] Stick to the project’s gitflow workflow


catroid/src/main/java/org/catrobat/catroid/content/actions/BaseFileStorageAction.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    protected fun getFile(fileName: String): File? {
+        val file = File(Constants.EXTERNAL_STORAGE_ROOT_EXPORT_DIRECTORY, fileName)
+        return if (file.exists()) file else null
+    }
+
+    protected fun getFileName(): String {
+        var fileName = Utils.sanitizeFileName(formula?.interpretString(scope))
+        if (!fileName.endsWith(fileExtension)) {
+            fileName += fileExtension
```
wslany: This looks like a behavior change for `ReadVariableFromFileAction` on legacy storage.

Previously the read action only added `.txt` when there was no extension at all. With the shared helper, `config.csv` now becomes `config.csv.txt`, which would break existing projects reading non-`.txt` files.

Could we keep the old extension handling for the read action?

catroid/src/main/java/org/catrobat/catroid/content/actions/BaseFileStorageAction.kt
```diff
@@ -0,0 +1,173 @@
...
+    internal abstract fun checkIfDataIsReady(): Boolean
+    protected abstract fun showDataMissingMessage()
+    abstract override fun getTargetIntent(): Intent
+
+    override fun act(delta: Float): Boolean {
+        if (checkIfDataIsReady()) {
+            startFileInteraction()
+        } else {
+            showDataMissingMessage()
+        }
```
wslany, line 66: I think this refactor removed the old null guard.

Before, these actions returned early when `formula == null` (and some also tolerated missing `scope`). Now `act()` can still reach `getFileName()` or `scope!!` with null state and crash instead of no-oping.

Could we restore a common `formula/scope` guard in the base class?

wslany (changes requested):
I think this PR needs regression coverage for the behavior moved into `BaseFileStorageAction`:

- a legacy read test for explicit non-`.txt` extensions like `config.csv`
- a null-safety test proving `act()` does not crash when `formula == null` (and ideally also with missing `scope` where relevant)

wslany (changes requested):
Please resolve copilot findings.

catroid/src/main/java/org/catrobat/catroid/content/actions/BaseFileStorageAction.kt
```diff
@@ -0,0 +1,180 @@
...
+            file
+        } else null
+    }
+
+    protected fun getFile(fileName: String): File? {
+        val file = File(Constants.EXTERNAL_STORAGE_ROOT_EXPORT_DIRECTORY, fileName)
+        return if (file.exists()) file else null
+    }
+
+    protected fun getFileName(): String {
```
wslany, line 173: I rechecked the latest version after the rebase/fix commits. The original `ReadVariableFromFileAction` extension regression and null guard issue look addressed, but I think one related filename regression remains.

`BaseFileStorageAction.getFileName()` now preserves any explicit extension for all actions. That matches the old read-variable behavior, but it changes the write actions: `SavePlotAction`, `SaveLaserAction`, `WriteEmbroideryToFileAction`, and `WriteVariableToFileAction` previously appended their required extension unless the filename already ended with that exact extension.

For example, `Write embroidery data to "foo.txt"` used to save `foo.txt.dst`; now it saves `foo.txt`, and `Save plot to "foo.txt"` no longer produces an `.svg` filename.

Could we keep the “preserve arbitrary extension” behavior only for reads, while write/export actions continue enforcing their own output extension (`.txt`, `.svg`, `.dst`)? A small regression test for a write action with a mismatched explicit extension would help lock this down.

wslany (changes requested):
Thanks for the update and the rebase. Would it be OK to add one more behavior-level regression test for the CATROID-1615 flow? Right now the new tests are useful, but they mainly exercise the extracted helper methods, and do not show that the original user-visible path is protected, I think. Could you add one happy-path action-level test that goes through the real flow end to end?

Also, `SimpleTest.kt` may be a leftover?

catroid/src/androidTest/java/org/catrobat/catroid/test/content/actions/BaseFileStorageActionTest.kt
```diff
@@ -0,0 +1,180 @@
...
+        val testContext = ApplicationProvider.getApplicationContext<Context>()
+        val cacheFile = File(testContext.cacheDir, fileName)
+
+        try {
+            if (action.checkIfDataIsReady()) {
+                action.handleFileWork(cacheFile)
+            }
+        } finally {
+            if (cacheFile.exists()) {
+                cacheFile.delete()
```
moe-threeface, line 127: Small test question: in testCoreFileLifecycle the cache file is deleted inside the test itself.

So the final assertion mostly checks the cleanup done by the test, not really the cleanup from BaseFileStorageAction. Would it make sense to run the actual BaseFileStorageAction flow here, so we really test the production cleanup path?

catroid/src/main/java/org/catrobat/catroid/utils/Utils.java
```diff
@@ -95,6 +95,10 @@ public final class Utils {
 
 	private static final String TAG = Utils.class.getSimpleName();
 
+	public static boolean isLegacyStorageManager() {
+		return android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.Q;
```
moe-threeface, line 99: isLegacyStorageManager() was added, but it looks like it is not used anywhere yet.

Since this PR still has direct SDK checks in Project.java and BaseFileStorageAction.kt, should those places use this helper, or should the helper be removed for now?

moe-threeface (commented):
I checked the CATROID-1615 changes locally.

The general approach with the shared BaseFileStorageAction looks good to me, and the added storage regression tests are nice to have. I only left two small questions around one test and one unused helper.

Looks good overall.

PR #5162, opened 2026-03-11
 CATROID-1656: Add ShowToastBrick

Danveyd:
Jira ticket: https://catrobat.atlassian.net/browse/CATROID-1656

### Description
This PR introduces the `ShowToastBrick` to the Device category. It allows users to display temporary on-screen messages, which is useful for debugging and user feedback.

 ### Changes
 Following the `AGENTS.md` guidelines and TDD approach:
 1. **Commit 1 [test]:** Registered `ShowToastBrick` in `BrickCategoryTest` to establish the requirement.
 2. **Commit 2 [feature]:** Implemented the brick and action in **Kotlin**. 
    - The `ShowToastAction` uses `Handler(Looper.getMainLooper()).post` to safely display the Toast on the Android UI thread.
    - Registered the brick in `XstreamSerializer`, `CategoryBricksFactory`, and `ActionFactory`.
    - Added RU and EN translations and layout XML.
 3. **Commit 3 [test]:** Added a Catrobat Language Test (CLT) in `assets/catrobatTests/bricks/device/` to verify runtime execution.


DahMaus:
Nice work man :D

Danveyd:
**Hello! I have finished updating the PR.**

 **Technical Summary:**
 * Migrated the `ShowToastBrick`  to Kotlin.
 * Ensured UI-thread safety by using `Handler(Looper.getMainLooper()).post` for displaying Toasts.
 * Updated `ActionFactory.java` (Java) and registered the brick in `XstreamSerializer`.
 * Added RU and EN translations in `strings.xml`.

 **Testing:**
 * Updated `BrickCategoryTest.java` to include the new brick in the Device category.
 * Added a **Catrobat Language Test (CLT)**: `testShowToast.catrobat` in the `bricks/device` directory to verify runtime behavior.
 * Verified that SonarCloud analysis is successful.

 **Blockers:**
 * **Jira:** Says "You don't have permission to create issues in this project" :/
 * **Slack:** I am an external contributor and do not have a `@tugraz.at` email to join.

wslany:
Thanks! While until now we did not commit translations into normal PRs (the translations were supposed to be done by humans via CrowdIn), the CrowdIn workflow is broken in our project, and we have decided to let LLMs do the translations in a systematic and eventually, automatic way. In the case of this PR, adding, e.g., a Russian translation where this is a language you know well, is fine from now on. This is meant as a special note, which I'll also announce internally on our Slack.

Danveyd:
Hello! I have updated the PR to match the official Jira ticket and address the latest feedback:

1. **Jira Integration:** Updated the PR title to `CATROID-1656: Add ShowToastBrick` and added the ticket link to the description.
2. **Brick Positioning:** 
   - In the **Looks** category, the brick is now placed after `ShowTextColorSizeAlignmentBrick`.
   - In the **Device** category, it is placed after `ResetTimerBrick`.
   - The **CLT** has been moved to the `bricks/looks/` directory accordingly.
3. **Safety & Logic:** 
   - Added a `try-catch` block in `ShowToastAction` to handle `InterpretationException` with a `"NaN"` fallback, preventing potential stage thread crashes.
   - Fixed `Brick.java` to include the `MESSAGE` field in `EXPECTS_STRING_VALUE`.
4. **Code Style & CI:**
   - Updated the layout XML to use `match_parent` for the message field and adjusted the style to match the light-green **Looks** category.
5. **Estimation:** I suggest an effort estimate of **2 (Peanuts)** for this story.

The commit history has been cleaned up into 3 logical steps

wslany:
Hi @Danveyd, thanks so much for this contribution and welcome! 👋 The TDD structure (test → implementation → integration test) is exactly what we like to see, and you didn't miss the easy-to-forget bits like `EXPECTS_STRING_VALUE` registration and the `XstreamSerializer` alias — nice work.

I’ve also added the brick placement and the 2-point estimate to the Jira ticket. The positioning after `ShowTextColorSizeAlignmentBrick` in Looks and after `ResetTimerBrick` in Device sounds reasonable to me. Thank you also for cleaning up the commit history and aligning the PR with the ticket.

I have a few suggestions before we merge. Don't worry, none of these are blockers in the "you did something wrong" sense — they're mostly about matching existing patterns in the codebase that aren't obvious from the outside.

**One likely runtime issue I'd love you to look at:**

`ShowToastAction` extends `TemporalAction`, whose `update(percent)` method gets called repeatedly across the action's duration by scene2d. As written, I think this will post a fresh `Toast` on every frame while the action runs, rather than once. For a fire-once side effect like a toast, it's usually better to extend `Action` directly and return `true` from `act()` so it completes immediately. Take a look at `BroadcastAction` or `SetVariableAction` for the pattern we use elsewhere — they're good references. Could you give that a try and see if the toast still behaves as expected?

**A few smaller things, take or leave:**

- Inside `update()`, `Handler(Looper.getMainLooper())` allocates a new handler each call. Once the above is fixed it's only allocated once, but it's still nice to hoist it to a `companion object` or a `val` so it's reused.
- The `"NaN"` fallback when `InterpretationException` is thrown will show up literally on screen, which might surprise users. Most other formula bricks fall back to an empty string in this case — would you mind matching that convention?
- Just to double-check: the PR description mentions the **Looks** category, but the `BrickCategoryTest` change places it under **Device**. Both are defensible (toast is a device/platform concept), but could you confirm `CategoryBricksFactory` and the test agree on one?
- `var scope: Scope?` and `var toast: Formula?` as public mutable nullable properties is a bit loose; sibling actions tend to use constructor params or `lateinit`. Minor style thing.

Happy to clarify any of this or pair on the `TemporalAction` change if it's helpful --- just ping me. Thanks again for picking this up! 🙌

Danveyd:
Hi @wslany ! Thank you so much for the warm welcome and the incredibly detailed review!
 
 You were absolutely right about `TemporalAction` — spamming 60 Toasts per second was definitely not the intention! That's a great catch. I have applied all of your suggestions:
 
 1. Changed inheritance to `Action`, implemented `act()`, and returned `true` so it fires exactly once.
 2. Hoisted the UI `Handler` into a `companion object` to avoid unnecessary allocations.
 3. Changed the `InterpretationException` fallback from `"NaN"` back to an empty string `""` to match existing conventions.
 4. Updated the `scope` and `toast` properties to use `lateinit var` instead of nullable `var`s.
 
 Regarding the categories: Wolfgang and I discussed in the Jira ticket (CATROID-1656) that the brick makes sense in **both** categories (Looks as the primary one, and Device as a platform concept). 
 I realized my first `[test]` commit was missing the `Looks` category update, which caused the confusion! I have interactively rebased the history to ensure `BrickCategoryTest` properly expects the brick in both categories right from the first commit. 
 
 I have force-pushed the updated commits. Let me know if everything looks good now! Thanks again for the guidance!

wslany (approved):
LGTM! 🎉

ActionFactory.createShowToastAction returns com.badlogic.gdx.scenes.scene2d.Action, consistent with createSetVariableAction and createSpeakAction. ✅

The CLT does the right thing: since a Catrobat script can't observe whether a system Toast actually rendered, a sentinel assertion that's only reached if the brick didn't throw is the right shape.

Approving. Thanks for the careful work and the responsiveness to feedback, really nice first contribution. Welcome to OldCatroid! 🙌

wslany (changes requested):
Thank you again for this contribution. I tested the rebased build locally, and the brick is visible and working nicely.

https://github.com/user-attachments/assets/5aad5344-d37e-4fd9-b891-5bf4ae7b10de

I noticed one small edge case that might be worth handling before merge: when `Show message` is placed inside a `forever` loop, the toast can be triggered many times very quickly. Android then queues several toast messages, so they may continue appearing for a short while even after leaving the stage.

https://github.com/user-attachments/assets/3c69e24f-f970-4fe3-be44-e35a837b2762

This is not a major issue, and on newer Android versions there seems to be a small system-side queue limit, so it does stop after a few messages, but on [older devices this queue had 50 entries](https://android.googlesource.com/platform/frameworks/base/+/f76a50c/services/java/com/android/server/NotificationManagerService.java#71). Therefore, I think it would be good if the brick itself avoided building up a long queue. One possible approach would be to keep a shared/current `Toast`, cancel or replace it before showing the next one, and maybe clear it when the stage stops.

So the behavior I’d suggest aiming for is: if the brick is executed repeatedly very fast, the user should only see the latest/current message, not a backlog of older messages.

Again, very nice work on the brick. This is a subtle runtime behavior, and it only showed up during manual testing with a tight loop.

Danveyd:
> Thank you again for this contribution. I tested the rebased build locally, and the brick is visible and working nicely.
> 
>  5162-1.webm 
> I noticed one small edge case that might be worth handling before merge: when `Show message` is placed inside a `forever` loop, the toast can be triggered many times very quickly. Android then queues several toast messages, so they may continue appearing for a short while even after leaving the stage.
> 
>  5162-2.webm 
> This is not a major issue, and on newer Android versions there seems to be a small system-side queue limit, so it does stop after a few messages, but on [older devices this queue had 50 entries](https://android.googlesource.com/platform/frameworks/base/+/f76a50c/services/java/com/android/server/NotificationManagerService.java#71). Therefore, I think it would be good if the brick itself avoided building up a long queue. One possible approach would be to keep a shared/current `Toast`, cancel or replace it before showing the next one, and maybe clear it when the stage stops.
> 
> So the behavior I’d suggest aiming for is: if the brick is executed repeatedly very fast, the user should only see the latest/current message, not a backlog of older messages.
> 
> Again, very nice work on the brick. This is a subtle runtime behavior, and it only showed up during manual testing with a tight loop.

Good point, thanks! I'll try to fix it tonight.

Danveyd:
Hi Wolfgang!

Thank you for testing the edge cases!

I have updated the `ShowToastAction` implementation to fix this. I added a static `currentToast` reference inside the companion object. Now, inside the UI thread handler, we call `currentToast?.cancel()` before creating and showing the new one.
This guarantees that the Toast queue never exceeds 1 message, even if the brick is executed 60 times a second. As a result, when the user leaves the stage, there is no backlog - at most, only the single currently active short Toast will naturally fade out.

I have force-pushed the updated [feature] commit.

harshsomankar123-tech (changes requested):
Hey @Danveyd, thanks for the feature! Ihave some concern PTAL.

* When both the Looks and Device variants of `ShowToastBrick` are added to the same script, they currently render with the same green color, which makes it hard to distinguish which brick belongs to which category. IMO, both bricks should have different category-specific colors.

* I also noticed that `ShowToastBrick` was added in this PR, but I could not find the corresponding help URL mapping in `BricksHelpUrlTest`. That test should normally fail if the mapping is missing, so could you please add it before merge?

harshsomankar123-tech:
@Danveyd one more thing these are some missing test in `ShowToastBrick` PTAL.
and add focused coverage for the new toast brick:

ShowToastBrickTest
* verifies `BrickField.MESSAGE` is registered with the toast edit-text id
* verifies the string/formula constructors store the default message formula
* verifies `addActionToSequence()` adds a `ShowToastAction`

ShowToastActionTest
* verifies the message formula is interpreted as a string
* verifies invalid formula interpretation does not crash
* verifies `act()` returns true
* ideally verifies previous toast cancellation behavior, if this can be isolated without brittle Android UI timing

Serialization round-trip test
* save/load a project containing `ShowToastBrick`
* verify the `MESSAGE` formula survives deserialization

Current category coverage only proves the brick appears in palettes; it does not verify execution or persistence.

If you need any help with the tests, feel free to ping me :)
Thanks!

PR #5168, opened 2026-03-27
IDE-318 Fix reflection helper null handling on Java 22

wslany:
## Summary

This PR fixes a unit test failure caused by changed reflection behavior on Java 22.

The shared test helper `Reflection.invokeMethod(Class<?>, Object, String, ParameterList)` now explicitly throws `NullPointerException` when it is asked to invoke a non-static method on a `null` receiver.

https://catrobat.atlassian.net/browse/IDE-318

Note: This PR is conceptually independent from #5170 but they currently affect the same test pipeline. Removing the obsolete Scratch Converter tests does not fix the separate Java 22 reflection failure, and #5168 does not remove the obsolete Scratch Converter tests. So each PR is reviewable on its own, but CI may stay red until both are merged.

## Background

In the current CI setup, the reflection-based unit test `ReflectionExceptionsTest.testInvokeMethodWithNullObject3` no longer behaves as expected on Java 22.

Previously, this test expected a `NullPointerException` when invoking an instance method with a `null` target through the helper. On Java 22, the reflective call path wraps that situation differently, which makes the test fail even though the underlying problem is still a null receiver.

## Change

Instead of relying on JDK-version-specific reflection internals, the helper now performs this check itself before calling `Method.invoke(...)`:

- if the target object is `null`
- and the reflected method is not static

then it throws `NullPointerException` directly.

This makes the helper behavior stable and keeps the existing test expectation meaningful across Java versions.

## Verification

The existing regression test `ReflectionExceptionsTest.testInvokeMethodWithNullObject3` failed before this change on Java 22 and passes again after the helper-side null check was added.

Verified locally with:

`bash`
`./gradlew -PenableCoverage jacocoTestCatroidDebugUnitTestReport --full-stacktrace --no-daemon`

### DIFF
--- a/catroid/src/test/java/org/catrobat/catroid/test/utils/Reflection.java
+++ b/catroid/src/test/java/org/catrobat/catroid/test/utils/Reflection.java
@@ -24,6 +24,7 @@ package org.catrobat.catroid.test.utils;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 public final class Reflection {
 	private Reflection() {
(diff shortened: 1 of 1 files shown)


harshsomankar123-tech:
@wslany Looks good to me 👍
We can proceed with this once PR #5179 is merged.

harshsomankar123-tech (approved):

ratschillerp (approved):

PR #5186, opened 2026-04-04
CATROID-1640 Reset embroidery, plot, and laser drawing state on fresh starts

wslany:
### Summary
This PR fixes stale drawing/stitch runtime state being reused across fresh executions.

https://catrobat.atlassian.net/browse/CATROID-1640

If a project ended while a running stitch, plot, cut, or engrave mode was still active, the next fresh run could inherit that old state and create an unintended initial segment from the previous endpoint to the new start position.

The fix resets that runtime state when execution is supposed to start from a blank slate, while preserving the existing behavior for resume-style flows.

### Why this approach
The underlying issue was not in the stitch/plot geometry itself, but in sprite runtime state surviving longer than intended.

Sprite keeps drawing-related runtime objects such as:

- runningStitch
- plot
- penConfiguration

Those objects should be preserved for resume behavior like pause/continue and Continue scene, but they must be reset for fresh starts such as project restart and Start scene.

To keep that distinction explicit, this PR introduces a dedicated drawing-state reset helper on Sprite and uses it only on fresh-start lifecycle paths.

Changes

- Added a regression test for sprite drawing-state reset
- Added Sprite.resetDrawingState()
- Reused that helper from resetSprite()

Explicitly reset drawing state on:
- project restart
- Start scene
- Left these flows unchanged:
- stage pause / continue
- Continue scene via scene backup/restore

### Result
Fresh runs now start with clean embroidery/plot/laser state, even if no explicit stop brick was used in the previous execution.

This covers:

- embroidery running stitches
- plotting
- laser cutting
- engraving

### Testing

Verified with:

`./gradlew :catroid:testCatroidDebugUnitTest --tests org.catrobat.catroid.test.content.SpriteDrawingStat
`

Test on phone: screen recording before and after fix: https://photos.app.goo.gl/yGBKrRXvmYrQhcev6

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
+++ b/catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
@@ -1,6 +1,6 @@
 /*
  * Catroid: An on-device visual programming system for Android devices
- * Copyright (C) 2010-2025  The Catrobat Team
+ * Copyright (C) 2010-2026 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
@@ -460,6 +460,11 @@ public class StageListener implements ApplicationListener {
 
 		Gdx.input.setInputProcessor(stage);
 
+		// Fresh starts must drop any persisted stitch/plot state before the scene is recreated.
+		for (Sprite sprite : scene.getSpriteList()) {
+			sprite.resetDrawingState();
+		}
+
 		scene.firstStart = true;
 		create();
 	}
@@ -490,6 +495,9 @@ public class StageListener implements ApplicationListener {
 		UserDataWrapper.resetAllUserData(ProjectManager.getInstance().getCurrentProject());
 
 		for (Scene scene : ProjectManager.getInstance().getCurrentProject().getSceneList()) {
+			for (Sprite sprite : scene.getSpriteList()) {
+				sprite.resetDrawingState();
+			}
 			scene.firstStart = true;
 		}
 		reloadProject = true;
--- a/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
+++ b/catroid/src/main/java/org/catrobat/catroid/content/Sprite.java
@@ -1,6 +1,6 @@
 /*
  * Catroid: An on-device visual programming system for Android devices
- * Copyright (C) 2010-2025  The Catrobat Team
+ * Copyright (C) 2010-2026 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
@@ -342,6 +342,12 @@ public class Sprite implements Nameable, Serializable {
 		}
 	}
 
+	public void resetDrawingState() {
+		penConfiguration = new PenConfiguration();
+		plot = new Plot();
+		runningStitch = new RunningStitch();
+	}
+
 	public void resetSprite() {
 		Brick.ResourcesSet resourcesSet = new Brick.ResourcesSet();
 		addRequiredResources(resourcesSet);
@@ -359,9 +365,7 @@ public class Sprite implements Nameable, Serializable {
 			look.setLookData(getLookList().get(0));
 		}
 
-		penConfiguration = new PenConfiguration();
-		plot = new Plot();
-		runningStitch = new RunningStitch();
+		resetDrawingState();
 	}
 
 	public void invalidate() {
--- a/catroid/src/test/java/org/catrobat/catroid/test/stage/StageListenerDrawingStateResetTest.kt
+++ b/catroid/src/test/java/org/catrobat/catroid/test/stage/StageListenerDrawingStateResetTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2026 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.catrobat.catroid.test.stage
+
+import android.graphics.PointF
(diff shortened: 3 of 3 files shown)


catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -460,6 +460,10 @@ public void startScene(String sceneName) {
 
 		Gdx.input.setInputProcessor(stage);
 
+		for (Sprite sprite : scene.getSpriteList()) {
+			sprite.resetDrawingState();
+		}
+
```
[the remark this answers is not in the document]
wslany: Thanks, fixed!

wslany:
@dorianpercic Should I rewrite the Java production code files to Kotlin here as well? They seem a bit broad for the narrow topic of this PR, and would require more extensive testing. Please let me know what you think.

dorianpercic:
> @dorianpercic Should I rewrite the Java production code files to Kotlin here as well? They seem a bit broad for the narrow topic of this PR, and would require more extensive testing. Please let me know what you think.

No I think our convention was that only new files should be implemented in Kotlin. Maybe in future, to accelerate refactoring to Kotlin, especially with today's LLM capabilities, we might change the policy to let rewrite each changed file immediately to Kotlin, but this might need some more testing on how good this works. Maybe there exists some tools already which alleviate this?

dorianpercic (approved):

wslany:
> No I think our convention was that only new files should be implemented in Kotlin. Maybe in future, to accelerate refactoring to Kotlin, especially with today's LLM capabilities, we might change the policy to let rewrite each changed file immediately to Kotlin, but this might need some more testing on how good this works. Maybe there exists some toold already which alleviates this?

OK. I've included in my AGENTS.md file the following lines:

> Prefer Kotlin over Java for Catroid source code.
> 
> When a PR needs changes in an existing Catroid Java source file, prefer rewriting that touched source file to Kotlin as part of the same PR, so a follow-up refactoring ticket is not needed again.
> 
> Apply that preference with judgment:
> - new source code files should normally be written in Kotlin, not Java
> - if a PR touches an existing Java production file, plan for a Kotlin rewrite of that file unless there is a strong reason not to
> - keep tests aligned with the same red/green workflow when a Java-to-Kotlin rewrite is part of the change
> - do not quietly skip the Kotlin rewrite expectation for touched Java files; call it out explicitly if a rewrite would be too large or risky

dorianpercic:
> > No I think our convention was that only new files should be implemented in Kotlin. Maybe in future, to accelerate refactoring to Kotlin, especially with today's LLM capabilities, we might change the policy to let rewrite each changed file immediately to Kotlin, but this might need some more testing on how good this works. Maybe there exists some toold already which alleviates this?
> 
> OK. I've included in my AGENTS.md file the following lines:
> 
> > Prefer Kotlin over Java for Catroid source code.
> > When a PR needs changes in an existing Catroid Java source file, prefer rewriting that touched source file to Kotlin as part of the same PR, so a follow-up refactoring ticket is not needed again.
> > Apply that preference with judgment:
> > 
> > * new source code files should normally be written in Kotlin, not Java
> > * if a PR touches an existing Java production file, plan for a Kotlin rewrite of that file unless there is a strong reason not to
> > * keep tests aligned with the same red/green workflow when a Java-to-Kotlin rewrite is part of the change
> > * do not quietly skip the Kotlin rewrite expectation for touched Java files; call it out explicitly if a rewrite would be too large or risky

In refactoring tickets (example: https://catrobat.atlassian.net/browse/CATROID-1548) we have these guidelines: 
1. Get rid of "!!" and make sure there can't be any NPE's. Since every Exception in KT is treated like a Runtime Exception be sure you catch them.
2. Take advantage of Null-Safety features in Kotlin (https://kotlinlang.org/docs/null-safety.html).
3. If you find any other smells in the code or unclean code parts feel free to improve them as well. But make sure that the behavior of the code is not changed.

maybe we can add this, and especially Null Safety handling, as well.

urkat (approved):
Lgtm

dorianpercic (changes requested):
Can I ask, what this PR actually really fixes? Since all there is done now is instead of doing:
```
penConfiguration = new PenConfiguration();
plot = new Plot();
runningStitch = new RunningStitch();
```
a new function is called, which does exactly the same. I might be confused, but I don't see what this really changes now. Or maybe only the PR description does not match the content; if it's only a small refactoring ticket though, then I get it.

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Catroid: An on-device visual programming system for Android devices
- * Copyright (C) 2010-2025  The Catrobat Team
+ * Copyright (C) 2010-2026 The Catrobat Team
```
dorianpercic, line 3: we can remove this, the content of the file did not change, so it should stay 2025.

wslany:
> Can I ask, what this PR actually really fixes? Since all there is done now is instead of doing:
> 
> ```
> penConfiguration = new PenConfiguration();
> plot = new Plot();
> runningStitch = new RunningStitch();
> ```
> 
> a new function is called, which does exactly the same. I might be confused, but I don't see what this really changes now. Or maybe only the PR description does not match the content; if it's only a small refactoring ticket though, then I get it.

You are right, thanks a lot for that save! The apk Katja and I tested was created before a "cleanup" in response to the “redundant reset” review by Copilot. Shouldn't have done that, the order was wrong...

The supporting evidence from reflog/history is:

- #763398e5 = last commit with the StageListener loops
- APK build happened before #fcb4dfd7
- #fcb4dfd7 is exactly where those loops disappeared

This also revealed the insufficiency of the current test. I will by fixing that in my next commit.

wslany:
@dorianpercic Thanks again. The PR has been updated, and it now contains the actual fix again.

What changed:
- the branch was rebuilt into a proper red/green history
- the old helper-only test was replaced with a lifecycle regression test
- the production fix in `StageListener` was restored

Why the old test was not good enough:
- the previous test only called the new helper directly on `Sprite`
- that proved the helper worked, but not that the real bug path was fixed
- so it could stay green even if project restart / `Start scene` still reused stale drawing state in practice

Why the new test is more meaningful:
- the new test exercises `StageListener.reloadProject()`, i.e. an actual fresh-start lifecycle path
- it first puts a sprite into a “dirty” state with active pen / plot / running-stitch data
- then it triggers the fresh-start flow
- before the fix, the same pen/plot/stitch objects were still there, so the test fails
- with the fix, the drawing state is replaced/reset during the fresh-start path, so the test turns green

So the two commits now tell the intended story:
- `98c4bc27` `test: add regression for fresh-start drawing reset`
  this is the red test against old behavior
- `bc710825` `fix: reset drawing state on fresh starts`
  this restores the `StageListener` resets and turns the test green

And in addition to the automated regression test, I also manually verified the rebuilt APK on-device just now: the embroidery start behavior is correct again, so eyeballs 1.0 also confirm the fix. :eyes:

dorianpercic (approved):
Thanks, now it's clear! :)

catroid/src/main/java/org/catrobat/catroid/stage/StageListener.java
wslany: It changed, no? See later added lines (463-467 and 498-500).

dorianpercic: Yes, this was a comment from before your recent commit, when code was again added, so it can be resolved.

PR #5199, opened 2026-04-18
Migrate to the new Catroweb - share community platform. 

dmetzner:
## Summary

https://catrobat.atlassian.net/browse/IDE-317

Migrate Catroid from the old `share.catrob.at` API (custom 32-char tokens, offset pagination, `/api/login/Login.json`-style endpoints) to the new `share.catrobat.org` API v2 (JWT access/refresh tokens, cursor-based pagination, OpenAPI-compliant endpoints).

Prerequisite for the next Catroid release — old API endpoints removed as part of Catroweb API v2 standardization ([ADR 2026-04-11](https://github.com/Catrobat/Catroweb/blob/develop/docs/architecture/adr-api-v2-standardization.md)).

### Authentication

- **JWT token management**: `JwtTokenStore` (EncryptedSharedPreferences), `AuthService` (Retrofit), `AuthInterceptor` (auto Bearer header + transparent 401 refresh), `LoginRepository`
- **Login flow**: `LoginDialogFragment` → coroutine-based `LoginRepository`; Google OAuth simplified to single API call; registration & password reset open share WebView
- **WebView auth**: JWT injected as `BEARER` cookie; cookie sync stores only access token; exit route (`/{flavor}/exit`) returns to app
- **Token validation**: `LoginRepository.validateToken()` checks server-side; upload flow verifies before starting

### API & Data

- **Project listings**: Cursor pagination with `{data, next_cursor, has_more}` envelope, `ImageVariants` for responsive images, `FeaturedProjectApi` split from Room entity
- **Upload**: Retrofit multipart `POST /api/projects` with Bearer auth; sends stored `project_id` so renamed projects update existing server entry instead of creating duplicates
- **Rate limit handling (HTTP 429)**: Upload and other write operations show a user-friendly "Too many requests — please wait and try again" message when the server returns 429, instead of a generic error
- **Downloads**: New `DownloadClient.kt` with shared `executeDownload`/`buildClientWithProgress` helpers; extracts server UUID from URL and writes share URL into `code.xml` `<header><url>` for Catroweb remix detection
- **WebView download fix**: `shouldInterceptRequest` now uses RFC 5987 `filename*=utf-8''` parser for Content-Disposition (v2 API returns unquoted filenames); blob URL downloads guarded in DownloadListener
- **Media library**: URLs updated to `share.catrobat.org/{flavor}/media-library`; proper Content-Disposition parsing (RFC 5987 `filename*` support)

### Domain & URL changes

- `share.catrob.at` → `share.catrobat.org` across all source files including intent filter
- `Constants.API_BASE_URL` derived from `Constants.MAIN_URL_HTTPS` — respects `WEB_TEST_FLAG`/`WEB_TEST_URL` for test/staging
- Community URLs use flavor name (`pocketcode`, `luna`, etc.) instead of hardcoded `/app/`

### Security & quality

- `ErrorInterceptor` fixed: `.string()` instead of `.toString()` for response body
- `JwtTokenStore.setAccessTokenOnly()` validates JWT format before storing (rejects non-JWT cookies)
- `LoginRepository` uses `safeAuthCall` helper to extract structured `ApiErrorResponse` from HTTP errors
- `DownloadClient.executeDownload` wraps response in `.use{}` block (prevents connection leak)
- `DownloadClient` restricts cleartext HTTP to `BuildConfig.DEBUG` only
- `LoginDialogFragment` cancels `LoginHelper` coroutine in `onDestroyView`
- `ProjectUploadActivity.verifyUserIdentity()` uses coroutine with job cancellation (replaces `Thread { runBlocking }`)
- Null-safe `listFiles()` calls in `ProjectUploadService` and `ProjectUpload`

### Code cleanup

- Extract shared UUID/URL regex to `ProjectIdUtils`
- Convert `getTags()` to coroutine
- Remove redundant `downloadedProjectIds` SharedPreferences
- Remove TOCTOU `exists()` checks
- Remove debug logging

### Session handling

- **Web logout ends the app session**: the app injects its JWT as `BEARER` cookie on every WebView visit, so logging out on the share site left the app signed in and the next visit restored the session. The WebView recognises the themed logout route (`/{theme}/logout`), revokes the token through `DELETE /api/authentication` and clears the local session with it.
- Removed the cookie-to-token sync, which adopted whatever `BEARER` cookie the web session carried. The server marks that cookie `HttpOnly`, so it was never readable.
- `tags` arriving as `[]` for empty and `{}` otherwise is handled by `FlexibleMapAdapter` instead of failing the upload with `JsonDataException`.

### Test stability

- `MainMenuFragmentTest` seeds the community sections straight into the database from the fixtures. The fragment renders what the database holds, so routing a fragment test through HTTP only made it depend on a reachable server.
- The two `assumeTrue` guards became assertions; they were turning an empty result into a silent pass.
- `testShowNoInternetMsg` pins the connection state, which the fragment's network callback used to overwrite on resume.
- `testBackButtonAfterTappingOnPlayButton` waits for the stage dialog between the two back presses. The first press was being swallowed by the system's full-screen cling, which shows once per device; the instrumentation runner confirms it up front so every test that leaves the stage stops depending on that device state.
- A featured-projects or categories sync that fails with a 4xx/5xx was silently ignored, leaving the sections empty with no trace. The status is logged now.

### Removed (17 old files)

`ServerAuthenticator`, `LoginTask`, `RegistrationTask`, 6 OAuth/check AsyncTasks, `CatrobatServerCalls`, `HttpRequests`, `GetTagsTask`, `GetSurveyTask`, `DeleteTestUserTask`, `RegistrationDialogFragment`, `OAuthUsernameDialogFragment`

Companion PR: Catrobat/Catroweb#6761 (adds `project_id` param to upload API, merged)

## Test plan

- [x] Build compiles clean (`assembleCatroidDebug`)
- [x] Unit tests pass: `AuthInterceptorTest`, `LoginRepositoryTest`, `DownloadClientTest`, `WebViewActivityLoginCookieTest`, `CookieTest`
- [x] Test fixtures updated for API v2 response format
- [x] CI green: Unit Tests, Code Analysis, Build APK, Instrumented Tests, Testrunner Tests, RTL Tests, SonarCloud
- [x] Test download from share with corrected Content-Disposition parsing
- [x] Test login end-to-end against share.catrobat.org
- [x] Test upload end-to-end (blocked by Catroweb tags fix)
- [ ] Test all build flavors (luna, create@school, etc.)
- [x] Test 429 rate limit response shows user-friendly message
- [x] `MainMenuFragmentTest` 6/6 and `CatroidWebServerTest` 11/11 pass locally on an Android 13 emulator, from a device state where the full-screen cling had not been confirmed

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dmetzner:
Review update (public-safe, no secrets): I found 4 actionable items to address in this PR.

1. [P1] `catroid/build.gradle` (line ~127)
`intentFilterHost` still points to `share.catrob.at`. New deep links from `share.catrobat.org/<flavor>/project/...` may no longer open in Catroid. Please update host (or support both during transition).

2. [P1] `catroid/src/main/java/org/catrobat/catroid/transfers/project/ProjectUpload.kt` (line ~77)
Upload call currently sends only `file` + `checksum`; no `flavor` part is provided. Catroweb API defaults missing flavor to `pocketcode`, so non-pocketcode builds can upload into the wrong flavor. Please pass `FlavoredConstants.FLAVOR_NAME` explicitly.

3. [P2] `catroid/src/main/java/org/catrobat/catroid/koin/CatroidKoinHelper.kt` (line ~68)
`API_BASE_URL` is hardcoded to production (`https://share.catrobat.org/api/`). This can mismatch test/staging web hosts (`WEB_TEST_URL`) and cause auth/session inconsistencies. Please derive API base URL from the same environment source as `Constants.MAIN_URL_HTTPS`.

4. [P2] `catroid/src/main/java/org/catrobat/catroid/ui/WebViewActivity.java` (line ~443)
Cookie sync writes `setTokens(bearerToken, bearerToken)`. If `BEARER` is access-token only, refresh token becomes invalid and refresh will fail after expiry, causing forced logout. Please avoid seeding refresh from BEARER cookie unless a true refresh token is available.

I can run another pass after updates land.

hannesweilharter:
The testcases `testProjectCategoriesWithDefaultValues` and `testProjectCategoriesResponeHasCorrectStructure` in `CatroidWebServerTest` are failing locally:

grafik 

`MainMenuFragmentTest.testCatrobatCommunitySectionIsDisplayed` is also failing:
grafik

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MainMenuFragmentTest.kt
hannesweilharter, line 1: `testDoesNotShowNoInternetMsg` and `testCatrobatCommunitySectionIsDisplayed` are failing here

hannesweilharter (commented):
please fix all warnings from github-advanced-security and detekt warnings, and have a look at the new SonarQube issues.

Also make sure to update the copyright year in every file.

dmetzner:
Rebased on the latest `develop` and addressed the review feedback in a single squashed `fix:` commit on top of the original `feat:` commit.

### What changed since the last review
- **Failing instrumented tests** now run against `MockWebServer` instead of the live `share.catrobat.org` API:
  - `CatroidWebServerTest` builds a local Retrofit pointing at the mock server. URL-format tests (`testFeaturedProjectsCallWithDefaultValues`, `testProjectCategoriesCallWithDefaultValues`) now inspect the `Request` URL instead of executing the call, so they no longer need a network at all.
  - `MainMenuFragmentTest` installs a Koin override module in `@Before` that swaps `WebService` with a cleartext Retrofit bound to `MockWebServer`. A `Dispatcher` serves the existing `featured_projects_success_response.json` and `projects_categories_response.json` fixtures, so `featuredProjectsSync` / `projectsCategoriesSync` always populate the DB even on a slow/offline emulator.
- Copyright year bumped to `2010-2026` on every file the PR touches (76 files).
- Earlier cleanup commits (PHP empty-array tags, cookie hardening, KTX `SharedPreferences.edit`, self-closing `<url/>` handling, `DownloadClient` null-body guard, lint/detekt/checkstyle/Copilot follow-ups) are folded into the same commit.

dmetzner:
@hannesweilharter rebased and fixed remaining code smells from sonarq, even if those were already in the code before. Hence, the PR contains some changes that might seem out of scope.

catroid/src/main/res/values-af/strings.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <!--
   ~ Catroid: An on-device visual programming system for Android devices
-  ~ Copyright (C) 2010-2025  The Catrobat Team
+  ~ Copyright (C) 2010-2026  The Catrobat Team
```
hannesweilharter: please remove the extra whitespace after the year - also in all other files

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MainMenuFragmentTest.kt
hannesweilharter, line 1: `testShowNoInternetMsg` and `testBackButtonAfterTappingOnPlayButton` are still failing locally

catroid/src/main/java/org/catrobat/catroid/web/DownloadClient.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    private fun buildClientWithProgress(url: String, progressCallback: (Long) -> Unit): OkHttpClient {
+        val builder = okHttpClient.newBuilder()
+        builder.networkInterceptors().add { chain ->
+            val originalResponse = chain.proceed(chain.request())
+            val body = ProgressResponseBody(originalResponse.body()) { progress ->
+                progressCallback(progress)
+            }
+            originalResponse.newBuilder().body(body).build()
+        }
```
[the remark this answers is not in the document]
hannesweilharter: https://github.com/Catrobat/Catroid/pull/5199/changes#r3154266722

please fix

catroid/src/test/java/org/catrobat/catroid/test/transfers/ProjectDownloaderTest.java
hannesweilharter, line 1: This test could be updated to the changes accordingly

hannesweilharter (changes requested):
1. Project uploads are not working. I tested it locally on a Pixel 7 API 35 emulator.
Image  



2. After trying to upload a project (which fails like above) and going back to the landing page, clicking on the `Explore our community` button results in the following:
Image  



3. Google login is not working properly. It always crashes or just returns to the login page.

https://github.com/user-attachments/assets/49db5924-fbb7-48f5-9ac0-9b663ef671b7

dmetzner:
Thanks for the detailed testing, that helped a lot. Rebased on develop and pushed fixes.

**1. Upload failing** — your test account isn't e-mail-verified. The API correctly answers `403 "Email verification required."`, but the app threw the response body away and always showed "the upload feature is currently unavailable", which is why it looked like an outage. It now parses the API error and shows what you actually have to do (verify e-mail / parental consent / suspended / session expired).

**2. `verify-pending` in Chrome** — same root cause: as an unverified user the web redirects you to the verification page, and the WebView handed that relative URL to an external browser, which then tried to resolve `verify-pending` as a host name. Only absolute `http(s)` URLs get handed off now.

While digging there I found a third one: `syncLoginStateFromCookies()` logged you out on *every* page load. The server marks `BEARER` as HttpOnly and `CookieManager.getCookie()` never returns HttpOnly cookies, so a missing value was read as "web session gone" and the local JWT got cleared. Removed.

**3. Google login** — that's config on our side, not code: the OAuth client for the debug SHA-1 / the server's `GOOGLE_ID` still needs an admin, so it can't work yet. What was definitely wrong is that it failed *silently* — in your video there's no error at all. It now honours a cancelled sign-in and reports the `ApiException` status code, so we can tell a misconfigured OAuth client (10) from a network error (7) or a cancel (12501). If you get a chance to run it again, the code in the toast/logcat would tell us exactly what to ask the admin for.

Also done:
- copyright header now has a single space after the year everywhere (that's what the checkstyle header rule expects too)
- `ProjectDownloaderTest` was a commented-out PowerMock block — replaced with real tests for the new URL/name extraction and the download queue
- `testShowNoInternetMsg` set the offline state *after* launching the activity, so it raced the observer; it's set before launch now, and the connection monitor is reset in `tearDown` so it can't leak into the next test. `testBackButtonAfterTappingOnPlayButton` got waits around the stage start. I couldn't run Espresso locally (only have an Android 37 image here and Espresso blows up on it), so please give those another try.
- the three Copilot comments (null response body, division by zero, self-closing `<url/>`) were already fixed in the pushed code, the threads are just stale.

catroid/src/androidTest/java/org/catrobat/catroid/uiespresso/ui/fragment/MainMenuFragmentTest.kt
hannesweilharter: they are still failing, ptal again

hannesweilharter:
Thank you for your updates. I tested it again and there are still some minor issues, but overall it looks good.
1. My test account is actually verified, but still the confirmation error shows up.
grafik  
  
2. It seems the **Logout** is not working properly. Even after trying to logout, the user actually doesn't get logged out and clicking on `Explore our community` redirects the user to `verify pending` in the web browser. So it is not possible for the user to open the share community page again. PTAL at the screen recording.
[logout_issue.webm](https://github.com/user-attachments/assets/2b730fa5-13c4-430f-831d-c7a567f3a581)

dmetzner:
Rebased on `develop` and pushed fixes for both points.

**1. Upload asked for e-mail confirmation on a verified account.** The API only sends that error for an account it reads as unverified, so the app was still authenticated as an older, unverified session. That is the same root cause as your second point.

**2. Logout did not log you out.** The app has no logout of its own; the one you used is the share site's, inside the WebView. It cleared the *web* session, but the app injects its JWT as a `BEARER` cookie on every WebView visit, so the next visit signed you straight back in — and since that session was unverified, `Explore our community` landed on `verify-pending` and the upload kept failing.

The WebView now recognises the themed logout route (`/{theme}/logout`), revokes the token server-side through `DELETE /api/authentication` and clears the local session with it. I also dropped the cookie-to-token sync in the same area: it adopted whatever `BEARER` cookie the web session carried, which is the same coupling seen from the other side, and the server marks that cookie `HttpOnly` so it was never readable anyway.

**On the failing tests.** I could reproduce all of them locally and they pass now — `MainMenuFragmentTest` 6/6 and `CatroidWebServerTest` 11/11 on an Android 13 emulator. Two real causes:

- `testBackButtonAfterTappingOnPlayButton`: the first back press out of the stage was being eaten by the system's full-screen cling ("Viewing full screen / To exit, swipe down from the top"). It shows once per device, so this fails on a fresh emulator and passes on CI images that have already seen it. The instrumentation runner now confirms it up front, which fixes it for every test that leaves the stage. The test also waits for the stage dialog between the two presses instead of sleeping.
- `testShowNoInternetMsg`: setting the monitor offline did not stick, because the fragment registers the network callback on resume and that immediately posts the real connectivity back. The forced value is pinned for the duration of the test now.

`MainMenuFragmentTest` no longer talks to a mock HTTP server at all. The fragment renders what the local database holds, so the fixtures are decoded and stored directly; going through HTTP only made a fragment test depend on a reachable server. The two `assumeTrue` guards are assertions now — they were turning an empty result into a silent pass, which is why `testCatrobatCommunitySectionIsDisplayed` looked green while showing nothing.

One thing worth flagging: a featured-projects or categories sync that comes back 4xx/5xx was silently ignored, so the app just showed empty sections. It logs the status now.

One caveat on my own runs: my emulator had a global HTTP proxy configured, which hijacked the mock-server traffic and produced a pile of misleading 404s until I spotted it. If you see mock-server tests failing with 404, check `settings get global http_proxy` before anything else.

Separately, `Catrobat/Catroweb#7152` fixes a 500 on the password reset page that I hit while looking into the account state. Not needed for this PR, but it does block anyone resetting a password on production right now.

PR #5222, opened 2026-06-27
Add MqttManager for broker connection lifecycle (CATROID-1671)

Paras-ydv:
## Summary

This PR implements the centralized MQTT Manager for Catroid, responsible for the full connection lifecycle of an MQTT broker client:

- [CATROID-1671](https://catrobat.atlassian.net/browse/CATROID-1671): MqttManager - connect, disconnect, state tracking, and failure handling

This ticket is scoped to **connection management only**. Message publishing, subscriptions, and MQTT bricks are deferred to separate tickets.

---

#### `org.catrobat.catroid.devices.mqtt`

- **`MqttClientInterface`** Abstracts the Paho client behind an interface, decoupling `MqttManager` from the third-party library and enabling pure JVM unit tests without an Android runtime or real network.

- **`PahoMqttClient`** Thin adapter wrapping `org.eclipse.paho.client.mqttv3.MqttClient` behind `MqttClientInterface`. Used in production.

- **`MqttClientFactory`** Fun interface for creating `MqttClientInterface` instances. `DefaultMqttClientFactory` is the production implementation that creates `PahoMqttClient`. Registered as a Koin `single` in `CatroidKoinHelper` and injected into `MqttManager`. Tests use `FakeMqttClientFactory` so no real Paho clients are created during unit tests.

- **`MqttConnectionConfig`** Data class grouping all six broker parameters (host, port, clientId, username, password, TLS flag). Includes a `fromContext()` factory that reads values from `SettingsFragment`. Keeps `connect()` to a single overload with one parameter.

- **`MqttManager`** Core manager class, registered as a Koin `single`:
  - Singleton via `by lazy`, following the same pattern as `RaspberryPiService`
  - `@Volatile` on `mqttClient` with `synchronized(this)` on `connect()` and `disconnect()` to handle concurrent access from background threads, UI thread, and Paho callbacks safely
  - `connect(config)` is idempotent if already connected, validates host, and auto-generates client ID if blank. If a stale disconnected client exists (e.g. after settings change), it is closed and nulled before a fresh one is created via the factory so updated broker URL or clientId are always picked up
  - `connectFromContext(context)` is a convenience entry point that reads config from `SettingsFragment`
  - `disconnect()` guards on `mqttClient == null` rather than `isConnected`, matching the `RaspberryPiService` pattern. This ensures cleanup runs even on dropped connections where the client exists but `isConnected` is false
  - `isConnected` exposes queryable connection state backed by the client
  - `buildServerUri()` produces a `tcp://` or `ssl://` URI based on the TLS flag
  - `buildConnectOptions()` configures clean session, connection timeout, and applies credentials when a username is provided
  - On connect failure, the client is closed and nulled so no stale reference is kept

### Testing

Added `MqttManagerTest` with 30 unit tests using `FakeMqttClient` and `FakeMqttClientFactory` as hand-written test doubles with no Mockito and no emulator needed:

- Singleton is non-null and returns the same object on repeated access
- `isConnected` is false before any call
- `connect()` returns true on success, calls the client, and sets the callback
- `connect()` returns false without crashing when `MqttException` is thrown, and calls `close()` for cleanup
- Already connected returns true without reconnecting; blank host returns false without touching the client; empty client ID auto-generates one
- Stale disconnected client is closed and a fresh one is created via the factory on reconnect; a live client is not closed
- `tcp://` vs `ssl://` scheme selection and full URI format verified
- `isCleanSession` is always true; credentials are set when username is provided, password can be empty
- `disconnect()` calls both `disconnect()` and `close()` on the client, cleans up dropped connections, skips safely when no client exists, and is safe to call twice

---

#### Your Checklist
- [x] Include the name of the Jira ticket in the PR's title
- [x] Confirm that the changes follow the project's coding guidelines
- [x] Check that the commits' message style matches the [project's guideline](https://github.com/Catrobat/Catroid/wiki/Branch-and-Commit-Message-Guidelines)
- [x] Stick to the project's gitflow workflow

[CATROID-1671]: https://catrobat.atlassian.net/browse/CATROID-1671

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttConnectionConfig.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttConnectionConfig.kt
@@ -1,0 +1,47 @@
+/*
+ * Catroid: An on-device visual programming system for Android devices
+ * Copyright (C) 2010-2026 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * An additional term exception under section 7 of the GNU Affero
+ * General Public License, version 3, is available at
+ * http://developer.catrobat.org/license_additional_term
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.catrobat.catroid.devices.mqtt
+
+import android.content.Context
+import org.catrobat.catroid.ui.settingsfragments.SettingsFragment
+
+data class MqttConnectionConfig(
+    val host: String,
+    val port: Int,
+    val clientId: String,
+    val username: String,
+    val password: String,
+    val useTls: Boolean
+) {
+    companion object {
+        fun fromContext(context: Context) = MqttConnectionConfig(
+            host = SettingsFragment.getMqttHost(context),
+            port = SettingsFragment.getMqttPort(context),
+            clientId = SettingsFragment.getMqttClientId(context),
+            username = SettingsFragment.getMqttUsername(context),
+            password = SettingsFragment.getMqttPassword(),
+            useTls = SettingsFragment.isMqttTlsEnabled(context)
+        )
+    }
+}
--- a/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
@@ -32,7 +32,10 @@ import org.eclipse.paho.client.mqttv3.MqttConnectOptions
 import org.eclipse.paho.client.mqttv3.MqttException
 import org.eclipse.paho.client.mqttv3.MqttMessage
 
-class MqttManager(private var mqttClient: MqttClientInterface? = null) {
+class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClientFactory) {
+
+    @Volatile
+    private var mqttClient: MqttClientInterface? = null
 
     val isConnected: Boolean
         get() = mqttClient?.isConnected == true
@@ -49,16 +52,25 @@ class MqttManager(private var mqttClient: MqttClientInterface? = null) {
 
     fun connectFromContext(context: Context) = connect(MqttConnectionConfig.fromContext(context))
 
-    fun connect(config: MqttConnectionConfig): Boolean {
+    fun connect(config: MqttConnectionConfig): Boolean = synchronized(this) {
         if (isConnected) return true
         if (config.host.isBlank()) {
             Log.e(TAG, "Cannot connect: host is blank")
             return false
         }
+        val currentClient = mqttClient
+        if (currentClient != null && !currentClient.isConnected) {
+            try {
+                currentClient.close()
+            } catch (e: MqttException) {
+                Log.e(TAG, "Failed to close stale client before reconnect", e)
+            }
+            mqttClient = null
+        }
         return try {
             val brokerUrl = buildServerUri(config.host, config.port, config.useTls)
             val resolvedClientId = config.clientId.ifEmpty { MqttClient.generateClientId() }
-            val client = mqttClient ?: PahoMqttClient(brokerUrl, resolvedClientId).also { mqttClient = it }
+            val client = mqttClient ?: clientFactory.create(brokerUrl, resolvedClientId).also { mqttClient = it }
             client.setCallback(callback)
             client.connect(buildConnectOptions(config.username, config.password))
             val connected = client.isConnected
@@ -77,15 +89,17 @@ class MqttManager(private var mqttClient: MqttClientInterface? = null) {
     }
 
     fun disconnect() {
-        if (!isConnected) return
-        try {
-            mqttClient?.disconnect()
-            mqttClient?.close()
-            Log.d(TAG, "Disconnected and closed client")
-        } catch (e: MqttException) {
-            Log.e(TAG, "Error during disconnect", e)
-        } finally {
-            mqttClient = null
+        synchronized(this) {
+            if (mqttClient == null) return
+            try {
+                mqttClient?.disconnect()
+                mqttClient?.close()
+                Log.d(TAG, "Disconnected and closed client")
+            } catch (e: MqttException) {
+                Log.e(TAG, "Error during disconnect", e)
+            } finally {
+                mqttClient = null
+            }
         }
     }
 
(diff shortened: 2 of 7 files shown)


catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return try {
+            val brokerUrl = buildServerUri(config.host, config.port, config.useTls)
+            val resolvedClientId = config.clientId.ifEmpty { MqttClient.generateClientId() }
+            val client = mqttClient ?: PahoMqttClient(brokerUrl, resolvedClientId).also { mqttClient = it }
+            client.setCallback(callback)
+            client.connect(buildConnectOptions(config.username, config.password))
+            val connected = client.isConnected
+            Log.d(TAG, "Connect result for clientId=$resolvedClientId at $brokerUrl: connected=$connected")
+            connected
```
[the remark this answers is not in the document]
Paras-ydv: Thanks for pointing this out. The current implementation assumes that configuration changes are preceded by a `disconnect()`, which is the intended lifecycle for this ticket. **CATROID-1671** focuses on connection lifecycle management only, and broker configuration changes are handled by reconnecting with a new client. I'll keep this in mind for future enhancements if dynamic reconfiguration becomes a requirement.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -0,0 +1,115 @@
...
+        return "$scheme://$host:$port"
+    }
+
+    internal fun buildConnectOptions(username: String, password: String) = MqttConnectOptions().apply {
+        isCleanSession = true
+        connectionTimeout = CONNECTION_TIMEOUT
+        if (username.isNotBlank()) {
+            this.userName = username
+            this.password = password.toCharArray()
+        }
```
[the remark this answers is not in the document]
Paras-ydv: The implementation intentionally allows an empty password when a username is provided, since this is supported by MQTT brokers. I'll update the PR description to reflect the actual behavior.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttConnectionConfig.kt
```diff
@@ -0,0 +1,47 @@
...
+    val password: String,
+    val useTls: Boolean
+) {
+    companion object {
+        fun fromContext(context: Context) = MqttConnectionConfig(
+            host = SettingsFragment.getMqttHost(context),
+            port = SettingsFragment.getMqttPort(context),
+            clientId = SettingsFragment.getMqttClientId(context),
+            username = SettingsFragment.getMqttUsername(context),
+            password = SettingsFragment.getMqttPassword(),
```
harshsomankar123-tech, line 43: It looks like we missed passing the context parameter here.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -0,0 +1,115 @@
...
+import android.content.Context
+import android.util.Log
+import org.eclipse.paho.client.mqttv3.IMqttDeliveryToken
+import org.eclipse.paho.client.mqttv3.MqttCallback
+import org.eclipse.paho.client.mqttv3.MqttClient
+import org.eclipse.paho.client.mqttv3.MqttConnectOptions
+import org.eclipse.paho.client.mqttv3.MqttException
+import org.eclipse.paho.client.mqttv3.MqttMessage
+
+class MqttManager(private var mqttClient: MqttClientInterface? = null) {
```
harshsomankar123-tech: Since MqttManager is accessed via the singleton instance from different threads (like background executors, UI thread, or callbacks), we should synchronize the mutable client state and connection methods to avoid potential race conditions.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -0,0 +1,115 @@
...
+            } catch (closeEx: MqttException) {
+                Log.e(TAG, "Failed to close client after connect error", closeEx)
+            }
+            mqttClient = null
+            false
+        }
+    }
+
+    fun disconnect() {
+        if (!isConnected) return
```
harshsomankar123-tech: If the connection gets dropped, Paho's isConnected will return false. Because of this early return, calling disconnect() won't actually close/clean up the client, which will leak Paho's background threads and keep the stale client reference active.

Following the pattern in RaspberryPiService.java where we disconnect and clean up based on whether the instance is null, we should change this check to: if (mqttClient == null) return


Image

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -0,0 +1,115 @@
...
+        private const val SSL_SCHEME = "ssl"
+        private const val CONNECTION_TIMEOUT = 5
+
+        @JvmStatic
+        val instance: MqttManager by lazy { MqttManager() }
+    }
+
+    fun connectFromContext(context: Context) = connect(MqttConnectionConfig.fromContext(context))
+
+    fun connect(config: MqttConnectionConfig): Boolean {
```
harshsomankar123-tech: Right now, if we have an existing client instance that is disconnected, we reuse it directly. However, if the user updated the broker URL or client ID in settings, those changes won't be picked up because the old client still points to the old configurations.

When starting a new connection attempt, we should check if we have a stale/disconnected client, close it, and instantiate a fresh one so the new settings are correctly applied.

harshsomankar123-tech (changes requested):
@Paras-ydv Thanks!

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttConnectionConfig.kt
Paras-ydv: `getMqttPassword()` intentionally takes no context, it uses `Koin` DI to inject `MqttPasswordRepository`, which wraps `EncryptedSharedPreferences`. This is consistent with how `DefaultMqttPasswordRepository` is registered in `CatroidKoinHelper`. I think no change is needed here.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: Fixed. Added `@Volatile` on `mqttClient` and wrapped both `connect()` and `disconnect()` in `synchronized(this)` to prevent races on concurrent calls.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: Fixed. Changed from `if (!isConnected)` `return` to `if (mqttClient == null)` `return`, matching the pattern in `RaspberryPiService.java`. This ensures cleanup runs even on dropped connections where `isConnected` is false but the client object still exists.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: Fixed. As we discussed: introduced `MqttClientFactory` (fun interface) and `DefaultMqttClientFactory` (object wrapping `PahoMqttClient`). `MqttManager` now takes the factory as a constructor param and is registered as a `Koin` single in `CatroidKoinHelper`. Stale client cleanup: before reconnecting, if mqttClient != null && !isConnected, we close and null it, then create a fresh one via the factory. Tests use `FakeMqttClientFactory` returning FakeMqttClient, no real Paho clients in unit tests.

wslany (approved):

PR #5220, opened 2026-06-21
Add MQTT dependencies, settings UI, and test coverage (CATROID-1670, CATROID-1672)

Paras-ydv:
## Summary

This PR introduces the initial MQTT foundation for Catroid by combining:

- [CATROID-1670](https://catrobat.atlassian.net/browse/CATROID-1670): Eclipse Paho MQTT dependency and Android configuration
- [CATROID-1672](https://catrobat.atlassian.net/browse/CATROID-1672): MQTT settings UI, persistence layer, validation, and automated tests

#### catroid/build.gradle

- Added `org.eclipse.paho.client.mqttv3:1.2.5`
- Added `org.eclipse.paho.android.service:1.1.1`
- Added `FEATURE_MQTT_ENABLED` build config flag

#### AndroidManifest.xml

- Registered `org.eclipse.paho.android.service.MqttService`
- Verified existing permissions:
  - INTERNET
  - ACCESS_NETWORK_STATE

### Testing

Added dependency sanity tests verifying:

- MqttClient
- MqttConnectOptions
- MqttMessage
- MqttException
- MqttCallback

are available on the test classpath.

### UI Demo

[Watch UI Demo](https://drive.google.com/file/d/1yLadrlKaI-Id3SxY3azNVey9TVCfbOPD/view?usp=sharing)

### Features Implemented

Added MQTT settings screen with:

- MQTT enable toggle
- Broker host
- Broker port
- TLS option
- Username
- Password
- Client ID

### Persistence

Implemented persistent storage using SharedPreferences.

Added accessor APIs:

- isMqttSharedPreferenceEnabled()
- getMqttHost()
- getMqttPort()
- isMqttTlsEnabled()
- getMqttUsername()
- getMqttPassword()
- getMqttClientId()

### Password Storage Refactor (MqttPasswordRepository)

Following a discussion with Harsh and Prof. Wolfgang's review feedback, the MQTT password is no longer stored in plain SharedPreferences.

Instead, a `MqttPasswordRepository` interface was introduced (modelled after `LocalHashVersionRepository`) with a
`DefaultMqttPasswordRepository` implementation backed by `EncryptedSharedPreferences`. It is registered as a Koin singleton and injected wherever the password is read or written.

This keeps `SettingsFragment.getMqttPassword(Context)` as the existing public API — it now delegates to the injected repository internally. Unit tests mock the interface via a Koin test module, so no Android
Keystore is required in the test environment.

### Validation

- Port range: 1–65535
[...]

PR #5225, opened 2026-07-05
Implement MQTT Message Publishing in MqttManager (CATROID-1673)

Paras-ydv:
## Summary

Extends `MqttManager` to support publishing MQTT messages to a broker using Eclipse Paho.

This PR previously carried the dependency and settings groundwork (CATROID-1670, CATROID-1672) and the connection lifecycle (CATROID-1671) as extra commits, because none of them were on `develop` yet. All three have since been merged ([#5220](https://github.com/Catrobat/Catroid/pull/5220), [#5222](https://github.com/Catrobat/Catroid/pull/5222)), so the branch has been rebased onto `develop` and those commits removed. What remains is a single commit containing only the publish work.

### Changes

- Added `publish(topic, message)` to `MqttClientInterface` and implemented it in `PahoMqttClient`
- `MqttManager.publish()` validates the topic (rejects blank, rejects the `#` and `+` wildcards, which are only legal in subscriptions) and the QoS range (0–2)
- Lazy-connects before publishing when no connection is active
- `publishFromContext()` convenience wrapper for the brick layer
- No forced unwraps; failures are logged and returned as `false` rather than thrown
- `connectionLost` now passes the `Throwable` to `Log.e` so the stacktrace is kept

### Tests

24 new tests covering publish and message building:

- Successful publish, and that the correct topic and payload reach the client
- QoS 0, 1 and 2 accepted; out-of-range QoS rejected
- Retained flag set both ways
- Blank topic rejected, and the client is not called
- Wildcard topics rejected (`#` and `+`)
- Lazy connect succeeds before publishing
- No reconnect when a connection already exists
- No publish when the lazy connect fails
- Client exception returns `false` without crashing
- `buildMessage()` sets payload, QoS and retained correctly

Two existing tests were updated in this rebase. `MqttManager` now takes an `MqttClientFactory` rather than a client instance, following the change made during the CATROID-1671 review, so tests that assumed an injected client had to connect first. Without that, `testPublishWhenAlreadyConnectedDoesNotReconnect` failed and `testSubscribeAfterDisconnectSucceedsAgain` did not compile.

Full suite: 4709 tests, 0 failures.

## Acceptance criteria

- [x] `publish()` API added to `MqttManager`
- [x] `publish()` added to `MqttClientInterface` and `PahoMqttClient`
- [x] Messages published successfully to a connected broker
- [x] Lazy connection established automatically if none exists
- [x] Configurable QoS and retained flag
- [x] Failures logged without crashing
- [x] Unit tests covering the success paths, the validation rejections and the failure modes
- [x] All existing and new tests pass

## Your Checklist

- [x] Include the name of the Jira ticket in the PR's title
- [x] Confirm that the changes follow the project's coding guidelines
- [x] Check that the commits' message style matches the [project's guideline](https://github.com/Catrobat/Catroid/wiki/Branch-and-Commit-Message-Guidelines)
- [x] Stick to the project's gitflow workflow

### DIFF
--- a/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttClientInterface.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttClientInterface.kt
@@ -25,6 +25,7 @@ package org.catrobat.catroid.devices.mqtt
 
 import org.eclipse.paho.client.mqttv3.MqttCallback
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions
+import org.eclipse.paho.client.mqttv3.MqttMessage
 
 interface MqttClientInterface {
     val isConnected: Boolean
@@ -32,4 +33,5 @@ interface MqttClientInterface {
     fun disconnect()
     fun close()
     fun setCallback(callback: MqttCallback)
+    fun publish(topic: String, message: MqttMessage)
 }
--- a/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
+++ b/catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
@@ -88,6 +88,36 @@ class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClie
         }
     }
 
+    fun publishFromContext(context: Context, topic: String, payload: String, qos: Int = 0, retained: Boolean = false) =
+        publish(MqttConnectionConfig.fromContext(context), topic, payload, qos, retained)
+
+    fun publish(config: MqttConnectionConfig, topic: String, payload: String, qos: Int = 0, retained: Boolean = false): Boolean {
+        if (topic.isBlank()) {
+            Log.e(TAG, "Cannot publish: topic is blank")
+            return false
+        }
+        if (topic.contains('#') || topic.contains('+')) {
+            Log.e(TAG, "Cannot publish: topic contains wildcard characters")
+            return false
+        }
+        if (qos !in 0..2) {
+            Log.e(TAG, "Cannot publish: invalid QoS value $qos")
+            return false
+        }
+        if (!isConnected && !connect(config)) {
+            Log.e(TAG, "Cannot publish: connection failed")
+            return false
+        }
+        val client = mqttClient ?: return false
+        return try {
+            client.publish(topic, buildMessage(payload, qos, retained))
+            true
+        } catch (e: MqttException) {
+            Log.e(TAG, "Failed to publish to '$topic'", e)
+            false
+        }
+    }
+
     fun disconnect() {
         synchronized(this) {
             if (mqttClient == null) return
@@ -117,13 +147,16 @@ class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClie
         }
     }
 
+    internal fun buildMessage(payload: String, qos: Int, retained: Boolean) = MqttMessage(payload.toByteArray()).apply {
+        this.qos = qos
+        isRetained = retained
+    }
+
     private val callback = object : MqttCallback {
         override fun connectionLost(cause: Throwable?) {
-            Log.e(TAG, "Connection lost: ${cause?.message}")
+            Log.e(TAG, "Connection lost", cause)
         }
-        // Message handling is implemented in a later ticket.
         override fun messageArrived(topic: String, message: MqttMessage) = Unit
-        // Delivery tokens are not used until publish is implemented in a later ticket.
         override fun deliveryComplete(token: IMqttDeliveryToken?) = Unit
     }
 }
--- a/catroid/src/test/java/org/catrobat/catroid/test/mqtt/MqttManagerTest.kt
+++ b/catroid/src/test/java/org/catrobat/catroid/test/mqtt/MqttManagerTest.kt
@@ -29,6 +29,7 @@ import org.catrobat.catroid.devices.mqtt.MqttConnectionConfig
 import org.catrobat.catroid.devices.mqtt.MqttManager
 import org.eclipse.paho.client.mqttv3.MqttCallback
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions
+import org.eclipse.paho.client.mqttv3.MqttMessage
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNotNull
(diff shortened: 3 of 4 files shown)


catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -88,6 +88,36 @@ class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClie
...
+        }
+        if (qos !in 0..2) {
+            Log.e(TAG, "Cannot publish: invalid QoS value $qos")
+            return false
+        }
+        if (!isConnected && !connect(config)) {
+            Log.e(TAG, "Cannot publish: connection failed")
+            return false
+        }
+        val client = mqttClient ?: return false
```
harshsomankar123-tech, line 111: add a quick Log.e here before returning false so we don't fail silently if mqttClient happens to be null.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -88,6 +88,36 @@ class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClie
...
+        return try {
+            client.publish(topic, buildMessage(payload, qos, retained))
+            true
+        } catch (e: MqttException) {
```
harshsomankar123-tech, line 115: Paho's internal validator also throws an unchecked IllegalArgumentException on malformed topics (like strings with null bytes \0). Add catch IllegalArgumentException here too so the app doesn't crash.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
```diff
@@ -88,6 +88,36 @@ class MqttManager(private val clientFactory: MqttClientFactory = DefaultMqttClie
         }
     }
 
+    fun publishFromContext(context: Context, topic: String, payload: String, qos: Int = 0, retained: Boolean = false) =
+        publish(MqttConnectionConfig.fromContext(context), topic, payload, qos, retained)
+
+    fun publish(config: MqttConnectionConfig, topic: String, payload: String, qos: Int = 0, retained: Boolean = false): Boolean {
```
harshsomankar123-tech, line 94: connect() and disconnect() are synchronized, but publish() isn't. If someone calls disconnect() while a publish is in progress, the client can get closed mid-call. we need to  make publish() synchronized(this) as well.

Also, heads up that client.publish() is a blocking network call. If this gets called from the UI/main thread, Android will throw a NetworkOnMainThreadException. We should ensure this runs on a background dispatcher (Dispatchers.IO).

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttClientInterface.kt
```diff
@@ -25,11 +25,13 @@ package org.catrobat.catroid.devices.mqtt
 
 import org.eclipse.paho.client.mqttv3.MqttCallback
 import org.eclipse.paho.client.mqttv3.MqttConnectOptions
+import org.eclipse.paho.client.mqttv3.MqttMessage
```
harshsomankar123-tech, line 28: Since this interface exists to decouple us from Paho, We need to avoid passing MqttMessage here. Passing (topic: String, payload: ByteArray, qos: Int, retained: Boolean) keeps Paho details out of the interface.

catroid/src/test/java/org/catrobat/catroid/test/mqtt/MqttManagerTest.kt
```diff
@@ -256,6 +257,168 @@ class MqttManagerTest {
...
+
+    @Test
+    fun testPublishReturnsFalseWhenClientThrows() {
+        fakeClient.connected = true
+        fakeClient.throwOnPublish = true
+        assertFalse(manager.publish(defaultConfig, "home/temp", "22"))
+    }
+
+    @Test
+    fun testPublishDoesNotCrashWhenClientThrows() {
```
harshsomankar123-tech, line 350: This test is an exact duplicate of testPublishReturnsFalseWhenClientThrows
Do we need this one???

harshsomankar123-tech (commented):
Hi @Paras-ydv, Thanks!  some blockers from my side.  PTAL

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: Added a Log.e there, thanks.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: Added. One thing worth mentioning: I tested this against a local Mosquitto and Paho 1.2.5 doesn't actually reject null bytes itself, it sends the topic and the broker drops the connection (MqttException 32109). Where it does throw IllegalArgumentException is a topic longer than 65535 bytes. So the catch is definitely needed, just for a slightly different case than we thought. I've also added a null character check to our own validation so it never gets that far.

catroid/src/test/java/org/catrobat/catroid/test/mqtt/MqttManagerTest.kt
Paras-ydv: You're right, it added nothing over testPublishReturnsFalseWhenClientThrows. Removed.

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttClientInterface.kt
Paras-ydv: Good point, changed it to publish(topic, payload, qos, retained) and moved building the MqttMessage into PahoMqttClient. connect() and setCallback() still take Paho types from the merged #5222, want me to clean those up here as well or keep it to a separate PR?

catroid/src/main/java/org/catrobat/catroid/devices/mqtt/MqttManager.kt
Paras-ydv: On the race: I tried to reproduce it by running publishes in a loop while another thread called disconnect() and close(), about 12,000 publishes across 300 rounds. Every failure came back as MqttException (32104, "Client is not connected"), which the existing catch already handles, so publish() just returns false. Nothing unchecked escaped. The client is also read into a local val first, so it can't become null halfway through.

I am thinking about locking is the other direction: disconnect() runs on the UI thread when the stage closes, so it would have to wait for any in-flight QoS 1 or 2 publish to be acknowledged by the broker before it could return. That felt like trading a crash that doesn't happen for a freeze that does. Happy to add it if you still want it, just wanted to show the numbers first.

On the threading: agreed it mustn't run on the main thread. One correction though, I instrumented the socket to check, and Paho does the connect and all the writes on its own internal threads (MQTT Con and MQTT Snd), so the calling thread never touches the network and NetworkOnMainThreadException won't fire. It does block waiting for the result, which is just as much of a problem on the main thread.

Nothing calls publish() yet in this PR. In the brick PR (#5240) it runs on a dedicated background executor. Would you rather I keep MqttManager synchronous and leave threading to the caller, or make it a suspend function on Dispatchers.IO? The second one is cleaner but StageResourceHolder is Java and can't call suspend functions, so I wanted to check before changing the API.

PR #4853, opened 2023-08-04
CATROID-1578 REFACTOR SetThreadColorAction to Kotlin

Hugoluk:
Refactored SetThreadColorAction class to kotlin and added Unit tests

https://catrobat.atlassian.net/browse/CATROID-1578

### Your checklist for this pull request
Please review the [contributing guidelines](https://github.com/Catrobat/Catroid/blob/develop/README.md) and [wiki pages](https://github.com/Catrobat/Catroid/wiki/) of this repository.

- [x] Include the name of the Jira ticket in the PR’s title
- [x] Include a summary of the changes plus the relevant context
- [x] Choose the proper base branch (*develop*)
- [x] Confirm that the changes follow the project’s coding guidelines
- [x] Verify that the changes generate no compiler or linter warnings
- [x] Perform a self-review of the changes
- [x] Verify to commit no other files than the intentionally changed ones
- [x] Include reasonable and readable tests verifying the added or changed behavior
- [x] Confirm that new and existing unit tests pass locally
- [x] Check that the commits’ message style matches the [project’s guideline](https://github.com/Catrobat/Catroid/wiki/Commit-Message-Guidelines)
- [x] Stick to the project’s gitflow workflow
- [x] Verify that your changes do not have any conflicts with the base branch
- [x] After the PR, verify that all CI checks have passed
- [ ] Post a message in the *catroid-stage* or *catroid-ide* [Slack channel](https://catrobat.slack.com) and ask for a code reviewer

PR #5234, opened 2026-08-09
CATROID-1578 REFACTOR SetThreadColorAction to Kotlin

hannesweilharter:
- Refactor `SetThreadColorAction` to Kotlin and simplify
- Added tests `SetThreadColorActionTest`

**Note**: There is an [older Pull request](https://github.com/Catrobat/Catroid/pull/4853) for this ticket which i also updated, but the Github Actions Workflow did not run for some reason, so I created a new one

https://catrobat.atlassian.net/browse/CATROID-1578


catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,62 @@
...
+    private var scope: Scope? = null
+    private var color: Formula? = null
+    private var sprite: Sprite? = null
+
+    override fun update(delta: Float) {
+        try {
+            val colorString = color?.interpretString(scope) ?: "#ff0000"
+
+            sprite?.embroideryThreadColor = Color.valueOf(colorString)
+        } catch (exception: RuntimeException) {
```
Paras-ydv: `RuntimeException` is too broad here. Please catch only the specific exception(s) that `Formula.interpretString()` can throw. Catching `RuntimeException` can silently take in unrelated bugs and makes it harder to reason about failure modes.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.catrobat.catroid.formulaeditor.Formula
+
+open class SetThreadColorAction : TemporalAction() {
+    private var scope: Scope? = null
+    private var color: Formula? = null
+    private var sprite: Sprite? = null
+
+    override fun update(delta: Float) {
+        try {
+            val colorString = color?.interpretString(scope) ?: "#ff0000"
```
Paras-ydv: The default color `"#ff0000"` is a magic string. Consider extracting it as a named constant to make the intent clear and keep it in sync with the test's `DEFAULT_COLOR = Color.RED`.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,62 @@
...
+
+            sprite?.embroideryThreadColor = Color.valueOf(colorString)
```
Paras-ydv: If `sprite` is `null` this silently does nothing with no `log` or `error`. This could hide misconfiguration. Consider adding a `log` warning when `sprite == null`, similar to how the exception is logged below.

catroid/src/test/java/org/catrobat/catroid/test/content/actions/SetThreadColorActionTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        assertEquals(INITIAL_COLOR, sprite.embroideryThreadColor)
+    }
+
+    @Test
+    fun testInvalidColorCodeDoesNotChangeColor() {
+        createAction(Formula("#XXXXXX")).act(1f)
+
+        assertEquals(INITIAL_COLOR, sprite.embroideryThreadColor)
+    }
+
```
Paras-ydv, line 89: There's no test case for when `sprite` is `null`. Given the silent null-safe call on `sprite`, a test documenting the expected behavior in that case would be valuable.

Paras-ydv (changes requested):
@hannesweilharter Thanks. PTAL.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,86 @@
...
+        } catch (exception: InterpretationException) {
+            Log.d(
+                javaClass.simpleName,
+                "Formula interpretation for this specific Brick failed.",
+                exception
+            )
+        }
+
+        try {
+            sprite?.embroideryThreadColor = Color.valueOf(colorString as String)
```
Paras-ydv: If `interpretString` throws an `InterpretationException`, the `catch` block returns `Unit` (return value of `Log.d`), making `colorString` type `Any`. Then this cast crashes with a `ClassCastException` at runtime  which is actually worse than the original Java that just logged and skipped silently. The fix is to `return` in the catch block so the color stays unchanged on failure, matching the original behavior.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,86 @@
...
+
+import android.util.Log
+import com.badlogic.gdx.graphics.Color
+import com.badlogic.gdx.scenes.scene2d.actions.TemporalAction
+import org.catrobat.catroid.content.Scope
+import org.catrobat.catroid.content.Sprite
+import org.catrobat.catroid.formulaeditor.Formula
+import org.catrobat.catroid.formulaeditor.InterpretationException
+
+open class SetThreadColorAction : TemporalAction() {
```
Paras-ydv: No subclasses of `SetThreadColorAction` exist in the codebase. Kotlin classes are `final` by default for a reason, so `open` should be removed unless there's a planned use for it.

Paras-ydv (changes requested):
Hey @hannesweilharter thanks for the earlier changes, I am mentioning some minor fixes otherwise LGTM.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
```diff
@@ -0,0 +1,87 @@
...
+            Log.d(
+                javaClass.simpleName,
+                "Formula interpretation for this specific Brick failed.",
+                exception
+            )
+            return
+        }
+
+        try {
+            sprite?.embroideryThreadColor = Color.valueOf(colorString)
```
moe-threeface, line 56: Just wondering: can Color.valueOf(colorString) also throw an IllegalArgumentException for some invalid color strings?

In the old Java version the whole conversion was inside a catch (Exception), so I’m a bit worried that this migration could accidentally let some invalid color inputs crash instead of only logging them.

Could be a misunderstanding on my side, just better to check.

moe-threeface (commented):
I only had one small concern about the exception handling during the Java to Kotlin migration. Other than that the change looks good to me.

catroid/src/main/java/org/catrobat/catroid/content/actions/SetThreadColorAction.kt
hannesweilharter: valid concern 👍
I replaced the `NumberFormatException` with `IllegalArgumentException`, since it is inherited, so it will catch both.

PR #5239, opened 2026-08-22
CATROID-1106: GlideToPositionBrick

rinaschedl:
https://catrobat.atlassian.net/browse/CATROID-1106

A new Brick (Feature) is created.
Brick Behavior: 
The Brick has 3 options:
-random position
	- the sprite glides to a random position 
-touch position
	- the player has to touch a position where the sprite then goes 
-other sprite position
	- the sprite glides (and at the moment follows) a other sprite till the other sprite stops.
	- this part is not finally discussed: may we change it so that the sprite goes to the startposition from the other sprite and don't follow at all.

Program changes:
create 3 Actions and 3 Actionstest, GlideToPositionBrick, brick_glide_to_position
changes in BrickValue, ActionFactory

Commitlist during coding:
TestGlideToRandomPositionDestination Test; flact:red
TestGlideToRandomPositionDestination flact:green; createGlidetoRandomPositionAction;
Refactoring and knew Test: GlideBehavior, bove Tests are green
show Brick in Bricklist with the correct attributs. Attributes don't work yet.
changed all glide_to_.... to glide_to_position_... so to better distinguish between.
last commit before PairProgramming Session with Thomas
glideToTouchPosition ready, brick test by hand and the tests are grenn
refactor the ActionTest: canceled all deprecated lines; correct misspellings;
add GlideToOtherSpritePosition
refactor GlideToRandomPositionAction.kt
Refactoring all files that changed in the ticket Catroid-1106


hannesweilharter:
please fix all warnings first and have a look at the Unit Test Report, there are failing testcases. Not sure if they are related to these changes. Then i can do a review :)

rinaschedl:
@hannesweilharter  please start with CodeReview.
the issue on the Quality Gate is discussed with Thomas and I shall ingnore it for now. There will be a ticket for the Action Faktory because to change the warnings, you have also some changing to the Unit Tests necessary and the workload is relativly high, therefor we changed the ActionFactory back, so you see the warnings but the tests mostly pass now. The 2 failing Tests has nothing to too with the changes I made in the code, so far I can see, so I guess there one of the flanky ones? Lokal all Tests passed, exept some of them who Thomas confirmed are broken. So we should be good.
Thanks
Katharina

.idea/.name
hannesweilharter, line 1: why deleting this file?

gradle/gradle-daemon-jvm.properties
hannesweilharter, line 1: whats the purpose of this file?

catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java
```diff
@@ -375,6 +379,35 @@ public Action createGlideToAction(Sprite sprite, SequenceAction sequence, Formul
...
+			Sprite destinationSprite, SequenceAction sequence,
+			Formula duration, int spinnerSelection) throws InterpretationException {
+		Scope scope = new Scope(ProjectManager.getInstance().getCurrentProject(), sprite, sequence);
+		switch (spinnerSelection) {
+			case BrickValues.GLIDE_TO_RANDOM_POSITION:
+				GlideToRandomPositionAction randomAction =
+						action(GlideToRandomPositionAction.class);
+				randomAction.setScope(scope);
+				randomAction.setDuration(duration.interpretFloat(scope));
+				return randomAction;
```
hannesweilharter, line 392: I would suggest setting the scope and duration after the switch statement to remove code duplication.

catroid/src/main/java/org/catrobat/catroid/content/Scope.kt
```diff
@@ -27,6 +27,6 @@ import com.badlogic.gdx.scenes.scene2d.actions.SequenceAction
 
 data class Scope(
     val project: Project?,
-    val sprite: Sprite,
+    var sprite: Sprite,
```
hannesweilharter, line 30: why changing `val` to `var`?

catroid/src/main/java/org/catrobat/catroid/content/actions/GlideToOtherSpritePositionAction.kt
hannesweilharter, line 1: IMO it would make sense to introduce a common `GlideToPositionAction` parent class to reduce duplicated code between the three position-based glide actions. They all store a start position and current position etc.
There is also an already existing `GlideToAction` class which might contain some overlapping gliding logic. I haven't looked into the concrete behavior in detail, so please check whether the existing class can be reused somehow.

catroid/src/main/java/org/catrobat/catroid/content/bricks/GlideToPositionBrick.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        if (spinnerSelection == BrickValues.GLIDE_TO_RANDOM_POSITION) {
+            spinner.setSelection(0)
+        }
+        if (spinnerSelection == BrickValues.GLIDE_TO_TOUCH_POSITION) {
+            spinner.setSelection(1)
+        }
+        if (spinnerSelection == BrickValues.GLIDE_TO_OTHER_SPRITE_POSITION) {
+            spinner.setSelection(destinationSprite)
+        }
```
hannesweilharter:
```suggestion
        when (spinnerSelection) {
            BrickValues.GLIDE_TO_RANDOM_POSITION -> 
                spinner.setSelection(0)
            BrickValues.GLIDE_TO_TOUCH_POSITION ->
                spinner.setSelection(1)
            BrickValues.GLIDE_TO_OTHER_SPRITE_POSITION ->
                spinner.setSelection(destinationSprite)
        }
```

catroid/src/main/java/org/catrobat/catroid/content/bricks/GlideToPositionBrick.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        if (string == context.getString(R.string.brick_glide_to_touch_position)) {
+            spinnerSelection = BrickValues.GLIDE_TO_TOUCH_POSITION
+            destinationSprite = null
+        }
+
+        if (string == context.getString(R.string.brick_glide_to_random_position)) {
+            spinnerSelection = BrickValues.GLIDE_TO_RANDOM_POSITION
+            destinationSprite = null
+        }
```
hannesweilharter:
```suggestion
        when (string) {
            context.getString(R.string.brick_glide_to_touch_position) -> {
                spinnerSelection = BrickValues.GLIDE_TO_TOUCH_POSITION
                destinationSprite = null
            }
            context.getString(R.string.brick_glide_to_random_position) -> {
                spinnerSelection = BrickValues.GLIDE_TO_RANDOM_POSITION
                destinationSprite = null
            }
        }
```

catroid/src/main/java/org/catrobat/catroid/ui/fragment/CategoryBricksFactory.kt
```diff
@@ -477,9 +479,9 @@ open class CategoryBricksFactory {
             GlideToBrick(
                 BrickValues.X_POSITION,
                 BrickValues.Y_POSITION,
-                BrickValues.GLIDE_SECONDS
-            )
+                GLIDE_SECONDS)
```
hannesweilharter:
```suggestion
                BrickValues.GLIDE_SECONDS
```
I would keep the BrickValues usage consistent here. `GLIDE_SECONDS` is imported directly, while the other constants use `BrickValues.*`. I’d suggest using `BrickValues.GLIDE_SECONDS` here as well, since the other constants are already referenced that way.

hannesweilharter (changes requested):
Basically the new feature is working well, but i have a few suggestions, ptal into the inline comments :)
There are a few other things I noticed, please have a look:

1. Small Bug: Move new brick into a Forever brick with "Touch position": This works only once. I guess the correct behavior should be that the sprite should always move to the touch position when the user touches the screen.

2. The two failing testcases in `BricksXmlSerializerTest` on the CI are actually related to the changes. The reason is: there is no alias registered for `GlideToPositionBrick`. To fix it just add this line in `XStreamSerializer`: `xstream.alias("brick", GlideToPositionBrick.class);`

3. Please make sure the copyright year is up to date in every file

4. Please fix all the code style warnings.

gradle/gradle-daemon-jvm.properties
rinaschedl: This file make sure that every coder, reviewer etc is using the same Java.Version in his IDE settings, so to reduce codeproblems according to different javaversions.

.idea/.name
rinaschedl: I can't remember that I removed a file. What does this file normaly do?

catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java
rinaschedl: There is no duplicated code, because in each case you call a different instance, there you set the scope etc.

catroid/src/main/java/org/catrobat/catroid/content/bricks/GlideToPositionBrick.kt
rinaschedl: applied changes;

catroid/src/main/java/org/catrobat/catroid/content/bricks/GlideToPositionBrick.kt
rinaschedl: applied changes

catroid/src/main/java/org/catrobat/catroid/content/ActionFactory.java
hannesweilharter: I agree that the three cases create different action instances. What I meant by duplication is that the same initialization is repeated for each instance: `setScope(scope)` and `setDuration(duration.interpretFloat(scope))`.

Since the three actions currently don't share a common `GlideToAction` parent and only extend `TemporalAction`, I understand that this can't simply be moved after the switch without introducing a common type. My suggestion was mainly to avoid repeating this initialization, e.g. by introducing a common parent/interface or a small helper if that fits the existing design.

.idea/.name
hannesweilharter: It's an Android Studio project metadata file which just contains the IDE project name. If there is no reason you removed it, please just revert the change.

